PR #324, opened 2018-06-08
Make helm chart support multiple namespaces

ramkumarvs:
Update helm chart to support multiple namespaces 
@rkarthik007 

In order to run helm charts to create multiple instances of YugaByte DB you would run the following command.
`helm install yugabyte --name dev`
That would bring up yugabyte in `dev` namespace

`helm install yugabyte --name staging`
That would bring up yugabyte in `staging` namespace

And you can independently cleanup
`helm delete dev`

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/values.yaml
+++ b/cloud/kubernetes/helm/yugabyte/values.yaml
@@ -15,16 +15,16 @@ persistentVolume:
 PodManagementPolicy: Parallel
 
 Services:
-  - name: "yb-masters"
-    label: "yb-master"
+  - name: "masters"
+    label: "master"
     replicas: 3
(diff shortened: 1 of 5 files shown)


cloud/kubernetes/helm/yugabyte/values.yaml
```diff
@@ -15,16 +15,16 @@ persistentVolume:
 PodManagementPolicy: Parallel
 
 Services:
-  - name: "yb-masters"
-    label: "yb-master"
+  - name: "masters"
+    label: "master"
```
rkarthik007: Why drop the yb prefix? Our docs for kubernetes will be out of sync with what the helm chart does.

PR #367, opened 2018-07-07
Bump helm chart version and add storageClassName

ramkumarvs:
Tested by running
`helm package`
and 
`helm install yugabyte-1.0.4.tgz --name demo --wait`

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/values.yaml
+++ b/cloud/kubernetes/helm/yugabyte/values.yaml
@@ -9,7 +9,7 @@ Image:
 
 persistentVolume:
   count: 2
-  storage: 100Mi
+  storage: 10Gi
   storageClass: standard 
 
 PodManagementPolicy: Parallel
--- a/cloud/kubernetes/helm/yugabyte/values.yaml
+++ b/cloud/kubernetes/helm/yugabyte/values.yaml
@@ -4,13 +4,13 @@
 Component: "yugabytedb"
 Image:
   repository: "yugabytedb/yugabyte"
-  tag: latest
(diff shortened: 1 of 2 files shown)


cloud/kubernetes/helm/yugabyte/values.yaml
```diff
@@ -4,13 +4,13 @@
 Component: "yugabytedb"
 Image:
   repository: "yugabytedb/yugabyte"
-  tag: latest
+  tag: 1.0.4.0-b24 
   pullPolicy: IfNotPresent
 
 persistentVolume:
   count: 2
   storage: 100Mi
```
rkarthik007: Is this the default value? Seems too low. We should make this at least 10GiB, otherwise its not a practical production value.

Also, this value can be overridden by the helm chart right?

cloud/kubernetes/helm/yugabyte/values.yaml
```diff
@@ -4,13 +4,13 @@
 Component: "yugabytedb"
 Image:
   repository: "yugabytedb/yugabyte"
-  tag: latest
```
rkarthik007: Can this be overridden by the helm chart?

cloud/kubernetes/helm/yugabyte/values.yaml
ramkumarvs: sure, will update to 10Gi, yeah this is overwritten from YugaWare

cloud/kubernetes/helm/yugabyte/values.yaml
ramkumarvs: yes

PR #372, opened 2018-07-09
Remove release name

ramkumarvs:
- Remove Release Name from the labels
- Use namespaces for helm install


ramkumarvs:
@rkarthik007

PR #450, opened 2018-08-24
Add loadbalancer option

ramkumarvs:
Validated by running

Expose master ui alone using loadbalancer (Default).
`helm install yugabyte --name demo --debug --dry-run`

Expose all the service (YQL, YEDIS) endpoints using one shared loadbalancer
`helm install yugabyte -f expose-all-shared.yaml --name demo --debug --dry-run`

Expose all the service (YQL, YEDIS) endpoints using one individual loadbalancers
`helm install yugabyte -f expose-all.yaml --name demo --debug --dry-run`

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/templates/service.yaml
+++ b/cloud/kubernetes/helm/yugabyte/templates/service.yaml
@@ -23,14 +23,23 @@ spec:
   selector:
     app: "{{ .label }}"
 
-{{- if .hasLoadBalancer }}
+{{- $service := . -}}
+
+{{ if eq $service.name "yb-masters" }}
+  {{- $metadata := $root.Values.serviceEndpoints.master }}
+{{ else }}
+  {{- $metadata := $root.Values.serviceEndpoints.tserver }}
+{{ end }}
+
 ---
+{{- range $type, $metadata :=  $root.Values.serviceEndpoints }}
+{{ if eq $service.label (printf "yb-%s" $type) }}
 apiVersion: v1
 kind: Service
 metadata:
-  name: "{{ .label }}-ui"
+  name: "{{ $service.label }}-service"
   labels:
-    app: "{{ .label }}"
+    app: "{{ $service.label }}"
     heritage: {{ $root.Release.Service | quote }}
     release: {{ $root.Release.Name | quote }}
     chart: "{{ $root.Chart.Name }}-{{ $root.Chart.Version }}"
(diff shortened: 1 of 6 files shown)


rkarthik007:
Had a few questions @ramkumarvs:
* Is the loadbalancer created by default? For example when you run the following is an lbs created:
```helm install yugabyte --name demo```

* In the command you had given in the diff description, were you missing the lbs name for the param `serviceEndpoints.tserver`?
```helm install yugabyte --set serviceEndpoints.tserver= --name demo --debug --dry-run```

File: cloud/kubernetes/helm/yugabyte/templates/service.yaml [hunk 1]
```diff
@@ -23,27 +23,34 @@ spec:
...
-      port: 7000
+    {{- range $label, $port :=  $metadata.ports }}
+    - name: {{ $label | quote }}
+      port: {{ $port }}
+    {{- end}}
   selector:
-    app: "{{ .label }}"
-  type: LoadBalancer
-{{- end }}
+    app: "{{ $service.label }}"
```
rkarthik007: By selecting this app, does the lbs get deployed into the correct namespace? Was not sure where we set the namespace for this lbs.

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -23,27 +23,34 @@ spec:
...
     release: {{ $root.Release.Name | quote }}
     chart: "{{ $root.Chart.Name }}-{{ $root.Chart.Version }}"
     component: "{{ $root.Values.Component }}"
 spec:
   clusterIP:
   ports:
-    - name: ui
-      port: 7000
+    {{- range $label, $port :=  $metadata.ports }}
+    - name: {{ $label | quote }}
```
rkarthik007: What are the lbs names going to be? Will they be `yql-port-service` and `yedis-port-service`? If so, can we change to `yb-cql-service` and `yb-redis-service`?

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -23,27 +23,34 @@ spec:
   selector:
     app: "{{ .label }}"
 
-{{- if .hasLoadBalancer }}
 ---
+{{- $service := . -}}
```
rkarthik007: Do we unconditionally create this? We should have a way of enabling/disabling this based on a flag. Ok to default to "enabled" in yw. Here maybe make it "disabled" default to make minikube deploys easier.

rkarthik007: Lets make all the three lbs separate flags in the helm chart, thats probably the safest. But lets always create the UI lbs by default - by setting that flag to true. We may have to discover if folks want lbs or node-port services in the context of various users.

PR #458, opened 2018-08-31
Add memory limit to ram ratio

ramkumarvs:
Tested by running helm install and validated 
`helm install yugabyte --name demo`

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/values.yaml
+++ b/cloud/kubernetes/helm/yugabyte/values.yaml
@@ -38,7 +38,7 @@ serviceEndpoints:
 Services:
   - name: "yb-masters"
     label: "yb-master"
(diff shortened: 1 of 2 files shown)


cloud/kubernetes/helm/yugabyte/values.yaml
```diff
@@ -49,6 +50,7 @@ Services:
       rpc-port: "7100"
       yql-port: "9042"
       yedis-port: "6379"
+    memory_limit_to_ram_ratio: 1
```
kmuthukk: The default (for TServers) is already 0.85 and we don't want to change that. We want to leave some room for other processes.

kmuthukk: In other words, let us not override the TServer setting for memory limit in this yaml file.

PR #481, opened 2018-09-25
Updated YAML to add partition field

Arnav15:
For staggered upgrade between master and tserver

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/templates/service.yaml
+++ b/cloud/kubernetes/helm/yugabyte/templates/service.yaml
@@ -88,7 +88,9 @@ spec:
           requests:
             storage: {{ $root.Values.persistentVolume.storage }}
     {{- end }}
-
+  {{ if $root.Values.waitTime }}
+  minReadySeconds: {{ $root.Values.waitTime }}
+  {{ end }}
   updateStrategy:
     type: RollingUpdate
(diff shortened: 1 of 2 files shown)


cloud/kubernetes/helm/yugabyte/values.yaml
```diff
@@ -26,6 +26,10 @@ replicas:
   master: 3
   tserver: 3
 
+Partition:
```
ramkumarvs: change this to lower case just like replicas.

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -88,9 +88,17 @@ spec:
           requests:
             storage: {{ $root.Values.persistentVolume.storage }}
     {{- end }}
-
+  {{ if $root.Values.waitTime }}
+  minReadySeconds: {{ $root.Values.waitTime }}
```
ramkumarvs: it is not clear on what the waitTime is for, may be we need to have it under updateStrategy or something?. That way in future we can even change the update to be non rolling if needed.

something like
```
updateStrategy:
    type: RollingUpdate
    delaySecond: 180
    masterPartition: 0
    tserverPartition: 3
```

PR #521, opened 2018-10-29
Add gflags override option for helm

ramkumarvs:
Tried to create a YB universe via helm with expose all override yaml. And validated the cassandra authentication was enabled.

`helm install yugabyte -f expose-all.yaml --namespace yb-demo --name yb-demo --wait`

### DIFF
--- a/cloud/kubernetes/helm/expose-all.yaml
+++ b/cloud/kubernetes/helm/expose-all.yaml
@@ -16,3 +16,8 @@ serviceEndpoints:
     app: "yb-tserver"
     ports:
       yedis-port: "6379"
+
+gflags:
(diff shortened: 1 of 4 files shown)


cloud/kubernetes/helm/expose-all.yaml
```diff
@@ -16,3 +16,8 @@ serviceEndpoints:
     app: "yb-tserver"
     ports:
       yedis-port: "6379"
+
+gflags:
+  tserver:
+    use_cassandra_authentication: True
```
rkarthik007: Lets keep the default here false. We can just add docs to try setting this to true.

PR #533, opened 2018-11-05
Updated Statefulset yaml

Arnav15:
1) Added GET_HOSTS_FROM field to prevent issues in minikube as mentioned in https://github.com/kubernetes/minikube/issues/314. Tested by running itests several times and checking that master addresses are always resolved.
2) Added parameters to support annotations from yugaware. Tested by creating a GKE cluster and creating a universe with the required annotations.
`helm install yugabyte --namespace <namespace> --name <name> --set annotations.master.loadbalancer=<annotation>` --wait`


cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -31,6 +31,18 @@ apiVersion: v1
 kind: Service
 metadata:
   name: "{{ $endpoint.name }}"
+{{- if $root.Values.annotations }}
+  annotations:
+  {{ if eq .name "yb-master-service" }}
+    {{- if $root.Values.annotations.master }}
```
ramkumarvs: can we add a check to make sure we have got a loadbalancer annotation, if not we don't need to add any annotation.

PR #583, opened 2018-11-14
Add Metric Node Name to helm chart

### LINKED ISSUE #586: K8s: Add metric node name to cluster so prometheus metrics can be filtered
Currently we don't add metric_node_name to the yb-admin command for clusters we bring up via K8s we need to add this in order for us to do prometheus metric filter from YugaWare.

ramkumarvs:
Add metric node name parameter for both master and tserver, this is needed if we want the correct exported_instance name to be populated for prometheus metrics

Tested by bring up a YB cluster
`helm install yugabyte -f expose-all-shared.yaml --namespace yb-demo --name yb-demo --wait`

And in the tserver service expose port 9000 as well. then hit the LB endpoint 
`curl http://<LB IP>:9000/prometheus-metrics`

This PR fixes #586


rkarthik007:
Would be good to create a github issue for this @ramkumarvs before merging.

Issue #628, opened 2018-12-04
Allow using both --memory_limit_hard_bytes and --default_memory_limit_to_ram_ratio together


bmatican:
Based on https://github.com/YugaByte/yugabyte-db/blob/efd44a194d887a47b91eca9a75ec55cda598227e/src/yb/util/mem_tracker.cc#L235

we should allow for a use case where we can specify total amount of memory allocated, as well as a hard  limit percentage, to be able to better cover issues such as #627.

kmuthukk:
We can move the multiplication by FLAGS_default_memory_limit_to_ram_ratio to outside the `if` block.

That is, change the following from:
```
  int64_t limit = FLAGS_memory_limit_hard_bytes;
  if (limit == 0) {
    int64_t total_ram;
    CHECK_OK(Env::Default()->GetTotalRAMBytes(&total_ram));
    limit = total_ram * FLAGS_default_memory_limit_to_ram_ratio;
  }
```
to something like:
```
  int64_t limit = FLAGS_memory_limit_hard_bytes;
  if (limit == 0) {
    int64_t total_ram;
    CHECK_OK(Env::Default()->GetTotalRAMBytes(&total_ram));
    limit = total_ram;
  }
  limit = limit * FLAGS_default_memory_limit_to_ram_ratio;

```

rahuldesirazu:
@bogdan @kmuthukk Is this still a relevant task?

kmuthukk:
No - I think we decided to keep these gflags separate. 

If explicit value is provide, we just use it. Otherwise, it is computed as a ratio of RAM. Closing this task out (as I don't believe we are planning to do this change).

PR #637, opened 2018-12-05
Updated helm charts to set memory limits to RAM

Arnav15:
Changes also include the namespace parameter for DNS resolution as well as gflag for logging to stderr.


cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -176,20 +181,24 @@ spec:
           - "/home/yugabyte/bin/yb-master"
           - "--fs_data_dirs={{range $index := until (int ($root.Values.persistentVolume.count))}}{{if ne $index 0}},{{end}}/mnt/disk{{ $index }}{{end}}"
           - "--rpc_bind_addresses=$(POD_IP)"
-          - "--server_broadcast_addresses=$(HOSTNAME).yb-masters:7100"
-          - "--master_addresses={{range $index := until (int ($root.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters:7100{{end}}"
+          - "--server_broadcast_addresses=$(HOSTNAME).yb-masters.$(NAMESPACE):7100"
+          - "--master_addresses={{range $index := until (int ($root.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$(NAMESPACE):7100{{end}}"
           - "--master_replication_factor={{ $root.Values.replicas.master }}"
           - "--metric_node_name=$(HOSTNAME)"
+          - "--memory_limit_hard_bytes={{ $root.Values.resource.master.limits.memory | regexFind "\\d+" | mul 1024 | mul 1024 | mul 870 }}"
```
kmuthukk: So we are assuming that "limits.memory" is in Gi, and does the regexFind drop the Gi part?

kmuthukk: 1) would be nice if there was a way to add comments to explain the math, and the above assumption.

2) The 870 would later become 1024 I suppose, once the enhancement for https://github.com/YugaByte/yugabyte-db/issues/628 lands, right?

cloud/kubernetes/helm/yugabyte/values.yaml
```diff
@@ -17,18 +17,24 @@ resource:
...
+      memory: 4Gi
 
 replicas:
   master: 3
   tserver: 3
 
 partition:
-  master: 3
-  tserver: 3
+  master: 0
```
kmuthukk: What does the "partition" field do? And is 0 fine?

Arnav15: This partition doesn't have anything to do with the master replication factor. This is used specifically just for rolling upgrades. We changed it to 0 since we cannot control rolling upgrades with custom intervals in CE for K8s.

cloud/kubernetes/helm/yugabyte/templates/service.yaml
Arnav15: Yes, we are assuming the memory is in GB (since that is how it is driven from YW). 
1) I'll add comments and create a separate PR.
2) Correct, currently the helm templates do not support floating point calculations and that is why I had to do it this way (could add a divide as well, but seemed like overkill)

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -196,6 +197,7 @@ spec:
           - "--tserver_master_addrs={{range $index := until (int ($root.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$(NAMESPACE):7100{{end}}"
           - "--tserver_master_replication_factor={{ $root.Values.replicas.master }}"
```
Arnav15: I also looked at the discussion in ENG-4215 and it seems the best way moving forward is to remove the flag for the tserver_master_replication_factor (it'll default to 0)

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -196,6 +197,7 @@ spec:
...
           - "--metric_node_name=$(HOSTNAME)"
+          - "--memory_limit_hard_bytes={{ $root.Values.resource.tserver.limits.memory | regexFind "\\d+" | mul 1024 | mul 1024 | mul 870 }}"
```
ramkumarvs: @Arnav15 so what i mentioned was to move and create a template function inside _helpers.tpl

```
{{/*
Derive the memory hard limit for each POD based on the memory limit.
Since the memory is represented in <x>GBi, we use this function to convert that into bytes.
*/}}
{{- define "yugabyte.memory_hard_limit" -}}
{{- printf "%d" .limits.memory | regexFind "\\d+" | mul 1024 | mul 1024 | mul 870 }}
{{- end -}}
```

Once you defined this, you could just inline it like this
`- "--memory_limit_hard_bytes={{ template "yugabyte.memory_hard_limit" $root.Values.resource.tserver }}"`

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -185,6 +185,7 @@ spec:
           - "--master_addresses={{range $index := until (int ($root.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$(NAMESPACE):7100{{end}}"
           - "--master_replication_factor={{ $root.Values.replicas.master }}"
           - "--metric_node_name=$(HOSTNAME)"
+          - "--memory_limit_hard_bytes={{ $root.Values.resource.master.limits.memory | regexFind "\\d+" | mul 1024 | mul 1024 | mul 870 }}"
```
ramkumarvs: if you follow the template function route, you can just replace this with
`- "--memory_limit_hard_bytes={{ template "yugabyte.memory_hard_limit" $root.Values.resource.master }}"`

Arnav15: Done!

cloud/kubernetes/helm/yugabyte/templates/service.yaml
Arnav15: Done!

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -185,6 +184,8 @@ spec:
           - "--master_addresses={{range $index := until (int ($root.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$(NAMESPACE):7100{{end}}"
           - "--master_replication_factor={{ $root.Values.replicas.master }}"
           - "--metric_node_name=$(HOSTNAME)"
+          # Assume memory is always mentioned in GBs. Also multiplied by 870 since 0.85 * 1024 ~ 870 (floating calculations not supported)
```
ramkumarvs: move this comment to the template function. so it is not repeated.

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -194,8 +195,10 @@ spec:
           - "--fs_data_dirs={{range $index := until (int ($root.Values.persistentVolume.count))}}{{if ne $index 0}},{{end}}/mnt/disk{{ $index }}{{end}}"
           - "--server_broadcast_addresses=$(HOSTNAME).yb-tservers.$(NAMESPACE):9100"
           - "--tserver_master_addrs={{range $index := until (int ($root.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$(NAMESPACE):7100{{end}}"
-          - "--tserver_master_replication_factor={{ $root.Values.replicas.master }}"
           - "--metric_node_name=$(HOSTNAME)"
+          # Assume memory is always mentioned in GBs. Also multiplied by 870 since 0.85 * 1024 ~ 870 (floating calculations not supported)
```
ramkumarvs: ditto, move it to the template function

PR #666, opened 2018-12-14
Fixed gflag to log to stderr as well as disk

Arnav15:
Changed the gflad logtostderr to stderrthreshold=0 so that k8s logs to disk as well as std err.

Tested by creating a universe and checking the logs are available through kubectl logs as well as through the pod disk


Arnav15:
Created a new PR from fork

PR #718, opened 2019-01-08
Split the volume claim override for master and tserver

### LINKED ISSUE #678: [k8s-deployment] Make the storage size for master and tserver configurable
Right now we use the same storage specification both for master and tserver. Ideally masters don't need to have the storage as a tserver. So make this overridable with sensible defaults for master as 40Gi and 100Gi for tserver.

ramkumarvs:
Make the volume claim to be overridable at master and tserver separately

Tested by overriding the default storage size for tserver and the storage count for master.
```
helm install yugabyte --set "storage.tserver.size=30Gi,storage.master.count=1" --namespace yb-demo --name yb-demo --wait
```
Tested by launch cluster with defaults.
```
helm install yugabyte --namespace yb-demo --name yb-demo --wait
```
fixes #678


ramkumarvs:
@Arnav15 yes working on that now.

PR #884, opened 2019-02-21
Helm add domain

### LINKED ISSUE #883: [k8s] Expose the cluster domain name so it can be overwritten
We need to expose the cluster domain name the default is `cluster.local`. So we can override the value if needed via helm charts.

ramkumarvs:
- Added an override param `domainName` at each of the service level. 
- Moved some common code into the _helpers.tpl file and use template calls

Tested by bring up a TLS enabled cluster.

`helm install yugabyte --namespace yb-demo --name yb-demo --set=tls.enabled=true --set=Image.repository=quay.io/yugabyte/yugabyte --set=Image.pullSecretName=yugabyte-k8s-pull-secret --wait`

Also tested by bring up non TLS enabled cluster.
`helm install yugabyte --namespace yb-demo --name yb-demo --set=Image.repository=quay.io/yugabyte/yugabyte --set=Image.pullSecretName=yugabyte-k8s-pull-secret --wait`

fixes #883

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
+++ b/cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
@@ -50,7 +50,13 @@ Create chart name and version as used by the chart label.
   Get YugaByte master addresses
 */}}
 {{- define "yugabyte.master_addresses" -}}
-{{range $index := until (int (.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$(NAMESPACE):7100{{end}}
+{{- $master_replicas := .Values.replicas.master | int -}}
+  {{- range .Values.Services }}
+    {{- if eq .name "yb-masters" }}
+      {{- $domain_name := .domainName -}}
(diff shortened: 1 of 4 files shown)


cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
```diff
@@ -38,3 +38,24 @@ Create chart name and version as used by the chart label.
...
+*/}}
+{{- define "yugabyte.fs_data_dirs" -}}
+{{range $index := until (int (.count))}}{{if ne $index 0}},{{end}}/mnt/disk{{ $index }}{{end}}
+{{- end -}}
+
+{{/*
+  Get YugaByte master addresses
+*/}}
+{{- define "yugabyte.master_addresses" -}}
+{{range $index := until (int (.Values.replicas.master))}}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$(NAMESPACE):7100{{end}}
```
ajcaldera1: Missing .svc and .domainName here -- should be specified to avoid DNS thrashing

ramkumarvs: sure let me fix this as well.

PR #1035, opened 2019-03-19
Added initdb support in helm

Arnav15:
Updated helm chart to add ability to initialize redis and postgres if required by the user. Verified by running the following helm install command:
`helm install yugabyte --namespace yb-demo --name yb-demo --set initDb=True --wait`

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
+++ b/cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
@@ -60,6 +60,19 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+  Get YugaByte master addresses for PostStart
+*/}}
+{{- define "yugabyte.master_addresses_ps" -}}
+{{- $master_replicas := .Values.replicas.master | int -}}
+  {{- range .Values.Services }}
+    {{- if eq .name "yb-masters" }}
+      {{- $domain_name := .domainName -}}
+      {{range $index := until $master_replicas }}{{if ne $index 0}},{{end}}yb-master-{{ $index }}.yb-masters.$NAMESPACE.svc.{{ $domain_name }}:7100{{end}}
+    {{- end -}}
+  {{- end -}}
+{{- end -}}
+
+{{/*
 Get the fully qualified server address
 */}}
 {{- define "yugabyte.server_address" -}}
--- a/cloud/kubernetes/helm/yugabyte/templates/service.yaml
+++ b/cloud/kubernetes/helm/yugabyte/templates/service.yaml
@@ -180,6 +180,22 @@ spec:
       containers:
       - name: "{{ .label }}"
         image: "{{ $root.Values.Image.repository }}:{{ $root.Values.Image.tag }}"
+        {{ if $root.Values.initDb}}
+        {{ if eq .name "yb-tservers" }}
+        lifecycle:
+          postStart:
+            exec:
+              command:
+                - "sh"
+                - "-c"
+                - >
+                  if [ "$HOSTNAME" == "yb-tserver-0" ]; then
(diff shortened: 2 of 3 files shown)


cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -180,6 +180,24 @@ spec:
       containers:
       - name: "{{ .label }}"
         image: "{{ $root.Values.Image.repository }}:{{ $root.Values.Image.tag }}"
+        {{ if $root.Values.initDb}}
+        {{ if eq .name "yb-tservers" }}
+        lifecycle:
```
ramkumarvs: `{{- $masterAddresses := template "yugabyte.master_addresses" $root  -}}` try the variable route so we don't duplicate the function

cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -180,6 +180,24 @@ spec:
...
+            exec:
+              command:
+                - "sh"
+                - "-c"
+                - >
+                  if ! ./bin/yb-admin -master_addresses {{ template "yugabyte.master_addresses_ps" $root }} list_tables | grep -q "postgres.pg_enum"; then
+                  if [ "$HOSTNAME" == "yb-tserver-0" ]; then
+                  while ! ./bin/yb-admin -master_addresses {{ template "yugabyte.master_addresses_ps" $root }} list_all_masters | grep -q "LEADER"; do true; done;
+                  YB_ENABLED_IN_POSTGRES=1 FLAGS_pggate_master_addresses={{ template "yugabyte.master_addresses_ps" $root }} /home/yugabyte/tserver/postgres/bin/initdb -D /tmp/yb_pg_initdb_tmp_data_dir -U postgres > /tmp/initdb.log;
+                  ./bin/yb-admin -master_addresses {{ template "yugabyte.master_addresses_ps" $root }} setup_redis_table;
```
rkarthik007: Are we tying `setup_redis_table` to the `initDb` step? We should not do this. Lets keep the creation of the redis table a separate step, and defaulted to false.

We want to reach a spot where we always run initDb by default and not the redis table setup.

cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
```diff
@@ -59,6 +59,19 @@ Create chart name and version as used by the chart label.
   {{- end -}}
 {{- end -}}
 
+{{/*
+  Get YugaByte master addresses for PostStart
```
rkarthik007: What does "for PostStart" mean? Should this just be "since we would need it post startup of the cluster"?

cloud/kubernetes/helm/yugabyte/templates/service.yaml
Arnav15: I think @bmatican wanted the setup redis to also be done as part of the initdb. We can remove it though if it isn't required.

cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
Arnav15: PostStart is the container lifecycle hook (so the k8s term for doing the operation PostStart of the container). The reason we need another template to get the master addresses is due to the fact that while running the command, we cannot have parenthesis around the env variables, but the parenthesis are required by the YB containers while setting the flags.

rkarthik007: Makes sense, thanks.

rdonegan09rr: Nide_js

rdonegan09rr: Nide_js

rdonegan09rr: Nide_js

rkarthik007:
Closing because this has been moved to another repo.

PR #1087, opened 2019-03-27
[yb-docker-ctl] Save initdb log

ndeodhar:


bin/yb-docker-ctl
```diff
@@ -418,11 +418,15 @@ class YBDockerControl():
                         init_db_cmd = "{} {}/initdb -U postgres -D {}".format(
                             init_db_envs, pg_bin_dir, tmp_data_dir)
                         # Setup the list based cmdline.
-                        print("Running initdb to initialize PostgreSQL metadata "
-                              "in the YugaByte cluster. This may take up to a minute.")
                         docker_cmd = ['docker', 'exec', '-it', self.YB_TSERVER_FORMAT.format(1)]
                         docker_cmd.extend(["bash", "-c", init_db_cmd])
+                        print(' '.join(docker_cmd))
+                        print("Running initdb to initialize PostgreSQL metadata "
```
bmatican:
```suggestion
                        print("Running initdb to initialize YSQL metadata "
```

Since @schoudhury was suggesting we turn our references to YSQL.

bin/yb-docker-ctl
```diff
@@ -418,11 +418,15 @@ class YBDockerControl():
                         init_db_cmd = "{} {}/initdb -U postgres -D {}".format(
                             init_db_envs, pg_bin_dir, tmp_data_dir)
                         # Setup the list based cmdline.
-                        print("Running initdb to initialize PostgreSQL metadata "
-                              "in the YugaByte cluster. This may take up to a minute.")
                         docker_cmd = ['docker', 'exec', '-it', self.YB_TSERVER_FORMAT.format(1)]
                         docker_cmd.extend(["bash", "-c", init_db_cmd])
+                        print(' '.join(docker_cmd))
```
bmatican: Do we really want to print out the whole command?

bin/yb-docker-ctl
```diff
@@ -418,11 +418,15 @@ class YBDockerControl():
...
+                        print("Running initdb to initialize YSQL metadata "
+                              "in the YugaByte cluster. This may take up to a minute.")
                         result = get_subprocess_result_as_str(docker_cmd)
+                        with open("initdb.log", "w") as initdb_log:
```
bmatican: I guess for now it's fine to put this log locally, but, for example, for yb-ctl, we might end up using the `--data_dir`. Unclear if there's a better option though for `yb-docker-ctl` though...

bin/yb-docker-ctl
ndeodhar: Yes, I want to keep it consistent with how we print out commands for yb-master, yb-tserver. Also, for now, provides more transparency to users.

bin/yb-docker-ctl
ndeodhar: Yea, I had thought about this. In case of docker, there's not a better option right now.

PR #1188, opened 2019-04-10
[#1063] Updated helm chart to support multi-az

Arnav15:
Added nodeAffinity spec to orchestrate pod deployments on the nodes in the required zones. Also updated the master addresses to be customizable for multi-az where the addresses can't be computed by the template.
Adds support for issue #1063

### DIFF
--- a/cloud/kubernetes/helm/yugabyte/templates/service.yaml
+++ b/cloud/kubernetes/helm/yugabyte/templates/service.yaml
@@ -121,6 +121,8 @@ spec:
     {{- range $index := until (int ($storageInfo.count )) }}
     - metadata:
         name: datadir{{ $index }}
+        annotations:
+          volume.beta.kubernetes.io/storage-class: {{ $storageInfo.storageClass }}
         labels:
           heritage: {{ $root.Release.Service | quote }}
           release: {{ $root.Release.Name | quote }}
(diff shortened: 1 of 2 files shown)


cloud/kubernetes/helm/yugabyte/templates/service.yaml
```diff
@@ -212,8 +224,13 @@ spec:
           - "--fs_data_dirs={{ template "yugabyte.fs_data_dirs" $storageInfo }}"
           - "--rpc_bind_addresses={{ template "yugabyte.server_address" . }}"
           - "--server_broadcast_addresses={{ template "yugabyte.server_address" . }}:7100"
+          {{ if $root.Values.isMultiAz }}
+          - "--master_addresses={{ $root.Values.masterAddresses}}"
+          - "--replication_factor={{ $root.Values.replicas.totalMasters }}"
```
rkarthik007: You would need the masters to come up with the correct placement info. See this step: https://docs.yugabyte.com/latest/deploy/public-clouds/aws/#create-yb-master1-s-config-file

There should be a corresponding thing done for the tservers also.

cloud/kubernetes/helm/yugabyte/values.yaml
```diff
@@ -60,6 +60,8 @@ PodManagementPolicy: Parallel
 
 enableLoadBalancer: True
 
+isMultiAz: False
```
rkarthik007: If this is set, at the end of the deploy you need to set the config in the master also, its an API call. See this step: https://docs.yugabyte.com/latest/deploy/public-clouds/aws/#7-configure-az-region-aware-placement

cloud/kubernetes/helm/yugabyte/templates/service.yaml
Arnav15: Those flags are passed while creating the universe through YW.

cloud/kubernetes/helm/yugabyte/values.yaml
Arnav15: I believe this is also done through YW during the create universe step.

cloud/kubernetes/helm/yugabyte/templates/service.yaml
rkarthik007: Got it!

rkarthik007:
This mostly looks good to me, will let @ramkumarvs take a look and accept.

@Arnav15 - can you also write the set of steps you used to being up a multi-zone universe? For example, creating the storage classes, args for each run of the helm chart, etc?

Issue #1262, opened 2019-04-25
[k8s] Implement PodDisruptionBudgets for single AZ / single cluster in kubernetes


ajcaldera1:
In order to make sure that customers can do upgrades of underlying infrastructure in Kubernetes, we should implement `PodDisruptionBudgets` to ensure that quorum is maintained for both masters and tservers.

ajcaldera1:
Closed with commit 0d6f119

PR #1263, opened 2019-04-25
issue #1262 - Implement PodDisruptionBudgets for single AZ / single cluster installations

ajcaldera1:
Tested against single AZ's at replication factor 3,5 and 7.
Verified no creation of PodDisruptionBudgets in multi-AZ case


cloud/kubernetes/helm/yugabyte/templates/_helpers.tpl
```diff
@@ -65,3 +65,12 @@ Get the fully qualified server address
 {{- define "yugabyte.server_address" -}}
 {{- printf "$(HOSTNAME).%s.$(NAMESPACE).svc.%s" .name .domainName }}
 {{- end -}}
+
+{{/*
+Compute the maximum number of unavailable pods based on the number of master replicas
```
ramkumarvs: can you add a comment on the formula itself.

Issue #1179, opened 2019-04-09
K8s: Add ability to specify network name in YW and YB helm charts


RashmiPanthangi:
We need to be able to create YB pods in tenants network. The deployment needs to support following annotation:
spec:
  template:
    metadata:
      annotations:
        <example.com>/endpoint0: '{" network":" << network_name >>"}'

Both YW and YB helm charts need to have the above configuration, which will then create the pods in the specified network and we can bring up the instances in specified network.

Arnav15:
Is it possible to get an actual YAML of a deployment with the network annotation specified? @RashmiPanthangi

ramkumarvs:
They way we have implemented is, given annotations can be at any level, we wanted to build a generic way for the customers to provide overrides, so what we have done is for the customer to provider a top level overrides yaml, which can have annotations for network, load balancer and in future we can introduce other items as well. 

A sample overrides yaml provided can be something like this where you can see the customer has a way to override the ports, loadbalancer type, add any annotations etc all in one.

in turn we have orchestrated our helm chart in such a way that it can apply the override to the correct section.

```
serviceEndpoints:
  - name: yb-master-service
    type: LoadBalancer
    app: yb-master
    annotations:
      annotation-1: foo
    ports:
      ui: 7000

  - name: yb-tserver-service
    type: LoadBalancer
    app: yb-tserver
    ports:
      ycql-port: 9042
      yedis-port: 6379

networkAnnotation:
  diamanti.com/endpoint0: '{network: ec, perfTier: high}'

```

PR #1282, opened 2019-04-29
Added support for Network annotations for the Statefulset deployements.

Arnav15:
Also modified the annotations for the services to now be populated with the service endpoints.

Tested the helm chart by creating a statefulset deployment with the network annotation and also checked the service annotations after editing the serviceEndpoint values in Values.yaml.
Fixes for: #1179 and #797


ramkumarvs:
@Arnav15 Did we have any documentation around how to set the load balancer annotations

PR #1286, opened 2019-04-30
Documentation for Yugabyte Installs

mwbutera:


docs/yugabyte-paths.md
```diff
@@ -0,0 +1,20 @@
+<h2>Cluster Install Directories</h2>
```
rkarthik007: This is very useful! But this information would be ideally added to the following page: https://docs.yugabyte.com/latest/admin/yb-ctl/
The above page corresponds to the following file: `content/latest/admin/yb-ctl.md`

Do you mind adding this info there? Otherwise please let us know and we can do that.

schoudhury:
Added this information to the docs repo and pushed the changes live: https://docs.yugabyte.com/latest/admin/yb-ctl/#default-cluster-configuration

rkarthik007:
This has been incorporated as a part of [7eb323b](https://github.com/YugaByte/docs/commit/7eb323bf8fd5d6faa919b4a55d3e59ffb6859656), thanks for your contribution @mwbutera!

Issue #1334, opened 2019-05-08
Generate kubeconfig script not fetching the context properly


ramkumarvs:
We actually have a script to generate the kubeconfig file with the user token, and it looks like that script is not pulling the context correctly.

When you run the following command

`kubectl --namespace kube-system config get-contexts gke_yugabyte_us-west1-b_yb-gke-demo`
returns this
```
CURRENT   NAME                                    CLUSTER                                 AUTHINFO                                NAMESPACE
*         gke_yugabyte_us-west1-b_yb-gke-demo   gke_yugabyte_us-west1-b_yb-gke-demo   gke_yugabyte_us-west1-b_yb-gke-demo
```
the cluster is in the third column, we are using the value from the fourth column instead.

PR #1335, opened 2019-05-08
Fix a bug with generate kubeconfig script

ramkumarvs:
Tested by running

`./cloud/kubernetes/helm/generate_kubeconfig.py -s yugabyte-helm`

fixes: #1334


ramkumarvs:
> LG! Surprised that the auth_info and cluster name could be the same!

yeah in GKE it seems the same but on diamanti it wasn't

PR #1593, opened 2019-06-20
PHP sample for ysql API

kmuthukk:


docs/content/latest/develop/build-apps/_index.html
```diff
@@ -132,7 +132,20 @@
     </a>
   </div>
 
-  
+  <div class="col-12 col-md-6 col-lg-12 col-xl-6">
+    <a class="section-link icon-offset" href="php/">
+      <div class="head">
+        <div class="icon">
+          <i class="icon-csharp"></i>
```
rkarthik007: Maybe use `icon-php` here?

docs/content/latest/develop/build-apps/php.md
```diff
@@ -0,0 +1,28 @@
+---
+title: PHP
+linkTitle: PHP
+description: Build a PHP App
+aliases:
```
rkarthik007: Don't think we need these now. This was from when the links were moved around from a different location.

kmuthukk: I'll defer this change because other files seem to have it also.

docs/content/latest/develop/build-apps/_index.html
kmuthukk: done

Issue #1757, opened 2019-07-10
[docs] Anchor links seem broken


bmatican:
I was using our docs search: https://docs.yugabyte.com/search/?q=sysctl

The one result pointed to: https://docs.yugabyte.com/latest/quick-start/install/#prerequisites

This apparently first loads the page, then blanks the body.

Finally, when you click on, say, the linux tab, it correctly loads, but does not change the link to the new anchor.

cc @rkarthik007 @schoudhury @vitpankin

yb-andrew:
The first issue if the docs search leading to a broken link is due to my change here: https://github.com/YugaByte/docs/pull/114
I think we'll have to update algolia or our config file

Second issue of the tab not changing the url from the anchor link is due to my change here: https://github.com/YugaByte/docs/pull/108
I have a solution for this issue.

yb-andrew:
Fixed anchor links issue but Algolia search may still be outdated

PR #1759, opened 2019-07-10
Fix anchor link not updating the URL and preventing the page from scrolling

yb-andrew:
Relates to #1757 
The page will not scroll on newer browsers; however, on older browsers you will still see the page jump to the section.

### DIFF
--- a/docs/layouts/partials/footer_js.html
+++ b/docs/layouts/partials/footer_js.html
@@ -146,11 +146,12 @@
           $('.nav-tabs li a[href="'+href+'"], .nav-tabs-alt li a[href="'+href+'"]').addClass('active');
 
           const hashValue = $.attr(this, 'href');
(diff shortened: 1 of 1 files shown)


docs/layouts/partials/footer_js.html
```diff
@@ -145,6 +145,12 @@
           //get all nav tabs matching the href and set to active
           $('.nav-tabs li a[href="'+href+'"], .nav-tabs-alt li a[href="'+href+'"]').addClass('active');
 
+          const hashValue = $.attr(this, 'href');
+          if(history.pushState) {
```
bmatican: nit: space

yb-andrew:
https://stackoverflow.com/questions/3870057/how-can-i-update-window-location-hash-without-jumping-the-document/14690177

PR #1915, opened 2019-07-26
Remove dual action Add Universe button and replace

### LINKED ISSUE #1583: Confusing Add Universe GUI
There is this large Add Universe button which is divided into top and bottom part. The top part is create new universe and the bottom part is import universe. Excepted that there is no visual indication to their meaning. First time user as I was find it confusing as I hit the bottom part all the time at first but I needed the top part.

Feature Request
When mouse hover over top part, the top part gets high lighted and a text displays create and when hover over bottom text shows import.

Though I still don’t like that button with the feature request. The graphics suggest a single action, but it is actually two different actions.

yb-andrew:
Fixes #1583

We want to replace with similar layout as List Universes page, where the header is on the navbar and the two action buttons are on the top right.

### DIFF
--- a/managed/ui/src/components/dashboard/Dashboard.js
+++ b/managed/ui/src/components/dashboard/Dashboard.js
@@ -15,14 +15,9 @@ export default class Dashboard extends Component {
 
     return (
       <div id="page-wrapper">
-        <Row className="header-row">
-          <Col xs={6}>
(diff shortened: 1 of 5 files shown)


yb-andrew:
Changed design to header is on the same row as the two action buttons. Update icon on Create Universe action button.

PR #1965, opened 2019-08-02
Update cluster status example

### LINKED ISSUE #1964: Replace cluster status example in Create Local Cluster page
Replace existing example in "2. Check cluster status with yb-ctl" with new one provided by Jason.

[cluster_status.txt](https://github.com/YugaByte/yugabyte-db/files/3462800/cluster_status.txt)

https://docs.yugabyte.com/latest/quick-start/create-local-cluster/

stevebang:
Replace existing cluster status example with new one, provided by Jason.

Closes #1964.


docs/content/latest/admin/cqlsh.md
```diff
@@ -16,11 +16,11 @@ showAsideToc: true
 
 ## Overview
 
-`cqlsh` is a command line shell for interacting with YugaByte DB through [YCQL](../../api/ycql/). 
+`cqlsh` is a command line shell for interacting with YugaByte DB through [YCQL](/latest/api/ycql/).
```
kmuthukk: Why are we replacing relative links with hard-coded links? This will make further branching harder.

docs/content/latest/admin/cqlsh.md
```diff
@@ -1,7 +1,7 @@
 ---
```
rkarthik007: Can we split into 2 separate diffs? Working with strings should go into its own diff ideally. Having different sets of changes in a single PR would slow down the review and merge process.

docs/content/latest/admin/cqlsh.md
stevebang: Fixed -- replaced with relative links.

PR #2009, opened 2019-08-09
Change Get Started button link and add responsive styling

yb-andrew:
Point Get Started button to download.yugabyte.com and add styling for social media links in header when it is less than ultra wide.


rkarthik007:
This looks good to me, but there are some checks that are failing above. Is this good to merge @ramkumarvs ?

ramkumarvs:
LG. I have disabled netlify builds for now, we can debug that later.

PR #2093, opened 2019-08-17
Add a config.yaml.example file and update_config script

ramkumarvs:
In order for us to host our docs in netlify, we would need to make the config yaml to pull keys from environment variables.

Added that logic into an update_config.sh script which can be run before running the build command.

Tested by running after setting the environment variables.
`./update_config.sh`


rkarthik007:
@ramkumarvs - this looks good, will let you merge to verify the continuous deployment works.

Issue #721, opened 2019-01-08
Unexpected error log though master change config succeeded.


bbaddepudi:
Noticed this output in one of the recent ops:
`$ ~/master/bin/yb-admin -master_addresses <ip1,ip2> change_master_config REMOVE_SERVER <old_ip> 7100 1
I0108 22:30:54.751463 17854 mem_tracker.cc:240] MemTracker: hard memory limit is 25.722462 GB
I0108 22:30:54.751554 17854 mem_tracker.cc:242] MemTracker: soft memory limit is 21.864092 GB
_E0108 22:30:56.275851 17854 client-internal.cc:1166]
Error: Invalid argument (yb/client/client-internal.cc:1167): Unable to change master config: master addresses cannot be empty_`

The master UI shows master quorum updated correctly to remove the `old_ip`. 

The log seems to be due to some kind of first try printing an error instead of a debug/warning, need to change that to not show up by default.

viswanathreddy:
@bbaddepudi can  you please assign this issue to me?

kmuthukk:
Thx @viswanathreddy for offering to look into this.

viswanathreddy:
@bbaddepudi I am trying to reproduce the issue. I have cloned and build the codebase successfully. But i don't see yb-admin file in the yugabyte-db/bin folder. Am i missing something?

ndeodhar:
@viswanathreddy It should be there in build/latest/bin folder.

kmuthukk:
@viswanathreddy --

if you are looking at the source tree for the binary -- https://github.com/YugaByte/yugabyte-db/tree/master/bin, please note that this is an executable that's generated binary when you do a build of the source, and as @ndeodhar mentioned it'll land in build/latest/bin folder.

The starting point / source code for yb-admin is:

https://github.com/YugaByte/yugabyte-db/blob/master/src/yb/tools/yb-admin_cli.cc

viswanathreddy:
@bbaddepudi Though UI doesn't show removed server, 'bin/yb-ctl status' command still shows removed server. Does it mean the server is not actually removed? Will have to check the inconsistency between UI results and yb-ctl status command results

bmatican:
Hey @viswanathreddy , I think `yb-ctl status` is based off of directory structure on disk, as it is a local testing tool.

[...]

PR #2117, opened 2019-08-21
[#721] fix in unexpected error log with change in master config

viswanathreddy:
Summary: with yb-admin util, when master ip addr is removed using change_master_config flag, unexpected error log is popping on the terminal. Unassigned master ip addr's list was passed to SetMasterAddresses() which caused the issue


src/yb/client/client-internal.cc
```diff
@@ -1273,10 +1273,13 @@ Status YBClient::Data::RemoveMasterAddress(const HostPort& addr) {
     if (it != master_server_addrs_.end()) {
       master_server_addrs_.erase(it);
     }
-  }
 
-  RETURN_NOT_OK(SetMasterAddresses(HostPort::ToCommaSeparatedString(new_list)));
+    for(const auto& a : master_server_addrs_) {
+        updated_addrs += a;
```
kmuthukk: Are we missing commas (",") here when concatenating. Previously, we were doing ToCommaSeparatedString?

viswanathreddy: @kmuthukk ya I missed  ',' separator. Do we have a way to do unit testing so that i can avoid these trivial errors?

viswanathreddy: @kmuthukk updated

viswanathreddy:
@kmuthukk updated

kmuthukk:
Thx @viswanathreddy .

Will let @spolitov or @bmatican do the final review/accept.

viswanathreddy:
@spolitov updated

viswanathreddy:
@bmatican @mbautin

bmatican:
@viswanathreddy Confirmed tests look fine on our end so good to land this. I did want to add the test steps you did to the commit message, if you don't mind posting those, I'll merge this in!

Thanks for your help on this!

viswanathreddy:
@bmatican 
remove_master_config

PR #2189, opened 2019-08-29
Design for YugabyteDB operator

samkulkarni20:
This is the initial draft of YugabyteDB operator design. `operator_design.md` contains overall design details along with sample CR spec. `ybcluster_crd.yaml` contains CR validation rules.

Though I am raising this PR against `yugabyte-db` repo, should there be a separate repo for operator alone? I can close this & recreate the design PR, if we do create a separate repo for operator.

### DIFF
--- a/cloud/kubernetes/design/operator_design.md
+++ b/cloud/kubernetes/design/operator_design.md
@@ -1,0 +1,153 @@
+# YugabyteDB Operator Design
+
+A cluster of YugaByte DB should be created using the `ybclusters.yugabyte.com` custom resource definition. Below are sample custom resource specs for creating 3-master, 3 t-servers cluster using the CRD. The [minimal spec](#Minimal-Spec-Sample) and the [full spec](#Full-Spec-Sample) are essentially same, with [minimal spec](#Minimal-Spec-Sample) stripping away all the default vaules. The samples are followed by an explanation of different configuration options available on the YugaByte DB CRD.
+
+
+## Minimal Spec Sample
+
+Below is an example of a spec with minimum attributes required to define a custom resorce of type `ybclusters.yugabyte.com`. The spec below uses all possible defaults & specifies only the mandatory attributes.
+
+```yaml
+apiVersion: yugabyte.com/v1alpha1
+kind: YBCluster
+metadata:
+  name: yugabytedb
+spec:
+  replicationFactor: 3
+  master:
+    replicas: 3
+    storage:
+      size: 10Gi
+  tserver:
+    replicas: 3
+    storage:
+      count: 1
+      size: 10Gi
+```
+
+The spec can be used as a starting point &  can be updated with any other attributes, as needed. All valid attributes & their description is as below.
+
+## Full Spec Sample
+
+Below is an example of the full spec for custom resource `ybclusters.yugabyte.com`. It shows all possible attributes that can be specified with it, along with the default value, if any.
+
+```yaml
+apiVersion: yugabyte.com/v1alpha1
+kind: YBCluster
+metadata:
+  name: yugabytedb
+  namespace: yugabytedb
+spec:
+  image:
+    repository: yugabytedb/yugabyte     # Optional. Default value "yugabytedb/yugabyte".
+    tag: 1.3.2.0-b19                    # Optional. Default value "1.3.2.0-b19".
+    pullPolicy: IfNotPresent            # Optional. Default value "IfNotPresent".
+  tls:
+    enabled: false                      # Optional. Default value "false".
+    # rootCA:
+      # cert:                           # Uncomment if tls.enabled is set to true. Specify the root certificate generated .
+      # key:                            # Uncomment if tls.enabled is set to true. Specify the root certificate key generated.
+  replicationFactor: 3                # Required. Data replication factor for the cluster. Should be >= 1.
+  master:
+    replicas: 3                         # Required. Pod replica count for Master.
+    # Mentioning network ports is optional. If some or all ports are not specified, then they will be defaulted to below-mentioned values.
+    masterUIPort: 7000                  # Optional. Default value "7000".
+    masterRPCPort: 7100                 # Optional. Default value "7100".
+    enableLoadBalancer: false           # Optional. Default value "false". Change it to true to be able to access YugabyteDB Master UI over the internet.
+    podManagementPolicy: Parallel       # Optional. Default value "Parallel", out of valid values "Parallel" and "OrderedReady". If "OrderedReady" value is specified, cluster will take more time coming up.
+    storage:
+      count: 1                          # Optional. Default value "1".
+      size: 10Gi
+      storageClass: standard            # Optional. Default value "standard". Provide storage class to use.
+    resources:                          # Optional. No resource requests/limits will be applied, if this property is omitted. You may also specify one or both of requests & limits
+      requests:
+        cpu: 2
+        memory: 1Gi
+      limits:
+        cpu: 2
+        memory: 1Gi
+    gflags:                             # Optional. No GFlags will be applied, if this property is omitted. List at least one, when specified.
+      - key: default_memory_limit_to_ram_ratio
+        value: 0.85
+  tserver:
+    replicas: 3                         # Required. Pod replica count for TServer.
+    # Mentioning network ports is optional. If some or all ports are not specified, then they will be defaulted to below-mentioned values, except for tserver-ui.
+    # For tserver-ui a cluster ip service will be created if the yb-tserver-ui port is explicitly mentioned. If it is not specified, only StatefulSet & headless service will be created for TServer. TServer ClusterIP service creation will be skipped. Whereas for Master, all 3 kubernetes objects will always be created.
+    tserverUIPort: 9000                 # Optional. ClusterIP service will not be created, if omitted.
+    tserverRPCPort: 9100                # Optional. No default value. ClusterIP service won't be created for tserver, if this is omitted.
+    ycqlPort: 9042                      # Optional. Default value "9042".
+    yedisPort: 6379                     # Optional. Default value "6379".
+    ysqlPort: 5433                      # Optional. Default value "5433".
+    enableLoadBalancer: false           # Optional. Default value "false". Change it to true to be able to access YugabyteDB TServer UI over the internet. Value will be ignored if "tserverUIPort" is omitted.
+    # Volume claim template for TServer
+    podManagementPolicy: Parallel       # Optional. Default value "Parallel", out of valid values "Parallel" and "OrderedReady". If "OrderedReady" value is specified, cluster will take more time coming up.
+    storage:
+      count: 1
+      size: 10Gi
+      storageClass: standard            # Optional. Provide storage class to use. Field will be defaulted to "standard", if omitted.
+    resources:                          # Optional. No resource requests/limits will be applied, if this property is omitted. You may also specify one or both of requests & limits
+      requests:
+        cpu: 2
+        memory: 1Gi
+      limits:
+        cpu: 2
+        memory: 1Gi
+    gflags:                             # Optional. No GFlags will be applied, if this property is omitted. List at least one, when specified.
+      - key: default_memory_limit_to_ram_ratio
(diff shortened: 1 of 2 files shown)


design/operator_design.md
```diff
@@ -0,0 +1,129 @@
...
+  master:
+    # Replica count for Master.
+    replicas: 3
+    # Mentioning network ports is optional. If some or all ports are not specified, then they will be defaulted to below-mentioned values.
+    masterUIPort: 7000                  # Optional. default value "7000".
+    masterRPCPort: 7100                 # Optional. default value "7100".
+    enableLoadBalancer: false           # Optional. default value "false". Change it to true to be able to access YugabyteDB Master UI over the internet.
+    podManagementPolicy: Parallel       # Optional. default value "Parallel", out of valid values "Parallel" and "OrderedReady". If "OrderedReady" value is specified, cluster will take more time coming up.
+    storage:
+      count: 2
```
rkarthik007: The `yb-master` does not need 2 disks, 1 should suffice here. cc @ramkumarvs @Arnav15

design/operator_design.md
```diff
@@ -0,0 +1,129 @@
...
+        value: 0.85
+```
+
+# Cluster Settings
+
+### Image
+Mention YugabyteDB docker image attributes such as `repository`, `tag` and `pullPolicy` under `image`.
+
+### TLS
+Enable TLS encryption for YugabyteDB (Enterprise Edition only). Before setting this up, you need to have the pull secret installed from Yugabyte private Enterprise Edition registry. You can use the TLS encryption with 3 GFlags, explained later. If you have set `enabled` to true, specify your `cert` & `key` under `rootCA`.
```
rkarthik007: This is no longer true because we have now open sourced the whole core DB. So everything is in the default docker image. So maybe replace `(Enterprise Edition only)` with `if desired`.

design/operator_design.md
```diff
@@ -0,0 +1,129 @@
...
+
+[comment]: <> (Not clear on its implementation yet. Q: How will the user know master address before creating a cluster? If current is the first cluster out of a multi-AZ deployment, then the user won't have any master addresses to specify.)
```
rkarthik007: They should use the service name if inside a k8s pod, or use the load balancer for outside.

design/operator_design.md
```diff
@@ -0,0 +1,129 @@
...
+### MultiAZ
+If the cluster spans multiple availability zones, set `multiAZ.enabled` to `true`.
```
rkarthik007: What does this setting do?

design/ybcluster_crd.yaml
```diff
@@ -0,0 +1,186 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: ybclusters.yugabytedb.yugabyte.com
```
rkarthik007: Not too happy with this resource name. What do other distributed databases use here?
cc @schoudhury

design/operator_design.md
```diff
@@ -0,0 +1,129 @@
...
+```yaml
+apiVersion: yugabytedb.yugabyte.com/v1alpha1
+kind: YBCluster
+metadata:
+  name: yugabytedb
+  namespace: yugabytedb
+spec:
+  image:
+    repository: yugabytedb/yugabyte     # Default.
+    tag: master                         # Default.
```
aegershman: [just be careful](https://helm.sh/docs/chart_best_practices/#images) about using `master` or [`latest` tags](https://kubernetes.io/docs/concepts/configuration/overview/#container-images); using `latest` is just a matter of time before it breaks someone who didn't pin it.

design/operator_design.md
```diff
@@ -0,0 +1,129 @@
...
+    pullPolicy: IfNotPresent            # Default
+  tls:
+    enabled: false
+    # rootCA:
```
aegershman: I'm unfamiliar with defining operators; do these have to be commented out even if `tls.enabled: false`? e.g. couldn't `rootCA.cert/key` be left uncommented, but left empty, as long as `tls.enabled: false`?

design/operator_design.md
```diff
@@ -0,0 +1,129 @@
...
+      storageClass: standard            # Optional. Provide storage class to use. Field will be defaulted to "standard", if omitted.
+    resources:
+      requests:
+        cpu: 2
+        memory: 1Gi
+      limits:
+        cpu: 2
+        memory: 1Gi
+    gflags:
+      - key: default_memory_limit_to_ram_ratio
```
aegershman: is `default_memory_limit_to_ram_ratio` required to be defined as an array element in `gflags`? [it looks like gflags must have minItems of 1](https://github.com/YugaByte/yugabyte-db/pull/2189/files#diff-aba4b15ebb06b64aee98d69a7c38e085R114);

the reason I ask is because if it's essential, then it might blow up [if someone overwrites gflags with _other_ key/value pairs for other gflags, and they forget or don't know to keep `default_memory_limit_to_ram_ratio: 0.85`](https://github.com/helm/helm/blob/master/docs/chart_best_practices/values.md#consider-how-users-will-use-your-values)

design/ybcluster_crd.yaml
```diff
@@ -0,0 +1,186 @@
...
+  group: yugabytedb.yugabyte.com
+  names:
+    kind: YBCluster
+    listKind: YBClusterList
+    singular: ybcluster
+    plural: ybclusters
+  scope: Namespaced
+  version: v1alpha1
+  validation:
+    openAPIV3Schema:
```
aegershman: in case you're interested I think you can put [`description` blocks in these definitions](https://github.com/coreos/prometheus-operator/blob/master/example/prometheus-operator-crd/alertmanager.crd.yaml#L17-L19) so that anything you'd otherwise leave as a `# comment` could be defined and described in the schema itself?

design/operator_design.md
```diff
@@ -0,0 +1,129 @@
+# YugabyteDB Operator Design
```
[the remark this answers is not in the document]
samkulkarni20: Do you mean with or without the design directory under `https://github.com/YugaByte/yugabyte-db/tree/master/cloud/kubernetes` ?

design/operator_design.md
samkulkarni20: Should we default the storage disk count to 1 for master & remove the attribute from CRD?

design/operator_design.md
samkulkarni20: ok. updating.

design/operator_design.md
samkulkarni20: Agreed. Though, the idea is user will override it, as needed. Pinning it to, probably current latest pinned tag (i.e. 1.3.2.0-b19), would mean user will have to override it always, to get the latest features.

design/operator_design.md
samkulkarni20: That's right. User will have to either give values for `cert/key` or omit/comment them. The values can't be null.
It's not because of how operators need to be defined but rather, the Open API V3 schema validations available for K8s CRDs. K8s CRDs validation schema doesn't support some keywords from Open API V3, including `nullable`.

design/operator_design.md
samkulkarni20: The `gflags` attribute itself can be omitted & no entries in it are required. If the user does specify `gflags` attribute then he/she must add min 1 entry, which can be any valid gflag key-value.

design/ybcluster_crd.yaml
samkulkarni20: Sure. I might add some later.

design/ybcluster_crd.yaml
samkulkarni20: How does `ybclusters.yugabyte.com` sound?
Some examples of other operator CR names:
Pod monitoring operator CR name by coreos.com: `podmonitors.monitoring.coreos.com`
Alert manager operator CR name by coreos.com: `alertmanagers.monitoring.coreos.com`
Istio control plane operator CR name by istio.io : `istiocontrolplanes.istio.io`

design/operator_design.md
samkulkarni20: Marks the cluster as multiAZ enabled/disabled.

design/operator_design.md
aegershman: Right, but in general it's preferable to have to override the tag rather than take `latest`:

- if YB releases a `latest` which involves changes which users need to be aware of or new ways of how the operator has to be configured, then they're broken. That's a much worse experience than having users update the `tag` manually. In terms of user experience, stability is going to be significantly more valuable than features.
- Since the default `pullPolicy` is `IfNotPresent`, it means the "true" `latest` tag won't represent what YB publishes as `latest`. Since `latest` is just a tag, if the kubernetes hosts' docker daemon sees "`yugabytedb/yugabyte:latest`" cached, then it won't try to reach out to the upstream source to retrieve the "true" `latest`. It'll see whatever it has locally as `latest` and just use that.

That can get you into a weird spot where `latest` on one k8s host is actually `1.3.0-b17`, and then `latest` will represent `1.3.0.0-b17` until the docker image is purged from the hosts' docker image cache. Then let's say YB publishes `1.3.2.0-b19` and marks it as `latest`. In a given k8s cluster, if a _different_ k8s host comes online which doesn't have `yugabytedb/yugabyte:latest` cached, it'll pull it from the upstream repo and consider it "`latest`", e.g. to _that_ host in the k8s cluster, "`1.3.2.0-b19` is `latest`, but to the _other_ host, `1.3.0.0-b17` is `latest`... Then even within the same k8s cluster, there's uncertainty about which image is being used, because `latest` is relative.

see also: https://vsupalov.com/docker-latest-tag/

I presume this is something that would be configurable about the operator, but just something to consider. You might consider having the default be opting-in to using `master` rather than set it as default out of the box, just something to consider 👍

samkulkarni20: Good point. Let's make `1.3.2.0-b19` as default.

design/operator_design.md
samkulkarni20: @rkarthik007 Please let me know on above follow-up question & also on the few others in the review.

design/ybcluster_crd.yaml
samkulkarni20: I added some descriptions in the CRD file.

design/operator_design.md
samkulkarni20: Done.

design/operator_design.md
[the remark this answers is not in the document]
samkulkarni20: Moved design directory under `https://github.com/YugaByte/yugabyte-db/tree/master/cloud/kubernetes`. Marking as resolved.

design/ybcluster_crd.yaml
samkulkarni20: Updated CR name to `ybclusters.yugabyte.com`. Marking as resolved.

cloud/kubernetes/design/operator_design.md
```diff
@@ -0,0 +1,129 @@
+# YugabyteDB Operator Design
+
+A cluster of YugaByte DB should be created using the `ybclusters.yugabyte.com` custom resource definition. Below is a sample custom resource spec for creating a 3-master, 3 t-servers cluster using the CRD. The sample is followed by an explanation of different configuration options available on the YugaByte DB CRD.
+
+
+```yaml
```
vishal-biyani: Let's also have a super minimal CR example - where rest of values are assumed to defaults

design/operator_design.md
rkarthik007: Let's leave it configurable but default to 1.

design/operator_design.md
samkulkarni20: Removed multi-AZ feature from version 1, as discussed.

design/operator_design.md
samkulkarni20: Removed multi-AZ feature from version 1, as discussed.

cloud/kubernetes/design/operator_design.md
samkulkarni20: Done

design/operator_design.md
samkulkarni20: Done.

Issue #1367, opened 2019-05-13
expose number of dead/unresponsive tservers as yb-master metric


kmuthukk:
The yb-master leader tracks yb-tserver heartbeats and knows which TServers haven't been able to heartbeat/have been unresponsive (determined by the`tserver_unresponsive_timeout_ms` setting).

The yb-master process exposes (via http://\<master-ip\>:7000/prometheus-metrics) a `num_tablet_servers_live` metric:

```
num_tablet_servers_live{ \
exported_instance="yb-1-xdc-user-identity-n1",\
metric_type="cluster",metric_id="yb.cluster"} \
<n> \
<timestamp> \
```

But a `num_tablet_servers_dead` (the inverse metric) might be more useful/easier for setting up alerts.  This number being greater than 0 could be set to result in an alert rather than basing the alert on the live number of nodes. The latter needs the alerting to be based on number of nodes in the system, and that itself could change over time.

Note: On yb-master followers this metric should simply return 0.

bmatican:
One note on this is that we also need a way to GC information about old TS that maybe have just been removed. Right now, our TSDescriptor info sticks around for the duration of a master being a leader, so if a TS is decomissioned and replaced, we might still have a reference to the old TS in our memory for a while. cc @kmuthukk

deeps1991:
@kmuthukk and @bmatican I was looking for a good first issue to get familiar with the contributing process, and this caught my eye. If this is not being actively worked on yet, please could you let me know if I could start on this one?

bmatican:
@deeps1991 Happy to see you excited about YugaByte! You are correct, this is not actively being worked on. Would definitely love having you contribute, if you're interested!

Let me know if you need any more pointers about how to approach this task.

Also, in case you have not already signed our CLA, please use https://cla-assistant.io/YugaByte/yugabyte-db?pullRequest=1638 cc @mbautin

kmuthukk:
+1 to what @bmatican said. Thanks @deeps1991 !

deeps1991:
@bmatican From @kmuthukk's snippet above, I found the code that creates and updates this metric. So I can try to figure out the rest from there, will get back if I have questions. But what would be the best way for me to test this out? I was thinking to setup a local cluster, simulate a node failure (or maybe there is a unit test that does this as well?) and monitor the metrics page.

kmuthukk:
@deeps1991 - that might sufficient for this metrics change. Your test plan looks good to me.

[...]

PR #2258, opened 2019-09-10
Issue #1367 expose number of dead/unresponsive tservers as yb-master metric

deeps1991:
The yb-master leader tracks yb-tserver heartbeats and knows which TServers haven't been able to heartbeat/have been unresponsive. The yb-master process exposes (via http://<master-ip>:7000/prometheus-metrics) a num_tablet_servers_live metric, but a num_tablet_servers_dead
might be more useful/easier for setting up alerts. This number being greater than 0 could be set to result in an alert.
    
To do this, expose another metric num_tablet_servers_dead from the catalog manager similar to the num_tablet_servers_live. This metric tracks the number of total tablet servers - live tablet servers.
    
Test plan:
1) Set up local yugabyte cluster with 3 nodes with prometheus monitoring
2) Verify that num_tablet_servers_live = 3 and num_tablet_servers_dead=0
on the :7000 prometheus metrics page
3) Use yb-ctl commands to stop one of the nodes
4) Verify that num_tablet_servers_live=2 and num_tablet_servers_dead=1


deeps1991:
Looks like I accidentally got the merge commit into this pull request. Will send another clean pull request.

PR #2260, opened 2019-09-10
Issue #1367 expose number of dead/unresponsive tservers as yb-master metric

deeps1991:
The yb-master leader tracks yb-tserver heartbeats and knows which
TServers haven't been able to heartbeat/have been unresponsive

The yb-master process exposes (via http://<master-ip>:7000/prometheus-metrics)
a num_tablet_servers_live metric, but a num_tablet_servers_dead
might be more useful/easier for setting up alerts. This number being greater
than 0 could be set to result in an alert.

To do this, expose another metric num_tablet_servers_dead from the catalog
manager similar to the num_tablet_servers_live. This metric tracks the
number of total tablet servers - live tablet servers.

Test plan:
1) Set up local yugabyte cluster with 3 nodes with prometheus monitoring
2) Verify that num_tablet_servers_live = 3 and num_tablet_servers_dead=0
   on the :7000 prometheus metrics page
3) Use yb-ctl commands to stop one of the nodes
4) Verify that num_tablet_servers_live=2 and num_tablet_servers_dead=1

### DIFF
--- a/src/yb/master/catalog_manager.cc
+++ b/src/yb/master/catalog_manager.cc
@@ -5930,7 +5930,7 @@ void CatalogManager::ReportMetrics() {
   master_->ts_manager()->GetAllLiveDescriptors(&ts_descs);
   const int32 num_live_servers = ts_descs.size();
   metric_num_tablet_servers_live_->set_value(num_live_servers);
-  
+
   master_->ts_manager()->GetAllDescriptors(&ts_descs);
   metric_num_tablet_servers_dead_->set_value(ts_descs.size() - num_live_servers);
 }
--- a/src/yb/master/catalog_manager.cc
+++ b/src/yb/master/catalog_manager.cc
@@ -205,6 +205,12 @@ METRIC_DEFINE_gauge_uint32(cluster, num_tablet_servers_live,
                            "in the time interval defined by the gflag "
                            "FLAGS_tserver_unresponsive_timeout_ms.");
 
+METRIC_DEFINE_gauge_uint32(cluster, num_tablet_servers_dead,
+                           "Number of dead tservers in the cluster", yb::MetricUnit::kUnits,
+                           "The number of tablet servers that have not responded or done a "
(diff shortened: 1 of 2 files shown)


iSignal:
This does seem like a good metric to alert on, very nice. To nitpick a bit though, it is possible this tablet server does not have any tablets allocated to it (for some reason, policy or blacklist or otherwise). The more accurate metric to alert on would be the number of tablets that are under replicated. Is that something that is easy enough to export out?

kmuthukk:
1) @iSignal -- yes, underreplicated tablets metric would be very useful also. But best done as a separate task.

2) @amitanandaiyer -- the changes look reasonable to me. Would you mind helping merge the PR into master?

src/yb/master/catalog_manager.cc
```diff
@@ -5919,7 +5928,11 @@ void CatalogManager::ReportMetrics() {
   // Report metrics on how many tservers are alive.
   TSDescriptorVector ts_descs;
   master_->ts_manager()->GetAllLiveDescriptors(&ts_descs);
-  metric_num_tablet_servers_live_->set_value(ts_descs.size());
+  const int32 num_live_servers = ts_descs.size();
+  metric_num_tablet_servers_live_->set_value(num_live_servers);
+
```
amitanandaiyer: nit: can you pls get rid of the white space on this line?

> Lint: Trailing Whitespace
> This line contains trailing whitespace. Consider setting up your editor to automatically remove trailing whitespace, you will save time.
> 
> Lint: whitespace/end_of_line
> Line ends in whitespace. Consider deleting these extra spaces.
>

deeps1991:
@amitanandaiyer Fixed the lint issue, thanks for pointing it out 👍

src/yb/master/catalog_manager.cc
deeps1991: Done.

PR #2339, opened 2019-09-17
[docs] Hash/range columns

ndeodhar:
Added documentation for:
1. Hash/range columns in CREATE TABLE and CREATE INDEX.
2. Nulls first/last syntax.
3. Temporary tables in CREATE TABLE.


docs/content/latest/api/ysql/commands/ddl_create_index.md
```diff
@@ -73,8 +73,60 @@ Specify the name of a column of the table.
...
+    "products_pkey" PRIMARY KEY, lsm (id HASH)
+    "products_code_idx" UNIQUE, lsm (code HASH)
+```
+
+### ASC ordered index
+
+Create an index with ascending ordered key.
+
+```sql
+yugabyte=# CREATE INDEX products_name ON products(name);
```
m-iancu: should this be `name ASC`, and then a `\d` command to check it?

docs/content/latest/api/ysql/syntax_resources/ysql_grammar.ebnf
```diff
@@ -343,9 +343,17 @@ column_constraint ::= [ 'CONSTRAINT' constraint_name ]
...
                      ( 'CHECK' '(' expression ')'
                        | 'UNIQUE' '(' column_names ')' index_parameters
-                       | 'PRIMARY' 'KEY' '(' column_names ')'
+                       | 'PRIMARY' 'KEY' '(' key_columns ')'
                        | 'FOREIGN' 'KEY' '(' column_names ')' references_clause ) ;
 
+key_columns ::= hash_columns [ ',' range_columns ]
+              | range_columns ;
+
+hash_columns ::= column_name [ 'HASH' ]
```
m-iancu: can we have `primary key (a HASH, b HASH, c HASH, d ASC ...)`? 
In that case maybe it should be:
```
hash_columns ::= ( column_name [ 'HASH' ] ) { ','  (column_name [ 'HASH' ] ) }
                 | '(' column_name { ',' column_name } ')' [ 'HASH' ] ;
```

Also I assume that something like `primary key ((a, b) HASH, c HASH)` or  `primary key (a HASH, (b,  c) HASH)` is not allowed.

docs/content/latest/api/ysql/commands/ddl_create_table.md
```diff
@@ -95,7 +99,29 @@ postgres=# CREATE TABLE sample(k1 int,
...
+Indexes:
+    "sample_pkey" PRIMARY KEY, lsm (k1 HASH, k2)
+```
+
+### Table with range primary key.
+```sql
+postgres=# CREATE TABLE range(k int,
+                              v1 int,
+                              v2 text,
+                              PRIMARY KEY (k ASC));
```
m-iancu: nit: Maybe have `k1 ASC, k2 DESC` for a more illustrative example.

docs/content/latest/api/ysql/syntax_resources/ysql_grammar.ebnf
ndeodhar: All of those are not allowed.
We only allow `(a HASH)` or `((a, b,...) HASH)`.

PR #2360, opened 2019-09-18
RBAC control commands

### LINKED ISSUE #2037: [YSQL][docs] RBAC
Grant, revoke, alter table, create role, drop role, create user etc.

ndeodhar:
This fixes #2037.

Commands added:
1. Set Role
2. Set session authorization
3. Alter default privileges
4. Drop owned
5. Reassign owned


docs/content/latest/api/ysql/commands/dcl_set_session_authorization.md
```diff
@@ -0,0 +1,76 @@
+---
+title: SET SESSION AUTHORIZATION
+description: SET SESSION AUTHORIZATION
+summary: Roles (users and groups)
```
m-iancu: Is it intentional that the summary is `Roles (users and groups)` for all pages?

PR #2583, opened 2019-10-11
[Docs] recover node,tserver,master

ddorian:


ddorian:
moved https://github.com/yugabyte/yugabyte-db/pull/2599

PR #2649, opened 2019-10-18
Update documentation for PK columns

ndeodhar:


docs/content/latest/api/ysql/commands/ddl_create_table.md
```diff
@@ -50,12 +50,15 @@ Create a table with *table_name*. An error is raised if `qualified_name` already
 ### Primary Key
 
 Primary key can be defined in either `column_constraint` or `table_constraint`, but not in both.
-There are 2 types of primary keys:
+There are 2 types of primary key columns:
 
-- `Hash primary keys`: This is the default and is used to hash-partition a table.
+- `Hash primary key columns`: The first column in Primary Key will default to HASH and is used to hash-partition a table.
```
stevebang:
```suggestion
- `Hash primary key columns`: The first column in Primary Key will default to HASH and is used to hash partition a table.
```

docs/content/latest/api/ysql/commands/ddl_create_table.md
```diff
@@ -50,12 +50,15 @@ Create a table with *table_name*. An error is raised if `qualified_name` already
...
+For example if primary key is `(a, b, c)`, column `a` will be HASH ordered.
 This is the recommended option to better scale read/write operations to the table.
```
stevebang:
```suggestion
This is the recommended option to better scale read and write operations to the table.
```

docs/content/latest/api/ysql/commands/ddl_create_table.md
kmuthukk: Suggested rewording:

`Hash primary key columns`: The primary key may have zero or more leading hash-partitioned columns. By default, only the first column is treated as the hash-partition column. But this behavior can be modified by explicit use of the HASH annotation. 

`Range primary key columns`: There can be one more range primary key columns and it controls the top-level ordering of rows within a table (if there are no hash partition columns) or the ordering of rows among rows that share a common set of hash partitioned column values. By default, the range primary key columns are stored in ascending order. But this behavior can be controlled by explicit use of the ASC/DESC annotation.


For example, if the primary key specification is `PRIMARY KEY (a HASH, b HASH, c DESC)` then columns `a` & `b` are used together to hash partition the table, and rows that share the same values for `a` and `b` are stored in descending order of their value for `c`.

PR #2652, opened 2019-10-19
yb-admin page

stevebang:
Adds yb-admin page to CLIs section.


stevebang:
We can publish as-is or add some meaningful groupings for the commands. Right now, the list is over 30 commands — a lot to read through without some categories.

schoudhury:
Merging for now, let us iterate on the groupings as a next step.

PR #2963, opened 2019-11-19
[ WIP ] Different packages for yugabyte added.

vp15591:
Hi, have added the following packages.

1. Deb package 
2. RPM package 
3. LinuxBrew Formula 
4. Brew Formula.


vp15591:
1. Added script which is added for creating the package and also updated README.md file for complete instruction on how to create deb and rpm package using `fpm` command-line utility.

2. Updated the brew formula for creating a symlink for bin/ysqlsh and bin/ycqlsh.

packages/script/yugabyte
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+# yugabyted daemon
+# chkconfig: 345 20 80
+# description: yugabyted daemon
+# processname: yugabyted
+
+DAEMON_PATH="/opt/yugabyte-2.0.5.2"
```
iSignal: ideally, we wouldn't need this separate daemon script because yugabyted supports all the functionality required. it has a start/stop/status command and puts its pid file in the root directory (yugaybted.pid)

packages/script/install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+chmod 775 -R /opt/yugabyte-2.0.5.2
+cd /opt/yugabyte-2.0.5.2/
+
+mkdir -p /var/log/yugabyte-logs/
+mv /opt/yugabyte-2.0.5.2/etc/yugabyte /etc
+
+rm -rf /opt/yugabyte-2.0.5.2/etc
+ln -s /opt/yugabyte-2.0.5.2/yugabyted /usr/bin/yugabyted
```
iSignal: as part of installation, could we run `/opt/yugabyte-<version>/bin/post_install.sh && touch /opt/yugabyte-<version>/post_install.sh.completed`

packages/Brew/yugabyte.rb
```diff
@@ -0,0 +1,55 @@
...
+    rm_f buildpath/"lib/cassandra-driver-internal-only-3.13.0.post0-743d942c.zip"
+    rm_f buildpath/"lib/futures-2.1.6-py2.py3-none-any.zip"
+    rm_f buildpath/"lib/six-1.7.3-py2.py3-none-any.zip"
+    prefix.install Dir["*"]
+    bin.install_symlink prefix/"yugabyte-db"
+    bin.install_symlink prefix/"bin/ysqlsh"
+    bin.install_symlink prefix/"bin/ycqlsh"
+  end
+
+  plist_options :manual => "yugabyte-db start"
```
iSignal: yugabyte-db was renamed to yugabyted in the latest package

packages/Brew/yugabyte.rb
```diff
@@ -0,0 +1,55 @@
...
+    <dict>
+      <key>Label</key>
+      <string>#{plist_name}</string>
+      <key>ProgramArguments</key>
+      <array>
+        <string>#{prefix}/yugabyte-db</string>
+        <string>start</string>
+      </array>
+      <key>RunAtLoad</key>
+      <true/>
```
iSignal: will this formula this work with `brew services`? so brew services list will show yugabyted?

packages/script/install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+chmod 775 -R /opt/yugabyte-2.0.5.2
+cd /opt/yugabyte-2.0.5.2/
+
+mkdir -p /var/log/yugabyte-logs/
```
iSignal: typically databases like mysql and postgres also create a non-root user to run the database as. Does fpm take care of that for us?

packages/script/yugabyte
vp15591: yugabyted support only start option there is no stop option available in yugabyte version 2.0.5.2. That's why I create a demon script. and when we start it directly, yugabyted create config in root directly, but I think i should create a config file in `/etc/yugabyte` directory, also redirect its logs to `/var/log/yugabyte-logs`
```
vaibhav@instance-3:~$ yugabyted --help
Usage: yugabyted [-h] {start,status,version,demo,connect} ...

To start YugabyteDB, run 'yugabyted start'.

Commands:
  {start,status,version,demo,connect}
    start               Start YugabyteDB.
    status              Print status of YugabyteDB.
    version             Version of YugabyteDB.
    demo                Load and interact with preset demo data.
    connect             Connect to an API through the CLI.

Flags:
  -h, --help            show this help message and exit

Run 'yugabyted [command] -h' for help with specific commands.

```

packages/Brew/yugabyte.rb
vp15591: yes in MacOS it supports brew service.

packages/script/install.sh
vp15591: fpm don't have such option so I have added a command to create a user called `yugabyte` and changed the ownership and permission of installed directory to `yuagbyte` user.

packages/script/install.sh
vp15591: added this in `install.sh` script.

packages/Brew/yugabyte.rb
vp15591: updated in brew formula

packages/Brew/yugabyte.rb
```diff
@@ -0,0 +1,82 @@
+class Yugabyte < Formula
```
mbautin: I think this should be contributed directly to https://github.com/Homebrew/homebrew-core

packages/RRADME.md
```diff
@@ -0,0 +1,61 @@
+### This directory contains all types of yugabyte packages.
```
mbautin: This file's name has a misspelling.

packages/linuxBrew/yugabyte.rb
```diff
@@ -0,0 +1,21 @@
+class Yugabyte < Formula
```
mbautin: I think this should be contributed directly to https://github.com/Homebrew/linuxbrew-core

packages/script/install.sh
mbautin: How about `/var/log/yugabyte` instead of `/var/log/yugabyte-logs`? Since it is in `/var/log`, the `-logs` part is redundant.

mbautin:
Please remove the actual generated .deb / .rpm packages from the pull request -- we don't want to check them into git.

packages/Brew/yugabyte.rb
vp15591: I have created homebrew tap for yugabyte, once tap PR is approved, We will plan to move it to Homebrew core repo.

mbautin: @vp15591 thanks! Would you mind linking to the tap PR? Also I don't think we should put this Homebrew code into the main yugabyte-db repo, even temporarily.

packages/script/install.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -e -x
+getent passwd yugabyte > /dev/null 
+
+if [ $? -eq 0 ]; then
```
iSignal: we never hit these lines because we have set -e on the script. when the getent fails, the script bails out immediately.

PR #3170, opened 2019-12-18
[Docs] Fix mobile responsive styling not sizing correctly and change fonts

yb-andrew:
- Fix h3 font-size being larger than h2 in mobile view.
- Fix definition list tags not having inline indents.
- Change fonts from Montserrat to Rubik and Open Sans to Inter.


schoudhury:
LGTM!

PR #3216, opened 2019-12-21
Making small fix to maintain invariance

ConcurrencyPractitioner:
There might be a few cases where lower bound might be higher than the upper bound. This condition should hold at any time, and thus, we can remove the need to check for the max of these bounds in later parts of the code which use these bounds.

To emphasize, we want to maintain the following invariance:
```yb_transaction_priority_upper_bound >= yb_transaction_priority_lower_bound```

### CODE REVIEW

**Comment by ConcurrencyPractitioner** [comment]
@spolitov This was part of your PR that was merged.

File: src/yb/yql/pggate/pg_txn_manager.cc [hunk 1]
```diff
@@ -49,10 +49,11 @@ extern "C" {
 
 void YBCAssignTransactionPriorityLowerBound(double newval, void* extra) {
   txn_priority_lower_bound = ConvertBound(newval);
+  txn_priority_upper_bound = std::max(txn_priority_lower_bound, txn_priority_upper_bound)
```
spolitov: Such behavior is undesired.
For instance, consider the following statement sequence:
```
SET yb_transaction_priority_upper_bound = 0.5
SET yb_transaction_priority_lower_bound = 0.75
SET yb_transaction_priority_lower_bound = 0.25
```
The user would expect to have (0.25, 0.5) range for priorities, but it will get (0.25, 0.75) range.

ConcurrencyPractitioner: Well, for the above example, it is ambiguous on exactly which bounds the user desires. Why would the user set the lower bound to 0.75, only to set it to 0.25 afterwards if he or she doesn't do something with the (0.75, 0.5) bound first? After the lower bound was first set, lower bound and upper bound was already inverted, and obviously, the user doesn't want this to be the case. Therefore, it is logical for us to set the bounds in that case to (0.75, 0.75). And after the final set, it is only logical that we have (0.25, 0.75) as our boundaries.

To make the bounds of what the user wants clear, we should just add a method which accepts two arguments i.e. a setter that sets both bounds with no room for ambiguity.

ConcurrencyPractitioner: To emphasize, we want to maintain the following invariance:
```yb_transaction_priority_upper_bound > yb_transaction_priority_lower_bound```

File: src/yb/yql/pggate/pg_txn_manager.cc [hunk 2]
```diff
@@ -49,13 +49,18 @@ extern "C" {
...
   txn_priority_lower_bound = ConvertBound(newval);
-  txn_priority_upper_bound = std::max(txn_priority_lower_bound, txn_priority_upper_bound)
+  txn_priority_upper_bound = std::max(txn_priority_lower_bound, txn_priority_upper_bound);
 }
 
 void YBCAssignTransactionPriorityUpperBound(double newval, void* extra) {
   txn_priority_upper_bound = std::max(txn_priority_lower_bound, ConvertBound(newval));
 }
 
+void YBCAssignTransactionPriorityLowerAndUpperBound(double newlowerbound, double newupperbound, void* extra) {
```
spolitov: What would not work with SET.

ConcurrencyPractitioner: @spolitov Would you mind explaining exactly what you mean by this? If this setter is not needed, then I can remove it. Just want some clarifications since this comment's intention is not clear.

spolitov: Those methods are not part of our API, our API is SQL.
And user could do:
```SET yb_transaction_priority_upper_bound = 0.5
SET yb_transaction_priority_lower_bound = 0.25```

That will result in the invocation of methods above.

`YBCAssignTransactionPriorityLowerAndUpperBound` is never called.

ConcurrencyPractitioner: Fair enough, will remove it then.

src/yb/yql/pggate/pg_txn_manager.cc
spolitov: We should not change settings unless user explicitly requests it.

PR #3351, opened 2020-01-15
[docs] Change WAL-related FAQ and add related gflags

stevebang:
- Change Karthik's code examples to use `yb-tserver` configuration options.
- Add the following three logging-related options (gflags) mentioned in the FAQ to the `yb-tserver` configuration page:

- `--log_min_segments_to_retain`
- `--log_min_seconds_to_retain`
- `--log_segment_size_mb`

### DIFF
--- a/docs/content/latest/faq/general.md
+++ b/docs/content/latest/faq/general.md
@@ -1,12 +1,12 @@
 ---
-title: Product FAQ
-linkTitle: Product FAQ
-description: Product FAQ
+title: General FAQ
+linkTitle: General FAQ
+description: General FAQ
 aliases:
   - /faq/product/
 menu:
   latest:
-    identifier: product
+    identifier: general-faq
     parent: faq
     weight: 2710
 isTocNested: false
--- a/docs/content/latest/faq/operations-faq.md
+++ b/docs/content/latest/faq/operations-faq.md
@@ -23,7 +23,7 @@ For YCQL, YugabyteDB provides automatic load balancing.
 
 ## Can write ahead log (WAL) files be cleaned up or reduced in size? I'm running out of disk space.
 
-For most YugabyteDB deployments, you should not need to adjust the log file configuration options for the write ahead log (WAL). While your data size is small and growing, the WAL files may seem to be much larger, but over time, the WAL files should reach their steady state while the data size continues to grow and become larger than the WAL files.
+For most YugabyteDB deployments, you should not need to adjust the configuration options for the write ahead log (WAL). While your data size is small and growing, the WAL files may seem to be much larger, but over time, the WAL files should reach their steady state while the data size continues to grow and become larger than the WAL files.
 
 WAL files are per tablet and the retention policy is managed by the following two `yb-tserver` configuration options:
 
--- a/docs/content/latest/faq/operations-faq.md
+++ b/docs/content/latest/faq/operations-faq.md
@@ -16,7 +16,7 @@ showAsideToc: true
 For YSQL, an external load balancer is recommended, but note the following:
 
 - If you use the [YugabyteDB JDBC driver (beta)](../../reference/drivers/yugabytedb-jdbc-driver) with an application regularly opens and closes connections to clients, then the YugabyteDB JDBC driver effectively provides basic load balancing by randomly using connections to your nodes.
-- If you use the[Spring Data Yugabyte driver (beta)](../../reference/drivers/spring-data-yugabytedb) with your Spring application, then the underlying YugabyteDB JDBC driver provides basic load balancing.
+- If you use the [Spring Data Yugabyte driver (beta)](../../reference/drivers/spring-data-yugabytedb) with your Spring application, then the underlying YugabyteDB JDBC driver provides basic load balancing.
 - If you have an application that is not in the same Kubernetes cluster, then you should use an external load balancing system.
 
 For YCQL, YugabyteDB provides automatic load balancing.
--- a/docs/content/latest/faq/operations-faq.md
+++ b/docs/content/latest/faq/operations-faq.md
@@ -23,13 +23,13 @@ For YCQL, YugabyteDB provides automatic load balancing.
 
 ## Can write ahead log (WAL) files be cleaned up or reduced in size? I'm running out of disk space.
 
-For most YugabyteDB deployments, you shouldn't need to adjust these values. While your data size is small and growing, the WAL files may seem to be much larger, but over time, the WAL files should reach their steady state while the data size continues to grow and become larger than the WAL files.
+For most YugabyteDB deployments, you should not need to adjust the log file configuration options for the write ahead log (WAL). While your data size is small and growing, the WAL files may seem to be much larger, but over time, the WAL files should reach their steady state while the data size continues to grow and become larger than the WAL files.
 
 WAL files are per tablet and the retention policy is managed by the following two `yb-tserver` configuration options:
 
-- [`log_min_segments_to_retain`](../../reference/configuration/tserver/#log-min-segments-to-retain)
-- [`log_min_seconds_to_retain`](../../reference/configuration/tserver/#log-min-seconds-to-retain)
+- [`log_min_segments_to_retain`](../../reference/configuration/yb-tserver/#log-min-segments-to-retain)
+- [`log_min_seconds_to_retain`](../../reference/configuration/yb-tserver/#log-min-seconds-to-retain)
 
-Also, the following `yb-tserver` configuration option is a factor in the size of each WAL before it is rolled into a new one:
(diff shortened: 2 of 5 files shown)


stevebang:
@rkarthik007 Do these three logging options only apply to WAL files or are they general logging options?

docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -155,6 +155,24 @@ The directory to write `yb-tserver` log files.
 
 Default: Same as [`--fs_data_dirs`](#fs-data-dirs)
 
+##### --log_min_seconds_to_retain
+
+The minimum number of seconds for which to keep log segments to keep at all times, regardless of what is required for durability. Logs may be retained for a longer amount of time if they are necessary for correct restart. This should be set long enough such that a tablet server which has temporarily failed can be restarted within the given time period. If a server is down for longer than this amount of time, it is possible that its tablets will be re-replicated on other machines.
```
bmatican: The latter part, of side effect if the node is down for longer, is not actually correct. There's 2 flags, which are defaulted to the same value of 15m, which maybe cause some confusion.

`follower_unavailable_considered_failed_sec` -- This flag dictates that if a TS is down for longer than this amount of seconds, then we will remove it from raft groups and re-replicate data somewhere else
`log_min_seconds_to_retain` -- This flag dictates that our GC policy will not remove wal files which contain records up to this time window. If a leader has GCed a WAL file that contains records needed for an active follower, then what will actually happen is that that follower will be forced to go into what we call `Remote Bootstrap`, where it will get all the files from the leader, in order to catch up.

However, in practice, if these flags have the same value, then once that threshold is reached, the unavailability flag kicks in and removes the peer, so the GC aspect of the second flag is irrelevant.

docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -155,6 +155,24 @@ The directory to write `yb-tserver` log files.
...
+
+##### --log_min_segments_to_retain
+
+The minimum number of past log segments to keep at all times, regardless of what is required for durability. The value must be at least 1.
+
+Default: `2`
+
+##### --log_segment_size_mb
+
+The default segment size, in megabytes (MB) for log rollovers.
```
bmatican: Not sure if it is clear for a user at this point what a log rollover is? Could maybe clarify that this is (loosely) the max size of each WAL file?

stevebang: Good point -- I'll add that.

docs/content/latest/reference/configuration/yb-tserver.md
stevebang: The content I used was directly from the online help — want me to add your revisions to a PR to correct that too?

docs/content/latest/faq/operations-faq.md
```diff
@@ -16,20 +16,20 @@ showAsideToc: true
...
-WAL files are per tablet and the retention policy is managed by the following two gflags:
+For most YugabyteDB deployments, you should not need to adjust the configuration options for the write ahead log (WAL). While your data size is small and growing, the WAL files may seem to be much larger, but over time, the WAL files should reach their steady state while the data size continues to grow and become larger than the WAL files.
 
-- `src/yb/consensus/log.cc:DEFINE_int32(log_min_segments_to_retain, 2,`
-- `src/yb/consensus/log.cc:DEFINE_int32(log_min_seconds_to_retain, 900,`
+WAL files are per tablet and the retention policy is managed by the following two `yb-tserver` configuration options:
 
-Also, the following gflag is a factor in the size of each WAL before it is rolled into a new one:
+- [`--log_min_segments_to_retain`](../../reference/configuration/yb-tserver/#log-min-segments-to-retain) – default is `2`.
+- [`--log_min_seconds_to_retain`](../../reference/configuration/yb-tserver/#log-min-seconds-to-retain) – default is `900`.
```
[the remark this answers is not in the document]
stevebang: At InfluxData, customers were told that the docs are the final authority — they learn that they can get everything they need to know about options (including important warnings/notes/tips, links to related options/topics) in one place. The defaults should definitely be included in the reference — any changes in the values should be tagged for docs. I will remove the defaults from the FAQ.

docs/content/latest/faq/operations-faq.md
```diff
@@ -16,20 +16,20 @@ showAsideToc: true
...
 
-- `src/yb/consensus/log_util.cc:DEFINE_int32(log_segment_size_mb, 64,`
+Also, the following `yb-tserver` configuration option is a factor in the size of each WAL file before it is rolled into a new one:
```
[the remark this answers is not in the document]
stevebang: Added to `yb-master` configuration options too.

docs/content/latest/faq/general.md
```diff
@@ -1,12 +1,12 @@
 ---
-title: Product FAQ
-linkTitle: Product FAQ
-description: Product FAQ
+title: General FAQ
+linkTitle: General FAQ
+description: General FAQ
```
bmatican: FYI: I'm not familiar with any infra issues, if we change these titles / id below, if these are cached anywhere

docs/content/latest/faq/operations-faq.md
```diff
@@ -21,15 +21,15 @@ For YSQL, an external load balancer is recommended, but note the following:
...
 
-## Can write ahead log (WAL) files for YCQL be cleaned up or reduced in size? I'm running out of disk space.
+## Can write ahead log (WAL) files be cleaned up or reduced in size?
 
-WAL files are per tablet and the retention policy is managed by the following two gflags:
+For most YugabyteDB deployments, you should not need to adjust the configuration options for the write ahead log (WAL). While your data size is small and growing, the WAL files may seem to be much larger, but over time, the WAL files should reach their steady state while the data size continues to grow and become larger than the WAL files.
 
-- `src/yb/consensus/log.cc:DEFINE_int32(log_min_segments_to_retain, 2,`
-- `src/yb/consensus/log.cc:DEFINE_int32(log_min_seconds_to_retain, 900,`
+WAL files are per tablet and the retention policy is managed by the following two `yb-tserver` configuration options:
```
bmatican: nit for both this and below, I believe we discussed offline, these also apply to the master, as it also has a raft group, but it's not in the critical path, so maybe that data size is never really a concern

docs/content/latest/faq/operations-faq.md
```diff
@@ -21,15 +21,15 @@ For YSQL, an external load balancer is recommended, but note the following:
...
 
-Also, the following gflag is a factor in the size of each WAL before it is rolled into a new one:
+- [`--log_min_segments_to_retain`](../../reference/configuration/yb-tserver/#log-min-segments-to-retain)
+- [`--log_min_seconds_to_retain`](../../reference/configuration/yb-tserver/#log-min-seconds-to-retain)
 
-- `src/yb/consensus/log_util.cc:DEFINE_int32(log_segment_size_mb, 64,`
+Also, the following `yb-tserver` configuration option is a factor in the size of each WAL file before it is rolled into a new one:
```
bmatican: see above nit on tserver & master

docs/content/latest/reference/configuration/yb-master.md
```diff
@@ -223,6 +219,78 @@ Default: `2`
...
+
+{{< note title="Note" >}}
+
+Ensure that values used for Raft and the write ahead log (WAL) in `yb-master` configurations match the values in `yb-tserver` configurations.
+
+{{< /note >}}
+
+##### --follower_unavailable_considered_failed_sec
+
+The duration, in seconds, after which a follower is considered to be failed because the leader has not received a heartbeat. The follower is then evicted from the configuration and the data is rereplicated elsewhere.
```
bmatican: Actually, the mechanic here is the other way around, from leader to follower. I would rephrase something like:

```The duration, in seconds, that a leader will continue trying to contact an unresponsive follower. If the leader has no successful communication with a follower, for this prolonged period of time, it will consider this follower failed and evict it from the configuration. Afterwards, a new follower will be automatically added to the configuration and continue to replicate the data.```

File: docs/content/latest/reference/configuration/yb-tserver.md [hunk 9]
```diff
@@ -199,6 +200,78 @@ Default: `false`
 
 ---
 
+### Raft options
+
+{{< note title="Note" >}}
+
+Ensure that values used for Raft and the write ahead log (WAL) in `yb-tserver` configurations match the values in `yb-master` configurations.
```
bmatican: Oh damn, I did not realize we have two separate copies of the settings, for master and tserver. In practice, the good portion of the flags should apply to both master and tserver, as they cover knobs for the storage engine, which both require.

Issue #3412, opened 2020-01-22
[docs] Add yb-ts-cli to CLIs section


stevebang:
Document the YugabyteDB tablet server tool, `yb-ts-cli`.

See:
- https://github.com/yugabyte/yugabyte-db/issues/2877#issuecomment-553682794
- command line help

stevebang:
Fixed by #3414

PR #3414, opened 2020-01-22
[docs] Add yb-ts-cli to CLIs reference

stevebang:
This PR adds a new page for the `yb-ts-cli` to the CLIs reference.

### DIFF
--- a/docs/content/latest/admin/yb-ctl.md
+++ b/docs/content/latest/admin/yb-ctl.md
@@ -33,7 +33,7 @@ $ ./bin/yb-ctl --help
 
 ## Commands
 
-### create
+##### create
 
 Creates a local YugabyteDB cluster. With no optional arguments, creates a 1-node cluster.
 
--- a/docs/content/latest/admin/yb-ts-cli.md
+++ b/docs/content/latest/admin/yb-ts-cli.md
@@ -100,20 +100,7 @@ yb-ts-cli [ --server_address=<host>:<port> ] dump_tablet <tablet_id>
 
 ##### list_tablets
 
-Lists the tablets on the specified tablet server, displaying the following information:
-
-- Tablet id
-- State
-- Table name
-- Partition
-- Schema
-- properties
-  - lumn_name
-  - Tablet id
-  - State
-  - Table name
-  - Partition
-  - Schema
+Lists the tablets on the specified tablet server, displaying the following properties: column name, tablet ID, state, table name, partition, and schema.
 
 **Syntax**
 
--- a/docs/content/latest/admin/yb-ts-cli.md
+++ b/docs/content/latest/admin/yb-ts-cli.md
@@ -4,14 +4,14 @@ linkTitle: yb-ts-cli
 description: yb-ts-cli
 menu:
   latest:
-    identifier: yb-ts-tool
+    identifier: yb-ts-cli
     parent: admin
     weight: 2466
 isTocNested: 3
 showAsideToc: true
 ---
 
-`yb-ts-cli` is a command line tool that can be used to perform an operation on tablet server (`yb-tserver`). Some of the commands perform operations similar to [`yb-admin` commands](../../../reference/configuration/yb-tserver).
+`yb-ts-cli` is a command line tool that can be used to perform an operation on a particular tablet server (`yb-tserver`). Some of the commands perform operations similar to [`yb-admin` commands](../yb-admin). The `yb-admin` commands focus on cluster administration, the `yb-ts-cli` commands apply to specific YB-TServer nodes.
 
 `yb-ts-cli` is a binary file installed with YugabyteDB and is located in the `bin` directory of the YugabyteDB home directory.
 
--- a/docs/content/latest/admin/yb-ts-cli.md
+++ b/docs/content/latest/admin/yb-ts-cli.md
@@ -1,0 +1,121 @@
+---
+title: yb-ts-cli
+linkTitle: yb-ts-cli
+description: yb-ts-cli
+menu:
+  latest:
+    identifier: yb-ts-tool
+    parent: admin
+    weight: 2466
+isTocNested: 3
+showAsideToc: true
+---
+
+## Syntax
+
+```sh
+./yb-ts-cli [--server_address=<addr>] <command> <options>
+```
+- *command*: asdfad
+- *options*: asfads
+
+## Command line help
+
+
+
+## Commands
+
+### list_tablets
+
+### are_tablets_running
+
+If all tablets are running, returns "All tablets are running".
+
+### set_flag
+
+Set the specified option (flag).
+
+### dump_tablet.
+
+Dump the tablet with the specified tablet ID (`tablet_id`).
+
+### delete_tablet
+
+Delete the tablet with the specified tablet ID (`tablet_id`).
+
+### current_hybrid_time
+
+
+
+### status
+
+Returns the status of the tablet servers.
+
+
+### count_intents
+
+## Options
+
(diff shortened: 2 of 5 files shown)


stevebang:
Fixes #3412

docs/content/latest/admin/yb-ts-cli.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Syntax**
+
+```sh
+yb-ts-cli [ --server_address=<addr> ] are_tablets_running
+```
+
+##### count_intents
+
+Print the count of uncompleted intents (or [provisional records](../../../architecture/transactions/ditributed-txns/#provisional-records)). Useful for debugging and troubleshooting.
```
bmatican: I think uncommitted is more appropriate than uncompleted

also maybe end with `Useful for debugging transactional workloads.`

docs/content/latest/admin/yb-ts-cli.md
```diff
@@ -0,0 +1,208 @@
...
+Deletes the tablet with the specified tablet ID (`tablet_id`) and reason.
+
+**Syntax**
+
+```sh
+$ ./bin/yb-ts-cli  [ --server_address=<addr> ] delete_tablet <tablet_id> "<reason-string>"
+```
+
+- *tablet_id*: The identifier (ID) for the tablet.
+- *reason-string*: Text string noting why the tablet was deleted.
```
bmatican: Maybe somehow mention that this is purely informative.

docs/content/latest/admin/yb-ts-cli.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+##### set_flag
+
+Sets the specified configuration option (flag) for the tablet server.
+
+**Syntax**
+
+```sh
+$ ./bin/yb-ts-cli set_flag [ --force ] <flag> <value>
```
bmatican: This one is actually one of the most operationally useful commands and might warrant a bit more info.

This essentially changes the in-memory value of this flag, atomically, for a running process. This means that it can alter behavior of the respective process! It also means it does NOT persist across restarts!

In practice, there are some flags that are runtime safe to change and some that are not. These are generally denoted probably only in the code, for now. For example, the bind address of the server cannot be changed at runtime, since the server binds just once at startup.

One typical operational flow is that we can use this to modify runtime flags in memory and then out of band also modify the conf file that the server uses to come up. This allows for flags to be changed on processes, without executing a restart of the process!

docs/content/latest/admin/yb-ts-cli.md
```diff
@@ -0,0 +1,208 @@
...
+
+##### --force
+
+If `true`, allows the [`set_flag`](#set-flag) command to set an option which is not explicitly marked as runtime-settable. The change may be ignored on the server or may cause the server to crash.
+
+Default: `false`
+
+##### --server-address
+
+The address of the tablet server to run against.
```
bmatican: as per above, this is host:port and is 9100 for TS and 7100 for master (some operations work on master too, such as setting flags)

It might be that it defaults to the TS rpc port of 9100

docs/content/latest/admin/ysql-dump.md
```diff
@@ -7,7 +7,7 @@ menu:
   latest:
     identifier: ysql-dump
     parent: admin
-    weight: 2466
+    weight: 2467
```
bmatican: are this and the next just for some explicit ordering?

docs/content/latest/admin/yb-ctl.md
```diff
@@ -33,7 +33,7 @@ $ ./bin/yb-ctl --help
 
 ## Commands
 
-### create
+##### create
```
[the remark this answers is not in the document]
stevebang: yes, unrelated — trying to use h5 for all commands and options. Intentional "misuse" of headings to make the content linkable. We did it at InfluxData too.

docs/content/latest/admin/yb-ts-cli.md
```diff
@@ -0,0 +1,229 @@
...
+
+Lists the tablets on the specified tablet server, displaying the following information:
+
+- Tablet id
+- State
+- Table name
+- Partition
+- Schema
+- properties
+  - lumn_name
```
bmatican: I suspect this should have been Column_name? If not, then maybe this is more confusing than useful, so could remove this?

docs/content/latest/admin/yb-ts-cli.md
```diff
@@ -0,0 +1,229 @@
...
+Sets the specified configuration option (flag) for the tablet server.
+
+**Syntax**
+
+```sh
+$ ./bin/yb-ts-cli [ --server_address=<host>:<port> ] set_flag [ --force ] <flag> <value>
+```
+
+- *host*:*port*: The *host* and *port* of the tablet server. Default is `localhost:9100`.
+- --force: Flag to to allow a change to a flag (option) that is not explicitly marked as runtime-settable. The change may be ignored on the server or may cause the server to crash. See [--force](#force).
```
bmatican: Where did we get this part from `may cause the server to crash`? I don't think that should be, unless we have some bad code that behaves badly on certain args, but also does not validate them...

docs/content/latest/admin/yb-ts-cli.md
```diff
@@ -0,0 +1,229 @@
...
+- *flag*: The `yb-tserver` configuration option (without the `--` prefix) to be set. See [`yb-tserver`](../../reference/configuration/yb-tserver/#configuration-options)
+- *value*: The value to be applied.
+
+{{< note title="Important" >}}
+
+The `set_flag` command changes the in-memory value of the specified flag, atomically, for a running process and can alter its behavior.  **The change does NOT persist across restarts.**
+
+In practice, there are some flags (options) that are runtime safe to change (runtime-settable) and some that are not. These are generally denoted probably only in the code, for now. For example, the bind address of the server cannot be changed at runtime, since the server binds just once at startup.
```
bmatican: > These are generally denoted probably only in the code

Not sure if this is useful to have in the docs :) sounds more confusing then useful..

Maybe instead we can explicitly mark them in the configuration flag section of our docs and call them out as the ones NOT runtime changeable. Obviously there might be some lag between code and doc changes, but I find it better to tell users to look at some other docs location, then tell them we might not have this info anywhere easily accessible for them.

docs/content/latest/admin/yb-ts-cli.md
bmatican: Perhaps we should at least put some note here, so it doesn't sound scary. Something like, `may cause the server to crash, if unsafe values are provided` ? That sounds a bit more defensive, even if we still do not define what is safe and what is not.

In practice, our code should already be defensive or have flag validation.

docs/content/latest/admin/yb-ts-cli.md
bmatican: I still don't like this part `These are generally denoted probably only in the code, for now.` I'd rather this say nothing at all about this and instead just put a WIP note on the configuration flag page.

stevebang: Removed that sentence. Added to note in `yb-ts-cli set_flag` that just added to the note in set_flags that most flags are probably runtime-settable, but that we need to review and update the docs accordingly

docs/content/latest/admin/yb-ts-cli.md
stevebang: Changed to say "Note that the change may be ignored on the server or may cause the server to crash, if unsafe values are provided."

PR #3555, opened 2020-02-05
Troubleshooting

bmatican:
As per offline discussion with @rkarthik007 took a pass at refactoring the troubleshooting section.

New format includes:
- basic checks (mostly ported from previous process checks)
- metrics (blank)
- query layer (ported from connectivity with ycql/yedis + new WIP page to cover /rpcz and traces)
- system stats (some porting of previous info + couple of new changes)
- logs (porting previous content)
- debug UIs (blank)
- runtime issues, split into past and current (added some content about a couple of previous and current issues)
- cluster level issues (kept old node decommission content, but I don't think it should be in this section! at best, it should live somewhere else, like the manage section and we should link to it from here...)
- platform (kept the same)

A bunch of pages I put in a skeleton WIP with comments to be expanded on in future work.


stevebang:
I will work with @bmatican to get these updates into the current structure.

PR #3596, opened 2020-02-07
delete unused case-sensitive files

haneefmubarak:
The uppercase versions of these filenames are not actually used (all repo references use the lowercase versions) and present a warning about filename case collisions when cloning the repo on macos.


stevebang:
@haneefmubarak — we'll welcome any other contributions you can make. Thanks again!

stevebang:
Oddly, when the PR was merged, the deletions were case-insensitive and removed both copies of each file. I created a new PR #3612 to add back the lowercased files. This seems an oddity of GitHub behavior.

PR #3597, opened 2020-02-08
Add a client to the docker-compose example

iSignal:
In the docker-compose example, this PR adds a client container (yb-sample-apps) that connects to the SQL service of the tserver. The client is on a separate network that is shared only with the tserver (not the master).

### DIFF
--- a/docs/content/latest/deploy/docker/docker-compose.md
+++ b/docs/content/latest/deploy/docker/docker-compose.md
@@ -126,7 +126,7 @@ CREATE TABLE
 yugabyte=#
 ```
 
-Here is an example of a client that is defined within the same docker-compose file.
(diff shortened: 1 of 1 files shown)


docs/content/latest/deploy/docker/docker-compose.md
```diff
@@ -126,7 +126,74 @@ CREATE TABLE
 yugabyte=#
 ```
 
-You can also follow the instructions in the [Quick Start](../../../quick-start/explore-ysql/#docker) section with Docker.
+Here is an example of a client that is defined within the same docker-compose file.
```
stevebang:
```suggestion
Here is an example of a client that is defined within the same Docker Compose file.
```

PR #3710, opened 2020-02-22
[WIP] Adding under replicated tablets to the health check.

psudheer21:
Summary: We add the the tablet id of all tablets that have replication lesser than the presribe replication factor.

Test Plan: Running the cluster and then checking the status on the 7000/api/v1/health-chek page

Reviewers: bogdan, sanketh

Subscribers: ybase, yugaware

### DIFF
--- a/.arclint
+++ b/.arclint
@@ -32,7 +32,7 @@
       "bin": "pycodestyle"
     },
     "googlecpplint": {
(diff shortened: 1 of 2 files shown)


kmuthukk:
In test plan -- 

s/health-chek/health-check

PR #3765, opened 2020-02-27
Add support for current version of the internet protocol (IPv6). 

### LINKED ISSUE #3644: Support IPv6 addresses across all YugabyteDB servers & CLIs
Today only IPv4 addresses are supported for all the various YugabyteDB servers and CLIs. This should be enhanced to support IPv6 addresses as well.

**Some context from https://en.wikipedia.org/wiki/IPv6_address**

IPv6 is the successor to the first addressing infrastructure of the Internet, Internet Protocol version 4 (IPv4). In contrast to IPv4, which defined an IP address as a 32-bit value, IPv6 addresses have a size of 128 bits. Therefore, IPv6 has a vastly enlarged address space compared to IPv4.

An IPv6 address is represented as eight groups of four hexadecimal digits, each group representing 16 bits (two octets, a group sometimes also called a hextet[6][7]). The groups are separated by colons (:). An example of an IPv6 address is: 2001:0db8:85a3:0000:0000:8a2e:0370:7334

synnack:
Also:
Fix incorrect preference for the legacy IP in resolver. 
Listen on IPv6 (and IPv4) by default.

Fixes #3644 
Possible shortcomings: 
1. Previously, the ParseString function accepted white space, not any more as I'm not sure that's the right place for it.
2. I'm not entirely familiar with the strings library and the use of Format vs Substitute and if they allocate memory. I've assumed they do the same thing and have standardized on where I've replaced calls.
3. ybmaster starts and works, but have not run the entire testsuite, nor extended the test suite with tests

### DIFF
--- a/src/yb/master/master-path-handlers.cc
+++ b/src/yb/master/master-path-handlers.cc
@@ -180,10 +180,10 @@ void MasterPathHandlers::CallIfLeaderOrPrintRedirect(
       }
 
       if (master.role() == consensus::RaftPeerPB::LEADER) {
-        // URI already starts with a /, so none is needed between $1 and $2.
-        redirect = Substitute("http://$0:$1$2$3",
-                              master.registration().http_addresses(0).host(),
-                              master.registration().http_addresses(0).port(),
+        // URI already starts with a /, so none is needed between $0 and $1.
+        redirect = Substitute("http://$0$1$2",
+                              HostPortToString(master.registration().http_addresses(0).host(),
+                                               master.registration().http_addresses(0).port()),
                               req.redirect_uri,
                               req.query_string.empty() ? "?raw" : "?" + req.query_string + "&raw");
         break;
@@ -265,9 +265,8 @@ void MasterPathHandlers::TServerDisplay(const std::string& current_uuid,
     if (desc->placement_uuid() == current_uuid) {
       const string time_since_hb = StringPrintf("%.1fs", desc->TimeSinceHeartbeat().ToSeconds());
       TSRegistrationPB reg = desc->GetRegistration();
-      string host_port = Substitute("$0:$1",
-                                    reg.common().http_addresses(0).host(),
-                                    reg.common().http_addresses(0).port());
+      string host_port = HostPortToString(reg.common().http_addresses(0).host(),
+                                          reg.common().http_addresses(0).port());
       *output << "  <tr>\n";
       *output << "  <td>" << RegistrationToHtml(reg.common(), host_port) << "</br>";
       *output << "  " << desc->permanent_uuid() << "</td>";
@@ -507,9 +506,8 @@ void MasterPathHandlers::HandleGetTserverStatus(const Webserver::WebRequest& req
     for (auto desc : descs) {
       if (desc->placement_uuid() == cur_uuid) {
         TSRegistrationPB reg = desc->GetRegistration();
-        string host_port = Substitute("$0:$1",
-                                      reg.common().http_addresses(0).host(),
-                                      reg.common().http_addresses(0).port());
+        string host_port = HostPortToString(reg.common().http_addresses(0).host(),
+                                            reg.common().http_addresses(0).port());
         jw.String(host_port);
 
         jw.StartObject();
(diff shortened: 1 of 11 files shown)


iSignal:
Adding @bmatican  for advice.

@synnack : could you run the unit tests and verify they pass? Could you also bring up a simple rf-3 local cluster with `yb-ctl start --rf 3` and verify that it can accept connections both on IPv4 and IPv6 on the local interface?

synnack:
Hmm, yb-ctl start --rf 3 won't really work to test this, because it explicitly binds to 127.0.0.1, 127.0.0.2 and 127.0.0.3.. which I guess is fair for a test cluster that will never have outside connectivity.. Interestingly... it only starts when I kill prometheus-node-exporter.. Seems like an annoying clash of using the same port number as prometheus uses for everything (9100). Might want to coexist with that one, given that you have prometheus metrics exports in everything.

If I kill prometheus-node-exporter, it works, but only on IPv4, obviously, as that is explicitly specified in the bind addresses. (This will never support dual stack as long as the outbound address is hardcoded to one single address)

I have, however, verified that the master listens on the right ports on IPv4 and IPv6:
yb-master 26471           root   32u  IPv6 70868974      0t0  TCP *:7100 (LISTEN)
yb-master 26471           root   33u  IPv6 70869413      0t0  TCP [2001:982:...]:38288->[2001:982:....]:7100 (ESTABLISHED)
yb-master 26471           root   34u  IPv6 70868985      0t0  TCP [2001:982:...]:7100->[2001:982:....]:38288 (ESTABLISHED)
yb-master 26471           root   35u  IPv6 70868975      0t0  TCP *:7000 (LISTEN)

I still have to test the tserver and run the test suite

synnack:
tservers work, but somehow in the web interface, the addresses get trashed.. looks like an HostPortPB encoding issue somewhere..

src/yb/master/master-path-handlers.cc
```diff
@@ -180,10 +180,10 @@ void MasterPathHandlers::CallIfLeaderOrPrintRedirect(
       }
 
       if (master.role() == consensus::RaftPeerPB::LEADER) {
-        // URI already starts with a /, so none is needed between $1 and $2.
-        redirect = Substitute("http://$0:$1$2$3",
-                              master.registration().http_addresses(0).host(),
-                              master.registration().http_addresses(0).port(),
+        // URI already starts with a /, so none is needed between $0 and $1.
+        redirect = Substitute("http://$0$1$2",
+                              HostPortToString(master.registration().http_addresses(0).host(),
```
spolitov: It worth to add HostPortToString version that accepts `HostPortPB`

src/yb/server/rpc_server.cc
```diff
@@ -57,7 +57,7 @@ using strings::Substitute;
 using std::unique_ptr;
 using std::make_unique;
 
-DEFINE_string(rpc_bind_addresses, "0.0.0.0",
+DEFINE_string(rpc_bind_addresses, "::",
```
spolitov: That would not work if we have only IPv4 addresses available.
If I recall correctly we have to listen IPv4 and IPv6 separately.

src/yb/util/net/dns_resolver.cc
```diff
@@ -136,12 +136,10 @@ Result<InetAddress> PickResolvedAddress(
   if (error) {
     return STATUS_FORMAT(NetworkError, "Resolve failed $0: $1", host, error.message());
   }
-  std::vector<InetAddress> addresses, addresses_v6;
+  std::vector<InetAddress> addresses;
```
spolitov: It was intentional to prefer IPv4 over IPv6 because we already seen users with setup when host resolves to both IPv4 and IPv6 but IPv6 actually was not properly configured.

src/yb/util/net/net_util.cc
```diff
@@ -140,24 +140,41 @@ Status HostPort::RemoveAndGetHostPortList(
...
+      host_ = str.substr(0, pos);
+      port_ = port;
+      return Status::OK();
+    }
   }
 
-  host_.swap(p.first);
-  port_ = port;
+  // Not in host:port or [host]:port format, assume it's a host
+  host_ = str;
```
spolitov: Should we remove `[]` from the host if present?

src/yb/master/master-path-handlers.cc
synnack: Yeah, that's a possibility.

src/yb/server/rpc_server.cc
synnack: This definitely works, only does not work when somebody engaged full-on legacy mode (current IP stack completely ripped out of the kernel), it may also not work if current IP sockets don't bind on legacy IP, which is true on OpenBSD and with a kernel setting on Linux. Though if we want separate sockets we need to bind with IPV6ONLY flag on the IPv6 socket and then bind with IPv4 separately.

src/yb/util/net/dns_resolver.cc
synnack: I don't feel like we should be enabling broken setups. If your current IP stack is not working, it should be fixed. Always prefer the current version instead of the legacy version in accordance with IETF Best Current Practices 177 (BCP177). Especially if we now support the current version properly..

src/yb/util/net/net_util.cc
synnack: At this point it sohuld not be present. If it was, it would have been caught by the [ ]:port if-statement above.

synnack: Unless it was given without a port, I guess.. then we *could* strip it, but it's not strictly neccesary

PR #3821, opened 2020-03-03
[docs] Add auto-generated, custom sitemap.xml to docs.yugabyte.com

stevebang:
This PR adds an auto-generated, custom `sitemap.xml` file that only includes the `/latest/` URLs. 
- As the existing Google Search cached files expire, users will see only the latest docs in their search results.
- Users can continue to find docs for older versions using the DocSearch search on our site.

During testing, Hugo generated 3559 pages. The auto-generated sitemap includes only 1443 files.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -7,6 +7,7 @@ canonifyurls: false
 googleAnalytics: UA-104956980-3
 pygmentsCodefences: true
 pygmentsStyle: "pygments"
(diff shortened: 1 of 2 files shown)


stevebang:
Attached here is a PDF of a  generated `sitemap.xml`:
[sitemap.pdf](https://github.com/yugabyte/yugabyte-db/files/4284272/sitemap.pdf)

PR #3831, opened 2020-03-04
[docs] Update robots.txt to exclude .diagram and .grammar files from Google Search indexing

stevebang:
Update `robots.txt` to:
- include link to `sitemap.xml`/latest/api/ysql/syntax_resources/commands/*.diagram
- disallow the following files:/latest/api/ysql/syntax_resources/commands/*.grammar

Based on discussion with @m-iancu

### DIFF
--- a/docs/static/robots.txt
+++ b/docs/static/robots.txt
@@ -1,4 +1,4 @@
 Sitemap: https://docs.yugabyte.com/sitemap.xml
 User-agent: *
-Disallow: /latest/api/ysql/syntax_resources/commands/*.diagram
-Disallow: /latest/api/ysql/syntax_resources/commands/*.grammar
+Disallow: /latest/api/ysql/syntax_resources/
+Allow: /latest/api/ysql/syntax_resources/grammar_diagrams/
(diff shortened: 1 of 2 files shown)


docs/static/robots.txt
```diff
@@ -1,2 +1,4 @@
+Sitemap: https://docs.yugabyte.com/sitemap.xml
 User-agent: *
-Disallow: /v1.0/
+Disallow: /latest/api/ysql/syntax_resources/commands/*.diagram
+Disallow: /latest/api/ysql/syntax_resources/commands/*.grammar
```
m-iancu: @stevebang I think the file name patterns are `*.diagram.md` and `*.grammar.md`.
Also, the logical rule would be to only allow `grammar_diagrams.md` from the syntax_resources folder.
Can we do something like:
```
Disallow: /latest/api/ysql/syntax_resources/
Allow: /latest/api/ysql/syntax_resources/grammar_diagrams.md
```

stevebang: Based on our discussion, changed to:
```
Disallow: /latest/api/ysql/syntax_resources/
Allow: /latest/api/ysql/syntax_resources/grammar_diagrams/
```

PR #3843, opened 2020-03-05
[docs] Correct prerequisites and cqlsh connection command

### LINKED ISSUE #3840: [Docs] Fix TLS "Connect to cluster" documentation
There are some mistakes in the "Connect to cluster" page for TLS: https://docs.yugabyte.com/latest/secure/tls-encryption/connect-to-cluster/

1. In prerequisites, generating yugabytedb.crt and yugabytedb.key links to "creating server certificates". Instead, it should link to "creating client certificates".

2. These prerequisites are only for YSQL. For YCQL, just the server certificates need to be generated.

3. cqlsh instructions for remote clusters: port should be 9042 and not 5433. Also, command is incorrect - cqlsh does not use -h, -p and -U options for host, port and username. It should just be `./bin/cqlsh 127.0.0.2 9042 --ssl`

stevebang:
- Fix links to client certificate files
- Edit and correct prerequisites
- Correct cqlsh connection command

Fixes #3840

### DIFF
--- a/docs/content/latest/secure/tls-encryption/connect-to-cluster.md
+++ b/docs/content/latest/secure/tls-encryption/connect-to-cluster.md
@@ -19,17 +19,17 @@ To connect CLIs, tools, and APIs to a remote YugabyteDB cluster when client-to-s
 
 ## Prerequisites
 
-Before you can use client certificates to connect to your YugabyteDB clusters, server-to-server encryption must be enabled. For details, see [Enable server-to-server encryption](../server-to-server).
+Before you can use client certificates to connect to your YugabyteDB clusters, server-to-server encryption and client-to-server encryption must be enabled. For details, see [Enable server-to-server encryption](../server-to-server/) and [Enable client-to-server encryption](../client-to-server).
 
 Also, you need to create the required client certificates. For details, see [Create client certificates](../client-certificates).
 
-Each client that connects to a YugabyteDB cluster needs the following three files to be accessible on the client computer.
+Each client that connects to a YugabyteDB cluster needs the following files to be accessible on the client computer.
 
-- `ca.crt` — root certificate file
+- `ca.crt` — root certificate file (YSQL and YCQL)
   - To generate, see [Generate the root certificate file](../server-certificates/#generate-the-root-certificate-file)
-- `yugabytedb.crt` — private node certificate
+- `yugabytedb.crt` — private node certificate (YSQL only)
   - To generate, see [Generate client private key and certificate](../client-certificates/#generate-private-key-and-certificate)
-- `yugabytedb.key` — private node key
+- `yugabytedb.key` — private node key (YSQL only)
   - To generate, see [Generate client private key and certificate](../client-certificates/#generate-private-key-and-certificate)
 
 All three files should be available in the `~/.yugabytedb`, the default location for TLS certificates when running the YSQL shell (`ysqlsh`) locally.
--- a/docs/content/latest/secure/tls-encryption/connect-to-cluster.md
+++ b/docs/content/latest/secure/tls-encryption/connect-to-cluster.md
@@ -19,16 +19,18 @@ To connect CLIs, tools, and APIs to a remote YugabyteDB cluster when client-to-s
 
 ## Prerequisites
 
-Before you can enable and use server-to-server encryption, you need to create and configure server certificates for each node of your YugabyteDB cluster. For information, see [Create client certificates](../client-certificates).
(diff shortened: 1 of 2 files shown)


docs/content/latest/secure/tls-encryption/connect-to-cluster.md
```diff
@@ -19,16 +19,18 @@ To connect CLIs, tools, and APIs to a remote YugabyteDB cluster when client-to-s
 
 ## Prerequisites
 
-Before you can enable and use server-to-server encryption, you need to create and configure server certificates for each node of your YugabyteDB cluster. For information, see [Create client certificates](../client-certificates).
+Before you can use client certificates to connect to your YugabyteDB clusters, server-to-server encryption must be enabled. For details, see [Enable server-to-server encryption](../server-to-server).
```
ndeodhar: server-to-server and client-to-server encryption must be enabled.

docs/content/latest/secure/tls-encryption/connect-to-cluster.md
```diff
@@ -19,16 +19,18 @@ To connect CLIs, tools, and APIs to a remote YugabyteDB cluster when client-to-s
...
 
-For each client that will connect to a YugabyteDB cluster, you need the following three files to be accessible on the client computer.
+Also, you need to create the required client certificates. For details, see [Create client certificates](../client-certificates).
 
-- `root.crt` — root certificate file
-  - To generate, see [Prepare root configuration file](../prepare-nodes/#generate-root-config)
+Each client that connects to a YugabyteDB cluster needs the following three files to be accessible on the client computer.
+
+- `ca.crt` — root certificate file
```
ndeodhar: Is this correct for YSQL? Or does YSQL filename need to be root.crt?

docs/content/latest/secure/tls-encryption/connect-to-cluster.md
```diff
@@ -19,16 +19,18 @@ To connect CLIs, tools, and APIs to a remote YugabyteDB cluster when client-to-s
...
+  - To generate, see [Generate the root certificate file](../server-certificates/#generate-the-root-certificate-file)
 - `yugabytedb.crt` — private node certificate
-  - To generate, see [Generate private key for each node](../prepare-nodes/#generate-private-key-for-each-node)
+  - To generate, see [Generate client private key and certificate](../client-certificates/#generate-private-key-and-certificate)
```
ndeodhar: These two are only needed if you are going to use YSQL. We should call that out explicitly.

Alternatively, we can change this to have different tabs for YCQL and YSQL and include instructions separately. In my opinion, that might be easier to understand.

docs/content/latest/secure/tls-encryption/connect-to-cluster.md
stevebang: Good catch. fixed

docs/content/latest/secure/tls-encryption/connect-to-cluster.md
stevebang: I just add "(YSQL and YCQL)" and "(YSQL only)" after the file names. I think that makes it clear.

docs/content/latest/secure/tls-encryption/connect-to-cluster.md
stevebang: Good question -- `root.crt` was my idea, but the code is looking for `ca.crt`.  I did a search and found two more instances with the wrong file name.

docs/content/latest/secure/tls-encryption/connect-to-cluster.md
```diff
@@ -19,16 +19,18 @@ To connect CLIs, tools, and APIs to a remote YugabyteDB cluster when client-to-s
...
 
-- `root.crt` — root certificate file
-  - To generate, see [Prepare root configuration file](../prepare-nodes/#generate-root-config)
-- `yugabytedb.crt` — private node certificate
-  - To generate, see [Generate private key for each node](../prepare-nodes/#generate-private-key-for-each-node)
-- `yugabytedb.key` — private node key
-  - To generate, see [Generate private key for each node](../prepare-nodes/#generate-private-key-for-each-node)
+Each client that connects to a YugabyteDB cluster needs the following files to be accessible on the client computer.
+
+- `ca.crt` — root certificate file (YSQL and YCQL)
```
ndeodhar: Did you confirm that naming the file as `ca.crt` works for YSQL?

Also, I still find this a bit hard to read (with YCQL and YSQL mentioned for every step), but it's okay to handle it later.

Issue #3882, opened 2020-03-09
[docs] Fix description field so it can be used for meta descriptions instead of page titles


stevebang:
Currently, the "description" field in the front matter is being misused for page titles. Let's fix this so that it is available for meta descriptions, which it is intended for. Here is a screenshot of a Google Search showing the topi results when filtering on our doc site. Note the repetitive and identical meta descriptions that appear below the titles in the search results:

![image](https://user-images.githubusercontent.com/8887198/76232344-cfccd700-61e3-11ea-8be7-635e6f90e286.png)

While not hurting our SEO rankings for the page, we're missing the opportunity to give users reasons to click on these pages.

PR #3962, opened 2020-03-14
[Docs][#3882] Add new 'headerTitle' param field to allow 'Description' to be used in meta tag

yb-andrew:
- Fix H1’s in HTML files to use front matter `headerTitle` if available, otherwise default to page `Title` which, for the most part, is the same. 
- Change section image in header to default to Title, which is accurate
- Change header to use page’s description if available, otherwise use Site param description, and if that fails, use a hardcoded string
- Change pagination link title to use `headerTitle` front matter if available, otherwise use the `Title` param
- Fix some cloud pages that we causing errors due to parser interpreting as html instead of markdown
- Update CONTRIBUTING.md to mention headerTitle as an override

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -19,7 +19,8 @@ All docs pages must start with a front matter as shown below.
 ---
 title: Browser-Title
 linkTitle: My-Left-Nav-Link
-description: Doc-Page-Title
+headerTitle: Doc-Page-Title (if different from Browser-Title)
+description: SEO-Meta-Description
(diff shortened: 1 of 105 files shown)


yb-andrew:
There are a ton of file changes but the core of this PR is in the HTML changes since anything else can be changed after.
Change Summary:
-Use `Title` as page Title (browser tab text)
-Use `headerTitle` when you want a custom page article or content title that is different from the `Title` param. If you want to use the same text, omit this field
-The `Description` param is now for SEO only. Please use 150-160 characters in your description

PR #4021, opened 2020-03-20
[docs] Add support for special characters in meta descriptions and titles

stevebang:
- Add fixes to Hugo to support special characters (for example, ampersand (`&`), `@`, and others)


stevebang:
Closing. Resolved by change in Markdown renderer (goldmark) and new settings.

PR #4068, opened 2020-03-25
Adding notes indicating that the benchmark binaries are compiled with…

psudheer21:
… JAVA 13

Reviewers: Neha

### DIFF
--- a/docs/content/latest/benchmark/tpcc-ysql.md
+++ b/docs/content/latest/benchmark/tpcc-ysql.md
@@ -39,6 +39,10 @@ $ tar -zxvf tpcc.tar.gz
 $ cd tpcc
 ```
 
+{{< note title="Note" >}}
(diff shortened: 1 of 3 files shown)


docs/content/latest/benchmark/tpcc-ysql.md
```diff
@@ -39,6 +39,10 @@ $ tar -zxvf tpcc.tar.gz
 $ cd tpcc
 ```
 
+{{< note title="Note" >}}
+The binaries are compiled with JAVA 13.
```
rkarthik007: Maybe be a bit more explicit? Say:
```
It is recommended to run these binaries with Java 13. Note that this package currently does not work with Java 14.
```

Also, would something compiled with Java 14 work with Java 13? If so we should just upload newer binaries and remove this note later.

PR #4129, opened 2020-04-01
Add a GitHub Actions build

mbautin:
Adding a basic build of non-test code (what's needed by the release package) in a CentOS Docker image on GitHub Actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,21 +25,11 @@ jobs:
     strategy:
       matrix:
         name:
-          #- macOS
           - "CentOS 7"
-          #- "Ubuntu 18.04"
         include:
-          #- name: "macOS"
-          #  os: macos-latest
-          #  docker_image: none
-
           - name: "CentOS 7"
             os: ubuntu-latest
             docker_image: yugabyteci/yb_build_infra_centos7:v2020-03-28T07_29_17
-
-          #- name: "Ubuntu 18.04"
-          #  os: ubuntu-latest
-          #  docker_image: yugabyteci/yb_build_infra_ubuntu1804:v2020-03-11T23_43_13
 
     if: >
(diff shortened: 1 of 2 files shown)


python/yb/download_and_extract_archive.py
```diff
@@ -191,7 +191,6 @@ def download_and_extract(url, dest_dir_parent, local_cache_dir, nfs_cache_dir):
             logging.info("Failed creating directory '%s': %s", local_cache_dir, ex)
 
     check_dir_exists_and_is_writable(local_cache_dir, "Local cache")
-    check_dir_exists_and_is_writable(nfs_cache_dir, "NFS cache")
```
rajukumaryb, line 194: With this change, `./yb_build.sh tsan --clean --dltp` was able to make progress. Thanks!

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+  # This workflow contains a single job called "build"
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        name:
+          #- macOS
+          - "CentOS 7"
+          #- "Ubuntu 18.04"
```
ttyusupov, line 29: Do we need to keep commented-out OS names in a final commit?

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+    branches:
+      - master
+    paths-ignore:
+      - README.md
+
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - README.md
```
ttyusupov, line 16: Do we need to ignore changes to docs, managed directories and might be something else?

mbautin: Added the ignore item for docs. Will add more items in a follow-up pull request once this pull request lands.

PR #4148, opened 2020-04-03
Add python-devel to list of packages

iSignal:
Needed for Python.h

Repros on vanilla CentOS 7 machine from GCP.


mbautin:
Let's add python3, python3-pip, and python3-devel as well.

PR #4337, opened 2020-04-29
[docs] add missing policies to AWS CloudFormation

### LINKED ISSUE #4336: [docs] AWS CloudFormation: user needs to have more policies
As reported by @Pranav-20186017 over Slack, the user who is creating the CFN stack via Console or CLI needs to have more permissions than just `AWSCloudFormationFullAccess`. Otherwise the stack creation fails with
```
An error occurred (ValidationError) when calling the CreateStack operation: AccessDenied. User doesn't have permission to call ssm:GetParameters
```
After adding above policy to the user we also need access to create EC2 resources,
```
AccessDenied. User doesn't have permission to call ec2:DescribeImages. Rollback requested by user.
AccessDenied. User doesn't have permission to call ec2:DescribeKeyPairs. Rollback requested by user.
API: ec2:CreateVpc You are not authorized to perform this operation.
API: ec2:CreateInternetGateway You are not authorized to perform this operation.
…
```

bhavin192:
These policies are required to create all the resources in the AWS account.
### Scenarios tested
- Created a new user and gave it following policies
  - `AWSCloudFormationFullAccess`
  - `AmazonEC2FullAccess`
  - `ssm:GetParameters`
  - `s3:CreateBucket`
  - `s3:PutObject`
  - `s3:GetObject`
- Switched to that user using switch role / or one can login directly using credentials
- Tried creating a new stack from the UI as mentioned in the document.
- Using credentials of same user, tried stack creation from CLI.

Fixes #4336

### DIFF
--- a/docs/content/latest/deploy/public-clouds/aws/cloudformation.md
+++ b/docs/content/latest/deploy/public-clouds/aws/cloudformation.md
@@ -37,10 +37,11 @@ showAsideToc: true
 
 ## Prerequisites
 
-1. You need to have an IAM user who has `AWSCloudFormationFullAccess` privilege.
-2. Create an SSH key pair that you want to attach to the nodes.
-3. In the region you want to bring up the stack, make sure you can launch new VPCs.  
-4. Download the template file.
+1. You need to have an IAM user who has `AWSCloudFormationFullAccess`, `AmazonEC2FullAccess` and `ssm:GetParameters` privilege.
+2. If you are going to create the stack via AWS Console then make sure the user also has `s3:CreateBucket`, `s3:PutObject` and `s3:GetObject` policies.
+3. Create an SSH key pair that you want to attach to the nodes.
+4. In the region you want to bring up the stack, make sure you can launch new VPCs.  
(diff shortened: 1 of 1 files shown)


Pranav-20186017:
If it helps here are the Polices that i used to deploy yugabyte on aws cloud formation without any errors
1. EC2 Full Access
2. S3 Full Access
3. VPC Full Access
4. EC2 SSM Full Access
5. Cloud Formation Full Access

bhavin192:
So the 3. `AmazonVPCFullAccess` is a subset of 1. `AmazonEC2FullAccess`, 5 is required. From 4th just `ssm:GetParameters` is enough as our template is not using any other permissions than that. And few policies from 2. are required, if creating the stack via Console. Thank you the list you used. I will update the PR accordingly.

rkarthik007:
Hi @Pranav-20186017 - thanks for the contributions! Could you please add yourself to the Yugabyte contributors list: https://github.com/yugabyte/yugabyte-db/blob/master/CONTRIBUTORS.md

PR #4456, opened 2020-05-12
listen_ip update for yb-ctl quick start and reference 

schoudhury:


docs/content/latest/quick-start/create-local-cluster/linux.md
```diff
@@ -71,9 +75,9 @@ To run a distributed SQL cluster locally for testing and development, you can qu
 $ ./bin/yb-ctl --rf 3 create
 ```
 
-You can now check `$HOME/yugabyte-data` to see `node-i` directories created where `i` represents the `node_id` of the node. Inside each such directory, there will be two disks, `disk1` and `disk2`, to highlight the fact that YugabyteDB can work with multiple disks at the same time. Note that the IP address of `node-i` is by default set to `127.0.0.i`.
+Note that in this 3-node mode, the bind IP address by default for all ports is the individual loopback addresses (that you setup in the previous step). In this mode you will not be able to externally access the database APIs and admin UIs because `0.0.0.0` remains unbound. Binding to `0.0.0.0` by even one set of servers would disallow the usage of any other address.
```
iSignal, line 78: would remove the `0.0.0.0 remains unbound` part - it doesn't sound right from a networking terminology pov. 

Separately, I feel that this explanation of why a 3 node yb-ctl doesn't bind on 0.0.0.0 is not useful to end users. For them, it is just a matter of knowing it doesn't work. The rest is our implementation reasons.

PR #4591, opened 2020-05-27
Update docker-swarm docs to fix communication between masters

baba230896:
- Added 'tasks' before the YB master service name in '--master_addresses' option to resolve communication issues between YB master containers.
- Removed extra hyphens from the initialize [YEDIS API](https://docs.yugabyte.com/latest/deploy/docker/docker-swarm/#yedis-api) command.

Tested Scenarios with this change -
- [APIs](https://docs.yugabyte.com/latest/deploy/docker/docker-swarm/#5-test-the-apis)
- [Test fault-tolerance with node failure](https://docs.yugabyte.com/latest/deploy/docker/docker-swarm/#6-test-fault-tolerance-with-node-failure)
- [Test auto-scaling with node addition](https://docs.yugabyte.com/latest/deploy/docker/docker-swarm/#7-test-auto-scaling-with-node-addition)

### DIFF
--- a/docs/content/latest/deploy/docker/docker-swarm.md
+++ b/docs/content/latest/deploy/docker/docker-swarm.md
@@ -167,7 +167,8 @@ $ docker service create \
 --publish 7000:7000 \
 yugabytedb/yugabyte:latest /home/yugabyte/bin/yb-master \
 --fs_data_dirs=/mnt/data0 \
---master_addresses=yb-master1:7100,yb-master2:7100,yb-master3:7100 \
+--master_addresses=tasks.yb-master1:7100,tasks.yb-master2:7100,tasks.yb-master3:7100 \
+--rpc_bind_addresses=0.0.0.0:7100 \
 --replication_factor=3
 ```
 
@@ -179,7 +180,8 @@ $ docker service create \
 --mount type=volume,source=yb-master2,target=/mnt/data0 \
 yugabytedb/yugabyte:latest /home/yugabyte/bin/yb-master \
 --fs_data_dirs=/mnt/data0 \
---master_addresses=yb-master1:7100,yb-master2:7100,yb-master3:7100 \
+--master_addresses=tasks.yb-master1:7100,tasks.yb-master2:7100,tasks.yb-master3:7100 \
+--rpc_bind_addresses=0.0.0.0:7100 \
 --replication_factor=3
 ```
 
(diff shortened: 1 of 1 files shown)


docs/content/latest/deploy/docker/docker-swarm.md
```diff
@@ -167,7 +167,8 @@ $ docker service create \
 --publish 7000:7000 \
 yugabytedb/yugabyte:latest /home/yugabyte/bin/yb-master \
 --fs_data_dirs=/mnt/data0 \
---master_addresses=yb-master1:7100,yb-master2:7100,yb-master3:7100 \
+--master_addresses=tasks.yb-master1:7100,tasks.yb-master2:7100,tasks.yb-master3:7100 \
+--rpc_bind_addresses=0.0.0.0:7100 \
```
iSignal, line 171: rpc_bind_addresses should be set to the node's IP not 0.0.00. For master1 that should be`tasks.yb-master1:7100`

baba230896: okay. I'll make this change.

docs/content/latest/deploy/docker/docker-swarm.md
```diff
@@ -227,7 +230,8 @@ $ docker service create \
 --publish 9000:9000 \
 yugabytedb/yugabyte:latest /home/yugabyte/bin/yb-tserver \
 --fs_data_dirs=/mnt/data0 \
---tserver_master_addrs=yb-master1:7100,yb-master2:7100,yb-master3:7100
+--rpc_bind_addresses=0.0.0.0:9100 \
```
stevebang, line 233: @iSignal Does the yb-tserver setting also need to be changed?

iSignal: yes, it does!

stevebang, line 233:
```suggestion
--rpc_bind_addresses=task.yb-master:9100 \
```

stevebang: @baba230896 That setting is applicable to both yb-master and yb-tserver. It's not explicitly stated in the docs, so I will fix that.

baba230896: No, this change isn't required. For every "yb-master" we execute its own create' command that's why we can give specific hostname or service name whereas in the case of "yb-tserver" we create multiple instances with a single command, therefore, giving hostname or service name is not possible in case of "yb-tserver".

baba230896: @stevebang If we pass this "tasks.yb-master1:9100" to the "yb-tserver" service's rpc_bind_address then deployment will fail because "tasks.yb-master1" points towards the Container IP of yb-master1 service and yb-tserver service instance doesn't have this IP in its network configuration.

baba230896:
The previous implementation is a little bit quite confusing, so I updated it with the easier one.

New implementation explanation:

- Added hostname parameter in every "yb-master" service' create command, which should be the same as the service name. Now, the hostname will help the rpc_bind_address option to bind with its container IP.
- All container communication will use Swarm VIP.
- rpc_bind_addresses for the "yb-tserver" service should remain "0.0.0.0:9100" because for every "yb-master" we execute its own create command and that's why can give specific hostname whereas in the case of "yb-tserver" we create multiple instances with a single command, therefore, giving hostname or service name is not possible in case of "yb-tserver".

PR #4684, opened 2020-06-04
Use index aggregate and limit information to EXPLAIN SELECT.

### LINKED ISSUE #4649: [YCQL] Explain plan for index scans omits limit clause
##### Set up a table.
```
create keyspace foo;
use foo;
create table test(h int, r int, v int, primary key(h,r)) with transactions = {'enabled' : true};
create index on test(v);
```
##### Insert some rows.
```
insert into test (h,r,v) values (1,1,1);
insert into test (h,r,v) values (2,2,1);
insert into test (h,r,v) values (2,3,1);
insert into test (h,r,v) values (2,4,2);
insert into test (h,r,v) values (2,5,1);
select * from test where v = 1;
```
##### Test explain plan for sequential scan.
```
explain select * from test where h = 1 limit 1;
```
```
 QUERY PLAN
---------------------------------
 Limit
   ->  Range Scan on foo.test
         Key Conditions: (h = 1)
```
##### Test explain plan for index scan.
```
explain select * from test where v = 1 limit 1;
```
```
 QUERY PLAN
--------------------------------------------------
 Index Only Scan using foo.test_v_idx on foo.test
   Key Conditions: (v = 1)
```
^ Limit component missing.

##### Actual select works as expected.
_This is an `EXPLAIN` issue only._
```
select * from test where v = 1 limit 1;
```
```
 h | r | v
---+---+---
 1 | 1 | 1

(1 rows)
```

#### Possible cause.
The function `AnalysisResultToPB` in [`pt_select.cc`](https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/yql/cql/ql/ptree/pt_select.cc#L428) should inspect the `child_select` instead of (or in addition to) the top level select when checking for a limit_clause (and possibly aggregate).

Sircular:
This PR modifies `EXPLAIN SELECT` to use limit and aggregate information from any index if it is present. Closes #4649 .

### DIFF
--- a/src/yb/yql/cql/ql/ptree/pt_select.cc
+++ b/src/yb/yql/cql/ql/ptree/pt_select.cc
@@ -423,15 +423,18 @@ ExplainPlanPB PTSelectStmt::AnalysisResultToPB() {
   string key_conditions = "  Key Conditions: ";
   string filter = "  Filter: ";
   size_t longest = 0;
+  // If index is being used, show limit clause and aggregation of the index.
+  const auto limit_clause = child_select_ ? child_select_->limit_clause_ : limit_clause_;
+  const auto aggregate = child_select_ ? child_select_->is_aggregate() : is_aggregate();
   // If overarching information( "Aggregate" | "Limit") then rest of the explain plan output needs
   // to be indented.
-  if (is_aggregate() || limit_clause_) {
-    string aggr = (is_aggregate()) ? "Aggregate" : "Limit";
-    select_plan->set_aggregate(aggr);
+  if (aggregate || limit_clause) {
+    string aggr_name = aggregate ? "Aggregate" : "Limit";
+    select_plan->set_aggregate(aggr_name);
     key_conditions = "      " + key_conditions;
     filter = "      " + filter;
     select_plan->set_select_type("  ->  " + select_plan->select_type());
-    longest = max(longest, aggr.length());
+    longest = max(longest, aggr_name.length());
   }
   longest = max(longest, select_plan->select_type().length());
   // If index is being used, change the split of key conditions and filters to that of the index.


Sircular:
@ndeodhar I would, except that whatever queries I write seem to trigger a sequential scan. Perhaps I'm not understanding the schema and how YugabyteDB determines whether to do a sequential scan or index-only scan, but the following query should be index-only, right?

```
EXPLAIN SELECT * FROM movie_stats
WHERE user_rank = 5 
AND movie_name = 'm2' 
LIMIT 1;
```

ndeodhar:
@Sircular It depends on what your index looks like. Do you have an index created on `user_rank` and `movie_name`?

See these examples:
```
CREATE TABLE imdb.movie_stats (
    movie_genre text,
    movie_name text,
    user_name text,
    user_rank int,
    last_watched timestamp,
    PRIMARY KEY (movie_genre, movie_name, user_name)
) WITH CLUSTERING ORDER BY (movie_name ASC, user_name ASC)
    AND default_time_to_live = 0
    AND transactions = {'enabled': 'true'};

CREATE INDEX best_rated ON imdb.movie_stats ((user_rank, movie_genre), movie_name, user_name)
    WITH transactions = {'enabled': 'true'};
```

Index scan example:
```
explain SELECT * FROM movie_stats WHERE user_rank = 5  AND movie_genre = 'comedy'  LIMIT 1;

 QUERY PLAN
----------------------------------------------------------------
 Index Scan using imdb.best_rated on imdb.movie_stats          
   Key Conditions: (user_rank = 5) AND (movie_genre = 'comedy')
```

Index only scan example (note that we are only selecting columns that are part of this index and not all columns from the table):
```
explain SELECT user_rank, movie_genre, movie_name FROM movie_stats WHERE user_rank = 5  AND movie_genre = 'comedy'  LIMIT 1;

 QUERY PLAN
----------------------------------------------------------------
 Index Only Scan using imdb.best_rated on imdb.movie_stats     
   Key Conditions: (user_rank = 5) AND (movie_genre = 'comedy')
```

PR #4698, opened 2020-06-07
Password cache

Sircular:
Implements a password cache to allow connections to be created more quickly from recently used accounts. Closes or ameliorates #4596 .

### DIFF
--- a/java/yb-cql/src/test/java/org/yb/cql/BaseAuthenticationCQLTest.java
+++ b/java/yb-cql/src/test/java/org/yb/cql/BaseAuthenticationCQLTest.java
@@ -66,8 +66,8 @@ public class BaseAuthenticationCQLTest extends BaseCQLTest {
 
   @BeforeClass
   public static void setUpBeforeClass() throws Exception {
-    // Setting verbose level for debugging.
     BaseMiniClusterTest.tserverArgs.add("--use_cassandra_authentication=true");
+    BaseMiniClusterTest.tserverArgs.add("--password_hash_cache_size=0");
     BaseCQLTest.setUpBeforeClass();
   }
 
--- a/src/yb/yql/cql/cqlserver/cql_processor.cc
+++ b/src/yb/yql/cql/cqlserver/cql_processor.cc
@@ -23,8 +23,6 @@
 #include "yb/rpc/messenger.h"
 #include "yb/rpc/rpc_context.h"
 
-#include "yb/util/crypt.h"
-
 #include "yb/yql/cql/cqlserver/cql_service.h"
 
 METRIC_DEFINE_histogram_with_percentiles(
@@ -505,7 +503,7 @@ CQLResponse* CQLProcessor::ProcessResult(const ExecutedResult::SharedPtr& result
                   "Provided username " + params.username + " and/or password are incorrect");
             }
             const auto& saved_hash = salted_hash_value.string_value();
-            if (bcrypt_checkpw(params.password.c_str(), saved_hash.c_str())) {
+            if (!service_impl_->CheckPassword(params.password, saved_hash)) {
               return new ErrorResponse(*request_, ErrorResponse::Code::BAD_CREDENTIALS,
                   "Provided username " + params.username + " and/or password are incorrect");
             } else if (!can_login) {
--- a/src/yb/yql/cql/cqlserver/cql_service.cc
+++ b/src/yb/yql/cql/cqlserver/cql_service.cc
@@ -13,8 +13,12 @@
 
 #include "yb/yql/cql/cqlserver/cql_service.h"
 
+#include <openssl/sha.h>
+
 #include <mutex>
 #include <thread>
+
+#include <boost/compute/detail/lru_cache.hpp>
 
 #include "yb/client/meta_data_cache.h"
 #include "yb/client/transaction_pool.h"
@@ -31,6 +35,8 @@
 #include "yb/tserver/tablet_server.h"
 
 #include "yb/util/bytes_formatter.h"
+#include "yb/util/crypt.h"
+
 #include "yb/util/mem_tracker.h"
 
 using namespace std::placeholders;
@@ -44,6 +50,8 @@ DEFINE_int64(cql_service_max_prepared_statement_size_bytes, 128_MB,
 DEFINE_int32(cql_ybclient_reactor_threads, 24,
              "The number of reactor threads to be used for processing ybclient "
              "requests originating in the cql layer");
+DEFINE_int32(password_hash_cache_size, 64, "Number of password hashes to cache. 0 or "
+             "negative disables caching.");
 
 namespace yb {
 namespace cqlserver {
@@ -65,6 +73,7 @@ CQLServiceImpl::CQLServiceImpl(CQLServer* server, const CQLServerOptions& opts)
     : CQLServerServiceIf(server->metric_entity()),
       server_(server),
       next_available_processor_(processors_.end()),
+      password_cache_(FLAGS_password_hash_cache_size),
       messenger_(server->messenger()) {
   // TODO(ENG-446): Handle metrics for all the methods individually.
   cql_metrics_ = std::make_shared<CQLMetrics>(server->metric_entity());
@@ -227,6 +236,45 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
   VLOG(1) << "DeletePreparedStatement: CQL prepared statement cache count = "
           << prepared_stmts_map_.size() << "/" << prepared_stmts_list_.size()
           << ", memory usage = " << prepared_stmts_mem_tracker_->consumption();
+}
+
+bool CQLServiceImpl::CheckPassword(
+    const std::string plain,
+    const std::string expected_bcrypt_hash) {
+  if (FLAGS_password_hash_cache_size <= 0) {
+    return util::bcrypt_checkpw(plain.c_str(), expected_bcrypt_hash.c_str()) == 0;
+  }
+
+  std::string sha_hash(SHA256_DIGEST_LENGTH, '\0');
+  {
+    SHA256_CTX ctx;
+    SHA256_Init(&ctx);
+    SHA256_Update(&ctx, plain.c_str(), plain.length());
+    SHA256_Final((unsigned char*) &sha_hash[0], &ctx);
+  }
+  // bcrypt can generate multiple hashes from a single key, since a salt is
+  // randomly generated each time a password is set. Using a compound key allows
+  // the same plaintext to be associated with different hashes.
+  std::string key = sha_hash + ":" + expected_bcrypt_hash;
+
+  {
+    std::lock_guard<std::mutex> guard(password_cache_mutex_);
+    auto entry = password_cache_.get(key);
+    if (entry) {
+      return true;
+    }
+  }
+
+  // bcrypt_checkpw has stringcmp semantics.
+  bool correct = util::bcrypt_checkpw(plain.c_str(), expected_bcrypt_hash.c_str()) == 0;
+  if (correct) {
+    std::lock_guard<std::mutex> guard(password_cache_mutex_);
+    // Boost's LRU cache interprets insertion of a duplicate key as a no-op, so
+    // even if two threads successfully log in to the same account, there should
+    // not be a race condition here.
+    password_cache_.insert(key, true);
+  }
+  return correct;
 }
 
 void CQLServiceImpl::InsertLruPreparedStatementUnlocked(const shared_ptr<CQLStatement>& stmt) {
--- a/src/yb/yql/cql/cqlserver/cql_service.h
+++ b/src/yb/yql/cql/cqlserver/cql_service.h
@@ -19,6 +19,8 @@
 
 #include <vector>
 
+#include <boost/compute/detail/lru_cache.hpp>
+
 #include "yb/client/client_fwd.h"
 
 #include "yb/yql/cql/cqlserver/cql_message.h"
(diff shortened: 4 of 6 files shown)


Sircular:
TODO: use a second hashing algorithm (most likely SHA-2) to prevent plaintext passwords from being stored in memory.

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -229,6 +234,29 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
...
+      return true;
+    }
+  }
+
+  // bcrypt_checkpw has stringcmp semantics
+  bool correct = util::bcrypt_checkpw(plain.c_str(), hash.c_str()) == 0;
+  if (correct) {
+    {
+      std::lock_guard<std::mutex> guard(password_cache_mutex_);
+      password_cache_.insert(plain, hash);
```
kmuthukk, line 254: Looks like if another thread managed to insert this mapping before this thread got here, it works just fine because boost's LRU cache seems to treat the "insert" as a no-op if the key already exists (and that's OK for our purposes).

Sircular: I can think of a couple race conditions that might happen if we change the password to be the same, since bcrypt might generate a different hash with a new salt. I'm going to change this to explicitly replace any duplicate entries. Thanks for catching this!

src/yb/yql/cql/cqlserver/cql_processor.cc
```diff
@@ -505,7 +503,7 @@ CQLResponse* CQLProcessor::ProcessResult(const ExecutedResult::SharedPtr& result
                   "Provided username " + params.username + " and/or password are incorrect");
             }
             const auto& saved_hash = salted_hash_value.string_value();
-            if (bcrypt_checkpw(params.password.c_str(), saved_hash.c_str())) {
+            if (!service_impl_->CheckPassword(params.password, saved_hash)) {
```
hectorgcr, line 506: We need a flag check here and leave the original code as the default value:

```
bool valid_password = false;
if (FLAGS_use_password_hash_cache) {
  valid_password = service_impl_->CheckPassword(params.password, saved_hash)
} else {
  valid_password = bcrypt_checkpw(params.password.c_str(), saved_hash.c_str());
}
if (valid_password) {
```

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -65,6 +69,7 @@ CQLServiceImpl::CQLServiceImpl(CQLServer* server, const CQLServerOptions& opts)
     : CQLServerServiceIf(server->metric_entity()),
       server_(server),
       next_available_processor_(processors_.end()),
+      password_cache_(16),
```
hectorgcr, line 72: let's replace this constant with some flag value

```
password_cache_(FLAGS_password hash_cache_size);

File: src/yb/yql/cql/cqlserver/cql_service.cc [hunk 4]
```diff
@@ -229,6 +234,29 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
...
+
+  {
+    std::lock_guard<std::mutex> guard(password_cache_mutex_);
+    auto cached_entry = password_cache_.get(plain);
+    if (cached_entry && *cached_entry == hash) {
+      return true;
+    }
+  }
+
+  // bcrypt_checkpw has stringcmp semantics
```
hectorgcr, line 249: nit: end comments with a period.

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -229,6 +234,29 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
...
+  bool correct = util::bcrypt_checkpw(plain.c_str(), hash.c_str()) == 0;
```
hectorgcr, line 250: This cache should also serve as a negative cache so we can still get its performance benefits when someone tries a lot of connections with a wrong password:

For example

For these cases let's assume that the username is `foo` and the password is `bar` with a bcypt hash `X` and a sha256 hash `A`. And for password `barbar` we get a sha256 hash`B` and a bcrypt hash `Y`.

Case 1: correct password, not found in the cache.
1. Client sends authentication request with foo/bar.
2. We retrieve password from the master for `foo`, the master returns `X`
3. pswd is checked with `CheckPassword`,  we run sha256 on `bar` and get `A` but find nothing on the cache.
4. We compare bcrypt(`bar`) with the returned value of `X` and they match, so we insert into the catch the pair (`A`, `X`).
5. we return OK because password match

Case 2. Correct password, found in the cache.
Same as above, except on step 3 we find the pair (`A`, `X`) and return OK because passwords match

Case 3. Wrong password, not found in the cache.
1. Client sends authentication request with foo/barbar.
2. We retrieve password from the master for `foo`, the master returns `X`
3. pswd is checked with `CheckPassword`,  we run sha256 on `barbar` and get `B` but find nothing on the cache.
4. We compare bcrypt(`barbar`) with the returned value of `X` and they don't match, but we still insert into the catch the pair (`B`, `Y`).

Case 4. Wrong password, found in the cache
Same as above, but in step 3 we find that the pair (`B`, `Y`) so we don't need to run bcrypt on `barbar`, and we return an error

src/yb/yql/cql/cqlserver/cql_service.h
```diff
@@ -147,6 +152,10 @@ class CQLServiceImpl : public CQLServerServiceIf,
   // Tracker to measure and limit memory usage of prepared statements.
   MemTrackerPtr prepared_stmts_mem_tracker_;
 
+  // Password and hash cache.
+  boost::compute::detail::lru_cache<std::string, std::string> password_cache_;
```
hectorgcr, line 156: We have started to use thread safety annotations:
`GUARDED_BY(password_cache_mutex_);`

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -229,6 +234,29 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
           << ", memory usage = " << prepared_stmts_mem_tracker_->consumption();
 }
 
+bool CQLServiceImpl::CheckPassword(const std::string plain, const std::string hash) {
```
hectorgcr, line 237: let's rename hash to `expected_bcrypt_hash`

src/yb/yql/cql/cqlserver/cql_processor.cc
Sircular: Where should we define these flags?

src/yb/yql/cql/cqlserver/cql_service.cc
Sircular: How should we handle passwords being changed while the existing password is in the cache? It's possible that the plaintext password is correct for the new hash, but the old hash is still stored in the cache.

src/yb/yql/cql/cqlserver/cql_processor.cc
Sircular: Additionally, which binaries need to declare these flags?

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -44,6 +48,8 @@ DEFINE_int64(cql_service_max_prepared_statement_size_bytes, 128_MB,
 DEFINE_int32(cql_ybclient_reactor_threads, 24,
              "The number of reactor threads to be used for processing ybclient "
              "requests originating in the cql layer");
+DEFINE_int32(password_hash_cache_size, 16, "Number of password hashes to cache. 0 or "
```
kmuthukk, line 51: Ok with making the default a bit higher like 64.

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -229,6 +236,40 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
...
+    }
+  }
+
+  // bcrypt_checkpw has stringcmp semantics.
+  bool correct = util::bcrypt_checkpw(plain.c_str(), expected_bcrypt_hash.c_str()) == 0;
+  if (correct) {
+    std::lock_guard<std::mutex> guard(password_cache_mutex_);
+    // Boost's LRU cache interprets insertion of a duplicate key as a no-op, so
+    // even if two threads successfully log in to the same account, there should
+    // not be a race conditino here.
```
kmuthukk, line 267: typo: condition

src/yb/yql/cql/cqlserver/cql_service.cc
hectorgcr: @kmuthukk So I'm assuming you want this cache to be enabled by default, correct?

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -229,6 +236,40 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
...
+  // bcrypt can generate multiple hashes from a single key, since a salt is
+  // randomly generated each time a password is set. Using a compound key allows
+  // the same plaintext to be associated with different hashes.
+  std::string key = sha_hash + ":" + expected_bcrypt_hash;
+
+  {
+    std::lock_guard<std::mutex> guard(password_cache_mutex_);
+    auto entry = password_cache_.get(key);
+    if (entry) {
+      return true;
```
hectorgcr, line 257: I have actually been thinking about this, and even with the current approach, it should still work as a negative cache. For example, let's say that r1/p1 is the right username password, and an attempt is made with r1/p2, so we would just insert p2:bcrypt(p1), false into the LRU. Next time another attempt is made to connect with the same combo r1/p2, we would return false.

Let's say that the password does change to p2, then next time a connection attempt is made with r1/p2, we would try to find the key p2:bcrypt(p2) (since bcrypt(p2) is what the master will return), and we won't find it, so we will insert a new entry into the LRU and everything works as expected

src/yb/yql/cql/cqlserver/cql_service.h
```diff
@@ -147,6 +152,11 @@ class CQLServiceImpl : public CQLServerServiceIf,
   // Tracker to measure and limit memory usage of prepared statements.
   MemTrackerPtr prepared_stmts_mem_tracker_;
 
+  // Password and hash cache. Stores each password-hash pair as a compound key;
+  // see implementation for rationale.
+  boost::compute::detail::lru_cache<std::string, bool> password_cache_ GUARDED_BY(password_cache_mutex_);
```
hectorgcr, line 157: nit: Our line limit is 100 characters.

src/yb/yql/cql/cqlserver/cql_service.cc
kmuthukk: Yes - that's correct. @hectorgcr -- the nice thing about your suggestion of this caching strategy is that we should be able to pull this off without on-disk changes. Therefore, would like the new model to be the default.

hectorgcr:
@Sircular new changes look good to me. The only items left is check that the lines are 100 characters or less, and add unit tests.
Currently most of our unit tests are in java/yb-cql/src/test/java/org/yb/cql/TestAuthentication.java. We can start by doing something similar to what we did for some of our jedis tests using junit parameters:
```@RunWith(YBParameterizedTestRunner.class)
public class TestYBJedis extends BaseJedisTest {
  private static final Logger LOG = LoggerFactory.getLogger(TestYBJedis.class);

  public TestYBJedis(JedisClientType jedisClientType) {
    super(jedisClientType);
  }

  // Run each test with Jedis, JedisCluster, and YBJedis clients.
  @Parameterized.Parameters
  public static Collection jedisClients() {
    return Arrays.asList(JedisClientType.JEDIS, JEDISCLUSTER,
        JedisClientType.YBJEDIS);
  }```

We would also need a test that changes the password while the LRU is not full, and verify that the connetion with the old password doesn't work and that the connection with the new password also works. Also some other test that creates connection with more users than elements in the LRU.

**Comment by Sircular** [comment]
Do we also need to add command-line flags in the relevant binaries to write those tests? I'm not sure which top-level functions need to include it.

File: src/yb/yql/cql/cqlserver/cql_service.cc (hunk 10 above)
Comment (Sircular) [reply src/yb/yql/cql/cqlserver/cql_service.cc:262]: I'm not sure of the value of this, since most automated systems should have the correct password, and will immediately error out if it is not correct. Also, this would push the correct password entries out of the cache, which could hypothetically be used to slow down connection creation if done maliciously. This wouldn't be a full DoS attack, since it would just regress back to previous behavior, but it's still worth considering.

File: java/yb-cql/src/test/java/org/yb/cql/BaseAuthenticationCQLTest.java [hunk 12]
```diff
@@ -66,8 +66,8 @@
 
   @BeforeClass
   public static void setUpBeforeClass() throws Exception {
-    // Setting verbose level for debugging.
-    BaseMiniClusterTest.tserverArgs.add("--use_cassandra_authentication=true");
+    BaseMiniClusterTest.tserverArgs.add("--use_cassandra_authentication=true " +
+        "--password_hash_cache_size=64");
```
hectorgcr, line 70: Shouldn't this be 4?

hectorgcr: Also, usually for each flag we add a new element in the array

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -229,6 +238,43 @@ void CQLServiceImpl::DeletePreparedStatement(const shared_ptr<const CQLStatement
           << ", memory usage = " << prepared_stmts_mem_tracker_->consumption();
 }
 
+bool CQLServiceImpl::CheckPassword(const std::string plain, const std::string expected_bcrypt_hash) {
```
hectorgcr, line 241: This line is over 100 chars

src/yb/yql/cql/cqlserver/cql_service.cc
```diff
@@ -16,6 +16,10 @@
 #include <mutex>
 #include <thread>
 
+#include <openssl/sha.h>
```
hectorgcr, line 19: This is a C header file, so it needs to be before C++ header files:
```
Found C system header after C++ system header. Should be: cql_service.h,
    c system, c++ system, other.
```

Issue #4775, opened 2020-06-13
Build failure on Ubuntu 20.04


rohan-puri:
Build fails on Ubuntu 20.04,

Reproducer on Ubuntu 20.04, {gcc (Ubuntu 9.3.0-10ubuntu2) 9.3.0}, warnings related to
"-Werror=format-truncation" & "-Werror=stringop-truncation" are generated in postgres code, which are treated as errors and hence the build fails.

Examples of both the errors,
```
[...]

PR #4776, opened 2020-06-13
[#4775] Fix build on Ubuntu 20.04 (gcc 9.3.0)

rohan-puri:
Summary:
    postgres build fails on Ubuntu 20.04 (gcc version 9.3.0), due to
    "-Werror=format-truncation" & "-Werror=stringop-truncation" warnings
    being generated.
    
    When $COMPILER_FAMILY is 'gcc8', these warnings are ignored. 'gcc8'
    compiler family is only for gcc with version 8.*
    
    Since Ubuntu 20.04, has gcc version 9.3.0, the $COMPILER_FAMILY is
    'gcc' and these warnings are not ignored and hence they are treated
    as errors and the build fails.
    
    Fix is to add the check to postgres/CMakeLists.txt for gcc versions
    9 & above.
    
    Test Plan: `./yb_build.sh release` on Ubuntu 20.04
    
    Reviewers:
    
    Reviewed By:
    
    Subcribers:
    
    Differential Revision:


mbautin:
@rohan-puri Thanks a lot for the fix!

rohan-puri:
Thanks for merging @mbautin :)

PR #4822, opened 2020-06-18
[YSQL] Use default db as yugabyte irrespective of user name 

### LINKED ISSUE #2484: [YSQL] Use default db as yugabyte irrespective of user name
If I connect to yugabyteDB via ysqlsh using a user name that's not `yugabyte`, I see this error:
```
13:40 $ ./bin/ysqlsh  -U neha
ysqlsh: FATAL:  database "neha" does not exist
```
Looks like it's using the user name as the default DB name. We should change this to always use yugabyte as the default database.

manhhiep92:
Fixes #2484

## Scenarios tested
- Before changes: 
```
./bin/ysqlsh -U xxx
ysqlsh: FATAL:  database "xxx" does not exis
```

- After changes:
```
./bin/ysqlsh -U xxx
ysqlsh (11.2-YB-2.1.8.0-b0)
Type "help" for help.

yugabyte=>
```

### DIFF
--- a/src/postgres/src/interfaces/libpq/fe-connect.c
+++ b/src/postgres/src/interfaces/libpq/fe-connect.c
@@ -1132,7 +1132,7 @@ connectOptions2(PGconn *conn)
 	{
 		if (conn->dbName)
 			free(conn->dbName);
-		conn->dbName = strdup(conn->pguser);
+		conn->dbName = strdup("yugabyte");
 		if (!conn->dbName)
 			goto oom_error;
 	}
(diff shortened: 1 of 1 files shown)


src/postgres/src/interfaces/libpq/fe-connect.c
```diff
@@ -1132,7 +1132,7 @@ connectOptions2(PGconn *conn)
 	{
 		if (conn->dbName)
 			free(conn->dbName);
-		conn->dbName = strdup(conn->pguser);
+		conn->dbName = strdup("yugabyte");
```
ndeodhar, line 1135: Can you also add this to `conninfo_add_defaults()` for consistency? - we are setting other default values like username in that method.

manhhiep92: @ndeodhar I have added default `dbname` to `conninfo_add_defaults()` as suggested. Thanks!

PR #4833, opened 2020-06-19
[docs] Remove mention of pg_ctl and add requirement to restart tservers

stevebang:
Remove mention of `pg_ctl` on the following page and add that `yb-tserver` servers must be restarted in a rolling manner for changes to take effect.


docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -129,17 +129,17 @@ Specifies the client machine addresses that this record matches. This field can
...
+If a host name is specified (anything that is not an IP address range or a special key word is treated as a host name), that name is compared with the result of a reverse name resolution of the client's IP address (e.g., reverse DNS lookup, if DNS is used). Host name comparisons are case-insensitive. If there is a match, then a forward name resolution (e.g., forward DNS lookup) is performed on the host name to check whether any of the addresses it resolves to are equal to the client's IP address. If both directions match, then the entry is considered to match. (The host name specified in the `--pg_hba_conf` flag should be the one that address-to-name resolution of the client's IP address returns, otherwise the line won't be matched. Some host name databases allow associating an IP address with multiple host names, but the operating system will only return one host name when asked to resolve an IP address.)
```
ndeodhar, line 138:
```suggestion
If a host name is specified (anything that is not an IP address range or a special key word is treated as a host name), that name is compared with the result of a reverse name resolution of the client's IP address (e.g., reverse DNS lookup, if DNS is used). Host name comparisons are case-insensitive. If there is a match, then a forward name resolution (e.g., forward DNS lookup) is performed on the host name to check whether any of the addresses it resolves to are equal to the client's IP address. If both directions match, then the entry is considered to match. (The host name specified in the `--ysql_hba_conf` flag should be the one that address-to-name resolution of the client's IP address returns, otherwise the line won't be matched. Some host name databases allow associating an IP address with multiple host names, but the operating system will only return one host name when asked to resolve an IP address.)
```

docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -129,17 +129,17 @@ Specifies the client machine addresses that this record matches. This field can
...
 
 A host name specification that starts with a dot (`.`) matches a suffix of the actual host name. So `.example.com` would match `foo.example.com` (but not just `example.com`).
 
-When host names are specified using the `--pg_hba_conf` flag, you should make sure that name resolution is reasonably fast. It can be of advantage to set up a local name resolution cache such as `nscd`. Also, you may wish to enable the configuration parameter `log_hostname` to see the client's host name instead of the IP address in the log.
+When host names are specified using the `--pg_hba_conf` flag, you should make sure that name resolution is reasonably fast. It can be advantageous to set up a local name resolution cache, such as `nscd`. Also, you may want to enable the configuration parameter `log_hostname` to see the client's host name instead of the IP address in the log.
```
ndeodhar, line 142:
```suggestion
When host names are specified using the `--ysql_hba_conf` flag, you should make sure that name resolution is reasonably fast. It can be advantageous to set up a local name resolution cache, such as `nscd`. Also, you may want to enable the configuration parameter `log_hostname` to see the client's host name instead of the IP address in the log.
```

docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -183,15 +183,21 @@ In addition to the method-specific options listed below, there is one method-ind
 
 Files included by `@` constructs are read as lists of names, which can be separated by either whitespace or commas. Comments are introduced by `#`, just as in the `--pg_hba_conf` flag, and nested `@` constructs are allowed. Unless the file name following `@` is an absolute path, it is taken to be relative to the directory containing the referencing file.
```
ndeodhar, line 184:
```suggestion
Files included by `@` constructs are read as lists of names, which can be separated by either whitespace or commas. Comments are introduced by `#`, just as in the `--ysql_hba_conf` flag, and nested `@` constructs are allowed. Unless the file name following `@` is an absolute path, it is taken to be relative to the directory containing the referencing file.
```

docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -183,15 +183,21 @@ In addition to the method-specific options listed below, there is one method-ind
...
+Because the `--pg_hba_conf` records are examined sequentially for each connection attempt, the order of the records is significant. Typically, earlier records will have tight connection match parameters and weaker authentication methods, while later records will have looser match parameters and stronger authentication methods. For example, you might want to use trust authentication for local TCP/IP connections, but require a password for remote TCP/IP connections. In this case, a record specifying `trust` authentication for connections from `127.0.0.1` would appear before a record specifying password authentication for a wider range of allowed client IP addresses.
```
ndeodhar, line 186:
```suggestion
Because the `--ysql_hba_conf` records are examined sequentially for each connection attempt, the order of the records is significant. Typically, earlier records will have tight connection match parameters and weaker authentication methods, while later records will have looser match parameters and stronger authentication methods. For example, you might want to use trust authentication for local TCP/IP connections, but require a password for remote TCP/IP connections. In this case, a record specifying `trust` authentication for connections from `127.0.0.1` would appear before a record specifying password authentication for a wider range of allowed client IP addresses.
```

docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -183,15 +183,21 @@ In addition to the method-specific options listed below, there is one method-ind
...
 
-The `--pg_hba_conf` flag is read on start-up and when the main server process receives a SIGHUP signal. If you edit the file on an active system, you will need to signal the postmaster (using `pg_ctl` reload, calling the SQL function `pg_reload_conf()`, or using `kill -HUP`) to make it re-read the file.
+The `--pg_hba_conf` flag is read on start-up of your cluster. If you edit the file on an active cluster, you need to restart your `yb-tserver` processes for changes to take effect.
```
ndeodhar, line 188:
```suggestion
The `-ysql_hba_conf` flag is read on start-up of your cluster. If you edit the file on an active cluster, you need to restart your `yb-tserver` processes for changes to take effect.
```

docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -183,15 +183,21 @@ In addition to the method-specific options listed below, there is one method-ind
...
+
+{{< note title="Important" >}}
+
+Changes to `--pg_hba_conf` should be applied to all `yb-tserver` servers in a rolling upgrade and restart, making sure that all YB-TServers are not stopped at the same time.
```
ndeodhar, line 192:
```suggestion
Changes to `--ysql_hba_conf` should be applied to all `yb-tserver` servers in a rolling upgrade and restart, making sure that all YB-TServers are not stopped at the same time.
```

docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -183,15 +183,21 @@ In addition to the method-specific options listed below, there is one method-ind
...
+
+{{< /note >}}
 
 The system view `pg_hba_file_rules` can be helpful for pre-testing changes to the `pg_hba.conf` file, or for diagnosing problems if loading of the file did not have the desired effects. Rows in the view with non-null error fields indicate problems in the corresponding lines of the file.
 
 {{< note title="Tip" >}}
 
-To connect to a particular database, a user must not only pass the `--pg_hba_conf` checks, but must have the `CONNECT` privilege for the database. If you wish to restrict which users can connect to which databases, it's usually easier to control this by granting or revoking `CONNECT` privilege than to put the rules in `pg_hba_conf` entries.
+To connect to a particular database, a user must not only pass the `--pg_hba_conf` checks, but must have the `CONNECT` privilege for the database. If you want to restrict which users can connect to which databases, it's usually easier to control this by granting or revoking `CONNECT` privilege than to put the rules in `pg_hba_conf` entries.
```
ndeodhar, line 200:
```suggestion
To connect to a particular database, a user must not only pass the `--ysql_hba_conf` checks, but must have the `CONNECT` privilege for the database. If you want to restrict which users can connect to which databases, it's usually easier to control this by granting or revoking `CONNECT` privilege than to put the rules in `ysql_hba_conf` entries.
```

docs/content/latest/secure/authentication/client-authentication.md
```diff
@@ -183,15 +183,21 @@ In addition to the method-specific options listed below, there is one method-ind
...
-The `--pg_hba_conf` flag is read on start-up and when the main server process receives a SIGHUP signal. If you edit the file on an active system, you will need to signal the postmaster (using `pg_ctl` reload, calling the SQL function `pg_reload_conf()`, or using `kill -HUP`) to make it re-read the file.
+The `--pg_hba_conf` flag is read on start-up of your cluster. If you edit the file on an active cluster, you need to restart your `yb-tserver` processes for changes to take effect.
+
+{{< note title="Important" >}}
+
+Changes to `--pg_hba_conf` should be applied to all `yb-tserver` servers in a rolling upgrade and restart, making sure that all YB-TServers are not stopped at the same time.
+
+{{< /note >}}
 
 The system view `pg_hba_file_rules` can be helpful for pre-testing changes to the `pg_hba.conf` file, or for diagnosing problems if loading of the file did not have the desired effects. Rows in the view with non-null error fields indicate problems in the corresponding lines of the file.
```
ndeodhar, line 196:
```suggestion
The system view `pg_hba_file_rules` can be helpful for pre-testing changes to the `ysql_hba_conf` flag, or for diagnosing problems if the flag did not have the desired effects. Rows in the view with non-null error fields indicate problems in the corresponding lines of the file.
```

PR #5263, opened 2020-07-30
YSQL Regression Tests for Aggregate Functions

ngov17:
This pull request adds regression tests for Aggregate functions from the postgres framework `(src/postgres/src/test/regress/)`  which include `yb_pg_aggregates` and `yb_aggregates`. 

In light of this new schedule for aggregate functions, the test for` yb_pg_aggregates` in` yb_pg_misc_serial_schedule` has been deleted in pull request https://github.com/yugabyte/yugabyte-db/pull/5264.

This pull request was originally part of the larger pull request https://github.com/yugabyte/yugabyte-db/pull/5190 (now edited) which has now been broken down into 4 separate pull requests:  https://github.com/yugabyte/yugabyte-db/pull/5190 (edited), https://github.com/yugabyte/yugabyte-db/pull/5262, https://github.com/yugabyte/yugabyte-db/pull/5263, https://github.com/yugabyte/yugabyte-db/pull/5264.

PR #5262, opened 2020-07-30
 [YSQL] Enabled USING clause in DELETE and FROM clause in UPDATE

ngov17:
This pull request enables USING clause in DELETE and FROM clause in UPDATE in gram.y for queries like:
```
DELETE FROM a_table USING another_table WHERE a_table.id = another_table.id;
UPDATE a_table SET col = xxx FROM another_table WHERE a_table.id = another_table.id;
```
Tests for such queries are present in TestPgUpdate.java and TestPgDelete.java.

This effectively closes Issue https://github.com/yugabyte/yugabyte-db/issues/738.

This pull request was originally part of the larger pull request https://github.com/yugabyte/yugabyte-db/pull/5190 (now edited) which has now been broken down into 4 separate pull requests:  https://github.com/yugabyte/yugabyte-db/pull/5190 (edited), https://github.com/yugabyte/yugabyte-db/pull/5262, https://github.com/yugabyte/yugabyte-db/pull/5263, https://github.com/yugabyte/yugabyte-db/pull/5264.

PR #5190, opened 2020-07-23
 YSQL Regression Tests for Join Clause

ngov17:
This pull request adds a test for join (`yb_pg_join`) from the  postgres framework `src/postgres/src/test/regress/`

This pull request was originally part of the larger pull request https://github.com/yugabyte/yugabyte-db/pull/5190 (now edited) which has now been broken down into 4 separate pull requests:  https://github.com/yugabyte/yugabyte-db/pull/5190 (edited), https://github.com/yugabyte/yugabyte-db/pull/5262, https://github.com/yugabyte/yugabyte-db/pull/5263, https://github.com/yugabyte/yugabyte-db/pull/5264.

PR #5264, opened 2020-07-30
YSQL Regression Tests for Copy, Groupingsets, Limit, Random

ngov17:
This pull request adds the following regression tests from the postgres framework` (src/postgres/src/test/regress/)`:
1) Copy(`yb_pg_copy2` and `yb_pg_copydml`). Separate test and schedule created: `yb_pg_copy_schedule`

The following tests don't have a separate schedule and are in `yb_pg_misc_serial_schedule`:

2) Groupingsets (`yb_pg_groupingsets.sql`)
3) Limit (`yb_pg_limit`)
4) Random `(yb_pg_random`)

_Note: Any changes to an original postgres sql (ex foo.sql) file will be explained in the comments of the new corresponding yb file(yb_pg_foo.sq)l which is the YB-compatible modified version of the original foo.sql._

Also, the test for` yb_pg_aggregates` was deleted in `yb_pg_misc_serial_schedule`  as it was added in a separate test and schedule in pull request https://github.com/yugabyte/yugabyte-db/pull/5263.

This pull request was originally part of the larger pull request https://github.com/yugabyte/yugabyte-db/pull/5190 which has now been broken down into 4 separate pull requests:  https://github.com/yugabyte/yugabyte-db/pull/5190 (edited), https://github.com/yugabyte/yugabyte-db/pull/5262, https://github.com/yugabyte/yugabyte-db/pull/5263, https://github.com/yugabyte/yugabyte-db/pull/5264.


src/postgres/src/test/regress/yb_pg_copy_schedule
```diff
@@ -0,0 +1,8 @@
+
+# src/test/regress/yb_pg_copy2_schedule
```
jaki, line 2:
```suggestion
# src/test/regress/yb_pg_copy_schedule
```

jaki (commented):
Here's a partial review.  Most notable is the item for moving code changes to PR #5263.

src/postgres/src/test/regress/expected/yb_pg_groupingsets.out
```diff
@@ -0,0 +1,1570 @@
...
+                                     ^
+-- min max optimization should still work with GROUP BY ()
+explain (costs off)
+  select min(unique1) from tenk1 GROUP BY ();
+       QUERY PLAN
+-------------------------
+ Aggregate
+   Group Key: ()
+   ->  Seq Scan on tenk1
+(3 rows)
```
jaki, line 469: The diff is different here.  This is an architectural difference (lsm hash vs btree).  If this difference arises in many other areas, we should consider modifying `yb_pg_create_index.sql` to create this index as a range index rather than hash index so that this query uses optimization and reduces diff with the original test.

If you don't want to look into that, feel free to just add a comment saying that this is expected behavior as I described.

src/postgres/src/test/regress/expected/yb_pg_groupingsets.out
```diff
@@ -0,0 +1,1570 @@
...
+   Sort Key: hundred
+     Group Key: hundred
+   Sort Key: ten
+     Group Key: ten
+   Sort Key: four
+     Group Key: four
+   ->  Sort
+         Sort Key: unique1
+         ->  Seq Scan on tenk1
+(16 rows)
```
jaki, line 1501: Can you look more into why this differs?

src/postgres/src/test/regress/expected/yb_pg_copy2.out
```diff
@@ -0,0 +1,584 @@
...
+-- check copy out
+COPY x TO stdout;
+9999	\N	\\N	NN	before trigger fired
+10000	21	31	41	before trigger fired
+10001	22	32	42	before trigger fired
+10002	23	33	43	before trigger fired
+10003	24	34	44	before trigger fired
+10004	25	35	45	before trigger fired
+10005	26	36	46	before trigger fired
+3000	\N	c	\N	before trigger fired
```
jaki, line 103: Why are some rows omitted here compared to the original test?  Please add comments if there's a legitimate reason so that I (and future people) don't have to dig through why you made this decision.

src/postgres/src/test/regress/expected/yb_pg_copy2.out
```diff
@@ -0,0 +1,584 @@
...
+---
+ x
+ y
+(2 rows)
+
+COMMIT;
+BEGIN;
+TRUNCATE vistest;
+COPY vistest FROM stdin CSV;
+COMMIT;
```
jaki, line 333: Why was `BEGIN` and `COMMIT` added?  Please add comments when you change things from the original test that are nonobvious.

src/postgres/src/test/regress/yb_pg_misc_serial_schedule
```diff
@@ -18,14 +17,13 @@ test: yb_pg_create_table
...
+# Dependency for `yb_pg_select`, 'yb_pg_groupingsets', 'yb_pg_random', 'yb_pg_limit'
 test: yb_pg_copy
 
 test: yb_pg_create_operator
 
-# Dependency for `yb_pg_aggregates`
-test: yb_pg_create_aggregate
-
 test: yb_pg_errors
 test: yb_pg_select
```
jaki, line 26:
```suggestion
test: yb_pg_select
test: yb_pg_errors
```

This is not part of your change but worth correcting.  Order using `serial_schedule`.

jaki (changes requested):
Sorry for the wait.  Left some comments.

PR #5289, opened 2020-07-31
Global replace of triple-backtick 'postgresql' lexer syntaxwith 'plpg…

bllewell:
**Global replace of triple-backtick 'postgresql' lexer syntaxwith 'plpgsql' in JSON, Arrays, and Window functions**

292 replacements made in 83 files.


bllewell:
Thanks, Steve. This change makes a **HUGE** improvement to the readability of the code blocks.

PR #5350, opened 2020-08-10
[issue-#5344] Add username fallback in gen version info

danwt:
Hope its OK.


danwt:
Most likely abandonware

PR #5357, opened 2020-08-10
DumpReplayStateToStrings should handle too many entries scenario

tedyu:
Summary:
In DumpReplayStateToLog(), DumpReplayStateToStrings is called to gather the logs in vector<string>.

It seems we can trim replay state entries in DumpReplayStateToLog() (by checking OpIndexToEntryMap.size() against kMaxLinesToDump) since the entries beyond kMaxLinesToDump would not be logged. This way memory pressure is reduced by not collecting entries unnecessarily.

Test Plan:
Run test suite.

Reviewers: sergei

### DIFF
--- a/src/yb/tablet/tablet_bootstrap.cc
+++ b/src/yb/tablet/tablet_bootstrap.cc
@@ -185,9 +185,9 @@ struct ReplayState {
   CHECKED_STATUS UpdateSplitOpId(const ReplicateMsg& msg, const TabletId& tablet_id);
 
   void AddEntriesToStrings(
-      const OpIndexToEntryMap& entries, std::vector<std::string>* strings) const;
+      const OpIndexToEntryMap& entries, std::vector<std::string>* strings, int half_limit) const;
 
-  void DumpReplayStateToStrings(std::vector<std::string>* strings) const;
+  void DumpReplayStateToStrings(std::vector<std::string>* strings, int half_limit) const;
 
   bool CanApply(log::LogEntryPB* entry);
 
@@ -316,13 +316,30 @@ Status ReplayState::UpdateSplitOpId(const ReplicateMsg& msg, const TabletId& tab
 }
 
 void ReplayState::AddEntriesToStrings(const OpIndexToEntryMap& entries,
-                                      std::vector<std::string>* strings) const {
-  for (const OpIndexToEntryMap::value_type& map_entry : entries) {
-    strings->push_back(Format("   [$0] $1", map_entry.first, map_entry.second.entry.get()));
+                                      std::vector<std::string>* strings,
+                                      int half_limit) const {
+  const int n = entries.size();
+  const bool overflow = n > 2 * half_limit;
+  size_t index = 0;
+  for (const auto& entry : entries) {
+    if (!overflow || (index < half_limit || index >= n - half_limit)) {
+      const auto& replicate = entry.second.entry.get()->replicate();
+      strings->push_back(Format(
+          "    [$0] op_id: $1 hybrid_time: $2 op_type: $3 committed_op_id: $4",
+          index + 1,
+          OpId::FromPB(replicate.id()),
+          replicate.hybrid_time(),
+          replicate.op_type(),
+          OpId::FromPB(replicate.committed_op_id())));
+    };
+    if (overflow && index == half_limit - 1) {
+      strings->push_back(Format("($0 lines skipped)", n - 2 * half_limit));
+    }
+    index++;
   }
 }
 
-void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings)  const {
+void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings, int half_limit)  const {
   strings->push_back(Substitute(
       "ReplayState: "
       "Previous OpId: $0, "
@@ -341,7 +358,7 @@ void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings)  c
   }
   if (!pending_replicates.empty()) {
     strings->push_back(Substitute("Dumping REPLICATES ($0 items):", pending_replicates.size()));
-    AddEntriesToStrings(pending_replicates, strings);
+    AddEntriesToStrings(pending_replicates, strings, half_limit);
   }
 }
 
@@ -961,23 +978,10 @@ class TabletBootstrap {
     // Dump the replay state, this will log the pending replicates, which might be useful for
     // debugging.
     vector<string> state_dump;
-    replay_state_->DumpReplayStateToStrings(&state_dump);
     constexpr int kMaxLinesToDump = 1000;
-    static_assert(kMaxLinesToDump % 2 == 0, "Expected kMaxLinesToDump to be even");
-    if (state_dump.size() <= kMaxLinesToDump) {
-      for (const string& line : state_dump) {
-        LOG_WITH_PREFIX(INFO) << line;
-      }
-    } else {
-      int i = 0;
-      for (const string& line : state_dump) {
-        LOG_WITH_PREFIX(INFO) << line;
-        if (++i >= kMaxLinesToDump / 2) break;
-      }
-      LOG_WITH_PREFIX(INFO) << "(" << state_dump.size() - kMaxLinesToDump << " lines skipped)";
-      for (i = state_dump.size() - kMaxLinesToDump / 2; i < state_dump.size(); ++i) {
-        LOG_WITH_PREFIX(INFO) << state_dump[i];
-      }
+    replay_state_->DumpReplayStateToStrings(&state_dump, kMaxLinesToDump / 2);
+    for (const string& line : state_dump) {
+      LOG_WITH_PREFIX(INFO) << line;
     }
   }
 


tedyu:
@spolitov 
Please take a look.

tedyu:
@mbautin 
I have updated the PR where each log line is simplified.
w.r.t. limit handling, we are dealing with:
```
 // two segments are: [1, first], [second, n]
```
the calculation of first and second would be dependent on whether the limit is smaller than number of entries or not.

The current formation is easy to understand.

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -316,13 +316,35 @@ Status ReplayState::UpdateSplitOpId(const ReplicateMsg& msg, const TabletId& tab
 }
 
 void ReplayState::AddEntriesToStrings(const OpIndexToEntryMap& entries,
-                                      std::vector<std::string>* strings) const {
+                                      std::vector<std::string>* strings,
+                                      int limit) const {
```
mbautin, line 320: limit -> half_limit

There is no need to support an odd limit here.

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -316,13 +316,35 @@ Status ReplayState::UpdateSplitOpId(const ReplicateMsg& msg, const TabletId& tab
...
+  int n = entries.size();
```
mbautin, line 321: Let's create a utility function (as a lambda -- only used in the same function).

```c++
auto append_metadata_for_range = [&](size_t start_index, size_t num_records) {
  for (auto i = start_index; i < start_index + num_records; ++i) {
    const auto& replicate = entries[i];
    strings->push_back(Format(
        "    [$0] op_id: $1 hybrid_time: $2 op_type: $3 committed_op_id: $4",
        i + 1,
        OpId::FromPB(replicate.id()),
        replicate.hybrid_time(),
        replicate.op_type(),
        OpId::FromPB(replicate.committed_op_id())));
  }
};
```

Note that I've replaced `id` -> `op_id` as this is a more common notation for Raft OpIds in our codebase. Also, note the use of `OpId::FromPB` which will convert the OpIdPB to the non-protobuf OpId struct that then gets automatically formatted as "<term>.<index>".

Then, I think the logic that follows (with "first", "second", "truncated", etc.) is still too complicated. Let's replace it with the following:

```c++
const auto n = entries.size();
if (n <= 2 * half_limit) {
  append_metadata_for_range(0, n);
  return;
}
append_metadata_for_range(0, half_limit);
strings->push_back(Format("($0 lines skipped)", n - 2 * half_limit));
append_metadata_for_range(n - half_limit, half_limit);
```

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -316,13 +316,35 @@ Status ReplayState::UpdateSplitOpId(const ReplicateMsg& msg, const TabletId& tab
...
+    if (truncated && i > first && i < second) continue;
+    const auto& replicate = map_entry.second.entry.get()->replicate();
+    strings->push_back(Format("   [$0] id: $1.$2 hybrid_time: $3 op_type: $4 committed_op_id: $5", map_entry.first,
+      replicate.id().term(), replicate.id().index(),
+      replicate.hybrid_time(), replicate.op_type(), replicate.committed_op_id()));
   }
 }
 
-void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings)  const {
+void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings, int kMaxLinesToDump)  const {
```
mbautin, line 347: We don't use the `kSomeConstant` naming style for function parameters.

src/yb/tablet/tablet_bootstrap.cc
tedyu: w.r.t. entries[I], I get the following compilation error:
```
src/yb/tablet/tablet_bootstrap.cc:340:22: error: non-const lvalue reference to type 'log::LogEntryPB' cannot bind to a temporary of type 'std::__1::unique_ptr<yb::log::LogEntryPB, std::__1::default_delete<yb::log::LogEntryPB> >::pointer' (aka 'yb::log::LogEntryPB *')
    log::LogEntryPB& entry = map_entry.second.entry.get();

------
src/yb/tablet/tablet_bootstrap.cc:324:38: error: no viable overloaded operator[] for type 'const yb::tablet::OpIndexToEntryMap' (aka 'const map<long long, yb::tablet::Entry>')
      const auto& replicate = entries[i];
                              ~~~~~~~^~
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/map:1126:18: note: candidate function not viable: 'this' argument has type 'const yb::tablet::OpIndexToEntryMap' (aka 'const map<long long, yb::tablet::Entry>'), but method is not marked const
    mapped_type& operator[](const key_type& __k);
                 ^
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/map:1128:18: note: candidate function not viable: 'this' argument has type 'const yb::tablet::OpIndexToEntryMap' (aka 'const map<long long, yb::tablet::Entry>'), but method is not marked const
    mapped_type& operator[](key_type&& __k);
```

src/yb/tablet/tablet_bootstrap.cc
tedyu: If you look at the call chain, the limit for AddEntriesToStrings can be odd in current code base.
I don't think one log line would make a difference in terms of the efficacy of the dump, so I can use half_limit.

src/yb/tablet/tablet_bootstrap.cc
tedyu: In order to use the [] syntax to access the map, I had to change the method declaration for AddEntriesToStrings (dropping the const):
```
void ReplayState::AddEntriesToStrings(OpIndexToEntryMap& entries,
                                      std::vector<std::string>* strings,
```
Now I have:
```
      const auto& replicate = entries[i].entry->replicate();
```
But there is still one compilation error:
```
src/yb/tablet/tablet_bootstrap.cc:365:25: error: binding value of type 'const map<...>' to reference to type 'map<...>' drops 'const' qualifier
    AddEntriesToStrings(pending_replicates, strings, (limit-cnt)/2);
                        ^~~~~~~~~~~~~~~~~~
src/yb/tablet/tablet_bootstrap.cc:318:58: note: passing argument to parameter 'entries' here
void ReplayState::AddEntriesToStrings(OpIndexToEntryMap& entries,
                                                         ^
```

mbautin: OK, sorry, it's a map. Of course `[]` won't work here, and changing it to non-const is not an option. Then I would use something like this, along the lines of your solution, but with fewer cases to reason about (with min/max, etc.)
```c++
const auto n = entries.size();
const bool overflow = n > 2 * half_limit;
size_t index = 0;
for (const auto& entry : entries) {
  if (!overflow || (index < half_limit || index >= n - half_limit)) {
    const auto& replicate = entry.second.entry->replicate();
    strings->push_back(Format(
        "    [$0] op_id: $1 hybrid_time: $2 op_type: $3 committed_op_id: $4",
        index + 1,
        OpId::FromPB(replicate.id()),
        replicate.hybrid_time(),
        replicate.op_type(),
        OpId::FromPB(replicate.committed_op_id())));
  };
  if (overflow && index == half_limit - 1) {
    strings->push_back(Format("($0 lines skipped)", n - 2 * half_limit));
  }
  index++;
}
```

tedyu:
ReplayState::AddEntriesToStrings fills up the vector of strings which would be later logged.

How about dropping the vector and log the chosen entries directly ?

Thanks

tedyu:
@mbautin 
Do you think this can be merged ?

Thanks

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -335,13 +352,16 @@ void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings)  c
       pending_replicates.size(),
       OpIdToString(regular_stored_op_id),
       OpIdToString(intents_stored_op_id)));
+  int cnt = 1;
```
mbautin, line 355: Remove this

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -335,13 +352,16 @@ void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings)  c
...
   if (num_entries_applied_to_rocksdb > 0) {
     strings->push_back(Substitute("Log entries applied to RocksDB: $0",
                                   num_entries_applied_to_rocksdb));
+    cnt++;
```
mbautin, line 359: Remove this

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -316,13 +316,30 @@ Status ReplayState::UpdateSplitOpId(const ReplicateMsg& msg, const TabletId& tab
...
+    };
+    if (overflow && index == half_limit - 1) {
+      strings->push_back(Format("($0 lines skipped)", n - 2 * half_limit));
+    }
+    index++;
   }
 }
 
-void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings)  const {
+void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings, int limit)  const {
```
mbautin, line 342: Parameterize this with half_limit instead of limit

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -335,13 +352,16 @@ void ReplayState::DumpReplayStateToStrings(std::vector<std::string>* strings)  c
...
   if (num_entries_applied_to_rocksdb > 0) {
     strings->push_back(Substitute("Log entries applied to RocksDB: $0",
                                   num_entries_applied_to_rocksdb));
+    cnt++;
   }
   if (!pending_replicates.empty()) {
     strings->push_back(Substitute("Dumping REPLICATES ($0 items):", pending_replicates.size()));
-    AddEntriesToStrings(pending_replicates, strings);
+    cnt++;
+    AddEntriesToStrings(pending_replicates, strings, (limit-cnt)/2);
```
mbautin, line 364: Remove `cnt` altogether and just pass half_limit here.

Also or code style for expressions is `(limit - cnt) / 2`

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -961,23 +981,11 @@ class TabletBootstrap {
...
-      int i = 0;
-      for (const string& line : state_dump) {
-        LOG_WITH_PREFIX(INFO) << line;
-        if (++i >= kMaxLinesToDump / 2) break;
-      }
-      LOG_WITH_PREFIX(INFO) << "(" << state_dump.size() - kMaxLinesToDump << " lines skipped)";
-      for (i = state_dump.size() - kMaxLinesToDump / 2; i < state_dump.size(); ++i) {
-        LOG_WITH_PREFIX(INFO) << state_dump[i];
-      }
+    replay_state_->DumpReplayStateToStrings(&state_dump, kMaxLinesToDump);
```
mbautin, line 986: Maybe replace constant name with `kHalfMaxLinesToDump = 500` and remove the assertion.

tedyu: I removed the assertion.
Having the kMaxLinesToDump would give readers clear idea what's the total number of lines.

Issue #5402, opened 2020-08-14
[docs] add stable release concept to docs infrastructure


schoudhury:
**Today with a single latest version that is used for both prod and non-prod**
v2.2 (latest version) - https://docs.yugabyte.com/latest/
v2.1 (earlier version) - https://docs.yugabyte.com/v2.1/
v2.0 (earlier version) - https://docs.yugabyte.com/v2.0/
v1.3 (earlier version) - https://docs.yugabyte.com/v1.3/
v1.2 (earlier version) - https://docs.yugabyte.com/v1.2/
v1.1 (earlier version) - https://docs.yugabyte.com/v1.1/
v1.0 (earlier version) - https://docs.yugabyte.com/v1.0/

**Required asap to support the new concept of latest (for non-prod) and stable (for prod) releases** 
v2.3 (latest version) - https://docs.yugabyte.com/latest/
v2.2 (stable version) - https://docs.yugabyte.com/stable/
v2.1 (earlier version) - https://docs.yugabyte.com/v2.1/
v2.0 (earlier version) - https://docs.yugabyte.com/v2.0/
v1.3 (earlier version) - https://docs.yugabyte.com/v1.3/

Additionally (nice to have)
-Delete all previous versions (as long as v1.0/v1.1/v1.2 links are auto redirected to https://docs.yugabyte.com/)

**Next major release**
v2.5 (latest version) - https://docs.yugabyte.com/latest/
v2.2 (stable version) - https://docs.yugabyte.com/stable/
v2.1 (earlier version) - https://docs.yugabyte.com/v2.1/
v2.0 (earlier version) - https://docs.yugabyte.com/v2.0/
v1.3 (earlier version) - https://docs.yugabyte.com/v1.3/

**1mo after Next major release**
v2.5 (latest version) - https://docs.yugabyte.com/latest/
v2.4 (stable version) - https://docs.yugabyte.com/stable/
v2.2 (earlier version) - https://docs.yugabyte.com/v2.2/
v2.1 (earlier version) - https://docs.yugabyte.com/v2.1/
v2.0 (earlier version) - https://docs.yugabyte.com/v2.0/
v1.3 (earlier version) - https://docs.yugabyte.com/v1.3/

cc @stevebang @rkarthik007

yb-andrew:
This can be done by creating a stable directory and adding some extra logic in the version dropdown. The old links can be redirected to the home page with a Netlify.toml configuration.

stevebang:
[...]

PR #5437, opened 2020-08-19
[#5402] Add stable content directory and modify version list

yb-andrew:
Copied contents of latest into stable directory and changed menu entries to be under stable instead of latest. Change version numbers in version_switcher to include two special entries (one for latest and another for stable).


yb-andrew:
> @yb-andrew Pleasse confirm here that the aliases and other links have been updated. Then this PR is approved.

Yes, the aliases and links inside /stable content directory now point to "stable" content instead of "latest"

Issue #3893, opened 2020-03-10
[YCQL] system_schema.functions undefined column 'body'


ybnelson:
Jira Link: [DB-4672](https://yugabyte.atlassian.net/browse/DB-4672)
For migrating Cortex, https://github.com/yugabyte/cortex, a project which uses CQL and Cassandra 3.11, it expects the following layout of system_schema.functions:

```
cqlsh> SELECT * FROM system_schema.functions;

 keyspace_name | function_name | argument_types | argument_names | body | called_on_null_input | language | return_type
---------------+---------------+----------------+----------------+------+----------------------+----------+-------------
cqlsh> DESCRIBE system_schema.functions;
CREATE TABLE system_schema.functions (
    keyspace_name text,
    function_name text,
    argument_types frozen<list<text>>,
    argument_names frozen<list<text>>,
    body text,
    called_on_null_input boolean,
    language text,
    return_type text,
    PRIMARY KEY (keyspace_name, function_name, argument_types)
) WITH CLUSTERING ORDER BY (function_name ASC, argument_types ASC)
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
    AND comment = 'user defined function definitions'
    AND compaction = {'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32', 'min_threshold': '4'}
    AND compression = {'chunk_length_in_kb': '64', 'class': 'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND crc_check_chance = 1.0
    AND dclocal_read_repair_chance = 0.0
    AND default_time_to_live = 0
    AND gc_grace_seconds = 604800
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 3600000
    AND min_index_interval = 128
    AND read_repair_chance = 0.0
    AND speculative_retry = '99PERCENTILE';
```
From YBDB 2.1.1 instance:
```
cqlsh> SELECT * FROM system_schema.functions;

 keyspace_name | function_name | argument_types | argument_names | called_on_null_input | language | return_type
---------------+---------------+----------------+----------------+----------------------+----------+-------------

Error generated:
[...]

PR #5493, opened 2020-08-25
#3893 system_schema.functions and system_schema.aggregates now match …

marcusbooyah:
…Cassandra

Added field `body` to `system_schema.functions`
Added field `return_type` to `system_schema.aggregates`
Added `argument_types` to the primary key for `system_schema.funtions` and `system_schema.aggregates`
`argument_types` is now a frozen list in `system_schema.functions` and `system_schema.aggregates`


OlegLoginov:
@marcusbooyah Please describe a test plan for the changes.

marcusbooyah:
@OlegLoginov I am not entirely sure of the blast radius of these changes. The [current tests](https://github.com/yugabyte/yugabyte-db/blob/13897978217ab4a04936a0f90d56688406136e48/java/yb-cql/src/test/java/org/yb/cql/TestSystemTables.java#L204) related to these system tables only check if they are empty - do you think more is required? I currently have a Cortex cluster running, using Yugabyte (with these changes) as the backing data store.

m-iancu:
@marcusbooyah Technically, we could add a new test function that looks into the cluster metadata for each system table and checks the column names&type. 
For instance something like:
```
    Map<String, DataType> expected_cols = new HashMap<>();
    // fill in expected_cols.
    TableMetadata tmd = cluster.getMetadata().getKeyspace("system_schema").getTable("aggregates");
    for (ColumnMetadata cmd : tmd.getColumns()) {
      assertTrue(expected_cols.containsKey(cmd.getName()));
      assertEquals(cmd.getType(), expected_cols.get(cmd.getName()));
    }
} 
```
However, that might be overkill for this change.
Maybe we can just change the `select * ..` in that test (https://github.com/yugabyte/yugabyte-db/blob/master/java/yb-cql/src/test/java/org/yb/cql/TestSystemTables.java#L207) to explicitly call out the columns to ensure they exist (at least for the two tables that you modified).
e.g.
```
    assertEquals(0, session.execute("SELECT keyspace_name, aggregate_name, argument_types ...
```

cc @OlegLoginov

Issue #3956, opened 2020-03-13
Allow specifying env variable to create a new db user / password for yugabyte's docker image


iSignal:
Feedback from @desmap on #3951 

Similar to mysql's user and password environment variables MYSQL_USER, MYSQL_PASSWORD documented at https://hub.docker.com/_/mysql, we could have YB_USER, YB_PASSWORD env variables to have yugabyted within the docker container create a new user/password on db create.

PR #5499, opened 2020-08-26
Added env variables functionality in yugabyted

baba230896:
### Summary:

Added env variables functionality in `yugabyted`

- Added env variables for YSQL: `YSQL_USER, YSQL_PASSWORD, YSQL_DB`
- Added env variables for YCQL: `YCQL_USER, YCQL_PASSWORD, YCQL_KEYSPACE`
-  Added docs to use environment variables

**Edited:**
-----

- Environment variables should exist with every execution to maintain idempotency. Authentication flags will be added to the tserver on the existence of environment variables.
- `yugabyted` won't execute environment variables based configuration snippet on `follower nodes`.

**Note:** 

- `bin/yugabyted status` command output will change with the existence of environment variables.
- For now, the default credentials haven't been removed.

**Test Plan:**

1. A single node cluster with and without environment variables.
2. Multinode cluster with and without environment variables.
3. Multiple combinations of environment variables as suggested in the docs.
4. Tested with the latest docker image.
5. Tested with Python v2.7.5 and v3.5.2

Fixes: https://github.com/yugabyte/yugabyte-db/issues/3956

### DIFF
--- a/bin/yugabyted
+++ b/bin/yugabyted
@@ -94,8 +94,6 @@ Please set 'PGPASSWORD' in environment variables to use 'bin/ysqlsh'."
 DEFAULT_YCQL_USER = "cassandra"
 DEFAULT_YCQL_PASSWORD = "cassandra"
 DEFAULT_YCQL_KEYSPACE = None
-DEFAULT_YSQL_AUTH = False
-DEFAULT_YCQL_AUTH = False
 VERSION_METADATA_PATH = os.path.join(YUGABYTE_DIR, "version_metadata.json")
 YUGABYTE_API_CLIENT_PROGRAMS = {
     "ysql": "ysqlsh",
@@ -133,7 +131,7 @@ class ControlScript(object):
         self.stop_callhome = False
         self.alerts = []
         self.script = None
-        self.setup_env_init = SetupEnvBasedCredentials()
+        self.setup_env_init = EnvBasedCredentials()
 
     # Starts YugabyteDB node.
     def start(self):
@@ -248,8 +246,7 @@ class ControlScript(object):
             Output.log_error_and_exit(
                 "{} is not running. Cannot connect to YCQL.".format(SCRIPT_NAME))
         ycql_proxy = YcqlProxy(ip=self.advertise_ip(),
-                        port=self.configs.saved_data.get("ycql_port"),
-                        enforced_auth=self.configs.saved_data.get("ycql_auth"))
+                        port=self.configs.saved_data.get("ycql_port"))
         ycql_proxy.connect()
 
     # Creates demo database and starts an interactive shell into it. Destroys the sample database
@@ -295,18 +292,14 @@ class ControlScript(object):
             "Initializing {} demo database. This may take up to a minute...".format(demo_db))
         # Create demo database.
         Output.log("Creating database {}...".format(db_name))
-        _, err = ysql_proxy.create_db(db_name)
-        if err:
-            Output.log_error_and_exit("Failed to create {} database: {}".format(demo_db, err))
+        ysql_proxy.create_db(db_name)
 
         # Populate demo database.
         Output.log("Populating {} with sample data...".format(db_name))
         files = []
         for name in Configs.get_demo_info()[demo_db]["files"]:
             files.append(os.path.join(YUGABYTE_DIR, "share", name))
-        _, err = ysql_proxy.load_files(files, db=db_name)
-        if err:
-            Output.log_error_and_exit("Failed to populate data to {}: {}".format(demo_db, err))
+        ysql_proxy.load_files(files, db=db_name)
 
         msg = "Successfully loaded sample database!"
         Output.print_and_log(msg)
@@ -344,9 +337,7 @@ class ControlScript(object):
         db_name = DEMO_DB_PREFIX + demo_db
         ysql_proxy = YsqlProxy(self.advertise_ip(), self.configs.saved_data.get("ysql_port"))
         if ysql_proxy.db_exists(db_name):
-            _, err = ysql_proxy.drop_db(db_name)
-            if err:
-                Output.log_error_and_exit("Failed to drop database {}: {}".format(demo_db, err))
+            ysql_proxy.drop_db(db_name)
         msg = "Deleted demo database {}.".format(demo_db)
         Output.print_and_log(msg)
 
@@ -474,14 +465,6 @@ class ControlScript(object):
                 "--start_redis_proxy=false",
             ]
 
-        # Add ysql_enable_auth flag to enforce authentication for YSQL
-        if self.configs.saved_data.get("ysql_auth"):
-            yb_tserver_cmd.extend(["--ysql_enable_auth=true"])
-
-        # Add use_cassandra_authentication flag to enforce authentication for YCQL
-        if self.configs.saved_data.get("ycql_auth"):
-            yb_tserver_cmd.extend(["--use_cassandra_authentication=true"])
-
         if self.configs.saved_data.get("tserver_flags"):
             yb_tserver_cmd.extend(
                 ["--{}".format(flag) for flag in \
@@ -546,19 +529,14 @@ class ControlScript(object):
             # Add enforced authentication flags on first installation
             # Add ysql_enable_auth, if YSQL_PASSWORD env variable exists
             # Add use_cassandra_authentication, if YCQL_USER or YCQL_PASSWORD env variables exists
-            if is_first_install:
-                # Add ysql_enable_auth flag to enforce authentication for YSQL
-                if self.setup_env_init.is_exists('YSQL_PASSWORD'):
-                    yb_tserver_cmd.extend(["--ysql_enable_auth=true"])
-                    self.configs.saved_data["ysql_auth"] = True
-                    self.configs.save_configs()
+            # Add ysql_enable_auth flag to enforce authentication for YSQL
+            if self.setup_env_init.is_exists('YSQL_PASSWORD'):
+                yb_tserver_cmd.extend(["--ysql_enable_auth=true"])
 
-                # Add use_cassandra_authentication flag to enforce authentication for YCQL
-                if self.setup_env_init.is_exists('YCQL_USER') or \
-                        self.setup_env_init.is_exists('YCQL_PASSWORD'):
-                    yb_tserver_cmd.extend(["--use_cassandra_authentication=true"])
-                    self.configs.saved_data["ycql_auth"] = True
-                    self.configs.save_configs()
+            # Add use_cassandra_authentication flag to enforce authentication for YCQL
+            if self.setup_env_init.is_exists('YCQL_USER') or \
+                    self.setup_env_init.is_exists('YCQL_PASSWORD'):
+                yb_tserver_cmd.extend(["--use_cassandra_authentication=true"])
 
             # Start or initialize yb-master and yb-tserver.
             if is_first_run:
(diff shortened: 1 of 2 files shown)


baba230896:
@iSignal I've commented the default user deletion section as we've discussed.

bin/yugabyted
```diff
@@ -566,6 +584,28 @@ class ControlScript(object):
                 (_, was_started) = self.maybe_start_yw(is_first_run, is_first_install)
                 should_callhome = should_callhome or was_started
 
+            if is_first_install and not join_ip:
+                if self.get_failed_node_processes():
```
iSignal, line 588: can we move this functionality to a separate function like `process_init_env_vars` or similar. as we add init sql etc it will get longer.

bin/yugabyted
```diff
@@ -566,6 +584,28 @@ class ControlScript(object):
...
+                    Output.log_error_and_exit(
+                        "{} is not running.".format(SCRIPT_NAME))
+
+                if is_valid(ysql_user) or is_valid(ysql_password) or is_valid(ysql_db):
+                    ysql_proxy = YsqlProxy(ip=self.advertise_ip(),
+                                    port=self.configs.saved_data.get("ysql_port"),
+                                    use_env_credentials=False)
+                    if retry_op(ysql_proxy.is_ysql_up):
```
iSignal, line 596: in the else case, let's print a log message so we can track it

bin/yugabyted
```diff
@@ -566,6 +584,28 @@ class ControlScript(object):
...
+                        Output.log("Setting up custom credentials for YSQL...")
+                        setup_credentials(ysql_proxy, ysql_user, ysql_password, ysql_db)
+
+                if is_valid(ycql_user) or is_valid(ycql_password) \
+                        or is_valid(ycql_keyspace):
+                    ycql_proxy = YcqlProxy(ip=self.advertise_ip(),
+                                    port=self.configs.saved_data.get("ycql_port"),
+                                    use_env_credentials=False)
+                    if retry_op(ycql_proxy.is_ycql_up):
```
iSignal, line 605: log msg in the else case

bin/yugabyted
```diff
@@ -1769,52 +1821,204 @@ class YBAdminProxy(object):
...
+            env={"PGUSER": self.username, "PGPASSWORD": self.password, "PGDATABASE": self.db})
         out, err = p.communicate()
         return out.strip() == db
 
     # Creates specified db.
     def create_db(self, db):
-        cmd = [self.path, "-p", self.port, "-c", "create database {}".format(db)]
-        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        cmd = [self.path, "-h", self.ip, "-p", self.port, "-c", "create database {}".format(db)]
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
```
iSignal, line 1854: make a utility function that does the subprocess.Popen with the env vars so create_db looks simple like `run_ysql_command("create database. ..")`

bin/yugabyted
```diff
@@ -1769,52 +1821,204 @@ class YBAdminProxy(object):
...
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, err = p.communicate()
+        return out.strip() == ycql_username
+
+    # Create specified superuser
+    def create_user(self, ycql_username, ycql_password):
+        cmd = [self.path, self.ip, self.port, "-u", self.username, "-p", self.password, "-e",
+                "CREATE ROLE {} WITH PASSWORD = '{}' AND LOGIN = true AND SUPERUSER = true;".
+                format(ycql_username, ycql_password)]
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
iSignal, line 1972: like the SQL case, it would be good to have a utility method to encapsulate running a cql command.

bin/yugabyted
```diff
@@ -2283,5 +2487,151 @@ def retry_op(func, timeout=180):
...
+        #_, err = proxy_class.delete_user(user, password if is_valid(password) else user)
+        #if err:
+        #   Output.log_error_and_exit("Failed to delete default user: {}".format(err))
+        
+        Output.log("{} user created successfully".format(user))
+
+    Output.log("Custom credentials configured successfully")
+
+
+def get_env_credentials(yb_api_client_program, use_env_credentials=True):
```
iSignal, line 2574: minor: the use_env_credentials=True parameter could be better named as get_default_credentials=False

iSignal (changes requested):
Looks good overall. Minor style and organization comments inline.

Also, please add a Test plan to the summary.

bin/yugabyted
baba230896: If the retry fails, then it exits the script with the error and logs all the events in the log file. We don't need to provide explicitly else condition here.

bin/yugabyted
baba230896: If the retry fails, then it exits the script with the error and logs all the events in the log file. We don't need to provide explicitly else condition here.

baba230896:
@iSignal I've resolved all the review comments and also added the mechanism to store the authentication flags in `yugabyted.conf`.
Please take a look at this https://github.com/yugabyte/yugabyte-db/pull/5499#issue-473890344.

bin/yugabyted
```diff
@@ -526,6 +543,23 @@ class ControlScript(object):
...
                 rmcontents(data_dir, exclude_names=[pid_file_name])
 
+            # Add enforced authentication flags on first installation
+            # Add ysql_enable_auth, if YSQL_PASSWORD env variable exists
+            # Add use_cassandra_authentication, if YCQL_USER or YCQL_PASSWORD env variables exists
+            if is_first_install:
+                # Add ysql_enable_auth flag to enforce authentication for YSQL
+                if self.setup_env_init.is_exists('YSQL_PASSWORD'):
+                    yb_tserver_cmd.extend(["--ysql_enable_auth=true"])
+                    self.configs.saved_data["ysql_auth"] = True
```
iSignal, line 553: can we instead add these to the cmd line flag tserver_flags, (which should automatically be persisted?)

bin/yugabyted
```diff
@@ -2283,5 +2481,140 @@ def retry_op(func, timeout=180):
...
+    def __init__(self):
+        self.ysql_user = os.environ.get('YSQL_USER')
+        self.ysql_password = os.environ.get('YSQL_PASSWORD')
+        self.ysql_db = os.environ.get('YSQL_DB')
+        self.ycql_user = os.environ.get('YCQL_USER')
+        self.ycql_password = os.environ.get('YCQL_PASSWORD')
+        self.ycql_keyspace = os.environ.get('YCQL_KEYSPACE')
+
+    def is_valid(self, var_name):
+        return True if var_name is not None and var_name else False
```
iSignal, line 2495: you can just say `return var_name` here instead of this boolean expression

bin/yugabyted
```diff
@@ -2283,5 +2481,140 @@ def retry_op(func, timeout=180):
...
+
+    def get_ysql_user(self):
+        return self.ysql_user if self.is_valid(self.ysql_user) else DEFAULT_YSQL_USER
```
iSignal, line 2498: can we just say `return self.ysql_user or DEFAULT_YSQL_USER`

bin/yugabyted
```diff
@@ -2283,5 +2481,140 @@ def retry_op(func, timeout=180):
...
+            _, err = proxy_class.create_db(self.get_ysql_db())
+            if err:
+                Output.log_error_and_exit("Failed to create {} database: \
+                    {}".format(self.get_ysql_db(), err))
+
+            Output.log("{} database created successfully".format(self.get_ysql_db()))
+
+        # Update password for default user
+        if self.get_ysql_user() == DEFAULT_YSQL_USER:
+            _, err = proxy_class.update_password(self.get_ysql_password())
```
iSignal, line 2548: i would make these methods throw exceptions of some sort and avoid the detailed logging before and after each call.

bin/yugabyted
```diff
@@ -2283,5 +2481,140 @@ def retry_op(func, timeout=180):
...
+            return self.get_ysql_user(), self.get_ysql_password(), self.get_ysql_db()
+
+    def get_ycql_credentials(self, get_default_credentials=False):
+        if get_default_credentials:
+            return DEFAULT_YCQL_USER, DEFAULT_YCQL_PASSWORD, DEFAULT_YCQL_KEYSPACE
+        else:
+            return self.get_ycql_user(), self.get_ycql_password(), self.get_ycql_keyspace()
+
+    def is_exists(self, var_to_check):
+        return self.is_valid(os.environ.get(var_to_check))
```
iSignal, line 2532: the pythonic way to do this is `return var_to_check in os.environ`

iSignal (changes requested):
Some more style changes.

bin/yugabyted
```diff
@@ -2283,5 +2481,140 @@ def retry_op(func, timeout=180):
     raise RuntimeError("Failed after retrying operation for {} secs.".format(
         now - start_time))
 
+
+class SetupEnvBasedCredentials(object):
```
iSignal, line 2485: minor: rename this as `EnvBasedCredentials`

baba230896: okay. will do

bin/yugabyted
baba230896: The motto of `is_exists()` is to verify the argument variable is exist or not and non-empty. And return a boolean value.
The suggested one only checks for the variable is exists or not.

bin/yugabyted
baba230896: The motto of `is_valid` is to verify the argument variable is `non-empty` and `not None`. And return a boolean value.
The suggested one returns the variable, but we need boolean value.

1. If a variable doesn't exist then `os.environ.get('VAR')` returns None.
2. If a variable is empty like 

```
export VAR=
os.environ.get('VAR')
```
then `VAR` value will be ''. And this can create some fail scenario.

**Scenario:**
If users set `export YSQL_PASSWORD=` then, they'll believe they're setting an empty password. In reality, because of the variable, ysqlsh works in enforced authentication mode, and ysqlsh doesn't support an empty password in enforced authentication mode.

Same case in other env variables.

bin/yugabyted
baba230896: Yeah, I thought the same, although there is a scenario in which if the user spins up the YB server from the stopped state with some manipulation in the `tserver_flags` option then, this event override the flags in `yugabyted.conf` and spin up the server with the latest flags.
With the current implementation, they can't change the authentication flags with `yugabyted` options. If in case they want to change then, they can change manually in `yugabyted.conf`. This implementation reduces the chances of unintentional removal of authentication flags.

If still, we want to persist the flags in `tserver_flags` then we can remove these code snippets in the current implementation.
1. Lines: 477-483
```py
        # Add ysql_enable_auth flag to enforce authentication for YSQL
        if self.configs.saved_data.get("ysql_auth"):
            yb_tserver_cmd.extend(["--ysql_enable_auth=true"])

        # Add use_cassandra_authentication flag to enforce authentication for YCQL
        if self.configs.saved_data.get("ycql_auth"):
            yb_tserver_cmd.extend(["--use_cassandra_authentication=true"])
```
2. Lines: 1396-1397
```py
            "ysql_auth": DEFAULT_YSQL_AUTH,
            "ycql_auth": DEFAULT_YCQL_AUTH,
```
3. Lines: 97-98
```py
DEFAULT_YSQL_AUTH = False
DEFAULT_YCQL_AUTH = False
```

bin/yugabyted
baba230896: Another way to do this 
```py
def is_exists(var_name1):
    return True if var_name1 in os.environ and os.environ.get(var_name1) else False
```

bin/yugabyted
baba230896: I've enabled the `log_cmd` for most of the commands except `create_user` and `update_password` because the `log_cmd` prints the whole command so it'll print the password for both the commands in the logs file.

bin/yugabyted
baba230896: In new commit, `is_valid()` has removed.

baba230896:
@iSignal I've updated the suggested changes. Please take a look at it.

iSignal (approved):
Looks good! Thanks for being patient with all the iterations!

Issue #5637, opened 2020-09-10
[docs] Version dropdown list should display version along with "latest" and "stable"


stevebang:
Rather than display "latest" or "stable" to the doc users, we should display something more understandable. Maybe this:
- v2.3 (latest)
- v2.2 (stable)

stevebang:
@chirag-yb brought this up again. I mentioned I'd filed this issue.

PR #5688, opened 2020-09-15
[Docs][#5637] Sidebar version indicator list displays the version number

yb-andrew:
When on 'latest' or 'stable' version the textbox does not display what version number when the dropdown in minimized. Show the version number along with the version text.


stevebang (approved):
Looks good -- thanks

Issue #5606, opened 2020-09-06
[Grafana_Dashboard]Listing all master and tserver nodes in both panels


ricemaster:
Hi guys,

I don't know if it's OK to open an issue here for [YugabyteDB Dashboard for Grafana](https://grafana.com/grafana/dashboards/12620). I'm sorry for all the inconvenience.

I'm using Revision 2 of YugabyteDB Dashboard in Grafana v7.1.5 with Prometheus v2.20.1 on CentOS7. 

It's a 6-node cluster, 3 master nodes and 3 tserver nodes. I deployed the cluster following the instruction in [Manual deployment](https://docs.yugabyte.com/latest/deploy/manual-deployment/). 

My `prometheus.yml` is like below:
```
global:
  scrape_interval:     5s # Set the scrape interval to every 5 seconds. Default is every 1 minute.
  evaluation_interval: 5s # Evaluate rules every 5 seconds. The default is every 1 minute.
  # scrape_timeout is set to the global default (10s).

# YugaByte DB configuration to scrape Prometheus time-series metrics
scrape_configs:
  - job_name: 'yugabytedb'
    metrics_path: /prometheus-metrics

    static_configs:
      - targets: ['10.211.55.101:7000', '10.211.55.102:7000', '10.211.55.103:7000']
        labels:
          group: 'yb-master'

      - targets: ['10.211.55.104:9000', '10.211.55.105:9000', '10.211.55.106:9000']
        labels:
          group: 'yb-tserver'
```

And in the YugabyteDB Dashboard, all 6 nodes are shown in both `Master Node Status` panal and `Tserver Node Status` panal.
<img width="1719" alt="dashboard" src="https://user-images.githubusercontent.com/10449002/92318197-5b141180-f03b-11ea-976e-7c777346bb87.png">

I don't know it's an issue, or I did something wrong. It would be super apprieciated if someone could help me get out of this.

kmuthukk:
cc: @bhavin192 ^^ for help with respect to the grafana question.

@ricemaster -- for VM or baremetal deployment, you can run yb-master processes on the same nodes as yb-tserver processes. The yb-master processes are not very resource intensive. For example, if you have 6-similar nodes, you can run yb-tserver on all 6-nodes and yb-master on just 3 of them. [The number of yb-master processes should be same as the replication factor.]

ricemaster:
@kmuthukk Got it! Thank you for the advice. That's really helpful.

[...]

PR #5707, opened 2020-09-17
Combine the platform and community Grafana dashboards

bhavin192:
- Uses `export_type` instead of `group` label.
- Changes Tserver to TServer.
- Fixes an issue with node status panels where Master nodes were appearing in the TServer panel and vise versa.

The diff of JSON is not understandable because it's generated by Grafana.

### Scenarios tested

-   Deployed YugabyteDB on Kubernetes using the Helm chart.
-   Deployed YugabyteDB on AWS using Terraform.

The Grafana dashboard shows the node status and other panels correctly.

This is how it looks with Terraform deployment where both YB-TServer and YB-Master are running on all the nodes.

Screenshot_2020-09-17 YugabyteDB - Grafana

Ref: #5606


iSignal (approved):
Looks good. I ran a workload against a Platform cluster and verified that the RPCs showed up. 

One issue that I saw (which shouldn't block this change) is that when nodes are removed from a cluster, they still show up in the master/tserver status because of course, there's still prometheus labels corresponding to those nodes. We may need to either make the master/tserver status section more accurate in the long run or just remove it altogether.

bhavin192:
Right, basically those nodes show up as the time series data for them is present in our selected time frame.

Issue #5736, opened 2020-09-18
[docs] "Go to the latest (v2.2) version" appears on v2.2 doc pages


stevebang:
When "v2.2 (stable)" is selected in the version dropdown list, the top of all documentation pages include the following message:

**Attention**
This page documents an earlier version. Go to the latest (v2.2)version.

This is the latest stable version, so the message is not applicable. Also, I noticed the missing space after "(v2.2)" in the  "Go to" message.

stevebang:
@chirag-yb Before the "stable" release series was added, all previous versions went to the "latest" version of the docs, which is now "v2.3 (latest)".  
- I'm think that clicking on the "Go to the latest (v2.2) version." link should actually still go to the latest release series `2.3`.
- Should the link say: "Go to the v2.3 (latest) release."?  That matches the usage in the dropdown list.
   - I don't think we need to go to the most recent "stable" release because when users end up on an earlier version, they usually want to get to the latest (or at least that's the content we want them to be looking at).

PR #5747, opened 2020-09-21
[Docs][#5736] Fix version listed when on page that is not the latest

yb-andrew:
The version listed should be "Go to the latest version (2.3)" instead of 2.2


stevebang (approved):
Looks good.

Issue #5721, opened 2020-09-18
Remove dependency of env variables to toggle the authentication flags on yugabyted's restart


baba230896:
Environment variables are necessitated with every restart to keep the cluster in enforced authentication mode. It would be better if yugabyted can store the flags in the conf file and use the conf file at the time of restart instead of depending on the environment variables to set the authentication flags.

PR #5896, opened 2020-10-01
[yugabyted] Added flags to toggle Authentication mode

baba230896:
### Summary:

- Added command-line flags to toggle the authentication mode.
- Added functionality to store the authentication flags state in `yugabyted.conf`.
- Remove the environment variables dependency to set the authentication flags with every execution.
- The environment variable can change the `yugabyted` configuration only in the first run.
- The environment variables have higher priority than command-line flags.

### Test Plan:

1. Restart `yugabyted` without flags & environment variables where first-run have environment variables.
**_Result:_** `yugabyted` spin up with the `true` authentication flags as per the `yugabyted.conf` because first-run have those environment variables which can enable the authentication mode. 
Use the following commands to replicate the scenario:
```sh
export YSQL_USER=admin
export YSQL_PASSWORD=password
export YSQL_DB=admindb
export YCQL_USER=cqladmin
export YCQL_PASSWORD=cqlpassword
export YCQL_KEYSPACE=cqlks
bin/yugabyted start --daemon=true --ui=false
# Edited
unset YSQL_USER YSQL_PASSWORD YSQL_DB YCQL_USER YCQL_PASSWORD YCQL_KEYSPACE
bin/yugabyted stop
bin/yugabyted start --daemon=true --ui=false
```

2. Restart `yugabyted` with `false` authentication flags where first-run have `true` authentication flags and its vice-versa.
**_Result:_** `yugabyted` spin up with authentication flags which provided in the command and override the conf because flags exist in the restart command.
Use the following commands to replicate the scenario:
```sh
bin/yugabyted start --daemon=true --ui=false --ysql_enable_auth=true --use_cassandra_authentication=true
bin/yugabyted stop
bin/yugabyted start --daemon=true --ui=false --ysql_enable_auth=false --use_cassandra_authentication=false
```
```sh
bin/yugabyted start --daemon=true --ui=false --ysql_enable_auth=false --use_cassandra_authentication=false
bin/yugabyted stop
bin/yugabyted start --daemon=true --ui=false --ysql_enable_auth=true --use_cassandra_authentication=true
```

3. Start `yugabyted` with `true` authentication flags and without environment variables.
**_Result:_** Start the `yugabyted` with default credentials and enforced authentication mode.
Use the following commands to replicate the scenario:
```sh
bin/yugabyted start --daemon=true --ui=false --ysql_enable_auth=true --use_cassandra_authentication=true
```

4. Start `yugabyted` with environment variables and `false` authentication flags.
**_Result:_** `yugabyted` spin up with enforced authentication mode because environment variables have higher priority than command-line flags.
Use the following commands to replicate the scenario:
```sh
export YSQL_USER=admin
export YSQL_PASSWORD=password
export YSQL_DB=admindb
export YCQL_USER=cqladmin
export YCQL_PASSWORD=cqlpassword
export YCQL_KEYSPACE=cqlks
bin/yugabyted start --daemon=true --ui=false --ysql_enable_auth=false --use_cassandra_authentication=false
```

### Fixes:
https://github.com/yugabyte/yugabyte-db/issues/5721


bin/yugabyted
```diff
@@ -545,17 +545,29 @@ class ControlScript(object):
                     " Removing...").format(data_dir_files, data_dir))
                 rmcontents(data_dir, exclude_names=[pid_file_name])
 
+            # Add enforced authentication flags on the basis of command-line flag
+            if self.configs.saved_data.get("ysql_enable_auth"):
```
iSignal, line 549: we should move this to outside this loop, otherwise we will keep adding to yb_tserver_cmd each time this loop runs. For ex: we could do this in start_processes

bin/yugabyted
```diff
@@ -545,17 +545,29 @@ class ControlScript(object):
...
-            if self.setup_env_init.is_exists('YSQL_PASSWORD'):
-                yb_tserver_cmd.extend(["--ysql_enable_auth=true"])
-
-            # Add use_cassandra_authentication flag to enforce authentication for YCQL
-            if self.setup_env_init.is_exists('YCQL_USER') or \
-                    self.setup_env_init.is_exists('YCQL_PASSWORD'):
-                yb_tserver_cmd.extend(["--use_cassandra_authentication=true"])
+            if is_first_install:
+                if self.setup_env_init.is_exists('YSQL_PASSWORD'):
+                    yb_tserver_cmd.extend(["--ysql_enable_auth=true"])
```
iSignal, line 561: can we do this as part of the first_install_init

iSignal: also, can we rename first_install_init to first_install_init_auth

iSignal: another possibility is to add this code to the Configs.parse_config_files - when we load config files, we could convert this option to cause an append to tserver_gflags

bin/yugabyted
```diff
@@ -1288,6 +1305,12 @@ class ControlScript(object):
             cur_parser.add_argument(
                 "--ui", choices=BOOL_CHOICES, default="false", metavar="BOOL",
                 help="Toggle enabling or disabling webserver UI. Default false.")
+            cur_parser.add_argument(
+                "--ysql_enable_auth", choices=BOOL_CHOICES, metavar="BOOL",
```
iSignal, line 1309: add an argparse.SUPPRESS so that it does not show up in the default help list.

iSignal (changes requested):

bin/yugabyted
baba230896: okay. will update.

bin/yugabyted
baba230896: > can we do this as part of the first_install_init

Ahh, that won't be the right approach because we should run this on all nodes but only on the first-run whereas `first_install_init` runs only on the first node and first-run.

baba230896: > also, can we rename first_install_init to first_install_init_auth

Okay. will update the `first_install_init` to `first_install_init_auth`.

baba230896: > another possibility is to add this code to the Configs.parse_config_files - when we load config files, we could convert this option to cause an append to tserver_gflags

I've checked after your suggestion. I guess the best place is `def validate_and_set_configs` (at 1111) to put this snippet because here we're doing verification and overriding the configuration.

bin/yugabyted
baba230896: Yeah, we can add this snippet at 491 just after, where extracting the extra YB tserver gflags from `tserver_flags`.

baba230896:
@iSignal 

I've resolved all the review comments. Please take a look at it.

docs/content/latest/reference/configuration/yugabyted.md
```diff
@@ -169,6 +169,19 @@ Enable or disable the "call home" feature that sends analytics data to Yugabyte.
...
+Enable or disable YSQL Authentication. Default is `false`.
+If the `YSQL_PASSWORD` environment variable exists then, authentication mode is automatically changed to enforced.
+
+##### --use_cassandra_authentication *bool*
+
+Enable or disable YCQL Authentication. Default is `false`.
+If the `YCQL_USER` or `YCQL_PASSWORD` environment variables exist then, authentication mode is automatically changed to enforced.
+
+**Note**
+- Environment variables have more priority than the command-line flag.
```
iSignal, line 183:
```suggestion
- The corresponding environment variables have higher priority than the command-line flags.
```

docs/content/latest/reference/configuration/yugabyted.md
```diff
@@ -433,7 +446,7 @@ Combinations of environment variables and their uses.
 
 **Note**
 - In the case of multi-node deployment, all nodes should have similar environment variables. 
-
+- Environment variables can change the configuration only on the first run.
```
iSignal, line 449:
```suggestion
- Changing the values of the environment variables after the first run has no effect.
```

iSignal (approved):
Minor docs comments, please change and merge.

Issue #5720, opened 2020-09-18
Messenger::ScheduleOnReactor in messenger.cc has a small flaw


wumingyuan:
```C++
  for (const auto& r : reactors_) {
    if (r->IsCurrentThread()) {
      chosen = r.get();
      break; // +++++++++ should be added
    }
  }
```

bmatican:
@wumingyuan that makes sense, would you mind just putting a PR to fix this?

wumingyuan:
[https://github.com/yugabyte/yugabyte-db/pull/5996](https://github.com/yugabyte/yugabyte-db/pull/5996)

PR #5996, opened 2020-10-09
Update messenger.cc[#5720]

wumingyuan:
There has a small flaw.
Since only one reactor is required, it should be quit immediately to save time.


tedyu:
@wumingyuan 
Can you reference the issue number in the title, [#XYZ], where XYZ is the issue number ?

Please update the description with why the short circuit can be made.

thanks

spolitov (approved):

PR #5997, opened 2020-10-09
Update yb-admin_client.cc

wumingyuan:
Here has a bug.


OlegLoginov:
Please add more details in the bug description. (E.g.: What's wrong? How to reproduce? What's (how's) affected by the issue? Etc.)

tedyu:
@wumingyuan 
Please take a look at src/yb/client/client-test.cc and see if you can come up with a test which reveals the defect.

This would help us guard against future regression.

Thanks

wumingyuan:
The original leader is removed, if the new leader is the same as the original leader, the removal operation fails.

PR #6106, opened 2020-10-20
Pure technical change. New structure for 'The SQL language' under YSQL

bllewell:
The changes made by the `git push origin` command that caused this PR are entirely technical. They achieve the effect of moving the accounts of the eighty-or-so SQL statements that, for a very long time, have been here:

```
/latest/api/ysql/commands/
```

to here:

```
/latest/api/ysql/the-sql-language/statements/
```

The purpose is to allow a new section on  "The Common Table Expression (WITH clause)":

```
/latest/api/ysql/the-sql-language/with-clause/
```

to be sited under "YugabyteDB's SQL language" as a sibling to "Categorized list of SQL statements". Other future  siblings (for example, "How to read execution plans") will now have a natural home. As of this PR, the new "WITH clause" `_index.md` has just a sketch of the functionality and a note:

>  Under construction. A future version of this section will explain everything and provide working code examples.

Each of the eighty-or-so SQL statement accouns has at least one syntax diagram. Some have more than one. Each diagram is defined by a pair of source files in the `syntax_resources`. And these must be located on a directory with a path down from there that exactly matches the path to the referring content `.md` file down from the `ysql` directory—in this case, this:

```
/latest/api/ysql/syntax_resources/the-sql-language/statements
```

The result is that about 240 files were moved. Because of how relative links work, this move didn't need any link changes within the new `/latest/api/ysql/the-sql-language/` tree. (However, I did find, and correct, some historically wrongly spelled links.) I also ensured that every moved content file had a properly formed path for its former location defined by the `aliases` page property in the frontmatter so that links from external content (like blog posts) will still work.

However, to avoid confusion for future doc writing efforts, I searched out all links to the old locations of the moved files in the source `.md` files within the whole of the `/latest/` tree and updated these explicitly. This resulted in 7 modified files within the `/latest/api/ysql/` tree and a further 15 other modified files elsewhere in the `\latest\` corpus.


jaki:
If this goes through, `docs/README.md` needs some path changes.

m-iancu:
@bllewell Did we set aliases for the files we moved (or confirm they have them already)? Otherwise this might break the old URLs.

stevebang (approved):
Looks good.

rkarthik007:
Hi @stevebang @bllewell - did we check to make sure the aliases are redirecting the old url's correctly?

bllewell:
Please re-read the essay that I wrote to describe the PR. Look for this:

> The result is that about 240 files were moved. Because of how relative links work, this move didn't need any link changes within the new /latest/api/ysql/the-sql-language/ tree. (However, I did find, and correct, some historically wrongly spelled links.) I also ensured that every moved content file had a properly formed path for its former location defined by the aliases page property in the frontmatter so that links from external content (like blog posts) will still work.
>
> However, to avoid confusion for future doc writing efforts, I searched out all links to the old locations of the moved files in the source .md files within the whole of the /latest/ tree and updated these explicitly. This resulted in 7 modified files within the /latest/api/ysql/ tree and a further 15 other modified files elsewhere in the \latest\ corpus.

Does this answer all of your questions?

m-iancu:
@bllewell From my side 
> so that links from external content (like blog posts) will still work.
answers it. If the old URLs still work than we are good (for my question at least). 
I didn't realize we have aliases for all/most files in there (so figured that'd be a lot of work), but I did not check recently.

jaki:
`docs/README.md` is now referencing incorrect paths.  Maybe it wasn't clear, but I was hoping the changes would be made here.

Issue #6041, opened 2020-10-14
[YSQL] Add a tserver flag to allow limiting the minimum value for sequence CACHE


m-iancu:
In a distributed context _sequences_ add more relative overhead as getting one (or more) new value(s) requires global locking.
Therefore, for practical performance we recommend using a higher `CACHE` value (or a type with a unique value generator like `UUID`, instead).

However, in some cases, large schemas are exported from other systems (e.g. vanilla Postgres) many sequences with `CACHE` explicitly set to low values (e.g. `1`). This is because, even when the cache option was not specified on sequence creation, its value (the vanilla Postgres default `1`) will be made explicit when exporting it (therefore being specified in the SQL dump file for each sequence).

To improve usability for such migration use-cases, we should add a tserver flag `ysql_sequence_cache_minval` (default `100`) that will override the chosen cache value for a sequence (and log a message to the user), if it is lower than the set flag.

m-iancu:
Fixed by 6d29fc6.

PR #6182, opened 2020-10-28
[Doc][YSQL][#6041] Adding documentation related to TServer flag "ysql_sequence_cache_minval"

emhna:
The documentation is related to this Github issue: #6041.
The new tserver flag sets the cache size of sequence objects in YSQL.
This change is not included in v2.2.

3 pages have been updated:

1. Path: /latest/reference/configuration/yb-tserver/#ysql-sequence-cache-minval
New section is added for --ysql_sequence_cache_minval under YSQL yb-tserver flag page.

2. Path: /latest/api/ysql/the-sql-language/statements/ddl_create_sequence/#cache-cache
Added more description under create sequence cache clause section.

3. Path: /latest/api/ysql/the-sql-language/statements/ddl_alter_sequence/#cache-cache
Added more description under alter sequence cache clause section.

flag
create seq
alter seq

### DIFF
--- a/docs/content/latest/reference/configuration/yb-tserver.md
+++ b/docs/content/latest/reference/configuration/yb-tserver.md
@@ -485,6 +485,16 @@ For details on how online index backfill works, see the [Online Index Backfill](
 
 Default: `true`
 
+##### --ysql_sequence_cache_minval
+
+Specify the minimum number of sequence values to cache in the client for every sequence object.
+
+To turn off the default size of cache flag, set the flag to `0`.
+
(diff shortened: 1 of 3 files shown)


stevebang (approved):
Looks good after the edits!

docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -485,6 +485,16 @@ For details on how online index backfill works, see the [Online Index Backfill](
 
 Default: `true`
 
+##### --ysql_sequence_cache_minval
+
+Specify how many numbers from the sequence to cache in the client for every sequence object.
```
m-iancu, line 490: Should this say something like `Specify the minimum number of sequence values to cache ...`?

emhna: Makes sense since it is not always guaranteed this value will be used as the cache size if cache clause has greater value. 
Made change as specified.

m-iancu (approved):
LGTM!

PR #6197, opened 2020-10-28
[WIP PATCH]

deeps1991:
This patch explores exposing the YSQL API to create Tablespaces.
The user will specify placement options as part of this API.

These placement options will be stored in the pg_tablespace catalog
table and will be parsed by the YB Master to decide how to place
the tablets of this table.

If the table itself has placement information, then that will be used.
Otherwise placement info associated with its tablespace will be used.
If neither are present we will use cluster level placement policy.

There are sanity tests (that do pass) in this patch, but this is
still very much WIP and is being used to evaluate any better
design/code flows.


deeps1991:
Will clean up this patch a bit further and send it out on Phabricator (because now I can!)

Issue #3522, opened 2020-02-01
[CDC] SIGSEGV in ProcessRecord encountered 


ajcaldera1:
Encountered during b23 testing:
```
#0  0x00007f5123160b70 in tc_newarray () from /home/yugabyte/yb-software/yugabyte-2.0.11.0-b23-centos-x86_64/lib/yb-thirdparty/libtcmalloc.so.4
(gdb) bt
#0  0x00007f5123160b70 in tc_newarray () from /home/yugabyte/yb-software/yugabyte-2.0.11.0-b23-centos-x86_64/lib/yb-thirdparty/libtcmalloc.so.4
#1  0x00007f511ea160c9 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_mutate(unsigned long, unsigned long, char const*, unsigned long) ()
   from /home/yugabyte/yb-software/yugabyte-2.0.11.0-b23-centos-x86_64/linuxbrew/lib/libstdc++.so.6
#2  0x00007f511ea17603 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_append(char const*, unsigned long) ()
   from /home/yugabyte/yb-software/yugabyte-2.0.11.0-b23-centos-x86_64/linuxbrew/lib/libstdc++.so.6
#3  0x00007f5122ab960e in append (__n=<optimized out>, __s=0x7f50812d8200 "    @     0x7f512b2eb4dd  yb::tserver::enterprise::CDCPoller::HandlePoll()", this=0x7f50812d8be0)
    at /home/yugabyte/yb-software/yugabyte-2.0.11.0-b23-centos-x86_64/linuxbrew-xxxxxxxxxxxx/Cellar/gcc/5.5.0_4/include/c++/5.5.0/bits/basic_string.h:1105
#4  StringAppendV (dst=0x7f50812d8be0, format=0x7f512373cf78 <yb::(anonymous namespace)::kStackTraceEntryFormat> "    @ %*p  %s", ap=ap@entry=0x7f50812d8648) at ../../../../../src/yb/gutil/stringprintf.cc:54
#5  0x00007f5122ab98e4 in StringAppendF (dst=dst@entry=0x7f50812d8be0, format=format@entry=0x7f512373cf78 <yb::(anonymous namespace)::kStackTraceEntryFormat> "    @ %*p  %s") at ../../../../../src/yb/gutil/stringprintf.cc:111
[...]

PR #6221, opened 2020-10-30
[CDC][#3522]fix SIGSEGV in ProcessRecord encountered

jiaolongxue:
Summary:
CDCPoller callback thread  calls ProcessRecord while CDCWriteToTablet rpc callback thread  calls WriteCDCRecordDone. Both functions modify the records_ .


jiaolongxue:
signed cla

PR #6328, opened 2020-11-10
Direct Platform license requests to website form.

dotyjim-work:
Yugabyte support does not currently control the license issuing process,
so direct folks following along in the documentation to the form that
leads them down the right path.


yb-andrew (approved):
LG

PR #6381, opened 2020-11-16
[Docs] Update system config documentation

tylarb:
Remove note about epel-release
    
If a node doesn't have internet access, then installing epel-release
won't be helpful, as it only provides access to epel and no packages
from it.

---

Add clarity to the ulimits section
    
Note that ulimits should be applied permanently

---

Update docs to indicate THP must be enabled
    
TCMalloc heavily relies on transparent hugepages, so we should document
that they must be enabled.
    
[TCMalloc tuning reference](https://github.com/google/tcmalloc/blob/master/docs/tuning.md)

### DIFF
--- a/docs/content/latest/deploy/manual-deployment/system-config.md
+++ b/docs/content/latest/deploy/manual-deployment/system-config.md
@@ -161,3 +161,49 @@ In `/etc/systemd/user.conf` and `/etc/systemd/system.conf`, add at the end of fi
 
 Something similar may be needed for other distributions.
 {{< /note >}}
+
+
+## transparent hugepages
+
+Transparent hugepages should be enabled for optimal performance. By default, they are enabled.
+
+You can check with the following command:
+
+
+```sh
+$ cat /sys/kernel/mm/transparent_hugepage/enabled
+[always] madvise never
+```
+
+It is generally not necessary to adjust the kernel command line if the above value is correct.
+
+However, if the value is set to "madvise" or "never", you should modify your kernel command line to set transparent hugepages to "always".
+
+
+You should consult your operating system docs to determine the best way to modify a kernel command line argument for your operating system, but on RHEL or Centos 7 or 8, using grub2, the following steps are one solution:
+
+
+Append "transparent_hugepage=always" to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`
+
+```sh
+GRUB_CMDLINE_LINUX="rd.lvm.lv=rhel/root rd.lvm.lv=rhel/swap ... transparent_hugepage=always"
+```
+
+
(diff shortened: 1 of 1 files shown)


docs/content/latest/deploy/manual-deployment/system-config.md
```diff
@@ -166,3 +161,49 @@ In `/etc/systemd/user.conf` and `/etc/systemd/system.conf`, add at the end of fi
...
+$ cat /sys/kernel/mm/transparent_hugepage/enabled
+[always] madvise never
+```
+
+It is generally not necessary to adjust the kernel command line if the above value is correct.
+
+However, if the value is set to "madvise" or "never", you should modify your kernel command line to set transparent hugepages to "always".
+
+
+You should consult your operating system docs to determine the best way to modify a kernel command line argument for your operating system, but on RHEL or Centos 7 or 8, using grub2, the following steps are one solution:
```
dotyjim-work, line 183:
```suggestion
You should consult your operating system docs to determine the best way to modify a kernel command line argument for your operating system, but on RHEL or Centos [7](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/performance_tuning_guide/sect-red_hat_enterprise_linux-performance_tuning_guide-configuring_transparent_huge_pages) or [8](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/virtualization_tuning_and_optimization_guide/sect-virtualization_tuning_optimization_guide-memory-huge_pages), using grub2, the following steps are one solution:
```

dotyjim-work (approved):

docs/content/latest/deploy/manual-deployment/system-config.md
```diff
@@ -166,3 +161,49 @@ In `/etc/systemd/user.conf` and `/etc/systemd/system.conf`, add at the end of fi
...
+
+
+Addend "transparent_hugepage=always" to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`
```
kmuthukk, line 186: Did you mean Append (instead of Addend).

docs/content/latest/deploy/manual-deployment/system-config.md
tylarb: you actually linked the RHEL 6 docs for the second one Jim - I'll make the addition though. thanks!

tylarb: Actually... I can't find an appropriate doc for RHEL 8, so I'll remove the links entirely.

docs/content/latest/deploy/manual-deployment/system-config.md
tylarb: Yes thank you.

PR #6498, opened 2020-11-30
[docs] Update macos build requirements.

svarnau:
Bison and flex do not need to be installed, since they are in thirdparty
dependencies.

### DIFF
--- a/docs/content/latest/contribute/core-database/build-from-src-macos.md
+++ b/docs/content/latest/contribute/core-database/build-from-src-macos.md
@@ -60,8 +60,8 @@ First, install [Homebrew](https://brew.sh/), if you do not already have it. Home
 Install the following packages using Homebrew:
 
 ```sh
-brew install autoconf automake bash ccache cmake  coreutils gnu-tar icu4c libtool \
-             maven ninja pkg-config pstree wget zlib python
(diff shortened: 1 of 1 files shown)


docs/content/latest/contribute/core-database/build-from-src-macos.md
```diff
@@ -60,17 +60,8 @@ First, install [Homebrew](https://brew.sh/), if you do not already have it. Home
...
-             zlib python
-```
-
-An older version of `bison` is required to correctly compile the code. The following command installs the required `3.4.1` version of bison.
-
-```sh
-brew install \
-https://raw.githubusercontent.com/Homebrew/homebrew-core/ee89a1d59df03f495a85c15e253b60299082ab9d/Formula/bison.rb
+brew install autoconf automake bash ccache cmake  coreutils gnu-tar icu4c libtool \
+             maven ninja pkg-config pstree wget zlib python
```
mbautin, line 64: We certainly need to remove icu4c. I am also pretty sure we can remove zlib (we build it ourselves too as part of yugabyte-db-thirdparty).

mbautin (changes requested):

mbautin (approved):

PR #6576, opened 2020-12-08
arch: detail tserver-postgres connection for backfill

jaki:
Detail that YSQL needs to do a tserver to postgres connection for
backfilling a tablet.


rkarthik007 (approved):
LGTM!

PR #6593, opened 2020-12-09
Additional default firewall ports

wmaddux:
Adds additional default ports needed for firewall rules when setting up a cluster managed by YB Platform.

### DIFF
--- a/docs/content/latest/reference/configuration/default-ports.md
+++ b/docs/content/latest/reference/configuration/default-ports.md
@@ -64,7 +64,7 @@ YugabyteDB servers expose time-series performance metrics in the [Prometheus exp
 <target>/prometheus-metrics
 ```
 
-You can access the Prometheus server on port `9090`, and you can see the list of targets at the `http://<yugaware-ip>:9090/targets`. In particular, note port `93000` for node level metrics.
+You can access the Prometheus server on port `9090` of the Platform node, and you can see the list of targets at the `http://<yugaware-ip>:9090/targets`. In particular, note port `93000` for node level metrics.
 
 ### Servers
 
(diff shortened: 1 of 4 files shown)


dotyjim-work:
I'll loop back with you @wmaddux to grab the ways that we can change those added ports (to follow the convention of some of the `Flag (default)` convention).

Also I think we need to add the `9090` for the Prometheus UI ( but only on the platform host) and we need the `9300` for scraping some metrics from the nodes.

dotyjim-work (changes requested):
I'll loop back with you @wmaddux to grab the ways that we can change those added ports (to follow the convention of some of the Flag (default) convention).

Also I think we need to add the 9090 for the Prometheus UI ( but only on the platform host) and we need the 9300 for scraping some metrics from the nodes.

dotyjim-work:
```
commit b1d5fdbedd5b72c7735a74b7ec1e2c45575f2a80 (HEAD -> wmaddux-master)
Author: Jim Doty <jdoty@yugabyte.com>
Date:   Sun Jan 31 21:35:06 2021 -0700

    Add default Prometheus target information

diff --git a/docs/content/latest/reference/configuration/default-ports.md b/docs/content/latest/reference/configuration/default-ports.md
index 0db973f37..24327cd5d 100644
--- a/docs/content/latest/reference/configuration/default-ports.md
+++ b/docs/content/latest/reference/configuration/default-ports.md
@@ -48,13 +48,13 @@ Admin web server UI can be viewed at these addresses.
 ## Firewall Rules
 Along with the above, include the following common ports in firewall rules. 
 
-| Service     | Port
+| Service     | Port |
 | ------- | ------------------------- |
-| SSH    | `tcp:22` |
-| HTTP   | `tcp:80` |
-| HTTPS   | `tcp:443` |
-| HTTP (alternate)   | `tcp:8080` |
-| HTTP (Replicated)    | `tcp:8800` |
+| SSH    | 22 |
+| HTTP   | 80 |
+| HTTPS   | 443 |
+| HTTP (alternate)   | 8080 |
+| HTTP (Replicated)    | 8800 |
 
 ## Prometheus monitoring
 
@@ -64,7 +64,7 @@ YugabyteDB servers expose time-series performance metrics in the [Prometheus exp
 <target>/prometheus-metrics
 ```
 
-Following is the list of targets available.
+You can access the Prometheus server on port `9090`, and you can see the list of targets at the `http://<yugaware-ip>:9090/targets`. In particular, note port `93000` for node level metrics.
 
 ### Servers
```

dotyjim-work (approved):

PR #6688, opened 2020-12-17
test

jitendra-12113:


jitendra-12113:
Setting up the volumeCount for aws provider to 1 whose instanceType starting with c5/c4.

gaurav061:
Kindly review the changes, 
Change: For all AWS instances starting with c4 and c5, default volume-info field will be 1. This includes c5d. .. instances as well,
we have added a property ```volumeCount``` which holds the default value.

SergeyPotachev (changes requested):
Please add junit tests for changes in InstanceTypeController.java.

managed/src/main/java/com/yugabyte/yw/controllers/InstanceTypeController.java
```diff
@@ -57,6 +57,17 @@ public Result list(UUID customerUUID, UUID providerUUID) {
 
     try {
       instanceTypeList = InstanceType.findByProvider(provider);
+      // setting up the volumeCount to 1 for instanceType starting with c5/c4
+      for (InstanceType instanceType: instanceTypeList) {
+        if (provider.code.equals(aws.toString()) &&
+          ((instanceType.idKey.instanceTypeCode.startsWith("c5")) ||
```
Arnav15, line 63: This is incorrect. `c5d` instances are similar to i3 instances in that they use NVMe drives. So we want to use the value present in the details.

Arnav15 (changes requested):

managed/src/main/java/com/yugabyte/yw/controllers/InstanceTypeController.java
jitendra-12113: Thanks @Arnav15  for confirming this. Will do this only for c4/C5 instance type  which won't include c5d.
Will do necessary changes and update CR.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -608,7 +609,7 @@ export default class ClusterFields extends Component {
     if (volumeDetail) {
       const deviceInfo = {
         volumeSize: volumeDetail.volumeSizeGB,
-        numVolumes: volumesList.length,
+        numVolumes: volumeCount,
```
sshev, line 612: could you please update logic here as well https://github.com/yugabyte/yugabyte-db/blob/b6bc73a25aa519dd879409bb1e69235e42a822f8/managed/ui/src/redesign/universe/wizard/fields/DeviceInfoField/DeviceInfoField.tsx#L65

jitendra-12113:
Will raise another PR.

PR #6699, opened 2020-12-17
arch: improve tablet split doc style

jaki:
Improve style of `docdb-automatic-tablet-splitting.md`.

### DIFF
--- a/architecture/design/docdb-automatic-tablet-splitting.md
+++ b/architecture/design/docdb-automatic-tablet-splitting.md
@@ -5,7 +5,7 @@ Tracking GitHub Issue: https://github.com/yugabyte/yugabyte-db/issues/1004
 Automatic tablet splitting enables changing the number of tablets (which are splits of data) at runtime. There are a number of scenarios where this is useful:
 
 ### 1. Range Scans
-In use-cases that scan a range of data, the data is stored in the natural sort order (also known as *range-sharding*). In these usage patterns, it is often impossible to predict a good split boundary ahead of time. For example: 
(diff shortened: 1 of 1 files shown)


architecture/design/docdb-automatic-tablet-splitting.md
```diff
@@ -63,65 +69,65 @@ When leader tablet server receives `SplitTablet` RPC it adds a special Raft reco
...
 
 - Snapshots (created by `Tablet::CreateSnapshot`)
-  - The simplest solution is to clone snapshot as well using hard-link, but snapshots are immutable and once they will 
-  be re-distributed across different tservers it will increase space amplification. Snapshots are currently only used 
-  for backup and are quickly deleted. But it would be good to implement cutting of RocksDB without full compaction - 
-  could be done later. 
+  - The simplest solution is to clone snapshot as well using hard-link, but snapshots are immutable and once they will
+  be re-distributed across different tservers it will increase space amplification. Snapshots are currently only used
+  for backup and are quickly deleted. But it would be good to implement cutting of RocksDB without full compaction -
+  could be done later.
```
jaki, line 112: Had some trouble understanding this part.  Should it be

"The simplest solution is to clone snapshots as well using hard-link _because_ snapshots are immutable"?

ttyusupov: I think this part is outdated and the current state is:
"
Tablet data could also contain snapshots (created by Tablet::CreateSnapshot).
We don’t need to copy snapshots from the parent tablet to post-split tablets, but we might need to keep them for the parent tablet even after the split is completed.
"

ttyusupov (approved):

Issue #6486, opened 2020-11-25
[YSQL] Separate out metrics for BEGIN, COMMIT, ROLLBACK statements from the OTHER category


kmuthukk:
Today the top-level YSQL statement metrics at the TServer level are categorized as:

For example:
http://yb-tserver:13000/metrics

See output below. It'll be good to break out from the OtherStmts category, separate metrics for Begin/Commit/Rollback transaction statements.  [For PREPARE statements, the handling is a bit different. So will fork that into its own issue.]

```
  {
        "type": "server",
        "id": "yb.ysqlserver",
        "metrics": [
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_SelectStmt",
                "count": 205539578,
                "sum": 1239986847119
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_InsertStmt",
                "count": 116057923,
                "sum": 827644852030
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_DeleteStmt",
                "count": 0,
                "sum": 0
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_UpdateStmt",
                "count": 0,
                "sum": 0
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_OtherStmts",
                "count": 3450,
                "sum": 27716148
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_Transactions",
                "count": 321593501,
                "sum": 2067614691134
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_AggregatePushdowns",
                "count": 0,
                "sum": 0
            }
        ]
    }
```

VishnuK007:
I have a patch that I would like to submit for this issue. Can you please assign it to me if no one has started working on it yet?

rkarthik007:
Thanks for your interest @VishnuK007 🔥 !

m-iancu:
Fixed by e297413.

Issue #1125, opened 2019-04-04
[YSQL] Feature Support - Transaction Control Commands


nocaway:
The following transaction control commands are not yet supported.
- SAVEPOINT
- RELEASE SAVEPOINT
- RELEASE
- ROLLBACK
- PREPARE
- COMMIT PREPARED
- ROLLBACK PREPARED

andrew-s:
I'd like to see this, my particular use case is for automated migrations.

danports:
Would be nice to have this for pgloader, which uses `SAVEPOINT`:
```
Database error 0A000: SAVEPOINT <transaction> not supported yet
>        HINT: See https://github.com/YugaByte/yugabyte-db/issues/1125. Click '+' on the description to raise its priority
>        QUERY: savepoint pgloader;
```

ddorian:
`SAVEPOINT` is also used in Elixir ecto_sql for [database migrations](https://hexdocs.pm/ecto_sql/Ecto.Adapters.SQL.html#module-migrations).  Mentioned in [slack](https://yugabyte-db.slack.com/archives/CJN8V5ZKJ/p1593028191136400?thread_ts=1593018598.129700&cid=CJN8V5ZKJ).

marcusbooyah:
`SAVEPOINT` is also used by [Harbor](https://goharbor.io/)

davidkellis:
+1'ed due to the Sequel (http://sequel.jeremyevans.net/) library's use of savepoints

venkatramanm:
+1'ed SavePoint is crucial for my application.

skcc321:
+1 SAVEPOINT is widely used in RoR test suites (Rspec for instance). So right now I have no confidence in Yugabyte compatibility with my legacy application (postgresql).

cyucelen:
I also need this to use [txdb](https://github.com/DATA-DOG/go-txdb) for my tests.

dzirg44:
https://mikro-orm.io/ migrations

filipeaguiars:
+1'ed SavePoint is crucial for my application.

> 05-11:53:23.642 | main                           | INFO  |  Created new number range with id 1 for seqName NumberRangeAssignment             - (AbstractNumberRange.java:480)
Error: org.postgresql.util.PSQLException: ERROR: SAVEPOINT <transaction> not supported yet
  Hint: See https://github.com/YugaByte/yugabyte-db/issues/1125. Click '+' on the description to raise its priority
[...]

PR #6729, opened 2020-12-22
Create separate metrics for BEGIN, COMMIT, and ROLLBACK statements in YSQL (#6486)

VishnuK007:
This patch separates out the metrics for transaction begin/commit/rollback statements. 

[**Note:** It doesn't break out PREPARE statements yet. The handling for that might require different logic, which I haven't been able to figure out yet. Might be best to separate the PREPARE handling into its own issue.]

**Testing:**
* I created a local YugabyteDB cluster using yb-ctl
* Used ysqlsh to create a sample table and perform of combination of begin, insert, commit, and rollback statements
* Visited the metrics page [http://127.0.0.1:13000/metrics] to ensure that these new metrics showed up and that their counters were incremented

**Sample output from page shown below:**

```
[
    {
        "type": "server",
        "id": "yb.ysqlserver",
        "metrics": [
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_SelectStmt",
                "count": 0,
                "sum": 0
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_InsertStmt",
                "count": 1,
                "sum": 9579
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_DeleteStmt",
                "count": 0,
                "sum": 0
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_UpdateStmt",
                "count": 0,
                "sum": 0
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_BeginStmt",
                "count": 1,
                "sum": 62
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_CommitStmt",
                "count": 1,
                "sum": 14
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_RollbackStmt",
                "count": 1,
                "sum": 41
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_OtherStmts",
                "count": 0,
                "sum": 0
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_Transactions",
                "count": 1,
                "sum": 9579
            },
            {
                "name": "handler_latency_yb_ysqlserver_SQLProcessor_AggregatePushdowns",
                "count": 0,
                "sum": 0
            }
        ]
    }
]
```

### DIFF
--- a/src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
+++ b/src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
@@ -51,6 +51,9 @@ typedef enum statementType
 	Insert,
 	Delete,
 	Update,
+	Begin,
+	Commit,
+	Rollback,
 	Other,
 	Transaction,
 	AggregatePushdown,
@@ -109,6 +112,9 @@ set_metric_names(void)
   strcpy(ybpgm_table[Insert].name, YSQL_METRIC_PREFIX "InsertStmt");
   strcpy(ybpgm_table[Delete].name, YSQL_METRIC_PREFIX "DeleteStmt");
   strcpy(ybpgm_table[Update].name, YSQL_METRIC_PREFIX "UpdateStmt");
+  strcpy(ybpgm_table[Begin].name, YSQL_METRIC_PREFIX "BeginStmt");
+  strcpy(ybpgm_table[Commit].name, YSQL_METRIC_PREFIX "CommitStmt");
+  strcpy(ybpgm_table[Rollback].name, YSQL_METRIC_PREFIX "RollbackStmt");
   strcpy(ybpgm_table[Other].name, YSQL_METRIC_PREFIX "OtherStmts");
   strcpy(ybpgm_table[Transaction].name, YSQL_METRIC_PREFIX "Transactions");
   strcpy(ybpgm_table[AggregatePushdown].name, YSQL_METRIC_PREFIX "AggregatePushdowns");
@@ -528,6 +534,29 @@ ybpgm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
   {
     instr_time start;
     instr_time end;
+    statementType type;
+
+    if (IsA(pstmt->utilityStmt, TransactionStmt)) {
+      TransactionStmt *stmt = (TransactionStmt *)(pstmt->utilityStmt);
+      switch (stmt->kind) {
+      case TRANS_STMT_BEGIN:
+        type = Begin;
+        break;
+      case TRANS_STMT_COMMIT:
+        type = Commit;
+        break;
+      case TRANS_STMT_ROLLBACK:
+      case TRANS_STMT_ROLLBACK_TO:
+	type = Rollback;
+	break;
+      default:
+        type = Other;
+        break;
+      }
+    } else {
+      type = Other;
+    }
+
     INSTR_TIME_SET_CURRENT(start);
 
     ++statement_nesting_level;
@@ -552,7 +581,7 @@ ybpgm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 
     INSTR_TIME_SET_CURRENT(end);
     INSTR_TIME_SUBTRACT(end, start);
-    ybpgm_Store(Other, INSTR_TIME_GET_MICROSEC(end));
+    ybpgm_Store(type, INSTR_TIME_GET_MICROSEC(end));
   }
   else
   {
--- a/src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
+++ b/src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
@@ -547,8 +547,8 @@ ybpgm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
         break;
       case TRANS_STMT_ROLLBACK:
       case TRANS_STMT_ROLLBACK_TO:
-	type = Rollback;
-	break;
+        type = Rollback;
+        break;
       default:
         type = Other;
         break;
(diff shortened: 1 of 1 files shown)


jaki:
Hi, @VishnuK007.  Looks like your commits don't use your proper email, so you aren't able to sign the CLA.  You should set `user.email` in your git config and set the same email to your GitHub account.  You might also have to amend your two commits to use that email.  Let me know if you need help with that.  You can reach out on our community slack yugabyte-db.slack.com as well.

VishnuK007:
Thanks @jaki. I have made the suggested changes. Can you please take a look again?

src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
```diff
@@ -528,6 +534,29 @@ ybpgm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
+      case TRANS_STMT_COMMIT:
+        type = Commit;
+        break;
+      case TRANS_STMT_ROLLBACK:
+      case TRANS_STMT_ROLLBACK_TO:
+        type = Rollback;
+        break;
+      default:
+        type = Other;
+        break;
```
jaki, line 554: This should be indented one more.

src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
```diff
@@ -528,6 +534,29 @@ ybpgm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
+        type = Begin;
+        break;
+      case TRANS_STMT_COMMIT:
+        type = Commit;
+        break;
+      case TRANS_STMT_ROLLBACK:
+      case TRANS_STMT_ROLLBACK_TO:
+        type = Rollback;
+        break;
+      default:
```
jaki, line 552: Handle all statement kinds, particularly `TRANS_STMT_START`.  I would shape the code so that there's a failure when not all cases are explicitly handled.  The rationale is that future additions to the enum should cause equal changes here.

src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
```diff
@@ -528,6 +534,29 @@ ybpgm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
   {
     instr_time start;
     instr_time end;
+    statementType type;
+
+    if (IsA(pstmt->utilityStmt, TransactionStmt)) {
```
jaki, line 539: I suggest moving the getting `type` logic to a separate function.

jaki (changes requested):
Thanks, @VishnuK007.  I'm not too familiar with `prepare`.  I don't see a problem with deferring it for later.

Notable issue is `START TRANSACTION` is regarded as `OtherStmt`.

src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
VishnuK007: Done! Addressed all the other recommended changes as well.

VishnuK007 (commented):
@jaki - Could you please take another look? Thanks.

src/postgres/contrib/yb_pg_metrics/yb_pg_metrics.c
```diff
@@ -528,6 +564,15 @@ ybpgm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
+      TransactionStmt *stmt = (TransactionStmt *)(pstmt->utilityStmt);
+      type = ybpgm_getStatementType(stmt);
```
jaki, line 571:
```suggestion
      type = ybpgm_getStatementType(stmt);
```

jaki (commented):
Sorry for the long wait.  Added one small comment.  I'll run this through internal testing.

jaki (approved):
This looks good.  One questionable thing is that a `COMMIT` on a transaction that went bad and actually rollbacks will register in metrics as a commit statement.  That may or may not be desired, but I think it doesn't need to be decided in the scope of this PR.

```sql
BEGIN;
abc; -- syntax error
COMMIT; -- rollback
```

jaki:
@VishnuK007 

> It doesn't break out PREPARE statements yet

is this outdated now?  Also, I don't think we support `PREPARE TRANSACTION` statements anyway (issue #1125).

```
yugabyte=# prepare transaction 'abc';
ERROR:  PREPARE TRANSACTION not supported yet
LINE 1: prepare transaction 'abc';
        ^
HINT:  See https://github.com/YugaByte/yugabyte-db/issues/1125. Click '+' on the description to raise its priority
```

kmuthukk:
> This looks good. One questionable thing is that a `COMMIT` on a transaction that went bad and actually rollbacks will register in metrics as a commit statement. That may or may not be desired, but I think it doesn't need to be decided in the scope of this PR.
> 
> ```sql
> BEGIN;
> abc; -- syntax error
> COMMIT; -- rollback
> ```

Yes - agreed. The metric is measuring the user level operation initiated-- rather than what might happen underneath. For example, this may happen implicitly on a DML statement also if the txn conflicts with a higher priority txn.

@jaki -- assuming/when your internal test runs also pass, can you squash & merge as well.

jaki:
I think I got it.  Since "prepare" is mentioned by @kmuthukk in issue #6486, it probably means https://www.postgresql.org/docs/11/sql-prepare.html rather than https://www.postgresql.org/docs/11/sql-prepare-transaction.html.

jaki:
And internal testing looks fine.  Merging.  Leaving issue #6486 open because of `PREPARE` statements.

VishnuK007:
> And internal testing looks fine. Merging. Leaving issue #6486 open because of `PREPARE` statements.

Thanks @jaki for your help!

Issue #6243, opened 2020-11-01
[platform] Ensure "node-to-node" tls is enabled in order to use "client-to-node" tls during create universe


streddy-yb:
Node-to-node tls is a pre-requisite in order to use "client-to-node" tls. Currently UI allows selecting "client-to-node" tls even if "node-to-node" tls is not selected. Both API and UI should do a pre-flight check for this.

nishantSharma459:
PR link: https://github.com/yugabyte/yugabyte-db/pull/6751

PR #6751, opened 2020-12-28
[#6243] Platform: Enable "node-to-client" tls if "node-to-node" tls is enabled.

nishantSharma459:


nishantSharma459:
The change ensures that the "ClientToNode" TLS  is disabled initially until nd unless "NodeToNode" TLS is enabled.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -463,6 +474,10 @@ public Result create(UUID customerUUID) {
       LOG.info("Create for {}.", customerUUID);
       // Get the user submitted form data.
       formData = (ObjectNode) request().body().asJson();
+      if (validateEncryption(formData))
+      {
+        return ApiResponse.error(BAD_REQUEST, "Node to node encryption must required to enable for client to node encryption.");
+      }
```
jaydeepkumara, line 480: For now, i just have added validation for creating api let me know if we need to add any other api.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -124,6 +124,17 @@ public UniverseController(YBClientService service) {
     this.ybService = service;
   }
 
+  private boolean validateEncryption(ObjectNode formData) {
```
SergeyPotachev, line 127: Hm... Before this I saw two ways to implement validate functions - without result which throws exceptions if something is wrong, another - which returns true if the arguments is successfully validated. Your implementation can confuse someone as it implements inverted logic - it returns `true` if something is wrong. Actually I prefer standard conventions.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -124,6 +124,17 @@ public UniverseController(YBClientService service) {
...
+    boolean returnValue = true;
+    for (JsonNode cluster : formData.get("clusters")) {
+      JsonNode nodeToNodeEncryption = cluster.get("userIntent").get("enableNodeToNodeEncrypt");
+      JsonNode clientToNodeEncryption = cluster.get("userIntent").get("enableClientToNodeEncrypt");
+
+      returnValue = returnValue && (!nodeToNodeEncryption.asBoolean() && clientToNodeEncryption.asBoolean());
+    }
```
SergeyPotachev, line 134: Despite the previous comment, this one is about the loop - you can exit the loop if the variable becomes false.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -463,6 +474,10 @@ public Result create(UUID customerUUID) {
       LOG.info("Create for {}.", customerUUID);
       // Get the user submitted form data.
       formData = (ObjectNode) request().body().asJson();
+      if (validateEncryption(formData))
+      {
+        return ApiResponse.error(BAD_REQUEST, "Node to node encryption must required to enable for client to node encryption.");
```
SergeyPotachev, line 479: Please rephrase. It is not completely clear what it wants to say.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
SergeyPotachev: Please verify that UI processes the possible error correctly. Often it doesn't.

SergeyPotachev (changes requested):

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -724,6 +725,12 @@ export default class ClusterFields extends Component {
       updateFormField('primary.enableNodeToNodeEncrypt', event.target.checked);
       updateFormField('async.NodeToNodeEncrypt', event.target.checked);
       this.setState({ enableNodeToNodeEncrypt: event.target.checked });
+    } 
+    
+    // This condition will make sure if enableNodeToNodeEncrypt is false
+    // then enableClientToNodeEncrypt will set to false.
+    if(!event.target.checked) {
+      this.setState({ enableClientToNodeEncrypt: false});
```
yb-andrew, line 733: This `setState` call can be moved into the one above. Just check that the `event.target.checked` is false.
```
this.setState({
  enableNodeToNodeEncrypt: event.target.checked,
  enableClientToNodeEncrypt: this.state.enableClientToNodeEncrypt && event.target.checked
});    
```

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1020,6 +1027,16 @@ export default class ClusterFields extends Component {
...
   }
 
+  /**
+   * This method is used to disable the ClientToNodeTLS field initially.
+   * 
+   * @param isFieldReadOnly If true then readonly access.
+   * @param enableNodeToNodeEncrypt NodeToNodeTLS state.
+   */
+  clientToNodeEncryptField(isFieldReadOnly, enableNodeToNodeEncrypt) {
+    return isFieldReadOnly ? isFieldReadOnly : !enableNodeToNodeEncrypt;
```
yb-andrew, line 1037: Can change to `return isFieldReadOnly || !enableNodeToNodeEncrypt;`

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1304,7 +1321,7 @@ export default class ClusterFields extends Component {
         <Field
           name={`${clusterType}.enableClientToNodeEncrypt`}
           component={YBToggle}
-          isReadOnly={isFieldReadOnly}
+          isReadOnly={ this.clientToNodeEncryptField(isFieldReadOnly, this.state.enableNodeToNodeEncrypt)}
```
yb-andrew, line 1324: Is the intention to only set the field initially? What about when `this.state.enableNodeToNodeEncrypt` changes value? Shouldn't the client-to-node field then become editable?

yb-andrew (changes requested):

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
nishantSharma459: Done.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
nishantSharma459: Done.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
nishantSharma459: I've updated the Doc against the same.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
jaydeepkumara: ok got it

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
jaydeepkumara: Done

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
jaydeepkumara: Done

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
jaydeepkumara: Done

yb-andrew (approved):
UI changes look good.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
jaydeepkumara: This specific scenario has been handled in the Front end.

jaydeepkumara (approved):

SergeyPotachev (approved):
Looks good.

Issue #5919, opened 2020-10-02
[Platform] change default storage option for AWS provider to 1x250GB


rao-vasireddy:

gaurav061:
Current Behaviour: Number of instance that is set as default is being determined by the length of volumeDetailsList, which is the subset of instance type list, for most of the instance type this length is 2, but for some of them it is 1(i3.large and i3.xlarge), So Should i make this as hard coded value 1 for all instance type?

jitendra-12113:
Back end changes are done.  Can some please review this code? 
https://github.com/yugabyte/yugabyte-db/pull/6688

UI has to consume **volumeCount** field to show volumeCountInfo - TO DO
cc: @gaurav061

jitendra-12113:
Setting up the volumeCount for aws provider to 1 whose instanceType starting with c5/c4.

gaurav061:
Please find the PR for this ticket:
https://github.com/yugabyte/yugabyte-db/pull/6688

jitendra-12113:
Raised another PR for this ticket. Unit tests are included. Please review.
https://github.com/yugabyte/yugabyte-db/pull/6742

jitendra-12113:
@sb-yb  As discussed, made the necessary changes. Created another CR for the same.
https://github.com/yugabyte/yugabyte-db/pull/6792/
Please review! Thanks!

PR #6765, opened 2020-12-29
[#5919] Platform: change default storage option for AWS provider to 1x250GB

jitendra-12113:
Heading:
[5919][Platform] change default storage option for AWS provider to 1x250GB
Description:
We wanted to show default volume count to 1 on the UI for all the instance types starting with "c5./c4."
Testing:
When we select aws provider while creating universe, volume count was showing more than one for some instances types starting with ""c5./c4.". Now with this fix it will show only one.
Added Unit test as well.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/cloud/AWSInitializer.java
+++ b/managed/src/main/java/com/yugabyte/yw/cloud/AWSInitializer.java
@@ -22,6 +22,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
+import com.typesafe.config.Config;
+
 import com.google.inject.Singleton;
 import com.yugabyte.yw.commissioner.Common;
 import com.yugabyte.yw.common.ApiResponse;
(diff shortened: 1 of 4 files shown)


managed/src/main/java/com/yugabyte/yw/cloud/AWSInitializer.java
```diff
@@ -516,8 +518,7 @@ private void storeInstanceTypeInfoToDB() {
           LOG.error(msg);
           throw new UnsupportedOperationException(msg);
         } else {
-          // TODO: hardcode me not?
-          volumeCount = 2;
+          volumeCount = YB_VOLUME_INFO_VOLUME_COUNT;
```
SergeyPotachev, line 521: It looks like you forgot to read the value from the config file.

SergeyPotachev (changes requested):

managed/src/main/java/com/yugabyte/yw/cloud/AWSInitializer.java
sb-yb: FYI check `TaskGarbageCollector` for good example of how to inject config and use it. Other places are using deprecated code and that will be fixed soon.

sb-yb (changes requested):
@jitendra-12113 Thanks for the change!
Like I commented on thread. I feel configurable way to do this is:
- AWS initializer sets the default to -1.
- InstanceTypes.java will check if the value in database is -1 use the default from configuration.

That way in future one can change configuration and the change takes effect without database migration. 
(You will  still need db migration `one time` to fix existing defaults and set it to -1)

IMHO if we do not do it this way then making it configurable has no value since default gets persisted in the db.
Let me know if this is clear or you want to discuss on VC. cc: @Arnav15

managed/src/main/java/com/yugabyte/yw/cloud/AWSInitializer.java
jitendra-12113: Oops! My bad! Updated the patch. 

@sb-yb  I am not sure on how are we going to migrate DB for older instance types, need to check. Let's connect whenever your time permits to bring this to closure. Thanks!

Issue #6540, opened 2020-12-03
Add ability to remove certificates from UI


tylarb:
Currently all certificates which have ever been created remain visible in Yugaware. 

It would be nice to have a button that will allow you to delete wrongs/expired certificates etc., from the Yugaware platform.

gaurav061:
PR: https://github.com/yugabyte/yugabyte-db/pull/6767

PR #6767, opened 2020-12-30
[#6540] Platform: Remove certificate feature.

mahendranbhat:
Remove certificate feature which will be added in the list view of certificates to delete unused certificates.
If the certificate is linked to existing universe then we will be disabling the button in the FE.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
@@ -129,37 +129,34 @@ public class CertificateController extends AuthenticatedController {
     }
   }
 
-  public Result list(UUID customerUUID) {
-    List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
-    JsonNode certificates = Json.toJson(certs);
-    HashMap<String, Boolean> removable = new HashMap<String, Boolean>();
-    
-    
-    List<Universe> universes = Universe.getAll();
-    for (Universe universe : universes) {
-      try {
-  
-        Universe universe_obj = Universe.get(universe.universeUUID);
-        UUID cert = universe_obj.getUniverseDetails().rootCA;
-        removable.put(cert.toString(), true);
-      } catch (NullPointerException a) {
-        continue;
-      }
-      }
-      for (JsonNode cert : certificates) {
-        JsonNode cert_uuid = cert.get("uuid");
-        Boolean value = removable.get(cert_uuid.asText());
-        if (value == null) {
-          ((ObjectNode) cert).put("removable", true);
-        } else {
-          ((ObjectNode) cert).put("removable", false);
-        }
-      }
-      if (certs == null) {
-        return ApiResponse.error(BAD_REQUEST, "Invalid Customer UUID: " + customerUUID);
-      }
-      return ApiResponse.success(certificates);
-  }
+	public Result list(UUID customerUUID) {
+		List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
+		HashMap<String, Boolean> removable = new HashMap<String, Boolean>();
+		List<Universe> universes = Universe.getAll();
+		for (Universe universe : universes) {
+			try {
+				Universe universe_obj = Universe.get(universe.universeUUID);
+				UUID cert = universe_obj.getUniverseDetails().rootCA;
+				removable.put(cert.toString(), true);
+			} catch (NullPointerException a) {
+				continue;
+			}
+		}
+		JsonNode certificates = Json.toJson(certs);
+		for (JsonNode cert : certificates) {
+			JsonNode cert_uuid = cert.get("uuid");
+			Boolean value = removable.get(cert_uuid.asText());
+			if (value == null) {
+				((ObjectNode) cert).put("removable", true);
+			} else {
+				((ObjectNode) cert).put("removable", false);
+			}
+		}
+		if (certs == null) {
+			return ApiResponse.error(BAD_REQUEST, "Invalid Customer UUID: " + customerUUID);
+		}
+		return ApiResponse.success(certificates);
+	}
 
   public Result get(UUID customerUUID, String label) {
     CertificateInfo cert = CertificateInfo.get(label);
@@ -170,38 +167,35 @@ public class CertificateController extends AuthenticatedController {
     }
   }
 
-  public Result delete(UUID certificate) {
-
-    CertificateInfo cert = CertificateInfo.get(certificate);
-
-    if (cert == null) {
-      return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
-    }
-    List<Universe> universes = Universe.getAll();
-
-    for (Universe universe : universes) {
-      // For the universes which does not have certificate attached to it.
-      try {
-    	  Universe universe_obj = Universe.get(universe.universeUUID);
-          UUID certificate_uuid = universe_obj.getUniverseDetails().rootCA;
-        if (certificate_uuid.toString() == cert.uuid.toString()) {
-          return ApiResponse.error(BAD_REQUEST, "The certificate is in use.");
-        }
-      } catch (NullPointerException a) {
-        continue;
-      }
-    }
-
-    if (cert.delete()) {
-      ObjectNode responseJson = Json.newObject();
-      responseJson.put("Successfully deleted the certificate", true);
-      Audit.createAuditEntry(ctx(), request());
-      LOG.info("Successfully deleted the certificate:" + certificate);
-      return ApiResponse.success(responseJson);
-    } else {
-      return ApiResponse.error(INTERNAL_SERVER_ERROR, "Unable to delete Certificate");
-    }
-  }
+	public Result delete(UUID certificate) {
+		CertificateInfo cert = CertificateInfo.get(certificate);
+		if (cert == null) {
+			return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
+		}
+		List<Universe> universes = Universe.getAll();
+		for (Universe universe : universes) {
+			// Try catch block for the universes which does not have certificate attached to
+			// it.
+			try {
+				Universe universe_obj = Universe.get(universe.universeUUID);
+				UUID certificate_uuid = universe_obj.getUniverseDetails().rootCA;
+				if (certificate_uuid.toString() == cert.uuid.toString()) {
+					return ApiResponse.error(BAD_REQUEST, "The certificate is in use.");
+				}
+			} catch (NullPointerException a) {
+				continue;
+			}
+		}
+		if (cert.delete()) {
+			ObjectNode responseJson = Json.newObject();
+			responseJson.put("Successfully deleted the certificate", true);
+			Audit.createAuditEntry(ctx(), request());
+			LOG.info("Successfully deleted the certificate:" + certificate);
+			return ApiResponse.success(responseJson);
+		} else {
+			return ApiResponse.error(INTERNAL_SERVER_ERROR, "Unable to delete Certificate");
+		}
+	}
 
   public Result updateEmptyCustomCert(UUID customerUUID, UUID rootCA) {
     Form<CertificateParams> formData = formFactory.form(CertificateParams.class)
--- a/managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
@@ -10,7 +10,6 @@ import com.yugabyte.yw.models.CertificateInfo;
 import com.yugabyte.yw.models.Customer;
 import com.yugabyte.yw.forms.CertificateParams;
 import com.yugabyte.yw.forms.ClientCertParams;
-
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.yugabyte.yw.models.Universe;
@@ -133,19 +132,23 @@ public class CertificateController extends AuthenticatedController {
   public Result list(UUID customerUUID) {
     List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
     JsonNode certificates = Json.toJson(certs);
-    List<Universe> universes = Universe.getAll();
     HashMap<String, Boolean> removable = new HashMap<String, Boolean>();
+    
+    
+    List<Universe> universes = Universe.getAll();
     for (Universe universe : universes) {
       try {
-        UUID cert = universe.getUniverseDetails().rootCA;
+  
+        Universe universe_obj = Universe.get(universe.universeUUID);
+        UUID cert = universe_obj.getUniverseDetails().rootCA;
         removable.put(cert.toString(), true);
       } catch (NullPointerException a) {
         continue;
-      }}
+      }
+      }
       for (JsonNode cert : certificates) {
         JsonNode cert_uuid = cert.get("uuid");
         Boolean value = removable.get(cert_uuid.asText());
-        System.out.println(removable+"hello");
         if (value == null) {
           ((ObjectNode) cert).put("removable", true);
         } else {
@@ -179,8 +182,9 @@ public class CertificateController extends AuthenticatedController {
     for (Universe universe : universes) {
       // For the universes which does not have certificate attached to it.
       try {
-        UUID universe_json_info = universe.getUniverseDetails().rootCA;
-        if (universe_json_info.toString() == cert.uuid.toString()) {
+    	  Universe universe_obj = Universe.get(universe.universeUUID);
+          UUID certificate_uuid = universe_obj.getUniverseDetails().rootCA;
+        if (certificate_uuid.toString() == cert.uuid.toString()) {
           return ApiResponse.error(BAD_REQUEST, "The certificate is in use.");
         }
       } catch (NullPointerException a) {
--- a/managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
@@ -167,35 +167,34 @@ public class CertificateController extends AuthenticatedController {
     }
   }
 
-	public Result delete(UUID certificate) {
-		CertificateInfo cert = CertificateInfo.get(certificate);
-		if (cert == null) {
-			return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
-		}
-		List<Universe> universes = Universe.getAll();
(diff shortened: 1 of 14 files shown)


managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -128,13 +129,34 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
...
 
-  public Result list(UUID customerUUID) {
-    List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
-    if (certs == null) {
-      return ApiResponse.error(BAD_REQUEST, "Invalid Customer UUID: " + customerUUID);
-    }
-    return ApiResponse.success(certs);
-  }
+	public Result list(UUID customerUUID) {
+		List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
```
jitendra-12113, line 133: extra spaces, enable auto format in your ide to fix such issues.

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+   * @param certificateUUID Unique id of certificate.
+   */
+  deleteRootCertificate = (certificateUUID) => {
+    api.deleteCertificate(certificateUUID).then(
+      () => {
+        this.getCertificateArray()
+      }
+    ).catch(
+      err => {
+        console.log("Error", err)
```
jaydeepkumara, line 134: In the case of 4xx we need to show the message to Frontend

jaydeepkumara (approved):

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -128,13 +129,34 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
...
+			}
+		}
+		JsonNode certificates = Json.toJson(certs);
+		for (JsonNode cert : certificates) {
+			JsonNode cert_uuid = cert.get("uuid");
+			Boolean value = removable.get(cert_uuid.asText());
+			if (value == null) {
+				((ObjectNode) cert).put("removable", true);
+			} else {
+				((ObjectNode) cert).put("removable", false);
```
sshev, line 152: I'd suggest renaming this prop as it's unclear why a cert could be marked as removable
smth like `inUse` will be self-explanatory

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+   */
+  componentDidMount() {
+    this.getCertificateArray();
+  }
+
+  /**
+   * Fetch certificates attched to current user.
+   */
+  getCertificateArray = () => {
+    api.getCertificates().then(
```
sshev, line 151: we are using `redux` at legacy pages and this change deviates from the existing data flow
certs routines are in `CertificatesContainer.js` so let's reuse it please

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+            expiryDate: cert.expiryDate,
+            certificate: cert.certificate,
+            creationTime: cert.startDate,
+            privateKey: cert.privateKey,
+            customCertInfo: cert.customCertInfo,
+            removable: cert.removable
+          };
+        })
+        : [];
+      this.setState({ certificateArray: certificateArray})
```
sshev, line 168: no need to store certs list in the local state as it comes as prop from the redux store

managed/ui/src/redesign/helpers/api.ts
```diff
@@ -112,6 +113,16 @@ class ApiService {
   isRequestCancelError(error: unknown): boolean {
     return axios.isCancel(error);
   }
+
+  /**
+   * Delete certificate which is not attched to any universe.
+   * 
+   * @param certUUID - certificate UUID
+   */
+  deleteCertificate = (certUUID: string): Promise<any> => {
```
sshev, line 122: if we don't expect any response from the delete API call - the return type should be `Promise<void>`

sshev (changes requested):

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/ui/src/redesign/helpers/api.ts
gaurav061: Done

sshev:
@mahendranbhat could you please address the `removable` renaming and code formatting comment?
@Arnav15 @sb-yb folks, please take look at the java code

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -153,6 +173,14 @@ class Certificates extends Component {
         >
           <i className="fa fa-download"></i> Download Root CA Cert
         </MenuItem>
+        <MenuItem
+          onClick={() => { 
+            !deleteDisabled && this.deleteRootCertificate(payload?.uuid)
+          }}
+          disabled={deleteDisabled}
+        >
+          <i className="fa fa-trash"></i> Delete Root CA Cert
```
sshev, line 182: missed this one - as long as it's a generic cert delete functionality `root CA` part is unnecessary

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +117,27 @@ class Certificates extends Component {
       .finally(() => this.setState({ showSubmitting: false }));
   };
 
+  /**
+   * Delete the root certificate if certificate is safe to remove,
+   * i.e - Certificate is not attached to any universe for current user.
+   * 
+   * @param certificateUUID Unique id of certificate.
+   */
+  deleteRootCertificate = (certificateUUID) => {
```
sshev, line 126:
```suggestion
  deleteCertificate = (certificateUUID) => {
```

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -128,13 +129,34 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
...
+	public Result list(UUID customerUUID) {
+		List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
+		HashMap<String, Boolean> removable = new HashMap<String, Boolean>();
+		List<Universe> universes = Universe.getAll();
+		for (Universe universe : universes) {
+			try {
+				Universe universe_obj = Universe.get(universe.universeUUID);
+				UUID cert = universe_obj.getUniverseDetails().rootCA;
+				removable.put(cert.toString(), true);
+			} catch (NullPointerException a) {
```
sb-yb, line 141: Do use exceptions for control flow.
`if (<null check>) continue;`
for where you want to check.
What can be a `null`?

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -128,13 +129,34 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
...
-  public Result list(UUID customerUUID) {
-    List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
-    if (certs == null) {
-      return ApiResponse.error(BAD_REQUEST, "Invalid Customer UUID: " + customerUUID);
-    }
-    return ApiResponse.success(certs);
-  }
+	public Result list(UUID customerUUID) {
+		List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
+		HashMap<String, Boolean> removable = new HashMap<String, Boolean>();
```
sb-yb, line 134: `HashSet<UUID> removableUUIDs = new ..`

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -128,13 +129,34 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
...
+				UUID cert = universe_obj.getUniverseDetails().rootCA;
+				removable.put(cert.toString(), true);
+			} catch (NullPointerException a) {
+				continue;
+			}
+		}
+		JsonNode certificates = Json.toJson(certs);
+		for (JsonNode cert : certificates) {
+			JsonNode cert_uuid = cert.get("uuid");
+			Boolean value = removable.get(cert_uuid.asText());
```
sb-yb, line 148: You never put false into the map in the above loop.
So when `value` exists it can only be `true`. 
Yet here you are saying if value exists always set removable to `false`.

I think there is a bug if meaning of  removable remains same between the two places.

Or did you mean to call the map inUse? Please clarify.

sb-yb: Okay check the description. The map above is actually saying removable true when certificate is attached to universe.
`HashSet<String> inUseCerts ...
<first loop to populate this set>
`
Then In second loop you just do:
`certs.put("inUse", inUseCerts.contains(cert.get("uuid").asText()))
`

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +167,38 @@ public Result get(UUID customerUUID, String label) {
     }
   }
 
+  public Result delete(UUID certificate) {
+	CertificateInfo cert = CertificateInfo.get(certificate);
+	if (cert == null) {
+		return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
+	}
+	List<Universe> universes = Universe.getAll();
```
sb-yb, line 175: no need for this variable. Just inline it.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +167,38 @@ public Result get(UUID customerUUID, String label) {
...
+	for (Universe universe : universes) {
+		// Try catch block for the universes which does not have certificate attached to it
+		try {
+			Universe universe_obj = Universe.get(universe.universeUUID);
+			UUID certificate_uuid = universe_obj.getUniverseDetails().rootCA;
+			if (certificate_uuid.toString() == cert.uuid.toString()) {
+				return ApiResponse.error(BAD_REQUEST, "The certificate is in use.");
+			}
+		} catch (NullPointerException a) {
```
sb-yb, line 184: Do not use exceptions for control flow.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +167,38 @@ public Result get(UUID customerUUID, String label) {
...
+	if (cert == null) {
+		return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
+	}
+	List<Universe> universes = Universe.getAll();
+	for (Universe universe : universes) {
+		// Try catch block for the universes which does not have certificate attached to it
+		try {
+			Universe universe_obj = Universe.get(universe.universeUUID);
+			UUID certificate_uuid = universe_obj.getUniverseDetails().rootCA;
+			if (certificate_uuid.toString() == cert.uuid.toString()) {
```
sb-yb, line 181: Do not equal java objects with `==`

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +167,38 @@ public Result get(UUID customerUUID, String label) {
...
+  public Result delete(UUID certificate) {
+	CertificateInfo cert = CertificateInfo.get(certificate);
+	if (cert == null) {
+		return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
+	}
+	List<Universe> universes = Universe.getAll();
+	for (Universe universe : universes) {
+		// Try catch block for the universes which does not have certificate attached to it
+		try {
+			Universe universe_obj = Universe.get(universe.universeUUID);
```
sb-yb, line 179: But you already have the universe. So why lookup again?

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
sb-yb: Also why convert to string? You can just compare the UUIDs.

managed/src/test/java/com/yugabyte/yw/controllers/CertificateControllerTest.java
```diff
@@ -117,12 +122,29 @@ public void testListCertificates() {
...
+      assertEquals(e.get("removable"), true);
     }
     assertEquals(test_certs, result_labels);
     assertEquals(test_certs_uuids, result_uuids);
     assertAuditEntry(0, customer.uuid);
   }
 
+  @Test
+  public void testDeleteCertificate() {
+	UUID cert_uuid = test_certs_uuids.get(0);
```
sb-yb, line 134: formatting

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +167,38 @@ public Result get(UUID customerUUID, String label) {
     }
   }
 
+  public Result delete(UUID certificate) {
```
sb-yb, line 170: rename reqCertUUID

sb-yb (changes requested):
Lot of change needed on java side.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: If the universe is not attached to any certificate ( without TLS enabled ) then we would get null for `universe_obj.getUniverseDetails()` It will not return Null instead raises Nullpointer exception.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: I was getting Null every time if I use `universe` but when I get the object like above, I am getting the correct `rootCA` . Not sure what was the issue even though both objects look the same.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: Same reason as list API exception block.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: In short 
`universe_obj.getUniverseDetails().rootCA;`  giving correct rootCA
but 
`univers.getUniverseDetails().rootCA;`   giving Null every single time.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: Moved it to hashset.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: Made it as String as `JsonNode cert_uuid = cert.get("uuid");` returning JsonNode and converting it to string for comparision.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: Added a comment in the code as well

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
Arnav15: @mahendranbhat That is incorrect. `getUniverseDetails()` should not have anything to do with the certs. The rootCA might be null, the universe details will not. Also, I am not sure I like the naming. Doesn't the cert belonging to a universe make it non removable? And as @sb-yb said, you should use a hashSet instead of a map. If the cert is not in the set, we then mark it as removable in the corresponding object that you create for it.

Arnav15 (changes requested):
I see you commented on a couple of things as done, but I don't see them. Is it possible that you did not update the commit? Or are you just commenting on them as you are fixing them to mark it for yourself? Or is something just wrong on my end?

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
sb-yb: Either way even if this was indeed throwing NPE that is still not a good reason to use exception for control flow. 
Based on what @Arnav15 says you need to investigate why `getUniverseDetails` would throw in such case.  May be write a unittest to repro.
My original comment has a type it should say "DONT use exception for control flow"

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
sb-yb: Dont see this. I still see it is HashMap above and still named removable.

sb-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: @Arnav15 @sb-yb my bad , I thought I am using` getCertificate()` method to get a certificate.  ` getCertificate()` would give a null pointer exception.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: The PR is still pointing to the old branch. Opening we are giving PR from Jaydeep's repo.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: also `getUniverseDetails().rootCA`  would be null.

mahendranbhat:
Closing this PR as the base branch is updated to point to a different repo. Opening a new PR later.

sshev:
@mahendranbhat delete the branch if not needed anymore pls

tylarb:
new PR opened is #6885

PR #6768, opened 2020-12-30
Modified creationTime in Universe list API

mahendranbhat:
As the creation time format is incorrect in the list API, FE was not able to parse it. 
Changed the format of the date in the API.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
@@ -5,6 +5,7 @@ package com.yugabyte.yw.controllers;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


jaydeepkumara (approved):

SergeyPotachev (changes requested):
Looks good, but add junit test for this, please.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -841,6 +842,8 @@ public Result list(UUID customerUUID) {
         UniverseResourceDetails details = UniverseResourceDetails.create(universe.getNodes(),
             universe.getUniverseDetails());
         universePayload.put("pricePerHour", details.pricePerHour);
+        SimpleDateFormat creationDate = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+        universePayload.put("creationDate", creationDate.format(universe.creationDate));
```
Arnav15, line 846: Are we sure this is the issue? Since we do add this [here](https://github.com/yugabyte/yugabyte-db/blob/master/managed/src/main/java/com/yugabyte/yw/models/Universe.java#L150). And the format doesn't throw errors after the universe has successfully been created. It only occurs while the universe is in a creating state. Am I missing something?

Arnav15 (changes requested):

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
mahendranbhat: Yes the creation time is coming in the invalid format as we have implemented custom list API here. Yeah we can update the UI code to support it but in other places, we are getting the date directly from postgres but here `creationDate: "Thu Jan 07 19:08:38 IST 2021"`  
other places : `creationDate: "2020-12-11 06:59:21"` this format.

mahendranbhat: you can check the universe API creation date format and the other place date format in the UI.
We could update the UI to support this format too but the back-end format should be consistent too.

mahendranbhat:
Changing the base repository and raising new PR

Issue #6772, opened 2020-12-30
Display TServer clock information in yb-master ui


kmuthukk:
For debugging purposes, it'll be useful to add a new view say -- http://yb-master:7000/tablet-server-clocks) to visually display the clocks of the various TServers, that provides the following data for each tablet server:

* time since heartbeat (this is already available from the heartbeat sent to master and it is shown here today).
* the physical time on the tserver at the time of heartbeat
* the hybrid time on the tserver at the time of the heartbeat (its physical & logical components).

kmuthukk:
Implementation pointers on this (based on discussions with @spolitov ):

* add two new fields in `TSHeartbeatRequestPB`  for `ts_physical_time` and `ts_hybrid_time`. The plan would be to set these two fields in this function which handles the heartbeats on the TServer side: https://github.com/yugabyte/yugabyte-db/blob/8bac374d032fd9a7bac73fab3391e0ee7596ac76/src/yb/tserver/heartbeater.cc#L345

We typically use `fixed64` type for these fields in proto over `int64`. The latter uses variant length encoding which isn't very useful for very large integer values.

* To obtain the hybrid time on the TServer side, the logic would involve getting the hybrid clock:

 `auto* hybrid_clock = down_cast<server::HybridClock*>(server_->Clock());`

and from there, the current value can be obtained thus:

 `hybrid_clock->Now().ToUint64()`

* To obtain the physical time on the TServer, the TEST_clock() function, currently only used for tests, can be repurposed and renamed to say `physical_clock()`

https://github.com/yugabyte/yugabyte-db/blob/fe4e4cc9f137a11e6005619210bf2339af884ce8/src/yb/server/hybrid_clock.h#L145
 
and the now timestamp can be obtained via:

`Result<PhysicalTime> now = hybrid_clock->physical_clock()->Now();`

and subsequently via `now->time_point`

* On the master side, when heartbeats are received, the new fields in the proto structures can be copied/save into two new fields (for physical & hybrid time) in the TSDescriptor

Here's the master side logic when a TServer heartbeat is received.
https://github.com/yugabyte/yugabyte-db/blob/e9e96610facfffca3added68eae556790c33e217/src/yb/master/master_service.cc#L97

These TSDescriptor fields can then be used to drive the UI logic.

VishnuK007:
I am interested in working on this issue. I will try to submit a patch for it in the next couple of days.

rkarthik007:
[...]

PR #6776, opened 2021-01-01
#6772 display tserver clock information in yb-master UI

VishnuK007:
This patch adds a `/tablet-server-clocks` page to the yb-master UI.

The physical and hybrid times of the tablet server are sent as a part of the heartbeat mechanism.

**Testing**
### 
- Created a local YugabyteDB cluster with using `yb-ctl --rf 3 start`
- Then verified that the patch worked by going to `http://127.0.0.1:7000/tablet-server-clocks`

Screenshots of the changes included below.

**Utilities View** - This view has a new panel for TServer Clocks.
Screen Shot 2021-01-01 at 1 23 35 PM

**Tablet Server Clocks View** - This is the newly added view which displays the physical and hybrid times.
Screen Shot 2021-01-10 at 5 16 54 PM

**Default Tablet Servers View** - This is the default tablet servers view, which should remain the same as before even after the re-factoring of code for the new view.
Screen Shot 2021-01-01 at 1 23 50 PM

### DIFF
--- a/src/yb/integration-tests/mini_cluster.cc
+++ b/src/yb/integration-tests/mini_cluster.cc
@@ -560,7 +560,8 @@ std::vector<server::SkewedClockDeltaChanger> SkewClocks(
     auto* tserver = cluster->mini_tablet_server(i)->server();
     auto* hybrid_clock = down_cast<server::HybridClock*>(tserver->clock());
     delta_changers.emplace_back(
-        i * clock_skew, std::static_pointer_cast<server::SkewedClock>(hybrid_clock->physical_clock()));
+        i * clock_skew, std::static_pointer_cast<server::SkewedClock>(
+            hybrid_clock->physical_clock()));
   }
   return delta_changers;
 }
--- a/src/yb/integration-tests/mini_cluster.cc
+++ b/src/yb/integration-tests/mini_cluster.cc
@@ -560,7 +560,7 @@ std::vector<server::SkewedClockDeltaChanger> SkewClocks(
     auto* tserver = cluster->mini_tablet_server(i)->server();
     auto* hybrid_clock = down_cast<server::HybridClock*>(tserver->clock());
     delta_changers.emplace_back(
-        i * clock_skew, std::static_pointer_cast<server::SkewedClock>(hybrid_clock->TEST_clock()));
+        i * clock_skew, std::static_pointer_cast<server::SkewedClock>(hybrid_clock->physical_clock()));
   }
   return delta_changers;
 }
--- a/src/yb/master/master-path-handlers.cc
+++ b/src/yb/master/master-path-handlers.cc
@@ -216,17 +216,18 @@ void MasterPathHandlers::CallIfLeaderOrPrintRedirect(
 }
 
 inline void MasterPathHandlers::TServerTable(std::stringstream* output,
-                                             bool tserver_clocks_view) {
+                                             TabletServersViewType viewType) {
   *output << "<table class='table table-striped'>\n";
   *output << "    <tr>\n"
           << "      <th>Server</th>\n"
           << "      <th>Time since </br>heartbeat</th>\n"
           << "      <th>Status & Uptime</th>\n";
 
-  if (tserver_clocks_view) {
+  if (viewType == kTServersClocksView) {
     *output << "      <th> Physical Time</th>\n"
             << "      <th> Hybrid Time</th>\n";
   } else {
+    DCHECK_EQ(viewType, kTServersDefaultView);
     *output << "      <th>User Tablet-Peers / Leaders</th>\n"
             << "      <th>RAM Used</th>\n"
             << "      <th>Num SST Files</th>\n"
@@ -240,7 +241,7 @@ inline void MasterPathHandlers::TServerTable(std::stringstream* output,
           << "      <th>Region</th>\n"
           << "      <th>Zone</th>\n";
 
-  if (!tserver_clocks_view) {
+  if (viewType == kTServersDefaultView) {
     *output << "      <th>System Tablet-Peers / Leaders</th>\n"
             << "      <th>Active Tablet-Peers</th>\n";
   }
@@ -326,7 +327,7 @@ void MasterPathHandlers::TServerDisplay(const std::string& current_uuid,
                                         TabletCountMap* tablet_map,
                                         std::stringstream* output,
                                         const int hide_dead_node_threshold_mins,
-                                        bool tserver_clocks_view) {
+                                        TabletServersViewType viewType) {
   // Copy vector to avoid changes to the reference descs passed
   std::vector<std::shared_ptr<TSDescriptor>> local_descs(*descs);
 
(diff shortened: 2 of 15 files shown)


src/yb/master/master-path-handlers.cc
```diff
@@ -311,7 +325,8 @@ void MasterPathHandlers::TServerDisplay(const std::string& current_uuid,
                                         std::vector<std::shared_ptr<TSDescriptor>>* descs,
                                         TabletCountMap* tablet_map,
                                         std::stringstream* output,
-                                        const int hide_dead_node_threshold_mins) {
+                                        const int hide_dead_node_threshold_mins,
+                                        bool tserver_clocks_view) {
```
spolitov, line 329: It would be better to use an enum for this purpose.
Or even a more abstract way with class/function that renders a single server view.

src/yb/master/ts_descriptor.h
```diff
@@ -177,6 +177,26 @@ class TSDescriptor {
...
+  void set_hybrid_time(int64_t hybrid_time) {
+    std::lock_guard<decltype(lock_)> l(lock_);
+    hybrid_time_ = hybrid_time;
+  }
+
+  int64_t hybrid_time() const {
+    SharedLock<decltype(lock_)> l(lock_);
+    return hybrid_time_;
+  }
+
```
spolitov, line 199: With this approach, we could render info from different heartbeats in a single view.
Would prefer to add the `TSDescriptorState` struct and update/read it with a single call.
Something like:
```
TSDescriptorState state = {
  .num_live_replicas = req->num_live_tablets(),
  .leader_count = req->leader_count(),
  .physical_time = req->ts_physical_time(),
  .hybrid_time = HybridTime(req->ts_hybrid_time()),
};
desc->set_state(state);
```

src/yb/master/ts_descriptor.h
```diff
@@ -337,6 +357,10 @@ class TSDescriptor {
   // The last time a heartbeat was received for this node.
   MonoTime last_heartbeat_;
 
+  // The physical and hybrid times on this node at the time of heartbeat
+  int64_t physical_time_;
+  int64_t hybrid_time_;
```
spolitov, line 362: Store it as `HybridTime`

src/yb/master/ts_descriptor.h
```diff
@@ -337,6 +357,10 @@ class TSDescriptor {
   // The last time a heartbeat was received for this node.
   MonoTime last_heartbeat_;
 
+  // The physical and hybrid times on this node at the time of heartbeat
+  int64_t physical_time_;
```
spolitov, line 361: Use `MicrosTime`

spolitov (changes requested):

src/yb/master/ts_descriptor.h
VishnuK007: I have not incorporated this change, since it seems to be a pre-existing issue. Many TSDescriptor fields are set using individual methods, and a larger refactoring may be needed. Let me know if that is okay. Thanks.

VishnuK007:
Thanks @spolitov! I have addressed the changes you suggested. Can you please take a look at them?

src/yb/master/master-path-handlers.h
```diff
@@ -84,6 +84,11 @@ class MasterPathHandlers {
     kNumTypes,
   };
 
+  enum TabletServersViewType {
```
spolitov, line 87: We have `YB_DEFINE_ENUM` for this purpose.

src/yb/master/ts_descriptor.h
spolitov: Yes, it is ok, since we already have this issue. But would be nice to fix in future.

src/yb/master/master-path-handlers.cc
```diff
@@ -339,22 +355,41 @@ void MasterPathHandlers::TServerDisplay(const std::string& current_uuid,
...
-      *output << "    <td>" << desc->num_sst_files() << "</td>";
-      *output << "    <td>" << HumanizeBytes(desc->total_sst_file_size()) << "</td>";
-      *output << "    <td>" << HumanizeBytes(desc->uncompressed_sst_file_size()) << "</td>";
-      *output << "    <td>" << desc->read_ops_per_sec() << "</td>";
-      *output << "    <td>" << desc->write_ops_per_sec() << "</td>";
+
+      if (viewType == kTServersClocksView) {
+        // Render physical time.
+        const Timestamp p_ts(desc->physical_time());
+        *output << "    <td>" << p_ts.ToFormattedString() << "</td>";
```
spolitov, line 362: I don't think that we need `+0000` in each entry.
Also not sure that the logical component is useful since it changes quite frequently.

VishnuK007: I made it so that the logical component is only displayed if it is non-zero. I have updated the screenshot in the original post to reflect this.

I haven't made the change for the time zone information, since I'm using the available ToFormattedString() method.

src/yb/integration-tests/mini_cluster.cc
```diff
@@ -560,7 +560,7 @@ std::vector<server::SkewedClockDeltaChanger> SkewClocks(
     auto* tserver = cluster->mini_tablet_server(i)->server();
     auto* hybrid_clock = down_cast<server::HybridClock*>(tserver->clock());
     delta_changers.emplace_back(
-        i * clock_skew, std::static_pointer_cast<server::SkewedClock>(hybrid_clock->TEST_clock()));
+        i * clock_skew, std::static_pointer_cast<server::SkewedClock>(hybrid_clock->physical_clock()));
```
spolitov, line 563: This line became too long (> 100 chars), should be wrapped.

src/yb/master/master-path-handlers.cc
```diff
@@ -521,15 +560,16 @@ void MasterPathHandlers::HandleTabletServers(const Webserver::WebRequest& req,
...
+                 viewType);
 
   for (const auto& read_replica_uuid : read_replica_uuids) {
     *output << "<h3 style=\"color:" << kYBDarkBlue << "\">Read Replica UUID: "
             << (read_replica_uuid.empty() ? kNoPlacementUUID : read_replica_uuid) << "</h3>\n";
-    TServerTable(output);
-    TServerDisplay(
-        read_replica_uuid, &descs, &tablet_map, output, hide_dead_node_threshold_override);
+    TServerTable(output, viewType);
+    TServerDisplay(read_replica_uuid, &descs, &tablet_map, output, hide_dead_node_threshold_override,
```
spolitov, line 571: This line too long (> 100 chars). Should be wrapped.

src/yb/tserver/heartbeater.cc
```diff
@@ -382,6 +382,19 @@ Status Heartbeater::Thread::TryHeartbeat() {
   req.set_config_index(server_->GetCurrentMasterIndex());
   req.set_cluster_config_version(server_->cluster_config_version());
 
+  // Include the hybrid time of this tablet server in the heartbeat.
+  auto* hybrid_clock = down_cast<server::HybridClock*>(server_->Clock());
```
spolitov, line 386: We have tests where `server_->Clock()` is not an instance of `HybridClock`.
This case should be handled.

spolitov (changes requested):
Added comments based on phabricator build.

VishnuK007:
I have incorporated the latest suggestions, @spolitov. Thanks!

src/yb/tserver/heartbeater.cc
```diff
@@ -382,6 +382,23 @@ Status Heartbeater::Thread::TryHeartbeat() {
...
+
+  // Also include the physical clock time of this tablet server in the heartbeat.
+  Result<PhysicalTime> now = hybrid_clock->physical_clock()->Now();
+  if (!now.ok()) {
+    YB_LOG_EVERY_N_SECS(WARNING, 10) << "Failed to read clock: " << now.status();
+    req.set_ts_physical_time(0);
+  } else {
+    req.set_ts_physical_time(now->time_point);
+  }
+
```
spolitov, line 401: It should be moved to the `if` above.

spolitov (changes requested):

spolitov (approved):

PR #6777, opened 2021-01-03
allow OPTIONS in CQL protocol prior to AUTHENTICATE

idioterna:
for backwards compatibility with DataStax PHP driver (and possibly their cpp driver, which PHP driver relies on).

### DIFF
--- a/src/yb/yql/cql/cqlserver/cql_processor.cc
+++ b/src/yb/yql/cql/cqlserver/cql_processor.cc
@@ -246,6 +246,8 @@ bool CQLProcessor::CheckAuthentication(const CQLRequest& req) const {
   return call_->ql_session()->is_user_authenticated() ||
       // CQL requests which do not need authorization.
       req.opcode() == CQLMessage::Opcode::STARTUP ||
+      // Some drivers issue OPTIONS prior to AUTHENTICATE
+      req.opcode() == CQLMessage::Opcode::OPTIONS ||
       req.opcode() == CQLMessage::Opcode::AUTH_RESPONSE;
 }
 


rkarthik007:
Thanks for your PR @idioterna! Is the scenario being fixed the following:
* Enable authentication on YugabyteDB
* Connect to the cluster using a simple PHP program
* The `OPTIONS` request is issued before the auth is done, and hence this fix is needed?

idioterna:
Yes, https://github.com/datastax/php-driver/ issues an `OPTIONS` request before authenticating (presumably because https://github.com/datastax/cpp-driver/ may have done this as well) and we're running tserver with `--use_cassandra_authentication=true`. I also have pcap files of conversations if you need them, but this fixes the problem for us and since no privileged information is revealed by an `OPTIONS` response, I did it this way - if you feel we should instead fix the php-driver, I can also try to refactor that, but DataStax claim their driver is not maintaned anymore so it might take a while.

rkarthik007:
> this fixes the problem for us and since no privileged information is revealed by an `OPTIONS` response, I did it this way - if you feel we should instead fix the `php-driver`, I can also try to refactor that, but DataStax claim their driver is not maintaned anymore so it might take a while.

I think this way is totally fine. I was just trying to understand in what scenario this showed up. Will let @m-iancu review the change. Thanks for contributing this!

m-iancu:
Thanks @idioterna, you're correct: according to the wire protocol specification drivers are allowed to sent an OPTIONS message earlier. This was an omission from our side. Thanks for finding and fixing it.

Issue #5594, opened 2020-09-04
[YW] Handle invalid certs/keys correctly


iSignal:
1. Creating a universe with an invalid certificate key/crt leads to the following error trace - however the universe itself shows up as "Ready" with no errors. Deleting this kind of a universe may also fail because it is not full set up.

2. A similar error happens if one goes to the Certificates list and clicks Actions -> Get Client certificate from the dropdown. One issue there is that there is no way to remove the invalid cert from the list (because we do not allow removing any certs once they are added).

```
2020-09-04 05:22:51.286 INFO  UniverseController.java:465 [application-akka.actor.default-dispatcher-11] Created universe 9f15fa61-5e3d-47b3-8621-85af2e93636b : sanketh-test-1.
2020-09-04 05:22:51.289 INFO  UniverseController.java:471 [application-akka.actor.default-dispatcher-11] Added universe 9f15fa61-5e3d-47b3-8621-85af2e93636b : sanketh-test-1 for customer [1].
2020-09-04 05:22:51.607 ERROR CertificateHelper.java:183 [application-akka.actor.default-dispatcher-11] Unable to create client CA for username yugabyte: {}
java.security.spec.InvalidKeySpecException: java.security.InvalidKeyException: invalid key format
	at sun.security.rsa.RSAKeyFactory.engineGeneratePrivate(RSAKeyFactory.java:252)
	at java.security.KeyFactory.generatePrivate(KeyFactory.java:372)
	at com.yugabyte.yw.common.CertificateHelper.createClientCertificate(CertificateHelper.java:181)
	at com.yugabyte.yw.controllers.UniverseController.create(UniverseController.java:494)
	at v1.Routes$$anonfun$routes$1$$anonfun$applyOrElse$67$$anonfun$apply$67.apply(Routes.scala:3112)
	at v1.Routes$$anonfun$routes$1$$anonfun$applyOrElse$67$$anonfun$apply$67.apply(Routes.scala:3112)
	at play.core.routing.HandlerInvokerFactory$$anon$3.resultCall(HandlerInvoker.scala:143)
	at play.core.routing.HandlerInvokerFactory$$anon$3.resultCall(HandlerInvoker.scala:142)
	at play.core.routing.HandlerInvokerFactory$JavaActionInvokerFactory$$anon$8$$anon$2$$anon$1.invocation(HandlerInvoker.scala:116)
	at play.core.j.JavaAction$$anon$1.call(JavaAction.scala:112)
[...]

PR #6780, opened 2021-01-04
[#5594] Platform:Handle invalid certs/keys correctly.

jitendra-12113:
**Heading:**
[5594][Platform] Handle invalid certs/keys correctly
**Description:**
We wanted to verify the cert at upload time itself.
**Testing:**
When we try to upload selfSigned cert, we were able to upload any
garbage files. Now with this fix, we have handled this scenrio. User
won't be able to upload garbage file. Incase of self signed cert upload
workflow, We are matching the modulus of
cert RSA Public key with the modulus of its private key if its matched
then it means certs are valid otheriwse Invalid.
For CA Cert, we are just validating the Cert format.
Added the unit Tests as well.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
+++ b/managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
@@ -36,12 +36,15 @@ import org.slf4j.LoggerFactory;
 
 import play.libs.Json;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.StringWriter;
+import java.io.StringReader;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.math.BigInteger;
@@ -57,8 +60,11 @@ import java.security.SignatureException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
+import java.security.interfaces.RSAPrivateKey;
+import java.security.interfaces.RSAPublicKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Calendar;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


Arnav15 (changes requested):
Hi level comment for indentation. We use the google style guide for formatting, and here is the guide for line breaks: https://google.github.io/styleguide/javaguide.html#s4.5-line-wrapping
So all the indentations in line breaks that have been changed to 2 spaces, should be changed back to atleast 4.
The change looks good overall, but just some minor comments to help cleanup the existing code as well.
The existing functions have some common code that can maybe be made into helper functions. So, we have code for reading a keyfile in the createClientCertificate function. We also use the JCAPemWriter class to write the certificates in both the createRootCA and createClientCertificate. Now, since in the upload functionality, we are creating an X509 cert object, maybe we can use the JCAPemWriter class for writing that to the file system too. This will help us in unifying the output of the certs in a particular format (ending with a newline or not), which will help us with the computation of the md5 of the file (Recently introduced: https://github.com/yugabyte/yugabyte-db/blob/master/managed/src/main/java/com/yugabyte/yw/models/CertificateInfo.java#L112)

jitendra-12113:
Fixed the indentations in line breaks. 
Moving common code to helper functions, and use JCAPemWriter for upload cert workflow - I think we can track this as separate PR. Thanks!

Arnav15 (commented):
Any particular reason why you don't want to do the refactor as part of this PR? It makes sense to modify it since it will essentially make the existing code that much more modular and you can use it directly for the purpose of this diff.

managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
```diff
@@ -275,12 +281,35 @@ public static UUID uploadRootCA(
...
+    InputStream in = null;
+    try {
+      X509Certificate cert = null;
+      byte[] certEntryBytes = certContent.getBytes();
+      in = new ByteArrayInputStream(certEntryBytes);
+      CertificateFactory certFactory = CertificateFactory.getInstance("X.509");
+      cert = (X509Certificate) certFactory.generateCertificate(in);
+      if (certType == CertificateInfo.Type.SelfSigned) {
+        boolean isValid = verifySignature(cert, keyContent);
+        if (!isValid) {
```
sb-yb, line 293: inline unnecessary variable

managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
```diff
@@ -275,12 +281,35 @@ public static UUID uploadRootCA(
...
+          throw new RuntimeException("Invalid certificate or Key.");
```
sb-yb, line 294: I think this will translate to 5XX? 
We would like 4XX here. No?

managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
```diff
@@ -275,12 +281,35 @@ public static UUID uploadRootCA(
...
+        }
+        keyPath = String.format("%s/certs/%s/%s/ca.key.pem", storagePath,
+            customerUUID.toString(), rootCA_UUID.toString());
+      } else {
+        boolean isValidCert = isValidCACert(cert);
+        if (!isValidCert) {
```
sb-yb, line 300: Again. Inline unnecessary variable.

managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
```diff
@@ -275,12 +281,35 @@ public static UUID uploadRootCA(
...
+          throw new RuntimeException("Invalid certificate.");
+        }
+      }
+    } catch (Exception e) {
+      throw new RuntimeException("Invalid certificate.");
```
sb-yb, line 305: If this is Runtime Exception: Don't catch if you are going to throw
If this is Checked Exception: wrap it in RuntimeException for better stack trace at calling site.

managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
```diff
@@ -275,12 +281,35 @@ public static UUID uploadRootCA(
...
+        }
+        keyPath = String.format("%s/certs/%s/%s/ca.key.pem", storagePath,
+            customerUUID.toString(), rootCA_UUID.toString());
+      } else {
+        boolean isValidCert = isValidCACert(cert);
+        if (!isValidCert) {
+          throw new RuntimeException("Invalid certificate.");
+        }
+      }
+    } catch (Exception e) {
```
sb-yb, line 304: Is there a specific checked exception that you must catch? If so list it explicitly.

managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
```diff
@@ -275,12 +281,35 @@ public static UUID uploadRootCA(
...
+      throw new RuntimeException("Invalid certificate.");
+    } finally {
+      if (in != null) {
```
sb-yb, line 307: Use this: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

jitendra-12113:
Created a separate PR to address above review comments. 
Review link : https://github.com/yugabyte/yugabyte-db/pull/7168/commits/5f8d39a0366fd8a0552f8d8d3d0b119509b82549
 Please review. Thanks!

Issue #1234, opened 2019-04-21
UNIQUE constraint error when using sequelize ORM


syedshah12tech:
Hi,

I am seeing the following error in the logs when using sequelize ORM with YugaByte:
`
sequelize:sql:pg Executing (default): CREATE TABLE IF NOT EXISTS "orderlines" ("id"   SERIAL , "orderId" UUID NOT NULL, "productId" INTEGER NOT NULL, "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL, UNIQUE ("orderId", "productId"), PRIMARY KEY ("id")); +3ms
Executing (default): CREATE TABLE IF NOT EXISTS "orderlines" ("id"   SERIAL , "orderId" UUID NOT NULL, "productId" INTEGER NOT NULL, "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL, UNIQUE ("orderId", "productId"), PRIMARY KEY ("id"));
  sequelize:pool connection released +2ms
Unhandled rejection SequelizeDatabaseError: UNIQUE constraint not supported yet
    at Query.formatError (/Users/syedshah/Documents/dev/orm-examples/node/sequelize/node_modules/sequelize/lib/dialects/postgres/query.js:354:16)
    at query.catch.err (/Users/syedshah/Documents/dev/orm-examples/node/sequelize/node_modules/sequelize/lib/dialects/postgres/query.js:71:18)
    at tryCatcher (/Users/syedshah/Documents/dev/orm-examples/node/sequelize/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/syedshah/Documents/dev/orm-examples/node/sequelize/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/Users/syedshah/Documents/dev/orm-examples/node/sequelize/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/Users/syedshah/Documents/dev/orm-examples/node/sequelize/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/Users/syedshah/Documents/dev/orm-examples/node/sequelize/node_modules/bluebird/js/release/promise.js:690:18)
[...]

PR #6792, opened 2021-01-05
[#5919] Platform: change default storage option for AWS provider to 1x250GB

jitendra-12113:
**Heading:**
[#5919] Platform: change default storage option for AWS provider to 1x250GB
**Description:**
The empty volume list will be stored in the db in that case we will be using default value(volume_count, volume_size configured in refernce.conf. 
**Testing:**
When we select aws provider while creating universe, volume count was showing more than one for some instances types starting with ""c5./c4.". Now with this fix it will show only one.
Will have to reset instanceTypeDetailsJson to empty for older YW
instances.
Added UT's as well. sbt test run results:

[info] Test com.yugabyte.yw.commissioner.TaskGarbageCollectorTest.testPurge_noneStale started
[info] Test com.yugabyte.yw.commissioner.TaskGarbageCollectorTest.testIgnorePromError started
[info] Test com.yugabyte.yw.commissioner.TaskGarbageCollectorTest.testPurge_invalidData started
[info] Test run finished: 0 failed, 0 ignored, 7 total, 0.589s
[info] Passed: Total 1092, Failed 0, Errors 0, Passed 1088, Skipped 4
[success] Total time: 575 s, completed 18 Jan, 2021 12:33:44 PM
hasher@LAP-LIN-908:~/yugabyte-db/managed$


managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -44,6 +49,13 @@
     NVME
   }
 
+  @Inject
+  public InstanceType(ConfigFactory config) {
```
sb-yb, line 53: `Config` not `ConfigFactory`
Like I said use TaskGarbageCollector as example.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -29,6 +31,9 @@
 
   public static List<String> AWS_INSTANCE_PREFIXES_SUPPORTED = ImmutableList.of(
     "m3.", "c5.", "c5d.", "c4.", "c3.", "i3.");
+  static final String YB_VOLUME_INFO_VOLUME_COUNT = "yb.volumeInfo.volume_count";
+  static final String YB_VOLUME_INFO_VOLUME_SIZE_GB = "yb.volumeInfo.volume_size_gb";
+  private static ConfigFactory config;
```
sb-yb, line 36: why `static`?

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -179,6 +191,13 @@ public static void deleteInstanceTypesForProvider(Provider provider) {
       entries = entries.stream()
         .filter(supportedInstanceTypes(AWS_INSTANCE_PREFIXES_SUPPORTED))
         .collect(Collectors.toList());
+      for (InstanceType instanceType : entries) {
+        if (instanceType.instanceTypeDetailsJson.isEmpty()) {
+          instanceType.instanceTypeDetails.setVolumeDetailsList(
+            config.load().getInt(YB_VOLUME_INFO_VOLUME_COUNT),
```
sb-yb, line 197: Inject config. Then no need to `load`.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -29,6 +31,9 @@
 
   public static List<String> AWS_INSTANCE_PREFIXES_SUPPORTED = ImmutableList.of(
     "m3.", "c5.", "c5d.", "c4.", "c3.", "i3.");
+  static final String YB_VOLUME_INFO_VOLUME_COUNT = "yb.volumeInfo.volume_count";
```
sb-yb, line 34: YB_VOLUME_INFO_VOLUME_COUNT_KEY

sb-yb (changes requested):

sb-yb:
Also bug description should say what we are doing here.
Basically the fact that empty volume list is stored in db when we want to use config default.
Also Fix the "Heading" 
It should be following format 
`[#1234] Platform: One line description of change`

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -179,6 +191,13 @@ public static void deleteInstanceTypesForProvider(Provider provider) {
       entries = entries.stream()
         .filter(supportedInstanceTypes(AWS_INSTANCE_PREFIXES_SUPPORTED))
         .collect(Collectors.toList());
+      for (InstanceType instanceType : entries) {
```
sb-yb, line 194: extract method `populateDefaultIfEmpty()`

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: With Config,  always getting NLP(Null pointers), hence used ConfigFactory instead of.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: done

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: changed config var to non-static, others kept as it is, since those are being used in static block.

sb-yb: yep. my comment was only about config.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
sb-yb: Thats odd. In tests or when running server? May be some ordering issue during binding.
Inject 
`private final Provider<Config> config`
@Inject
InstanceType(Provider<Config> config) {
...
}`

then where you use it do 

`config.get()`

This delays the resolution to runtime.

Let me know if that fixes the problem. If not will patch and see whats wrong.

The config factory is not an option here. That code is buggy. 
Another way is to get to typesafe config through play but lets try this first.

managed/src/main/resources/reference.conf
```diff
@@ -13,4 +13,12 @@ yb {
     # For how long do we let the task be in database after it has completed
     task_retention_duration = 120 days
   }
+
+  volumeInfo {
+      # Volume count needed for AWS instance types
```
sb-yb, line 29: `# default volume count for aws instance types with EBS Only  storage info`

managed/src/main/resources/reference.conf
```diff
@@ -13,4 +13,12 @@ yb {
...
+      volume_count = 1
```
sb-yb, line 30: `default_`

managed/src/main/resources/reference.conf
```diff
@@ -13,4 +13,12 @@ yb {
...
+
+      # volume size in GB
+      volume_size_gb = 250
```
sb-yb, line 33: `default_`

managed/src/main/resources/reference.conf
```diff
@@ -13,4 +13,12 @@ yb {
     # For how long do we let the task be in database after it has completed
     task_retention_duration = 120 days
   }
+
+  volumeInfo {
```
sb-yb, line 28: `aws {
...
}`

sb-yb (changes requested):
Few more changes.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: Its happening while running server. 
Getting below error:
Caused by: com.google.inject.CreationException: Unable to create injector, see the following errors:
1) Error injecting constructor, java.lang.NullPointerException
  at AppInit.<init>(AppInit.java:47)
  at Module.configure(Module.java:48) (via modules: com.google.inject.util.Modules$OverrideModule -> Module)
  while locating AppInit
  
  Tried above things as well but that didn't work out. 
  Updated a patch, using config only. Please suggest! Thanks!

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: done

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: Done.

managed/src/main/resources/reference.conf
jitendra-12113: Done

managed/src/main/resources/reference.conf
jitendra-12113: Done

managed/src/main/resources/reference.conf
jitendra-12113: Done

managed/src/main/resources/reference.conf
jitendra-12113: Thanks! Done.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
sb-yb: So I saw a `provider` for `Config` in `Module`. But looks like that is a different Config type not the `typesafe.Config`. 
So it is not going to work this way. Provider trick also does not delay the resolution because we need it during AppInit.

So way to do this gracefully is 
 - Inject `Application` in `InstanceType`
  - Then you can get to typesafe config using `app.config()`  (NOT `app.configuration()` which is deprecated)
   `this.config = app.config()`
   
   Let me know if that helps.

sb-yb: Just noticed that InstanceType is a Entity bean. So it wont use guice injection. It is using the default ctor (that you added) followed by the field mapping by the ORM  and thus config remains null.
So you can just pass the config value into findByProvider as parameter. No config field.

Sorry for confusion. Right way to do this is not use so much static factory methods. Usually there is a singleton helper class like InstanceTypeHelper. That will have non-static findByProvider method. then you can actually do proper guice injection in such class.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
sb-yb: If this is the only config  value then pass a parameter in to this method `awsDefaultVolumeCount` 
Do config lookup in caller.
I will do a refactor later which should fix this DI fail as described above - using singleton helper classes.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: Done. Added this for now, later we will refactor the code once we have Helper method in place as you suggested above. Thanks!

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: Done!

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -185,6 +196,22 @@ public static void deleteInstanceTypesForProvider(Provider provider) {
       entry.getInstanceTypeCode())).collect(Collectors.toList());
   }
 
+  /**
+   * TODO: Please remove this method when we have helperMethod in place to get config details
+   * Query Helper to find supported instance types for a given cloud provider.
+   */
+  public static List<InstanceType> findByProvider(Provider provider, Config config) {
```
sb-yb, line 203: Do not make separate method. Just update same method.

sb-yb (changes requested):
Add unit tests

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: @sb-yb  I thought of doing this way but we will have to update it every place like AppInit.java and other places wherever  this method is being used. 
We know this is temp fix, anyway we are going to refactor the code. That's why created overloaded method to avoid changing everywhere.

managed/src/main/java/com/yugabyte/yw/controllers/InstanceTypeController.java
```diff
@@ -26,13 +27,20 @@
 public class InstanceTypeController extends AuthenticatedController {
 
   public static final Logger LOG = LoggerFactory.getLogger(InstanceTypeController.class);
+  private Config config;
 
   @Inject
   FormFactory formFactory;
 
   @Inject
```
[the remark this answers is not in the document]
jitendra-12113: Here we will have to use Inject, Without inject we are getting below error.

1) No implementation for com.yugabyte.yw.controllers.InstanceTypeController (with no qualifier annotation) was bound, and could not find an injectable constructor. Injectable classes must have either one (and only one) constructor annotated with @Inject or a zero-argument constructor that is not private.

managed/src/main/java/com/yugabyte/yw/controllers/InstanceTypeController.java
```diff
@@ -26,13 +27,20 @@
 public class InstanceTypeController extends AuthenticatedController {
 
   public static final Logger LOG = LoggerFactory.getLogger(InstanceTypeController.class);
+  private Config config;
```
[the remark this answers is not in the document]
jitendra-12113: done

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -185,6 +196,22 @@ public static void deleteInstanceTypesForProvider(Provider provider) {
       entry.getInstanceTypeCode())).collect(Collectors.toList());
   }
 
+  /**
```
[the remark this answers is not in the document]
jitendra-12113: Updated the same method instead of having overloaded method.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
jitendra-12113: Updated! Thank you!

sb-yb (changes requested):
@jitendra-12113 anything holding up this change? Do you need help with any part?

jitendra-12113:
> @jitendra-12113 anything holding up this change? Do you need help with any part?

@sb-yb  Have updated the patch. I think we are good to go ahead and merge this change. Need your approval though. Thanks!

managed/src/main/java/com/yugabyte/yw/controllers/InstanceTypeController.java
```diff
@@ -26,13 +27,20 @@
...
   @Inject
   FormFactory formFactory;
 
   @Inject
   CloudAPI.Factory cloudAPIFactory;
 
+  // TODO: Remove this when we have HelperMethod in place to get Config details
+  @Inject
+  public InstanceTypeController(Config config) {
+    this.config = config;
```
sb-yb, line 41: Do not mix ctor injections and field injections in same class.
Either fix all existing fields to be parameters here or make the new field use field injection.

FWIW, using ctor injection with private fina fields is the best practice. But i ll leave it to you which option you want for now.

managed/src/main/java/com/yugabyte/yw/models/InstanceType.java
```diff
@@ -179,6 +190,9 @@ public static void deleteInstanceTypesForProvider(Provider provider) {
       entries = entries.stream()
```
sb-yb, line 190: Instead consider moving the whole code (line 190 to 196) in populateDefaultIfEmpty() 
And name the method populateDefaultsIfEmpty (plural)

managed/src/test/java/com/yugabyte/yw/models/InstanceTypeTest.java
```diff
@@ -29,6 +31,9 @@
   private Customer defaultCustomer;
   private InstanceType.InstanceTypeDetails defaultDetails;
 
+  @Mock
```
sb-yb, line 34: where is a test where you have to setup this mock?
`when(mockConfig.get(..._COUNT)).thenReturn(1);`
`when(mockConfig.get(..._GB)).thenReturn(250);`

sb-yb (changes requested):
Final lap:
small code change and missing unit test

managed/src/main/java/com/yugabyte/yw/controllers/InstanceTypeController.java
jitendra-12113: Done. Thanks!

managed/src/test/java/com/yugabyte/yw/models/InstanceTypeTest.java
jitendra-12113: Ah! Didn't realise this is for entity class. We can't mock config in this file. Most of the values are being read from in memory.
Instead used ConfigFactory.load() just to test our code. I think, this should be fine if not please suggest! 
Also due to static method used in Entity class, getting difficulties in writing Unit test cases, need to refactor the code(Entity class). But for now I think this should be fine. 

Thanks!

managed/src/test/java/com/yugabyte/yw/models/InstanceTypeTest.java
```diff
@@ -90,7 +94,7 @@ public void testFindByProvider() {
     instanceType.setActive(false);
     instanceType.save();
     InstanceType.upsert(newProvider.code, "bar", 2, 10.0, defaultDetails);
```
sb-yb, line 96: This is incorrect. Not sure what is your confusion.
Here is what you have to do in great detail:
1. Revert to using mock.Config 
2. See what is in defaultDetails  - line 44 populates it. But we do not want to test that code path. To excercise our code path the defaultDetails has to be empty. Thats when the populateEmpty will do something and try to lookup config values.
3. Like line 88 upsert another instancetype  such that its details have volumeList empty.
4. Now the test will fail complaining that you have not mocked config.getInt or whatever you use. 
5. Add mock setup. For example volume count config return 3 and size returns 100GB.
6.  assert that the volume list for this instance as returned by findPyProvider is size 3 and each one is 100GB.

Hope it helps.

sb-yb (changes requested):
Given detailed instructions inline. LMK if it helps.

managed/src/test/java/com/yugabyte/yw/models/InstanceTypeTest.java
jitendra-12113: Had to refactor the code lil bit. Forgot to apply @RunWith(MockitoJUnitRunner.class) on top to use mock. Now its working fine. I really learnt a lot while working on this change.  Please let me know if you still thinks, some changes are required. Thanks for all your suggestions:)

sb-yb: 💯  for right attitude!

sb-yb (approved):
This looks great now! 
Important tips:
Make sure to run the unittests before submit. If any fail take it seriously and rerun failed tests. 
If you run sbt shell then you can do `testQuick` and most of the times that only re-runs failed tests.  Or you can use testOnly for specific tests. There are few flaky tests so you can ignore if they pass in subsequent run. If not then take it seriously and post on slack channel for any debugging help. Sometime there are broken tests at HEAD. Not ideal situation, I know.  But given that it is easier to not spend too much time debugging and just ask around (I learned painfully :( ..)

Issue #6460, opened 2020-11-23
Preflight checks for cloud provider configuration


chirag-yb:
Add pre-flight checks for cloud provider configuration
 -- ssh key
-- cloud credentials

In the case of validation failure, notify the user via the TASK window, and disable universe creation for that cloud provider.

jitendra-12113:
@chirag-yb  Could you please provide us the steps to repro this issue?
I think, it should be applicable to all cloud provider configuration? 
Appreciate your help! Thanks!

cc: @streddy-yb

chirag-yb:
@jitendra-12113 Yes this will apply to all cloud providers. This can be easily reproduced by 

- Use invalid cloud credentials (AWS access keys, GCP config JSON, Azure client Id/secret)  OR
- Use loud credentials without proper access control. User may not have enough roles/permissions in a cloud provider to create VMs/VPCs OR
- Provide an ssh key which is the incorrect format (AWS only)

Right now there is no validation and a user can continue to the universe creation which may fail due to lack of access control or invalid data provided. 

New proposed experience

1. During the cloud provider creation add a validation.
2. If validation fails, report the failure on the Tasks page
3. Cloud provider is not available for creating new universes

jaydeepkumara:
For now, we will be fixing this for AWS.

jitendra-12113:
Have handled only for AWS cloud provider. Please review. 
https://github.com/yugabyte/yugabyte-db/pull/6808

UI changes also need to be done accordingly.

Thanks!

jitendra-12113:
Created another PR to track this change. Please review
https://github.com/yugabyte/yugabyte-db/pull/6948/

chirag-yb:
@jitendra-12113 how are we tracking the UI change? 
Also, does your PR also cover the scenario when the cloud provider fails to create an ssh key?

`ybops.common.exceptions.YBOpsRuntimeError: Runtime error: KeyPair already exists yb-dev-aws-key
2021-02-10 05:16:41,866 [ERROR] from com.yugabyte.yw.commissioner.SubTaskGroup in TaskPool-0 - Failed to execute task java.util.concurrent.FutureTask@3498b47d, hit error java.lang.RuntimeException: YBCloud command access (add-key) failed to execute..
java.util.concurrent.ExecutionException: java.lang.RuntimeException: YBCloud command access (add-key) failed to execute
 `

jitendra-12113:
[...]

PR #6808, opened 2021-01-07
[#6460]Platform: Preflight checks for cloud provider configuration

jitendra-12113:
**Heading:**
[#6460]Platform: Preflight checks for cloud provider configuration
**Description:**
We need to validate the AWS credentials at the time of cloud provider configuration
**Testing:**
Added Unit tests. We can test this by passing invalid aws access key or secret key.
For now, we have handled only for AWS provider.
TODO - need to add this check for all the cloud provider

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/cloud/CloudAPI.java
+++ b/managed/src/main/java/com/yugabyte/yw/cloud/CloudAPI.java
@@ -44,4 +44,12 @@ public interface CloudAPI {
     Provider provider,
     Map<Region, Set<String>> azByRegionMap,
     Set<String> instanceTypesFilter);
+
+  /**
+   * Check whether cloud provider's credentials are valid or not.
+   *
+   * @param config The credentials info.
+   * @return true if credentials are valid otherwise return false.
+   */
+  boolean isValidCredentials(Map<String, String> config);
 }
--- a/managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
+++ b/managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
@@ -2,13 +2,15 @@ package com.yugabyte.yw.cloud.aws;
 
 import com.amazonaws.auth.*;
 import com.amazonaws.services.ec2.AmazonEC2;
+import com.amazonaws.services.ec2.AmazonEC2Client;
 import com.amazonaws.services.ec2.AmazonEC2ClientBuilder;
+import com.amazonaws.services.ec2.model.DescribeRegionsResult;
+import com.amazonaws.regions.Regions;
 import com.amazonaws.services.ec2.model.*;
 import com.google.common.base.Strings;
 import com.yugabyte.yw.cloud.CloudAPI;
 import com.yugabyte.yw.models.Provider;
 import com.yugabyte.yw.models.Region;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -16,10 +18,13 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import static java.util.stream.Collectors.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 //TODO - Better handling of UnauthorizedOperation. Ideally we should trigger alert so that
 // site admin knows about it
 class AWSCloudImpl implements CloudAPI {
+  public static final Logger LOG = LoggerFactory.getLogger(AWSCloudImpl.class);
 
   // TODO use aws sdk 2.x and switch to async
   public AmazonEC2 getEcC2Client(Provider provider, Region r) {
@@ -32,6 +37,25 @@ class AWSCloudImpl implements CloudAPI {
       .withCredentials(credentialsProvider)
       .build();
   }
+
+  @Override
+  public boolean isValidCredentials(Map<String, String> config) {
+    BasicAWSCredentials awsCreds = new BasicAWSCredentials(config.get("AWS_ACCESS_KEY_ID"),
+      config.get("AWS_SECRET_ACCESS_KEY"));
+    try {
+      AmazonEC2 ec2Client = AmazonEC2ClientBuilder
+        .standard()
+        .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
+        .withRegion(Regions.US_EAST_1)
+        .build();
+      DescribeRegionsResult regions_response = ec2Client.describeRegions();
+      return true;
+    } catch (Exception e) {
+      LOG.error(e.getMessage());
+      return false;
+    }
+  }
+
 
   // TODO: move to some common utils
   private static AWSCredentialsProvider getCredsOrFallbackToDefault(String accessKeyId,
--- a/managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.yugabyte.yw.cloud.AWSInitializer;
 import com.yugabyte.yw.cloud.GCPInitializer;
+import com.yugabyte.yw.cloud.CloudAPI;
 import com.yugabyte.yw.commissioner.Commissioner;
 import com.yugabyte.yw.commissioner.Common;
 import com.yugabyte.yw.commissioner.tasks.CloudBootstrap;
@@ -93,6 +94,9 @@ public class CloudProviderController extends AuthenticatedController {
   @Inject
   private play.Environment environment;
 
+  @Inject
+  CloudAPI.Factory cloudAPIFactory;
+
   /**
    * GET endpoint for listing providers
    * @return JSON response with provider's
@@ -172,6 +176,10 @@ public class CloudProviderController extends AuthenticatedController {
       if (!config.isEmpty()) {
         switch (provider.code) {
           case "aws":
+            CloudAPI cloudAPI = cloudAPIFactory.get(provider.code);
+            if (cloudAPI != null && !cloudAPI.isValidCredentials(config)) {
+              return ApiResponse.error(BAD_REQUEST, "Invalid AWS Credentials.");
+            }
             String hostedZoneId = provider.getAwsHostedZoneId();
             if (hostedZoneId != null) {
               return validateAwsHostedZoneUpdate(provider, hostedZoneId);
--- a/managed/src/test/java/com/yugabyte/yw/controllers/CloudProviderControllerTest.java
+++ b/managed/src/test/java/com/yugabyte/yw/controllers/CloudProviderControllerTest.java
@@ -10,6 +10,7 @@ import static com.yugabyte.yw.common.AssertHelper.assertValues;
 import static com.yugabyte.yw.common.AssertHelper.assertAuditEntry;
 import static com.yugabyte.yw.common.TestHelper.createTempFile;
 import static com.yugabyte.yw.common.ModelFactory.createUniverse;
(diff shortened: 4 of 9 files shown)


managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
```diff
@@ -33,6 +35,24 @@ public AmazonEC2 getEcC2Client(Provider provider, Region r) {
...
 
+  @Override
+  public boolean isAWSValidCredentials(Map<String, String> config) {
+    BasicAWSCredentials awsCreds = new BasicAWSCredentials(config.get("AWS_ACCESS_KEY_ID"),
+      config.get("AWS_SECRET_ACCESS_KEY"));
+    try {
+      AmazonEC2 ec2Client = AmazonEC2ClientBuilder
+        .standard()
+        .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
+        .withRegion(Regions.US_EAST_1)
```
jaydeepkumara, line 46: why specific to US_EAST_1?

managed/src/main/java/com/yugabyte/yw/cloud/CloudAPI.java
```diff
@@ -44,4 +44,12 @@ public CloudAPI get(String code) {
...
     Map<Region, Set<String>> azByRegionMap,
     Set<String> instanceTypesFilter);
+
+  /**
+   * Check whether AWS credentials(AWS access key and Secret Key) are valid or not.
+   *
+   * @param config The AWS credentials info.
+   * @return true if credentials are valid otherwise return false.
+   */
+  boolean isAWSValidCredentials(Map<String, String> config);
```
SergeyPotachev, line 54: As CloudAPI.java is a common class (it is situated outside of the `aws` folder) then it is assumed to have no specific methods (related to exact cloud types). Let's imagine that this interface is implemented for Azure, as example, which also has a check for valid credentials.
My proposal - to rename the function, to remove AWS mentions from the comment.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
```diff
@@ -33,6 +35,24 @@ public AmazonEC2 getEcC2Client(Provider provider, Region r) {
       .build();
   }
 
+  @Override
+  public boolean isAWSValidCredentials(Map<String, String> config) {
+    BasicAWSCredentials awsCreds = new BasicAWSCredentials(config.get("AWS_ACCESS_KEY_ID"),
```
SergeyPotachev, line 40: See how credentials are taken in `AWSCloudImpl.getEcC2Client()` (using AWSCredentialsProvider).

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
SergeyPotachev, line 46: Why the region is hardcoded?

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
```diff
@@ -33,6 +35,24 @@ public AmazonEC2 getEcC2Client(Provider provider, Region r) {
...
+    try {
+      AmazonEC2 ec2Client = AmazonEC2ClientBuilder
+        .standard()
+        .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
+        .withRegion(Regions.US_EAST_1)
+        .build();
+      DescribeRegionsResult regions_response = ec2Client.describeRegions();
+      return true;
+    } catch (Exception e) {
+      return false;
```
SergeyPotachev, line 51: What if we get an exception not related to incorrect credentials?

SergeyPotachev (changes requested):

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: I am assuming that won't be the case here. Could you please point me to such scenarios? Anyway, if we get any exceptions for whatever reasons;that means we are unable to validate the credentials and we won't  let user to proceed(This will be very rare).

cc: @sb-yb

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: Idea here is to validate the credentials by connecting to ec2. If we are able to connect it means credentials are valid else we will get some exceptions(Returning false for such scenarios)

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: Here we will have credentials before hand, we don't need to fallbackToDefault. 
BasicAWSCredentials serves the purpose here.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
SergeyPotachev: Simple example, I think it could happen for any cloud, - connection timeout (AWS network is unreachable or something like this). So, at least, you should log the exception (reason why the credentials are failed). But logging the exception be careful - the used credentials should not go into logs (otherwise it is a security hole).

managed/src/main/java/com/yugabyte/yw/cloud/CloudAPI.java
jitendra-12113: Make sense! I had seen CloudAPI.java was not being used anywhere other than AWSCloudImpl.java.

Will rename the method as you suggested! Thanks!

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: Sounds good! Will log a exception by considering security hole. Thanks!

managed/src/main/java/com/yugabyte/yw/cloud/CloudAPI.java
jitendra-12113: Done

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: Done

jitendra-12113: This log.Error message won't contain provided credentials so we are good at that.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
SergeyPotachev: OK, but anyway extract constants "AWS_ACCESS_KEY_ID" and "AWS_SECRET_ACCESS_KEY".

sb-yb: I think @SergeyPotachev has a point here.
Shouldn't the purpose of preflight check be to ensure that AWSCloudImpl.getEcC2Client() will work?
i.e. it should actually use the exact same way of getting creds and then make sure it works. For example if user has not setup access key and selected IAM profile option. This preflight check will fail even though the actual code is going to work. 
What am I missing?

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
SergeyPotachev: This particular region could be disabled for the customer's account (see description of describeRegions). So the check will fail, I suppose. But actually the credentials could be good for another region used by the customer.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
sb-yb: i.e. should the method be more generic `isValidCreds` or even `canConnectToCloudProvider` as opposed to are these access keys valid. That leads to reusable and maintainable code where your preflight checks will use exact same code to create awsclient  that you are going to use for actual work during flight.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
sb-yb: Yep. IIUC we should run against all the regions that have been selected in configuration for the provider and make sure creds work  for all. But need to understand how it all fits in the UI.  
i.e. we can validate config as a whole before saving it. Not just access key alone.

sb-yb:
Also ideally we should also check that these creds have permissions that we need. Maybe sending `dryrun` requests. But that can happen in other PR.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
sb-yb: I think here you should use dry run. 
See https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/ec2/AmazonEC2Client.html#dryRun-com.amazonaws.services.ec2.model.DryRunSupportedRequest-

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
sb-yb: If you remove withRegions it will use default region. That is better than explicitly selecting a region. 
But like I said ideally run dry run commands against all the regions in the provider config - may be that can happen in a separate PR.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: Yeah! My bad, missed IAM things in that case, we should fallback to default like mentioned above. Thanks for this catch.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: @sb-yb  I see, we don't have regions in the provider config(they are calling /bootstrap API to save these details. )
Running dry run commands against all the regions in the provider config may increase API response time(Performance issue)

Trying to perform some other operations without regions.

jitendra-12113: Without specifying any regions, getting below error.
2021-01-12 00:18:35.966 ERROR AWSCloudImpl.java:56 [application-akka.actor.default-dispatcher-10] Unable to find a region via the region provider chain. Must provide an explicit region in the builder or setup environment to supply a region.

We needed regions in the backend.  UI has to send the regions details as part of POST /providers API payload. 
Please suggest!

sb-yb: These are my suggestions:
1. Pick one region from provider/universe config (depending on context)
2. Basically the preflight check is holistic like I described above. 
3. Try to write code so as to reuse production client creation instead of doing something different.
4. Make DryRun request. Then you can also make mutate requests and get concrete permissions failure information that you can then expose to the user. So that they know whats wrong rather than "something is wrong" - without us having to look at the logs.

jitendra-12113:
This has been addressed here #6948. Closing this PR. 
created separate PR(#6948) on forked branch instead of master.

PR #6810, opened 2021-01-07
modified the add node platform page

epratt-yb:
The add node platform page refers to more of a replace node rather than adding a node fresh.  Adjusted the page to reflect how to add a node via the platform.


yb-andrew:
@epratt-yb The changes look fine but I think the original content is useful to be preserved on a different page.

epratt-yb:
> @epratt-yb The changes look fine but I think the original content is useful to be preserved on a different page.

epratt-yb:
@yb-andrew yea I think it would be better moved to another page that has a replace node title I think.  I can adjust all of it and commit back again later if we think that's the better route

Issue #6814, opened 2021-01-08
No Warning/Error if selected node is greater than available nodes for On-Premises provider


gaurav061:
1. Setup a Provider for on-Premises with 1 Node.
2. Try creating a universe with the provider as on-prem(By Default Replication factor and node will be set to 3).
3. AZ configuration will not appear as selected nodes are greater than available nodes.

gaurav061:
PR: https://github.com/yugabyte/yugabyte-db/pull/6815

Issue #6760, opened 2020-12-29
[platform] Avaibility zone waring message and error message not reseting if changed provider.


jaydeepkumara:
Steps:-
1. Configure 2 providers.
2. Go and try to create the universe.
3. Try to get an error message/warning message for the Availability Zone.
4. Change Provider.
5. Error message or warning message is still showing without any Availability Zone.

### Warning message is still showing
![Screenshot from 2020-12-29 20-53-13](https://user-images.githubusercontent.com/16913155/103294564-1cee9380-4a18-11eb-9458-9d233a0f8274.png)

### Error message is still appearing 
![Screenshot from 2020-12-29 20-54-09](https://user-images.githubusercontent.com/16913155/103294633-3d1e5280-4a18-11eb-981f-19f7566508d1.png)

nishantSharma459:
I'm putting this ticket on the ToDo list. I've figured there're lots of issues with the Availability Zone. I need to get deep into the code to debug the issue. So, in that regard will take it up next week.

gaurav061:
PR:https://github.com/yugabyte/yugabyte-db/pull/6815

Issue #6759, opened 2020-12-29
[platform]Create universe is disabled if we try invalid then valid configuration


jaydeepkumara:
Steps:-
1. Setup AWS provider properly
2. Setup On-premise provider but don't attach any instance to this provider.
3. Got Universe page and click on create Universe.
4. Select provider as On-premise and select region, change node to 1, and Replication Factor as 1. Availability Zone will not appear and the create button is disabled (valid scenario)
5. Change provider to AWS and change the region, Now this is a valid configuration still create button is disabled.

### Image1
![Screenshot from 2020-12-29 20-41-19](https://user-images.githubusercontent.com/16913155/103293683-57573100-4a16-11eb-9580-b0edec6878a5.png)

### Image2
![Screenshot from 2020-12-29 20-41-56](https://user-images.githubusercontent.com/16913155/103293714-663de380-4a16-11eb-94aa-c2d948445680.png)

### Image3
![Screenshot from 2020-12-29 20-41-38](https://user-images.githubusercontent.com/16913155/103293700-5e7e3f00-4a16-11eb-9b93-b87c89921b58.png)

gaurav061:
PR: https://github.com/yugabyte/yugabyte-db/pull/6815

PR #6815, opened 2021-01-08
[#6814,#6759,#6760] Platform: Fixed availability zone error scenarios

gaurav061:


gaurav061:
Please have a look and add your comments.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -574,13 +575,13 @@ export default class ClusterFields extends Component {
     //hook from parent universeForm to check if any fields was changed
     const nodeDetailsSet =
       getPromiseState(currentUniverse).isSuccess() &&
-      getPromiseState(universeConfigTemplate).isSuccess()
+        getPromiseState(universeConfigTemplate).isSuccess()
```
sshev, line 578: no need for extra indentation here

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -574,13 +575,13 @@ export default class ClusterFields extends Component {
...
         ? universeConfigTemplate.data.nodeDetailsSet
         : [];
     if (type === 'Edit' || (this.props.type === 'Async' && this.state.isReadOnlyExists)) {
       this.props.handleHasFieldChanged(
         this.hasFieldChanged() ||
-          !_.isEqual(currentUniverse.data.universeDetails.nodeDetailsSet, nodeDetailsSet)
+        !_.isEqual(currentUniverse.data.universeDetails.nodeDetailsSet, nodeDetailsSet)
```
sshev, line 584: let's keep original formatting

managed/ui/src/components/universes/UniverseForm/AZPlacementInfo.js
```diff
@@ -53,48 +53,53 @@ export default class AZPlacementInfo extends Component {
...
+          if (replicationFactor % 2 === 0) {
+            if (replicationFactor / 2 < az.replicationFactor) {
+              multiAz = false;
+            }
+          } else {
+            if ((replicationFactor - 1) / 2 < az.replicationFactor) {
+              multiAz = false;
+            }
+          }
+        });
```
sshev, line 88: @daniel-yb pls take a look, I remember it's you added the logic above

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1496,23 +1497,40 @@ export default class ClusterFields extends Component {
         });
     }
 
-    let placementStatus = <span />;
+    let placementStatus, placementStatusOnprem = null;
```
sshev, line 1500: let's declare single var per lane

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1496,23 +1497,40 @@ export default class ClusterFields extends Component {
...
           placementCloud={placementCloud}
+          providerCode={currentProvider.code}
+        />
+      );
+    } else if (currentProvider?.code === 'onprem' &&
+      !(this.state.numNodes <= this.state.maxNumNodes)
+      && self.props.universe.currentPlacementStatus
+      && isNonEmptyArray(formValues[clusterType].regionList)
+      && isNonEmptyString(formValues[clusterType].provider)) {
+      placementStatusOnprem = (
```
sshev, line 1527: move `) {` to the new line pls for better readability:
```suggestion
      && isNonEmptyString(formValues[clusterType].provider)
    ) {
      placementStatusOnprem = (
```

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1707,7 +1725,8 @@ export default class ClusterFields extends Component {
               )}
             </Col>
             <Col md={6} className={'universe-az-selector-container'}>
-              {azSelectorTable}
+              {(isNonEmptyArray(formValues[clusterType]?.regionList) && isNonEmptyString(formValues[clusterType]?.provider))
+                && placementStatusOnprem ? placementStatusOnprem : (isNonEmptyArray(formValues[clusterType]?.regionList) && isNonEmptyString(formValues[clusterType]?.provider)) ? azSelectorTable : null}
```
sshev, line 1729: poor readability, pls reformat

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1496,23 +1497,42 @@ export default class ClusterFields extends Component {
...
           placementInfo={self.props.universe.currentPlacementStatus}
           placementCloud={placementCloud}
+          providerCode={currentProvider.code}
+        />
+      );
+    } else if (currentProvider?.code === 'onprem' &&
+      !(this.state.numNodes <= this.state.maxNumNodes)
+      && self.props.universe.currentPlacementStatus
+      && isNonEmptyArray(formValues[clusterType].regionList)
+      && isNonEmptyString(formValues[clusterType].provider)
```
yb-andrew, line 1527: Would be better to move these two lines into a separate variable because I see this code being repeated below.

yb-andrew (changes requested):

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1707,7 +1727,14 @@ export default class ClusterFields extends Component {
...
             <Col md={6} className={'universe-az-selector-container'}>
-              {azSelectorTable}
+              {(isNonEmptyArray(formValues[clusterType]?.regionList)
+                && isNonEmptyString(formValues[clusterType]?.provider)
+              )
+                && placementStatusOnprem ? placementStatusOnprem
+                : (isNonEmptyArray(formValues[clusterType]?.regionList)
+                  && isNonEmptyString(formValues[clusterType]?.provider)
+                ) ? azSelectorTable : null
+              }
```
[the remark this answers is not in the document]
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/AZPlacementInfo.js
gaurav061: @daniel-yb Please add your inputs here.

daniel-yb: @yb-andrew / @sshev you'd know this better than I do, but aren't all numbers in javascript effectively a double? I guess you could use `Math.floor(replicationFactor / 2) < az.replicationFactor` or something like this to achieve the same behavior while simplifying the two cases of even/odd here. Also now that I think of it, since az.replicationFactor will always be a whole number, we don't even really need the if/else at all and can go with Andrew's recommendation. Otherwise this change looks fine to me; it shouldn't change the results of the calculations here, it should only skip them if the regionList is not null. You could also just do `const regionList = _.get(placementCloud, "regionList", []);` above instead of adding this new `if (regionList) {...}` case (which IMO would be cleaner since it doesn't indent a whole block of code under another if statement).

yb-andrew: @daniel-yb Numbers in Javascript can be int, floats, unsigned ints, unsigned floats, or big ints. You don't need the Math.floor if the `replicationFactor` is an int because dividing by another int will truncate the value (e.g. 5 / 2 = 2).
@gaurav061 You can use the suggestion of simplifying the even/odd logic cases but just add a code comment for clarity that `replicationFactor` is an int.

yb-andrew: @daniel-yb I'm incorrect here. Numbers in JS default to double like you said unless initializing with a different class (unsigned, bigint).

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1512,23 +1513,43 @@ export default class ClusterFields extends Component {
...
         );
     const placementCloud = getPlacementCloud(cluster);
-    if (self.props.universe.currentPlacementStatus && placementCloud) {
+    const regionAndProviderDefined = isNonEmptyArray(formValues[clusterType]?.regionList)
+      && isNonEmptyString(formValues[clusterType]?.provider);
+
+    // For onprem provider type if numNodes < maxNumNodes then show the AZ error.
+    if ((
+      self.props.universe.currentPlacementStatus && placementCloud
+      && regionAndProviderDefined)
```
sshev, line 1531: it's `if ((...))` - no need in second `()` here

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1512,23 +1513,43 @@ export default class ClusterFields extends Component {
...
       placementStatus = (
         <AZPlacementInfo
           placementInfo={self.props.universe.currentPlacementStatus}
           placementCloud={placementCloud}
+          providerCode={currentProvider.code}
+        />
+      );
+    } else if (currentProvider?.code === 'onprem' &&
+      !(this.state.numNodes <= this.state.maxNumNodes)
+      && self.props.universe.currentPlacementStatus
```
sshev, line 1542: let's be consistent on where to put `&&`

sshev (approved):
minor formatting notes, otherwise looks good to me

managed/ui/src/components/universes/UniverseForm/AZPlacementInfo.js
```diff
@@ -53,13 +54,16 @@ export default class AZPlacementInfo extends Component {
...
+    let currentStatusType;
+
+    // If placementInfo and placementCloud is empty and provier is not onprem then return
+    if ((!isNonEmptyObject(placementInfo) || !isNonEmptyObject(placementCloud)) && providerCode !== "onprem") {
       return <span />;
     }
 
     const replicationFactor = placementInfo.replicationFactor;
-    const regionList = placementCloud.regionList;
+    const regionList = _.get(placementCloud, "regionList", []);
```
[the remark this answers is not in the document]
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
gaurav061: Done

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
gaurav061: Done

sshev (approved):
👍

PR #6885, opened 2021-01-15
[#6540]  Platform: Remove certificate feature.

### LINKED ISSUE #6540: Add ability to remove certificates from UI

mahendranbhat:
**Description**: Remove certificate feature which will be added in the list view of certificates to delete unused certificates.
If the certificate is linked to the existing universe then we will be disabling the button in the FE.

**Testing**: :

1. Go to certificates 
2. Try to delete the certificate which is not linked to any universe.
3. Also we can not see the delete option for certs which is linked to the universe.

**sbt run**
```
[info] Passed: Total 1186, Failed 0, Errors 0, Passed 1182, Skipped 4
```


managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -128,12 +129,35 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
     }
   }
 
-  public Result list(UUID customerUUID) {
+	public Result list(UUID customerUUID) {
```
Arnav15, line 132: @daniel-yb Recently gave a great suggestion on one of my diffs to check the `inUse` of a certificate in the model itself. 
So, add a function as following in the CertificateInfo model class:
```public boolean getInUse() {
   // Do the check for the cert being in use
}
```
What this will do is that it will automatically populate the `isUse` field for all objects of this class.
So we won't need all this excess code in both list and delete. You can directly check the field.

Arnav15 (changes requested):
Another thing, we use the google style guide for code formatting, which means [line breaks need atleast a 4 space indentation](https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent)

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: @Arnav15 agree,  if this gonna be used in multiple places then moving it to the model makes sense. If not, Is the helper method is sufficient in the controller itself? @daniel-yb

Arnav15: Hey @mahendranbhat, look at this [commit](https://github.com/yugabyte/yugabyte-db/commit/9b879be1d169d90131dfb62814b138d9256cad48). This is essentially achieving a very similar goal to what this PR is for. So I'd definitely recommend going with an approach similar to what I did. Move the `inUse` check to the model, add a new function in the Universe model to find universes with that certificate, and return if it exists, and then accordingly mark it as `inUse` or not. Ideally, we'd want to add foreign keys for all such kinds of checks, but I guess for now this should be good enough.

Arnav15 (changes requested):
Also, it is important to correctly add the description as I had mentioned earlier.
Add the sections:
Description: <You already have that>
Testing: How you tested it, and the summary lines of the `sbt test` command. Should look like: 
```
[info] Passed: Total 1115, Failed 0, Errors 0, Passed 1111, Skipped 4
```

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+   * Delete the root certificate if certificate is safe to remove,
+   * i.e - Certificate is not attached to any universe for current user.
+   * 
+   * @param certificateUUID Unique id of certificate.
+   */
+  deleteRootCertificate = (certificateUUID) => {
+    api.deleteCertificate(certificateUUID).then(
+      () => {
+        this.getCertificateArray()
+      }
```
sshev, line 131:
```suggestion
      () => this.getCertificateArray()
```

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+    ).catch(
+      err => {
+        console.log("Error", err)
+      }
```
sshev, line 135:
```suggestion
      err => console.error(`Failed to delete certificate ${certificateUUID}`, err)
```

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+    )
+  };
+
+  /**
+   * Lifecycle method to fetch iniial data required by compoenent
```
sshev, line 140: typos

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+   */
+  componentDidMount() {
+    this.getCertificateArray();
+  }
+
+  /**
+   * Fetch certificates attched to current user.
+   */
+  getCertificateArray = () => {
+    api.getCertificates().then(
```
sshev, line 151: we are using `redux` at legacy pages and this change deviates from the existing data flow
certs routines are in `CertificatesContainer.js` so let's reuse it please

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -116,8 +118,61 @@ class Certificates extends Component {
...
+            expiryDate: cert.expiryDate,
+            certificate: cert.certificate,
+            creationTime: cert.startDate,
+            privateKey: cert.privateKey,
+            customCertInfo: cert.customCertInfo,
+            inUse: cert.inUse
+          };
+        })
+        : [];
+      this.setState({ certificateArray: certificateArray})
```
sshev, line 168:
```suggestion
      this.setState({ certificateArray });
```
and actually no need to store certs list in the local state as it comes as prop from the redux store

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -153,32 +208,26 @@ class Certificates extends Component {
         >
           <i className="fa fa-download"></i> Download Root CA Cert
         </MenuItem>
+        <MenuItem
+          onClick={() => { 
+            !deleteDisabled && this.deleteRootCertificate(payload?.uuid)
```
sshev, line 213: I feel like `!deleteDisabled` check should happen in `deleteRootCertificate()`
and it's actually redundant as a button will be disabled anyways

managed/ui/src/components/certificates/Certificates.js
```diff
@@ -153,32 +208,26 @@ class Certificates extends Component {
...
+          }}
+          disabled={deleteDisabled}
+        >
+          <i className="fa fa-trash"></i> Delete Root CA Cert
```
sshev, line 217: it's generic cert delete, so `root CA` shouldn't be in the label

sshev (changes requested):
need some refinements on UI side

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -130,10 +131,27 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
 
   public Result list(UUID customerUUID) {
     List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
+    HashSet<String> inUseUUIDs = new HashSet<String>();
+    for (Universe universe : Universe.getAll()) {
+      Universe universe_obj = Universe.get(universe.universeUUID);
```
daniel-yb, line 136: variables should be in camel case (i.e. universeObj).

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -130,10 +131,27 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
...
+      if (universe_obj.getUniverseDetails().rootCA != null) {
+        UUID cert = universe_obj.getUniverseDetails().rootCA;
+        inUseUUIDs.add(cert.toString());
+      }
+    }
+    JsonNode certificates = Json.toJson(certs);
+    for (JsonNode cert : certificates) {
+      JsonNode cert_uuid = cert.get("uuid");
```
daniel-yb, line 144: cert_uuid => certUuid

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +163,34 @@ public Result get(UUID customerUUID, String label) {
...
 
+  public Result delete(UUID reqCertUUID) {
+    CertificateInfo cert = CertificateInfo.get(reqCertUUID);
+    if (cert == null) {
+      return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
+    }
+    for (Universe universe : Universe.getAll()) {
+      Universe universe_obj = Universe.get(universe.universeUUID);
+      if (universe_obj.getUniverseDetails().rootCA != null) {
+        UUID certificate_uuid = universe_obj.getUniverseDetails().rootCA;
```
daniel-yb, line 174: certificate_uuid => certificateUUID

managed/src/main/resources/v1.routes
```diff
@@ -81,11 +81,13 @@ PUT     /customers/:cUUID/releases/:name                                       c
 # Certificate API
 POST    /customers/:cUUID/certificates                                         com.yugabyte.yw.controllers.CertificateController.upload(cUUID: java.util.UUID)
 GET     /customers/:cUUID/certificates                                         com.yugabyte.yw.controllers.CertificateController.list(cUUID: java.util.UUID)
+DELETE  /certificates/:rUUID                                                   com.yugabyte.yw.controllers.CertificateController.delete(rUUID: java.util.UUID)
```
daniel-yb, line 84: /certificates/:rUUID => /customers/:cUUID/certificates/:rUUID

(you should validate the customer with the request too).

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/ui/src/components/certificates/Certificates.js
gaurav061: For some reason, the button click action handler was being called even when the button was 'Disabled', 
Done

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/ui/src/components/certificates/Certificates.js
gaurav061: Removed this code, It is no longer needed as we use redux

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/ui/src/components/certificates/Certificates.js
gaurav061: Done

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -129,11 +131,18 @@ public Result getRootCert(UUID customerUUID, UUID rootCA) {
   }
 
   public Result list(UUID customerUUID) {
-    List<CertificateInfo> certs = CertificateInfo.getAll(customerUUID);
-    if (certs == null) {
+    List<CertificateInfo> certificates = CertificateInfo.getAll(customerUUID);
+    ArrayNode certList = Json.newArray();
+    certificates.forEach((certificate) -> {
```
Arnav15, line 136: You don't need to do this. Since we added a getter method in the CertificateInfo class, it will automatically populate that in the object's json.

Arnav15 (changes requested):
Also, please use 4 spaces at least when doing line breaks.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: I can not see this field in the response without this. Am I missing anything here? any annotations required for the method? @Arnav15

daniel-yb: @mahendranbhat two things here.
1) For the getInUse method, the customerUUID should not be passed in as a parameter. Since CertificateInfo already has customerUUID as a class field, just use that in the function instead of having a customerUUID parameter. 

2) Once you do this and the getter does not have a parameter, it should automatically add inUse to the json object without you needing to do it here as @Arnav15 mentioned.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +146,29 @@ public Result get(UUID customerUUID, String label) {
     }
   }
 
+  public Result delete(UUID customerUUID, UUID reqCertUUID) {
```
Arnav15, line 149: Add a check to verify the certificate belongs to the customer.

managed/src/main/java/com/yugabyte/yw/models/Universe.java
```diff
@@ -823,4 +823,15 @@ public void run(Universe universe) {}
     };
     Universe.saveDetails(universeUUID, updater);
   }
+
+  public static boolean existsCertificate(UUID certUUID, UUID customerUUID) {
+    Set<Universe> universeList = Customer.get(customerUUID).getUniverses();
+    for (Universe universe : universeList) {
```
Arnav15, line 829: Nit: Maybe use the java stream functionality. (If unaware, take a look [here](https://github.com/yugabyte/yugabyte-db/blob/master/managed/src/main/java/com/yugabyte/yw/models/Schedule.java#L132))

Arnav15 (changes requested):
Looks good, just these minor fixes and then it should be good to go.

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
```diff
@@ -145,9 +146,32 @@ public Result get(UUID customerUUID, String label) {
...
+    if (certificate == null) {
+      return ApiResponse.error(BAD_REQUEST, "Invalid certificate.");
+    }
+    if (!certificate.customerUUID.equals(customerUUID)) {
+      return ApiResponse.error(BAD_REQUEST, "Certificate doesn't belong to customer");
+    }
+    if (!certificate.getInUse()) {
+      if (certificate.delete()) {
+        ObjectNode responseJson = Json.newObject();
+        responseJson.put("Successfully deleted the certificate", true);
```
Arnav15, line 160: Any reason we need the message here? Can't we just return the success as empty? Doesn't it implicitly mean the task was successful?

Arnav15 (approved):

managed/src/main/java/com/yugabyte/yw/controllers/CertificateController.java
mahendranbhat: Nope , we are not showing it in FE . I can remove this.

mahendranbhat: Done

PR #6894, opened 2021-01-16
Convert an old internal doc on some perf debugging …

robertsami:
…tips into a doc

### DIFF
--- a/docs/content/latest/benchmark/performance-troubleshooting.md
+++ b/docs/content/latest/benchmark/performance-troubleshooting.md
@@ -105,19 +105,23 @@ W0325 06:47:13.033341 116514816 inbound_call.cc:208] Trace:
 0325 06:47:13.032168 (+    62us) inbound_call.cc:125] Queueing success response
 ```
 
-## Viewing real-time logs
+## Viewing real-time metrics
 
-You can view logs of Yugabyte processes at a particular node (e.g. 127.0.0.1) at various ports:
+You can view metrics of various Yugabyte processes at a particular node (e.g. 127.0.0.1) at these ports:
 
-| Description | Address
+| Process | Address
 -------------|-----------|
-Master Metrics | 127.0.0.1:7000
-TServer Metrics | 127.0.0.1:9000
-Yedis Metrics | 127.0.0.1:11000
-YCQL Metrics | 127.0.0.1:12000
-YSQL Metrics | 127.0.0.1:13000
-Per-Tablet, JSON Metrics | 127.0.0.1/metrics
-Per-Table, Prometheus Metrics | 127.0.0.1/prometheus-metrics
+Master | 127.0.0.1:7000
+TServer | 127.0.0.1:9000
+Yedis | 127.0.0.1:11000
+YCQL | 127.0.0.1:12000
+YSQL | 127.0.0.1:13000
+
+For each process, you can see the following types of metrics:
+| Description | Endpoint
+-------------|-----------|
+Per-Tablet, JSON Metrics | /metrics
+Per-Table, Prometheus Metrics | /prometheus-metrics
 
 ## Turning RPC tracing on
 
--- a/docs/content/latest/benchmark/performance-troubleshooting.md
+++ b/docs/content/latest/benchmark/performance-troubleshooting.md
@@ -125,7 +125,11 @@ Per-Table, Prometheus Metrics | /prometheus-metrics
 
 ## Turning RPC tracing on
 
-To turn tracing on for all RPCs, not just the slow ones, use the `rpc_dump_all_traces` gflag.
+To turn on tracing, you should set the `enable_tracing` flag:
+
+    $ ./yb-ts-cli --server_address=localhost:9100 set_flag enable_tracing 1
+
+To turn tracing on for all RPCs, not just the slow ones, including the `enable_tracing` flag, you may also set the `rpc_dump_all_traces` gflag:
 
     $ ./yb-ts-cli --server_address=localhost:9100 set_flag rpc_dump_all_traces 1
 
--- a/docs/content/latest/benchmark/performance-troubleshooting.md
+++ b/docs/content/latest/benchmark/performance-troubleshooting.md
@@ -7,9 +7,9 @@ image: /images/section_icons/architecture/concepts.png
 headcontent: Performance Troubleshooting
 menu:
   latest:
-    identifier: large-datasets-2-ysql
+    identifier: performance-troubleshooting
     parent: benchmark
-    weight: 7
+    weight: 22
 showAsideToc: true
 isTocNested: true
 ---
@@ -17,6 +17,8 @@ isTocNested: true
 Use this page for general guidance and steps you can take when troubleshooting the performance of your Yugabyte cluster.
 
 ## Location of various files on a YB cluster
+
+Note that these locations refer to the default for clusters intalled via Yugabyte Platform.
 
 ### YB Software/Binaries on the cluster
 
@@ -47,7 +49,7 @@ drwxr-xr-x. 2 yugabyte yugabyte   25 Jan 15 20:18 conf
 You can find config files for the master and tserver at `/home/yugabyte/{master|tserver}/conf/server.conf`.
 
 ### Transaction Logs a.k.a WAL (write-ahead logs)
-You can find WAL files at `/mnt/d*/yb-data/{master|tserver}/wals`.
+Assuming Yugabyte was run with `--fs_data_dirs=/mnt/d0,/mnt/d1`, you can find WAL files at `/mnt/d*/yb-data/{master|tserver}/wals`.
 
 To pretty print the contents of the WAL files, use the `log-dump` utility:
 
@@ -81,16 +83,11 @@ These go to e.g. `/home/yugabyte/tserver/tserver.{err|out}.
 
 ## Setting gflags dynamically
 
-The `yb-ts-cli` utility allows changing gflags dynamically. _Note: For this step you must identify the server using its RPC port (not the HTTP port).
+A cautionary note -- setting string flags dynamically is not a good idea, as it is not thread safe. That being said, the `yb-ts-cli` utility allows changing gflags dynamically. _Note: For this step you must identify the server using its RPC port (not the HTTP port).
 
 For example, to increase the verbose logging level to 2:
 
     $ ./yb-ts-cli --server_address=localhost:9100 set_flag v 2
-
-Or, to turn tracing on for all RPCs:
-
-    $ ./yb-ts-cli --server_address=localhost:9100 set_flag rpc_dump_all_traces 1
-
 
 ## Slow Response Logs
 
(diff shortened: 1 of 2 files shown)


docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+    identifier: large-datasets-2-ysql
+    parent: benchmark
+    weight: 7
+showAsideToc: true
+isTocNested: true
+---
+
+Use this page for general guidance and steps you can take when troubleshooting the performance of your Yugabyte cluster.
+
+## Location of various files on a YB cluster
```
bmatican, line 19: I would probably also note that these default locations, relative to `/home/yugabyte`, are true for platform, otherwise, it's up to the user where they install YB, vs where they point their data dirs to

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+lrwxrwxrwx. 1 yugabyte yugabyte   65 Jan 15 19:28 ui -> /home/yugabyte/yb-software/yugabyte-2.5.1.0-b187-centos-x86_64/ui
+lrwxrwxrwx. 1 yugabyte yugabyte   84 Jan 15 19:28 version_metadata.json -> /home/yugabyte/yb-software/yugabyte-2.5.1.0-b187-centos-x86_64/version_metadata.json
+lrwxrwxrwx. 1 yugabyte yugabyte   66 Jan 15 19:28 www -> /home/yugabyte/yb-software/yugabyte-2.5.1.0-b187-centos-x86_64/www
+-rw-rw-r--. 1 yugabyte yugabyte    0 Jan 15 19:29 master.out
+-rw-rw-r--. 1 yugabyte yugabyte 2200 Jan 15 20:18 master.err
+drwxr-xr-x. 2 yugabyte yugabyte   25 Jan 15 20:18 conf
+```
+
+### Config Files
+You can find config files for the master and tserver at `/home/yugabyte/{master|tserver}/conf/server.conf`.
```
bmatican, line 47: platform defaults to that, else users picks where they store their confs

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Transaction Logs a.k.a WAL (write-ahead logs)
+You can find WAL files at `/mnt/d*/yb-data/{master|tserver}/wals`.
```
bmatican, line 50: for wals/rocksdb, maybe we can prefix both with a note about, assume `--fs_data_dirs=/mnt/d0,/mnt/d1`

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+
+To pretty print the contents of the WAL files, use the `log-dump` utility:
+
+```
+$ pwd
+/home/yugabyte
+
+$ ./tserver/bin/log-dump /mnt/d0/yb-data/tserver/wals/table-e85a116bc557403e82f57037e7b13879/tablet-05bef5ed6fb74cabb420b648b6f850e3/
+
+# use -print_entries=pb to print the entire contents of each record
```
bmatican, line 60: nice

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+### Database (SSTable) Files for various tablets
+You can find SSTable files at `/mnt/d*/yb-data/{master|tserver}/data`.
+
+To pretty print the contents of these SSTable files, 
+
+```
+$ pwd
+/home/yugabyte
+
+$ ./tserver/bin/ldb dump --compression_type=snappy --db=/mnt/d0/yb-data/tserver/wals/table-e85a116bc557403e82f57037e7b13879/tablet-05bef5ed6fb74cabb420b648b6f850e3/
```
bmatican, line 72: oh nice, this still works?? awesome

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+### Debug Logs (.INFO, .WARNING, .FATAL, etc.)
+Debug logs are output to `/home/yugabyte/{master|tserver}/logs`.
+
+### View Standard Output/Standard Error (stderr/stdout) of yb-tserver & yb-master process
+
+These go to e.g. `/home/yugabyte/tserver/tserver.{err|out}.
+
+## Setting gflags dynamically
```
bmatican, line 82: I would also add a note here that changing string flags dynamically is not a good idea, as it is not thread safe (I guess note to selves, we should probably disallow that...)

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+
+The `yb-ts-cli` utility allows changing gflags dynamically. _Note: For this step you must identify the server using its RPC port (not the HTTP port).
+
+For example, to increase the verbose logging level to 2:
+
+    $ ./yb-ts-cli --server_address=localhost:9100 set_flag v 2
+
+Or, to turn tracing on for all RPCs:
+
+    $ ./yb-ts-cli --server_address=localhost:9100 set_flag rpc_dump_all_traces 1
```
bmatican, line 92: I probably wouldn't mention tracing here, if we have a separate section right below, about tracing

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+Note: On mac, the environment variable DYLD_FALLBACK_LIBRARY_PATH needs to be set for pbc-dump to work. Add the following to ~/.bash_profile
+```
+export DYLD_FALLBACK_LIBRARY_PATH=~/code/yugabyte/build/latest/rocksdb-build
+```
+
+## yb-ts-cli 
+You can run various tablet related commands with yb-ts-cli by pointing at the master:
+```
+./yb-ts-cli list_tablets --server_address=localhost:8101
+./yb-ts-cli dump_tablet --server_address=localhost:8101 e1bc59288ee849ab850ae0a40bd88649
```
bmatican, line 132: nit: let's change 8101 to 9100

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+## yb-admin
+You can run various commands with `yb-admin`:
```
bmatican, line 136: I'd probably change the master addresses here to `127.0.0.1:7100,127.0.0.2:7100,127.0.0.3:7100` and make a note in advance that this should point to the full set of master ip:ports

bmatican (changes requested):
Awesome stuff @robertsami !!

Maybe one other piece that's missing from here is metrics, particularly, two cuts
- by port: 7000 (master), 9000 (tserver), 11000 (yedis), 12000 (ycql), 13000 (ysql)
- by type: `/metrics` (per tablet, JSON format), `/prometheus-metrics` (aggregated up to table level, prometheus format)

But if you just want this PR to be about converting the wiki, can followup separately with the metrics

docs/content/latest/benchmark/performance-troubleshooting.md
robertsami: added a higher level note above, pertaining to the whole section here

docs/content/latest/benchmark/performance-troubleshooting.md
robertsami: https://github.com/yugabyte/yugabyte-db/issues/6997

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -108,10 +105,26 @@ W0325 06:47:13.033341 116514816 inbound_call.cc:208] Trace:
 0325 06:47:13.032168 (+    62us) inbound_call.cc:125] Queueing success response
 ```
 
+## Viewing real-time logs
```
bmatican, line 108: I assume you meant metrics not logs?

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -108,10 +105,26 @@ W0325 06:47:13.033341 116514816 inbound_call.cc:208] Trace:
...
+
+You can view logs of Yugabyte processes at a particular node (e.g. 127.0.0.1) at various ports:
+
+| Description | Address
+-------------|-----------|
+Master Metrics | 127.0.0.1:7000
```
bmatican, line 114: note: the metrics themselves will still be at `127.0.0.1:7000/metrics`, it's just that the port changes depending on which you want to look at...so really the split should be
- for these components, here are their webserver ports
- for each of them, you can view JSON vs prometheus metrics, on these endpoints

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -108,10 +105,26 @@ W0325 06:47:13.033341 116514816 inbound_call.cc:208] Trace:
...
+YCQL Metrics | 127.0.0.1:12000
+YSQL Metrics | 127.0.0.1:13000
+Per-Tablet, JSON Metrics | 127.0.0.1/metrics
+Per-Table, Prometheus Metrics | 127.0.0.1/prometheus-metrics
+
 ## Turning RPC tracing on
 
 To turn tracing on for all RPCs, not just the slow ones, use the `rpc_dump_all_traces` gflag.
 
+    $ ./yb-ts-cli --server_address=localhost:9100 set_flag rpc_dump_all_traces 1
```
bmatican, line 126: cc @amitanandaiyer don't we also need some `--enable_tracing` ?

bmatican (approved):
Accepting with one comment around rewording the metrics port and endpoint tables

docs/content/latest/benchmark/performance-troubleshooting.md
acube123: should we mention that some of the flags require you to use -force?

acube123: Also, common question is whether vmodule can be set at runtime. This cannot be set at runtime. only during start

docs/content/latest/benchmark/performance-troubleshooting.md
```diff
@@ -0,0 +1,150 @@
...
+0325 06:47:10.450697 (+ 55838us) raft_consensus.cc:1026] Updating replica for 0 ops
+0325 06:47:13.032064 (+2581367us) raft_consensus.cc:1242] Filling consensus response to leader.
+0325 06:47:13.032106 (+    42us) spinlock_profiling.cc:233] Waited 2.58 s on lock 0x108dc3d40. stack: 0000000103d63b0c 0000000103d639fc 00000001040ac908 0000000102f698ec 0000000102f698a4 0000000102f93039 0000000102f7e124 0000000102fdcf7b 0000000102fd90c9 00000001\
+02504396 00000001032959f5 0000000103470473 0000000103473491 00000001034733ef 000000010347338b 000000010347314c
+0325 06:47:13.032168 (+    62us) inbound_call.cc:125] Queueing success response
+```
+
+## Turning RPC tracing on
+
+To turn tracing on for all RPCs, not just the slow ones, use the `rpc_dump_all_traces` gflag.
```
acube123, line 128: there is probably also a sampling based print_every_n option.

will be expanding with the tracing diff I have in progress. Probably just mention other Gflags to look up?

acube123 (commented):
nice. rob.

I've added a few more comments/suggesions in case you want to expand on them.

docs/content/latest/benchmark/performance-troubleshooting.md
bmatican: if this in itself is ok for now, we can also just expand on this after we land this PR, so we don't block all of it

acube123: i'm ok with landing as is.

docs/content/latest/benchmark/performance-troubleshooting.md
acube123: yes. We need enable_tracing to be turned on for "collecting traces".

Once the collection is turned on. We can print them out based on latency/sampling or all. I've personally never used all. I tend to use the latency based, or the sampling based printing.

acube123: we don't want to print so many traces ... that printing the traces itself slows down the system/logging.

depending on the scenario, one must adjust the latency-threshold/print_every_n frequency to make sure we are not disturbing the state of the system too much.

docs/content/latest/benchmark/performance-troubleshooting.md
robertsami: @acube123 which flags require -force and what are the semantics of that?

acube123: Basically any flag that is not explictly marked TAG_FLAG(xxx, runtime) will need a -force flag.

Ideally, we should be updating the flags that are *safe* to be updated on the fly -- some that use GetAtomicFlag(&) and fetch the value for each use rather than using the flag to initialize an object once; as runtime; so that no one needs to use the -force flag. We are not there yet. So, may need the -force flags for flags that are not safe to update ... but not yet marked as runtime.

Issue #6899, opened 2021-01-16
[Platform] Fix broken unit tests


daniel-yb:
https://github.com/yugabyte/yugabyte-db/commit/7a8ddfacd120918ba8d71895f84e3eb04847b75c introduced a couple of unit test regressions in UniverseControllerTest.

PR #6910, opened 2021-01-18
[#6243] Ensure "node-to-node" tls is enabled in order to use "client-to-node" tls during create universe testcase

jaydeepkumara:


SergeyPotachev:
Sorry, @jaydeepkumara,
We already have a fix for this problem (#6899) prepared by Daniel (it is already accepted but not landed yet).

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
```diff
@@ -2205,6 +2205,40 @@ public void testExpandDiskSizeSuccess() {
...
+    String accessKeyCode = "someKeyCode";
+    AccessKey.create(p.uuid, accessKeyCode, new AccessKey.KeyInfo());
+    Region r = Region.create(p, "region-1", "PlacementRegion 1", "default-image");
+    AvailabilityZone.create(r, "az-1", "PlacementAZ 1", "subnet-1");
+    InstanceType i = InstanceType.upsert(p.code, "c3.xlarge", 10, 5.5,
+      new InstanceType.InstanceTypeDetails());
+
+    ObjectNode bodyJson = Json.newObject();
+    ObjectNode userIntentJson = Json.newObject().put("universeName", "SingleUserUniverse")
+      .put("instanceType", i.getInstanceTypeCode()).put("replicationFactor", 3).put("numNodes", 3)
```
sb-yb, line 2224: Make sure line length does not exceed 100.

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
```diff
@@ -2205,6 +2205,40 @@ public void testExpandDiskSizeSuccess() {
...
+    userIntentJson.set("regionList", regionList);
+    userIntentJson.put("accessKeyCode", accessKeyCode);
+    ArrayNode clustersJsonArray = Json.newArray()
+      .add(Json.newObject().set("userIntent", userIntentJson));
+    bodyJson.set("clusters", clustersJsonArray);
+    bodyJson.set("nodeDetailsSet", Json.newArray());
+
+    String url = "/api/customers/" + customer.uuid + "/universes";
+    Result result = doRequestWithAuthTokenAndBody("POST", url, authToken, bodyJson);
+    String expectedResult = String.format("It is imperative that the NodeToNode TLS encryption should be enabled for enabling the ClientToNode TLS encryption.");
```
sb-yb, line 2237: same.

sb-yb (approved):
> Sorry, @jaydeepkumara,
In this case its ok since he actually added two new test cases.
If Daniel can add his fix on top of this if it is any different or adds some value.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -121,11 +121,15 @@ public UniverseController(YBClientService service) {
   }
 
   private boolean validateEncryption(ObjectNode formData) {
+    JsonNode clusters = formData.get("clusters");
+    if (clusters == null) {
+      return true;
+    }
     for (JsonNode cluster : formData.get("clusters")) {
       JsonNode nodeToNodeEncryption = cluster.get("userIntent").get("enableNodeToNodeEncrypt");
```
daniel-yb, line 129: This can still result in a NPE if a cluster does not have a userIntent field. This method should not assume the data is in any specific shape since we don't have upstream validation of the shape of the JSON request payload before this method is called.

daniel-yb:
I just landed my diff fixing the NPE in UniverseController.java. If you want to change this diff to only include the extra unit tests that would still be useful though.

daniel-yb (changes requested):
See my last comment. As @SergeyPotachev mentioned, we definitely should figure out a way to give you guys more visibility on the PRs that we have up internally somehow to avoid duplicate work in the future. I should have better communicated that I had a fix up for review already for this, my bad.

Issue #5172, opened 2020-07-21
[Platform] Health check time is always in the Pacific timezone


iSignal:
Repro steps:

Change timezone on client browser device. Health check timestamps do not change though other timestamps like metrics / backups do.

It seems to be hardcoded here - https://github.com/yugabyte/yugabyte-db/blob/1403b2b4d3986fbc401dd14ce9cbc956eaaf2938/managed/devops/bin/cluster_health.py#L620

Aha! Link: https://yugabyte-test.aha.io/features/PLATFORM-107

jaydeepkumara:
PR:- https://github.com/yugabyte/yugabyte-db/pull/6933

PR #6933, opened 2021-01-20
[#5172] Health check time is always in the Pacific timezone 

jaydeepkumara:
Backend:- Removed hardcoded US timezone and changed to UTC.
Frontend:- Getting UTC time and converting to the local timezone.


SergeyPotachev (approved):
Looks good to me. But I'm not UI developer, don't know a lot of details about js.

managed/ui/src/components/universes/UniverseDetail/compounds/UniverseHealthCheckList.js
```diff
@@ -225,7 +224,7 @@ const detailsFormatter = (cell, row) => {
 function prepareData(data) {
   return data.map((timeDataJson) => {
     const timeData = JSON.parse(timeDataJson);
-    const timestampMoment = moment(timeData.timestamp);
+    const timestampMoment = momentTimezone.utc(timeData.timestamp).local();
```
sshev, line 227: @jaydeepkumara are you sure about `momentTimezone`?
it seems `moment.utc()` is what we need
https://momentjs.com/docs/#/parsing/utc/

jaydeepkumara: changed to a moment. Both libraries support UTC fun. The earlier moment had given an error so had used the moment timezone but I changed the import statement and it worked now.

sshev (approved):

sshev:
Make sure to add `Platform:` prefix into the PR title pls

Issue #6164, opened 2020-10-27
[Platform] manual backup of ysql task  is not showing up in the universe tasks page


rao-vasireddy:
It is showing in the YW Tasks page (Tasks in the left nav).
Restore task of ysql is showing up in the universe tasks page correctly.

jaydeepkumara:
PR:- https://github.com/yugabyte/yugabyte-db/pull/6935/files

PR #6935, opened 2021-01-20
[#6164] Manual backup of ysql task is not showing up in the universe tasks page

jaydeepkumara:
Universe task was not showing if the user clicks on the task tab after creating a manual backup.
The issue happens because we are not fetching latest task. Added API call when the user clicks on universe task tab.


yb-andrew (approved):
LG

sshev (approved):

PR #6948, opened 2021-01-21
[#6460]Platform: Preflight checks for cloud provider configuration

jitendra-12113:
**Heading:**
[#6460]Platform: Preflight checks for cloud provider configuration
**Description:**
We need to validate the AWS credentials at the time of cloud provider configuration
**Testing:**
Added Unit tests. We can test this by passing invalid aws access key or secret key.
For now, we have handled only for AWS provider.


managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
```diff
@@ -172,6 +176,11 @@ public Result create(UUID customerUUID) throws IOException {
       if (!config.isEmpty()) {
         switch (provider.code) {
           case "aws":
+            CloudAPI cloudAPI = cloudAPIFactory.get(provider.code);
```
daniel-yb, line 179: I'm not sure we want to make some kind of blocking call to AWS in the main API request thread. IMO this should be done as an asynchronous subtask for creating the cloud provider which fails the task if the credentials are invalid.

If it is moved to an async subtask, I would also argue that we should stick with using boto for now since we have not flushed out how exactly we want to use the Java layer to interact with cloud providers yet cc: @sb-yb on if you agree with this or not. My concern is that you implement it for AWS at the Java layer and then have to either implement CloudAPI for the other cloud provider types (GCP, Azure, etc) or that you have the AWS solution in Java and all the other solutions in python.

Since we haven't had time to plan out how we want to properly interact with cloud providers (i.e. whether we want to move everything to the Java layer or not), it's best to stick with the current pattern of using python for now (as discussed in the Slack channel for the approach to take for pausing a universe).

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
```diff
@@ -82,4 +87,28 @@ private static AWSCredentialsProvider getCredsOrFallbackToDefault(String accessK
...
+      config.get("AWS_ACCESS_KEY_ID"),
+      config.get("AWS_SECRET_ACCESS_KEY"));
+    try {
+      AmazonEC2 ec2Client = AmazonEC2ClientBuilder
+        .standard()
+        .withCredentials(credentialsProvider)
+        .withRegion(Regions.valueOf(region.toUpperCase()))
+        .build();
+      DryRunResult<DescribeInstancesRequest> dryRunResult = ec2Client
+        .dryRun(new DescribeInstancesRequest());
```
daniel-yb, line 103: Is testing whether the user can describe instances sufficient? Or do we want the pre-flight checks to comprehensively do a dry run of a set of actions such that it covers the list of required IAM permissions needed for a cloud provider?

Also you can just re-use the "getEcC2Client" method by creating local models of provider + region from the request payload without saving these models to the DB and then passing them to "getEcC2Client".

daniel-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
sb-yb: Actually both of these were my suggestions in first review :)

sb-yb: Basically we need a framework where if you add request Foo - we can easily add a dry run for that in here. 
But I felt this is a good start and more can be added later. 
I was planning to hold the line on getEcC2Client. You should switch to using that.
So either do what daniel suggested 
OR
Another way is to extract getEC2ClientInternal or something to so it becomes usable from getEc2Client and from here.

managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
sb-yb: So about async part - I think we are not using async handler in our controller. So no matter what we do behind the scene, the incoming API call will block. This is a general issue that I mentioned to @yb-andrew during one of his reviews. 
But we decided not to do any optimization at this point. It is really not-worth for the qps that we serve in platform. 

The later problem is something I have to solve for missing instance types GCP and Azure anyway. So they can implement after that.

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: @sb-yb @daniel-yb  Yup  I think more can be added later. For now, I am keeping this as it is(Stick to only DescribeInstancesRequest). 
Not really sure on how to ensure the list of required IAM permissions needed for a cloud provider. Please let me know if its ok else would like to discuss about this. 

Regarding to use getEcC2Client, here we have provider config but I will have to form Region to fit region details in it and then pass in the param - Need to figure out how, will update the PR with this changes shortly. Thanks!

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
```diff
@@ -14,12 +16,15 @@
...
 
 import static java.util.stream.Collectors.*;
 
 //TODO - Better handling of UnauthorizedOperation. Ideally we should trigger alert so that
 // site admin knows about it
 class AWSCloudImpl implements CloudAPI {
+  public static final Logger LOG = LoggerFactory.getLogger(AWSCloudImpl.class);
 
   // TODO use aws sdk 2.x and switch to async
   public AmazonEC2 getEcC2Client(Provider provider, Region r) {
```
sb-yb, line 30: @jitendra-12113 change this to:
```
public AmazonEC2 getEcC2Client(Provider provider, Region r) {
    return getEC2ClientInternal(provider.getConfig()), r.code);
}

private getEC2ClientInternal(config , regionCode) {
    ...
}
```

Then in your code just call 
`getEC2ClientInternal`

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
sb-yb: Answered above.

sb-yb (changes requested):

Arnav15 (changes requested):
Please add output of the complete sbt test command.

jitendra-12113:
sbt test run output:

2021-01-24 01:11:09.046 ERROR QueryAlerts.java:121 [application-akka.actor.default-dispatcher-10] Error querying for alerts
javax.persistence.PersistenceException: java.sql.SQLException: HikariDataSource HikariDataSource (HikariPool-1037) has been closed.
	at io.ebeaninternal.server.transaction.TransactionFactoryBasic.createQueryTransaction(TransactionFactoryBasic.java:35)
	at io.ebeaninternal.server.transaction.TransactionManager.createQueryTransaction(TransactionManager.java:360)
	at io.ebeaninternal.server.core.DefaultServer.createQueryTransaction(DefaultServer.java:2306)
	at io.ebeaninternal.server.core.OrmQueryRequest.initTransIfRequired(OrmQueryRequest.java:275)
	at io.ebeaninternal.server.core.DefaultServer.findList(DefaultServer.java:1595)
	at io.ebeaninternal.server.core.DefaultServer.findList(DefaultServer.java:1574)
	at io.ebeaninternal.server.querydefn.DefaultOrmQuery.findList(DefaultOrmQuery.java:1474)
	at com.yugabyte.yw.models.Customer.getAll(Customer.java:124)
	at com.yugabyte.yw.commissioner.QueryAlerts.scheduleRunner(QueryAlerts.java:107)
	at akka.actor.LightArrayRevolverScheduler$$anon$2$$anon$1.run(LightArrayRevolverScheduler.scala:104)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:41)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:49)
	at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at akka.dispatch.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
Caused by: java.sql.SQLException: HikariDataSource HikariDataSource (HikariPool-1037) has been closed.
	at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:96)
	at play.db.ebean.DefaultEbeanConfig$EbeanConfigParser$WrappingDatasource.getConnection(DefaultEbeanConfig.java:153)
	at io.ebeaninternal.server.transaction.TransactionFactoryBasic.createQueryTransaction(TransactionFactoryBasic.java:28)
	**... 15 common frames omitted
[info] Passed: Total 1142, Failed 0, Errors 0, Passed 1138, Skipped 4
[success] Total time: 806 s, completed 24 Jan, 2021 1:11:15 AM**

managed/src/main/java/com/yugabyte/yw/cloud/aws/AWSCloudImpl.java
jitendra-12113: Done. Thanks!

sb-yb (approved):

Issue #6320, opened 2020-11-09
[Platform] Create Universe does not validate encryption-at-rest configs exist


WesleyW:
Users should not be able to create a universe if no certificate is selected after enabling encryption at rest. This can happen if users do not notice the dropdown or if the config itself doesn't exist.

gaurav061:
PR:https://github.com/yugabyte/yugabyte-db/pull/6950

chirag-yb:
Should this be closed?

Issue #6862, opened 2021-01-13
[Platform] Pre-flight checks for encryption-at-rest keys


streddy-yb:
At the time of creating a new key configuration, do a pre-flight check to verify the key config is valid.

In the create universe workflow, if the encryption-at-rest is enabled, verify the selected key config is still valid.

jitendra-12113:
Please review:
https://github.com/yugabyte/yugabyte-db/pull/7701

PR #6950, opened 2021-01-21
[#6320] Platform: Create Universe does not validate encryption-at-rest

gaurav061:


gaurav061:
Fix: Dummy option had value as 0 which was causing this issue. PTAL

jaydeepkumara (approved):

daniel-yb (changes requested):
Please add validation to the UniverseController that the kms config UUID passed in as part of the API payload is valid and corresponds to an existing KmsConfig model in the DB.

We need more than just UI validation since it's possible a customer may interact with the platform through API as opposed to UI.

gaurav061:
@daniel-yb  We have a separate backend ticket to address this issue.https://github.com/yugabyte/yugabyte-db/issues/6862

sshev (approved):

Issue #6909, opened 2021-01-18
[docdb] Intent metadata_write_time is incorrectly set


triump2020:
Hi ,  folllowing code block comes from [line222](https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/tablet/transaction_participant.cc) ,  the 'data_copy ' seems that useless ?

`     auto data_copy = data;
      // We use hybrid time only for backward compatibility, actually wall time is required.
      data_copy.set_metadata_write_time(GetCurrentTimeMicros());`

triump2020:
Above code block comes from : [line222](https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/tablet/transaction_participant.cc#L222)

tedyu:
I wonder if Sergei meant the following:
```
diff --git a/src/yb/tablet/transaction_participant.cc b/src/yb/tablet/transaction_participant.cc
index 6a37f42fa..4d5d63966 100644
--- a/src/yb/tablet/transaction_participant.cc
+++ b/src/yb/tablet/transaction_participant.cc
@@ -222,7 +222,7 @@ class TransactionParticipant::Impl
       auto data_copy = data;
       // We use hybrid time only for backward compatibility, actually wall time is required.
       data_copy.set_metadata_write_time(GetCurrentTimeMicros());
-      auto value = data.SerializeAsString();
+      auto value = data_copy.SerializeAsString();
       write_batch->Put(key.AsSlice(), value);
     }
     return true;
```

triump2020:
Should be , but seems no impact on txn-correctness

triump2020:
@bmatican   Currently , the hlc timestamp  of write intents host in  intent-db  is useless , since snapshot reads must wait for the safe time to past completely ,  right ?

bmatican:
Thanks for digging into this! Seems like this is a bug, but indeed it should not affect txn correctness. I spoke to @spolitov and he mentioned that we would end up using the `metadata_write_time` when checking for aborted txns on compactions, but we would not actually be aborting anything that we shouldn't, so should be safe there.

The suggestion that @tedyu had above should be a good enough fix for this. @triump2020 would you mind putting a small PR for this? Figured to have the commit to your name, given you did the hard part of finding it.

triump2020:
@bmatican   I ' m  very happy to do it  ,  though it 's my firt time !

bmatican:
@triump2020 Hehe, even better :) always happy to take up new contributions!

[...]

PR #6953, opened 2021-01-21
fixed bug : Intent metadata_write_time is incorrectly set

triump2020:


triump2020:
Fix bug : [#6909] Intent metadata_write_time is incorrectly set.

Issue #6758, opened 2020-12-29
[platform] Unable to reuse On-premise instance incase previouse universe failed / deleted


jaydeepkumara:
Steps:-
1. Configure On-premise provider
2. Create the universe.
3. Delete the previously created universe.
4. Goto On-premise provider manage instance page, the previous instance will have the universe as n/a and it not allowed to delete.
5. Goto create Universe, The previous instance is not appearing while creating.

### Manage instance screenshot, after deleting universe, Delete is disabled for unattched intance
![Screenshot from 2020-12-29 20-24-00](https://user-images.githubusercontent.com/16913155/103292823-9c7a6380-4a14-11eb-89d3-933f9dc9af86.png)

### Availability zone is not appearing even if one instance is not attached to any universe.
![Screenshot from 2020-12-29 20-26-44](https://user-images.githubusercontent.com/16913155/103292906-cd5a9880-4a14-11eb-9cbd-3e5bde4d5f50.png)

mahendranbhat:
https://github.com/yugabyte/yugabyte-db/pull/8263
2.4 backport PR

mahendranbhat:
Backported to 2.4

PR #6970, opened 2021-01-22
[#6758] Platform: Reuse on-premise instance once the instance got delete.

mahendranbhat:
**Description:** Once we delete the onpremis universe, we have to clear the node instance so that we can use the node to create another universe.

**Testing:** : 

1. Add instance under onpremis
2. Created the universe with the instance (failed)
3. Delete the universe .
4. Used the same instance to create another universe.

`[info] Passed: Total 1141, Failed 0, Errors 0, Passed 1137, Skipped 4
[success] Total time: 707 s, completed 20 Jan, 2021 12:15:09 PM
`


managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UniverseTaskBase.java
```diff
@@ -316,6 +316,9 @@ public SubTaskGroup createDestroyServerTasks(Collection<NodeDetails> nodes,
       if (node.cloudInfo.private_ip == null) {
         LOG.warn(String.format("Node %s doesn't have a private IP. Skipping node delete.",
                                node.nodeName));
+        // Free up the node so that the client can use the instance to create another universe.
```
Arnav15, line 319: `NodeInstance` objects only exist for on-prem providers. We can't always run this command. We need to verify if the provider is an on-premise one or not, and if so, then only clear the state of the node instances.

Arnav15 (changes requested):

managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UniverseTaskBase.java
```diff
@@ -314,11 +315,19 @@ public SubTaskGroup createDestroyServerTasks(Collection<NodeDetails> nodes,
       // Check if the private ip for the node is set. If not, that means we don't have
       // a clean state to delete the node. Log it and skip the node.
       if (node.cloudInfo.private_ip == null) {
-        LOG.warn(String.format("Node %s doesn't have a private IP. Skipping node delete.",
-                               node.nodeName));
+        LOG.warn(String.format("Node %s doesn't have a private IP. Skipping node delete.", node.nodeName));
         // Free up the node so that the client can use the instance to create another universe.
-        NodeInstance providerNode = NodeInstance.getByName(node.nodeName);
-        providerNode.clearNodeDetails();
+        Universe universe = Universe.get(taskParams().universeUUID);
```
Arnav15, line 320: You don't need the universe. The node details class itself has a field `cloudInfo.cloud` that you can use for this check.

Arnav15 (changes requested):

managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UniverseTaskBase.java
```diff
@@ -314,8 +315,16 @@ public SubTaskGroup createDestroyServerTasks(Collection<NodeDetails> nodes,
       // Check if the private ip for the node is set. If not, that means we don't have
       // a clean state to delete the node. Log it and skip the node.
       if (node.cloudInfo.private_ip == null) {
-        LOG.warn(String.format("Node %s doesn't have a private IP. Skipping node delete.",
-                               node.nodeName));
+        LOG.warn(String.format("Node %s doesn't have a private IP. Skipping node delete.", node.nodeName));
```
SergeyPotachev, line 318: 1. We have line length limited to 100 characters. Please update this line and check link 320.
2. Not sure that this text fully reflects the updated behaviour.

managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UniverseTaskBase.java
```diff
@@ -314,8 +315,16 @@ public SubTaskGroup createDestroyServerTasks(Collection<NodeDetails> nodes,
       // Check if the private ip for the node is set. If not, that means we don't have
       // a clean state to delete the node. Log it and skip the node.
```
SergeyPotachev, line 316: Update the comment according to the updated behaviour.

SergeyPotachev (changes requested):
1. Cover these changes with junit tests.
2. When this problem occurs? Only when on-prem instance doesn't have public IP? Or in both cases (with and w/o public IP)? I want to know the root cause because not long time ago this functionality worked.

managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UniverseTaskBase.java
mahendranbhat: > Not sure that this text fully reflects the updated behaviour.

We are skipping node delete for other providers but if it is onprem then we are clearing the existing table entry.
Please suggest a better warning.

mahendranbhat:
> 1. Cover these changes with junit tests.
> 2. When this problem occurs? Only when on-prem instance doesn't have public IP? Or in both cases (with and w/o public IP)? I want to know the root cause because not long time ago this functionality worked.

1. As mentioned in the description, checked with @Arnav15  about the test cases.
2. This happens only when the node doesn't have private ip. Currently, our on-prem universe creation is failing in our local. When you delete the universe it didn't clear the onprem instance ( to reuse it to create the new universe as shown in the ticket )

SergeyPotachev:
> > 1. Cover these changes with junit tests.
> 
> @SergeyPotachev
> 
> 1. As mentioned in the description, checked with @Arnav15  about the test cases.
Initially I didn't pay attention on this part of comments, but now I disagree with this:

You can't write junit test for complete check of a universe deletion, but (!) you can write a test (and I can prove that it is not too difficult) to check this one function only. One test is for a case when your code is executed, another test - when it is not. 
Do call the function and then check that according node details are cleaned or not cleaned. The only thing here which can be not easy - to prepare a correctly filled universe.

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
```diff
@@ -2321,4 +2322,28 @@ public void testRemoveEnclosingDoubleQuotes() {
     // Null string
     assertEquals(null, UniverseController.removeEnclosingDoubleQuotes(null));
   }
+
+  @Test
+  public void testClearNodeDetails() {
```
SergeyPotachev, line 2327: I think we have a kind of misunderstanding. Let me explain once again.
We need functions `testCreateDestroyServerTasks_<Some condition>`. In the first function - check the path when your code doesn't work (call the function and check that there were no invocations of clearNodeDetails).
In the second function - check the path when your code works (call the function and check that there were invocations of clearNodeDetails).

SergeyPotachev (changes requested):

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
mahendranbhat: @SergeyPotachev yeah , the code I have added here is in subtask. So finding it difficult to create the data for this subtask. And do we any examples where we have written test case for subtasks so that I can refer and add one here?

SergeyPotachev: To speedup the review I've prepared a prototype of the requested test.

[UniverseTaskBaseTest.java.txt](https://github.com/yugabyte/yugabyte-db/files/5914784/UniverseTaskBaseTest.java.txt)

SergeyPotachev: And remove changes from UniverseControllerTest.java, they are useless. Please.

mahendranbhat: Added the testcases.

SergeyPotachev (approved):

Issue #6858, opened 2021-01-13
[Platform] Preflight checks for NFS backup storage configuration


streddy-yb:
When user initiates a manual backup -or- a scheduled backup, check whether NFS path is a valid path on each of the database nodes before starting the backup task . Fail the task right away before creating the snapshots for the backup.

jitendra-12113:
Please review:
https://github.com/yugabyte/yugabyte-db/pull/6971/

PR #6971, opened 2021-01-22
[#6858]Platform: Preflight checks for NFS backup storage configuration

jitendra-12113:
**Heading:** 
[#6858]Platform: Preflight checks for NFS backup storage configuration
**Description:**
When user initiates a manual backup -or- a scheduled backup, check whether
NFS path is a valid path on each of the database nodes before starting the backup task .
Fail the task right away before creating the snapshots for the backup.
**Testing:**
Configured the NFS backup storage PATH and created a AWS cloud provider
Then using cloud provider, created a multinode/single node universe
Ran sample Apps, which created table for backup.
Then after tried to create backup.
It will fail immediately if the storage path is not mounted on anyone of
the node.

Added unit Tests as well. It's working fine.

**Test Results:**

[info] Test run started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateS3Backup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateGcsBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testBulkImportWithSpecificInstanceCount started
[info] Test com.yugabyte.yw.common.TableManagerTest.testRestoreGcsBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testBulkImportWithDefaultInstanceCount started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateBackupWithSSHUser started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateNfsBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateBackupKubernetesWithTLS started
[info] Test com.yugabyte.yw.common.TableManagerTest.testRestoreNfsBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testRestoreUniverseBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateS3BackupVerbose started
[info] Test com.yugabyte.yw.common.TableManagerTest.testBulkImportWithIncorrectYBVersion started
[info] Test com.yugabyte.yw.common.TableManagerTest.testRestoreS3Backup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateTLSBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateUniverseBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testDeleteUniverseBackup started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateS3BackupWithSSE started
[info] Test com.yugabyte.yw.common.TableManagerTest.testCreateBackupKubernetes started
[info] Test run finished: 0 failed, 0 ignored, 18 total, 21.681s
[info] Passed: Total 18, Failed 0, Errors 0, Passed 18
[success] Total time: 72 s, completed 22 Jan, 2021 11:25:30 AM

### DIFF
--- a/managed/devops/bin/yb_backup.py
+++ b/managed/devops/bin/yb_backup.py
@@ -709,6 +709,8 @@ class YBBackup:
             '--restore_keys_destination', required=False,
             help="Location to download universe encryption keys backup file to"
         )
+        parser.add_argument(
+            '--nfs_storage_path', required=False, help="NFS storage mount path")
         logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s: %(message)s")
         self.args = parser.parse_args()
 
@@ -1610,6 +1612,30 @@ class YBBackup:
         else:
             self.run_program(del_cmd)
 
+    def find_nfs_storage(self):
+        """
+        check whether the nfs backup storage path mounted on each table server
+        or not if we don't find storage path mounted on any one of them then we
+        raise exception
+        :param nfs_storage_path: provided nfs storage path
+        """
+        output = self.run_yb_admin(['list_all_tablet_servers'])
+        for line in output.splitlines():
+            if LEADING_UUID_RE.match(line):
+                fields = split_by_space(line)
+                ip_port = fields[1]
+                state = fields[3]
+                (ip, port) = ip_port.split(':')
+                if state == 'ALIVE':
+                    try:
+                         isNFSStoragePathExists = self.run_ssh_cmd(
+                             ['find', self.args.nfs_storage_path],
+                             ip)
+                    except Exception as ex:
+                        raise BackupException(
+                            ('Did not find nfs backup storage path: %s mounted on tablet server %s'
+                            % (self.args.nfs_storage_path, ip)))
+
     def upload_metadata_and_checksum(self, src_path, dest_path):
         """
         Upload metadata file and checksum file to the target backup location.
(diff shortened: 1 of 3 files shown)


managed/devops/bin/yb_backup.py
```diff
@@ -1610,6 +1612,30 @@ def delete_bucket_obj(self):
         else:
             self.run_program(del_cmd)
 
+    def find_nfs_storage(self):
```
Arnav15, line 1615: As I had mentioned in the our initial discussion, we already allow for running commands in parallel, so there is no point in running these commands sequentially. Check how the `find_data_dirs` function is utilized by the script and do something similar for finding the NFS storage too.

Arnav15 (changes requested):

managed/devops/bin/yb_backup.py
jitendra-12113: Done. Thanks!

managed/devops/bin/yb_backup.py
```diff
@@ -1762,6 +1778,23 @@ def backup_table(self):
         backup location.
         """
 
+        if self.args.storage_type == 'nfs':
```
Arnav15, line 1781: We can't do this check only while backing up. We need to do it for restore as well. We have a function called `post_process_arguments` that we can utilize to do this check before running any commands. Further down the line, we can also expand this post processing to eventually verify the validity of the s3, gcs, and azure credentials.

Arnav15 (changes requested):
Sorry for not catching this in my last pass over your diff.

managed/devops/bin/yb_backup.py
jitendra-12113: Yup. Make sense. Thanks for catching this! Updated the PR.

Arnav15 (approved):
Looks good! Can you also once check the success (if the mount point is present) of the task before merging it in?

jitendra-12113:
> Looks good! Can you also once check the success (if the mount point is present) of the task before merging it in?

Yeah already tested both the scenarios(success and error out). It's working fine. Thanks!

Issue #6857, opened 2021-01-13
[Platform] Preflight checks for S3 backup storage configuration


streddy-yb:
- At the time of creating S3 backup configuration, do a pre-flight check whether the backup configuration is valid before saving the config.

- When user initiates a manual backup -or- a scheduled backup, check whether S3 bucket is still accessible. This check can be performed on the platform side.

jitendra-12113:
Please review:
https://github.com/yugabyte/yugabyte-db/pull/7028

jitendra-12113:
Created ticket: https://github.com/yugabyte/yugabyte-db/issues/8824(to address update config details)

shashwat-yb:
Validated the S3 backup storage configuration by passing valid credentials, invalid credentials, created backups and verified the appropriate responses.

cc @VijiYB

PR #7028, opened 2021-01-29
[#6857]Platform: Preflight checks for S3 backup storage configuration

jitendra-12113:
**Heading:**
[#6857]Platform: Preflight checks for S3 backup storage configuration
**Description:**
We need to validate the S3 backup config before saving it.
1. Validate AWS credentials
2. Validate whether bucketname already exists or not

Note: If user selects IAM then we are not validating the config details.
**Testing:**
1. Create aws cloud provider
2. Go to config and the backup tab, select Amazon S3 config
3. Enter either wrong AWS aceess key/secret key or the random bucket
name, we will get appropriate error message

**TODO:** Have to add UT before merging if this approach looks good.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
@@ -39,6 +39,13 @@ public class CustomerConfigController extends AuthenticatedController {
     if (errorJson.size() > 0) {
       return ApiResponse.error(BAD_REQUEST, errorJson);
     }
+    if (formData.get("name").asText().equals("S3") &&
+      formData.get("data").get("AWS_ACCESS_KEY_ID") != null) {
+      errorJson = configValidator.validateS3DataContent(formData, customerUUID);
+    }
+    if (errorJson.size() > 0) {
+      return ApiResponse.error(BAD_REQUEST, errorJson);
+    }
 
     CustomerConfig customerConfig = CustomerConfig.createWithFormData(customerUUID, formData);
     Audit.createAuditEntry(ctx(), request(), formData);
--- a/managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
+++ b/managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
@@ -15,6 +15,18 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.inject.Singleton;
 import com.yugabyte.yw.models.CustomerConfig;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.regions.Regions;
+import com.yugabyte.yw.models.Provider;
+import com.yugabyte.yw.models.Region;
+import com.yugabyte.yw.commissioner.Common.CloudType;
+import com.amazonaws.services.s3.model.Bucket;
+import com.amazonaws.services.s3.model.AmazonS3Exception;
+import java.util.UUID;
 
 import play.libs.Json;
 
@@ -177,6 +189,46 @@ public class CustomerConfigValidator {
     ObjectNode errorJson = Json.newObject();
     validators.forEach(v -> v.validate(formData.get("type").asText(), formData.get("name").asText(),
         formData.get("data"), errorJson));
+    return errorJson;
+  }
+
+  /**
+   * Validates data which is contained in formData.
+   * During the procedure it checks all the S3 config parameters whether they are valid or not.
+   * Errors are collected and
+   * returned back as a result. Empty result object means no errors.
+   *
+   * Currently is checked:
+   *  - S3/AWS - S3 Bucket, whether it exists or not.
+   *
+   * @param formData, customerUUID
+   * @return Json filled with errors
+   */
+  public ObjectNode validateS3DataContent(JsonNode formData, UUID customerUUID) {
+    ObjectNode errorJson = Json.newObject();
+    try {
+      AWSCredentials credentials = new BasicAWSCredentials(
+        formData.get("data").get("AWS_ACCESS_KEY_ID").asText(),
+        formData.get("data").get("AWS_SECRET_ACCESS_KEY").asText()
+      );
+      String region = Region.getByProvider(Provider.get(customerUUID, CloudType.aws).uuid)
+        .get(0).code.replace('-', '_').toUpperCase();
+      String bucketname = formData.get("data").get("BACKUP_LOCATION").asText();
+      AmazonS3 s3client = AmazonS3ClientBuilder
+        .standard()
+        .withCredentials(new AWSStaticCredentialsProvider(credentials))
+        .withRegion(Regions.valueOf(region))
+        .build();
+      List<Bucket> buckets = s3client.listBuckets();
+      for (Bucket bucket : buckets) {
+        if (bucket.getName().equals(bucketname))
+          return errorJson;
+      }
+      errorJson.set("S3BucketException:", Json.newArray().add("Bucket name " + bucketname +
+          " doesn't exist"));
+    } catch (AmazonS3Exception s3Exception) {
+      errorJson.set("BackupConfigException:", Json.newArray().add(s3Exception.getErrorMessage()));
+    }
     return errorJson;
   }
 }
--- a/managed/build.sbt
+++ b/managed/build.sbt
@@ -34,6 +34,7 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-kms" % "1.11.638",
   "com.amazonaws" % "aws-java-sdk-iam" % "1.11.670",
   "com.amazonaws" % "aws-java-sdk-sts" % "1.11.678",
+  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.931",
   "com.cronutils" % "cron-utils" % "9.0.1",
   "io.prometheus" % "simpleclient" % "0.8.0",
   "io.prometheus" % "simpleclient_hotspot" % "0.8.0",


managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
```diff
@@ -179,4 +191,44 @@ public ObjectNode validateDataContent(JsonNode formData) {
...
+   * Currently is checked:
+   *  - S3/AWS - S3 Bucket, whether it exists or not.
+   *
+   * @param formData, customerUUID
+   * @return Json filled with errors
+   */
+  public ObjectNode validateS3DataContent(JsonNode formData, UUID customerUUID) {
+    ObjectNode errorJson = Json.newObject();
+    try {
+      AWSCredentials credentials = new BasicAWSCredentials(
```
sb-yb, line 210: Extract method getS3Client.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
```diff
@@ -179,4 +191,44 @@ public ObjectNode validateDataContent(JsonNode formData) {
...
+        formData.get("data").get("AWS_ACCESS_KEY_ID").asText(),
+        formData.get("data").get("AWS_SECRET_ACCESS_KEY").asText()
+      );
+      String region = Region.getByProvider(Provider.get(customerUUID, CloudType.aws).uuid)
+        .get(0).code.replace('-', '_').toUpperCase();
```
sb-yb, line 215: Check how I convert to aws region in getEC2Client.
I do not recall having to do any string manipulation.  But may be I misremember

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
sb-yb, line 210: Extract method getS3Client.

managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
```diff
@@ -39,6 +39,13 @@ public Result create(UUID customerUUID) {
     if (errorJson.size() > 0) {
       return ApiResponse.error(BAD_REQUEST, errorJson);
     }
+    if (formData.get("name").asText().equals("S3") &&
+      formData.get("data").get("AWS_ACCESS_KEY_ID") != null) {
+      errorJson = configValidator.validateS3DataContent(formData, customerUUID);
+    }
```
sb-yb, line 45: Nit: Change this and all the above validate errors to append to errorJsn or errorjson[]
Will the frontend handle error Json  with multiple errors?
If it can, that is a better validation experience. This is because then user can see and fix all errors together not one by one. 
If frontend cannot be changed easily to do this then leave it.

sb-yb (changes requested):

sb-yb:
Other than minor review comments this generally looks good. You can add unit tests.

managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
jitendra-12113: I checked with FE folks, as of now there is no handling of errors in the UI. Everything is logged into the browser console. If that is the case then browser can handle it pretty well.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
jitendra-12113: Here I had to do this because from the backend we are getting region something like this **ap-south-1** which ideally should be like "ap_south_1".

In your case, we might be getting region details in the correct format i.e **ap_south_1**

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
jitendra-12113: Done. Thanks!

jitendra-12113: Done

managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
```diff
@@ -39,6 +43,15 @@ public Result create(UUID customerUUID) {
     if (errorJson.size() > 0) {
       return ApiResponse.error(BAD_REQUEST, errorJson);
     }
+    if (formData.get("name").asText().equals("S3") &&
+      formData.get("data").get("AWS_ACCESS_KEY_ID") != null) {
+      String region = Region.getByProvider(Provider.get(customerUUID, Common.CloudType.aws).uuid)
```
SergeyPotachev, line 48: This doesn't work if we don't have the AWS provider configured (will raise NPE). The page with Backup settings displays all the settings despite the fact that some of providers are not configured yet.

SergeyPotachev (changes requested):
Needs to be reworked.

managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
sb-yb: He should not be using the region from provider though.  He should use the s3Region he got in the formData.

jitendra-12113: Actually we don't have **s3Region** in form Data. 
The idea here was to take a backup for that we needed **universe** and to create a universe we need a provide right? 
So one provider should be configured to take a backup that could be of any type instead of hard coding it to AWS.
@sb-yb  Any thoughts on this?

managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
jitendra-12113: Its not checking that even if you give any name it would be take that. We made sure that bucket name should start with "s3://", hence made changes accordingly.

managed/src/test/java/com/yugabyte/yw/models/helpers/CustomerConfigValidatorTest.java
```diff
@@ -91,6 +91,21 @@ public void testValidateDataContent_Storage_TwoParamsToCheck(String storageType,
...
+    data.put("BACKUP_LOCATION", "abc");
+    data.put("AWS_ACCESS_KEY_ID", "xyz");
+    data.put("AWS_SECRET_ACCESS_KEY", "secret");
+    ObjectNode result = validator.validateS3DataContent(createFormData(
+      "STORAGE",
+      "S3",
+      data),
+      "AP_SOUTH_1"
+    );
+    assertEquals(1, result.size());
```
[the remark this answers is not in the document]
jitendra-12113: I think we need to pass valid creds to check  **bucket not found** error or whether bucket already exists or not. 
Hence added UT's for negative test cases.

managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java
sb-yb: I am looking for something like:
```
    validators.add(new PreflightS3BucketValidator(STORAGE_TYPE, NAME_S3, BACKUP_LOCATION_FIELDNAME, ...));
```
and the logic for preflight checks should be in there. 
lets talk if you think it is not possible.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
```diff
@@ -179,4 +194,48 @@ public ObjectNode validateDataContent(JsonNode formData) {
...
+      errorJson.set("BackupConfigException", Json.newArray().add("Bucket name " + bucketname +
+          " doesn't exist"));
+    } catch (AmazonS3Exception s3Exception) {
+      errorJson.set("BackupConfigException", Json.newArray().add(s3Exception.getErrorMessage()));
+    }
+    return errorJson;
+  }
+
+  // TODO: move to some common utils or AWSCloudImpl file
+  public static AmazonS3 getS3Client(JsonNode formData) {
```
sb-yb, line 230: take access key and secret params here.

sb-yb, line 230: create 
interface S3ClientFactory {
   AmazonS3 create(String key, String secret);
}

Then implement that factory (code from this method)
Then  take S3ClientFactory in ctor (DI)

In tests you can inject mocks3clientfactory that gives mock s3client 
you can then test based on various listBuckets behaviors failures as well as success cases.

The comment is so you know how to unit test this type of change. 
It would be nice if you add such test but its okay if you just do manual testing by bringing up the server.

sb-yb (approved):
2 comments 
one about test is nit.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
```diff
@@ -61,6 +72,10 @@ public void validate(String type, String name, JsonNode data, ObjectNode errorsC
       if (this.type.equals(type) && this.name.equals(name)) {
         JsonNode value = data.get(fieldName);
         doValidate(value == null ? "" : value.asText(), errorsCollector);
+        if (this.name.equals("S3") &&
```
SergeyPotachev, line 75: Can't understand the reason why it is here. You can create one mare validating class and add it to `validators`. Already inside it you can check whether "AWS_ACCESS_KEY_ID" is present or not and to do the checks.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
```diff
@@ -179,4 +194,48 @@ public ObjectNode validateDataContent(JsonNode formData) {
...
+      AmazonS3 s3client = getS3Client(formData);
+      List<Bucket> buckets = s3client.listBuckets();
+      for (Bucket bucket : buckets) {
+        if (bucket.getName().equals(bucketname))
+          return errorJson;
+      }
+      errorJson.set("BackupConfigException", Json.newArray().add("Bucket name " + bucketname +
+          " doesn't exist"));
+    } catch (AmazonS3Exception s3Exception) {
+      errorJson.set("BackupConfigException", Json.newArray().add(s3Exception.getErrorMessage()));
```
SergeyPotachev, line 224: Here is another question - in other places we are using a pattern errorJson.set("fieldname", "error string"). Using the field name UI can select the proper field and mark it as erroneous (later). Your usage is different and doesn't look meaningful.

SergeyPotachev (changes requested):
Some comments.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
sb-yb: Thats what I had suggested in my original suggestion.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
jitendra-12113: We can move this AmazonS3 create(String key, String secret); to CloudAPI.java interface.
And will provide there implementation in AwsCloudImpl.java.
Does this sound OK? 

I don't think we should create an interface for this since we already have an interface created.
Please suggest! Thanks!

https://github.com/yugabyte/yugabyte-db/blob/master/managed/src/main/java/com/yugabyte/yw/cloud/CloudAPI.java

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
jitendra-12113: I thought of going that route but will have to create another nested class and extends ConfigValidator and do the validations there. 
This sounds ok now. Will do this changes. Thanks!

In general, for all the API's can we have some kind of design patterns implemented in place may be TemplateMethod design pattern where we will have 

Abstract class AbstractCommand {

 public void execute() {
  prepareToExecute(); - for all the input validations
  doExecute(); - to process the inputs
  doGetResults(); - format the result
  }

@sb-yb @SergeyPotachev  Any thoughts of above? Thanks!

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
jitendra-12113: @SergeyPotachev  @sb-yb  How are we going to handle n/w failure, timed out or something erroneous which is not specific to any field. In that case do we want to throw error message in the form of string instead of errorJson Array?

sb-yb: It is still a network error during validation of that field. So you should still put it under fieldname.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
sb-yb: That seems reasonable.

managed/src/test/java/com/yugabyte/yw/models/helpers/CustomerConfigValidatorTest.java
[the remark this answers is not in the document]
sb-yb: Answered above. Use mock s3 impl that pretends that creds are valid and gives success response for positive cases.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
sb-yb: There is already a way to do this sort of separation cleanly in play framework. We do not use that as of now. But we can. One obstacle was lack of exception handler. Now that we have that may be we can simplify this further. 
But that is independent. What @SergeyPotachev has added here is plugging in small validation logic specific to a single field and easily add data driven unit tests for that. That will be useful no matter what's the high level invocation of validation logic pans out.

sb-yb:
Whats the status of this PR?

jitendra-12113:
@sb-yb  This is on hold because of this https://github.com/yugabyte/yugabyte-db/issues/6460(Had to revert due to itest failure). There will be merge conflicts. So wanted #6460 to go first. 
Still not sure who will make changes into itest file. 
 Please suggest!

mjoshi0923 (approved):
Other then minor comment, LGTM for UI.

managed/ui/src/components/config/Storage/StorageConfigurationContainer.js
```diff
@@ -26,6 +27,18 @@ const mapDispatchToProps = (dispatch) => {
   return {
     addCustomerConfig: (config) => {
       return dispatch(addCustomerConfig(config)).then((response) => {
+        if (response.error) {
+          const errorMessageObject = response.payload?.response?.data?.error || response.payload.message;
+          Object.keys(errorMessageObject).forEach((errorKey) => {
```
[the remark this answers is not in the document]
gaurav061: Since all the other instance of error payload is different than this I have not done this, we can create a generic component to handle all categories.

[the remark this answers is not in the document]
gaurav061: For that, we must have some definite cases

Issue #7024, opened 2021-01-29
[Platform][Bug] Unable to edit number of nodes in Availability Zones section


sshev:
STR:
- open AWS portal or run platform UI with API pointed to AWS portal
- open any existing universe for editing
- try to increase nodes number in the Availability Zones section
- observe UI app crashed in browser with js error in console `Uncaught TypeError: Cannot read property 'replicationFactor' of null`
![image](https://user-images.githubusercontent.com/11736711/106225500-1f204b00-619a-11eb-8ea9-a7a8458d88ad.png)

Bug above was introduced in https://github.com/yugabyte/yugabyte-db/commit/ad3a72ae2f48f5f9641305bc565e1345169903a1 commit.

cc @streddy-yb @chirag-yb

chirag-yb:
@sshev Which releases are impacted by this bug? Has this made into 2.4 by any chance?

gaurav061:
@chirag-yb  No this was not backported so it is not in 2.4

gaurav061:
PR: https://github.com/yugabyte/yugabyte-db/pull/7029

gaurav061:
@sshev This issue was due to null currentPlacementStatus prop. Earlier we were not getting the AZ info on the initial load it was only coming if you refresh the page or make any changes which result to call the cluster_resources and cluster_configure API's and thus setting the above prop to a valid data.

gaurav061:
Fixed PR: https://github.com/yugabyte/yugabyte-db/pull/7029

PR #7029, opened 2021-01-29
[#7024] Platform: Fixed edit number of nodes in AZ section

gaurav061:
Description: The user was not able to edit the numbers of nodes from the AZ section while editing the universe.

### DIFF
--- a/managed/ui/src/components/universes/UniverseForm/AZSelectorTable.js
+++ b/managed/ui/src/components/universes/UniverseForm/AZSelectorTable.js
@@ -110,23 +110,9 @@ export default class AZSelectorTable extends Component {
       universe: { currentPlacementStatus, universeConfigTemplate },
       clusterType
     } = this.props;
-    console.log("This props currentPlacementStatus", currentPlacementStatus)
     const universeTemplate = _.clone(universeConfigTemplate.data);
     const currentAZState = [...this.state.azItemState];
     const replicationFactor = currentPlacementStatus?.replicationFactor;
-    // if (!replicationFactor) {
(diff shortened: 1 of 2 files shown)


gaurav061:
Fix: currentPlacementStatus prop was null initially when we land to edit universe page, so in mounting lifecycle i fetched this info if it is undefined or null.
Please add your comments

sshev (changes requested):
Before https://github.com/yugabyte/yugabyte-db/commit/ad3a72ae2f48f5f9641305bc565e1345169903a1 commit the `currentPlacementStatus` prop was filled here https://github.com/yugabyte/yugabyte-db/blob/0857a10eed13435447efd21f651368f9b84dd028/managed/ui/src/components/universes/UniverseForm/AZSelectorTable.js#L469
Now this never happens due to extra conditions added at ClusterFields component:
`{placementStatusOnprem  || regionAndProviderDefined && azSelectorTable}`

This seems to be the real root cause, so let's investigate in that direction pls.

gaurav061:
After adding the above condition `{placementStatusOnprem` || regionAndProviderDefined && azSelectorTable}` AzSelector component is being rendered with this `_.isEqual(universeConfigTemplate, this.props.universe.universeConfigTemplate)` as true and thus it is blocking this `this.props.setPlacementStatus(placementStatusObject); `  So, we need this call to be made in `UNSAFE_componentWillMount` lifecycle for 1st time rendering for subsequent rending the call will be made from `UNSAFE_componentWillReceiveProps` lifecycle. 
Please add to it.

sshev (approved):

Issue #6784, opened 2021-01-05
ysql: use all rather than 0.0.0.0/0, ::0/0 in hba conf


jaki:
From https://www.postgresql.org/docs/11/auth-pg-hba-conf.html,

> 0.0.0.0/0 represents all IPv4 addresses, and ::0/0 represents all IPv6 addresses.

> You can also write all to match any IP address

"all" is simpler and only takes one line, so we should switch to that.  This involves the default HBA config in `src/yb/yql/pgwrapper/pg_wrapper.cc` `WritePgHbaConfig` and docs.  **Docs should be done separately** (maybe create a separate issue to track it).

pratikSethi:
Hi @jaki, I would like to work on this. Can I take this up?

jaki:
Hi, @pratikSethi.  Yes, you can.

jaki:
Hi, @pratikSethi.  Have you made any progress on this or plan on doing so in the next few days?  If not, I'd like others to feel welcome picking this up.

pratikSethi:
Hi @jaki, please give me time till Monday. I intend to work on it tomorrow or day after as I usually don't get time during the weekdays. If there is no update from me by then feel free to assign it to someone else. Thanks for your patience.

m-iancu:
Fixed by 784f61b.

PR #7046, opened 2021-02-01
[#6784] ysql: use all rather than 0.0.0.0/0, ::0/0 in hba conf

pratikSethi:
Replaced the occurrences of 0.0.0.0/0 and ::0/0 for ipv4 and ipv6 addressing with **all** in hba.conf


src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -246,8 +244,8 @@ Result<string> WritePgHbaConfig(const PgProcessConf& conf) {
 
   // Enforce a default hba configuration so users don't lock themselves out.
   if (lines.empty()) {
-    lines.push_back("host all all 0.0.0.0/0 trust");
-    lines.push_back("host all all ::0/0 trust");
+    lines.push_back("host all all all trust");
+    lines.push_back("host all all all trust");
```
iSignal, line 248: we don't need both lines now

src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -232,10 +232,8 @@ Result<string> WritePgHbaConfig(const PgProcessConf& conf) {
     const auto host_type =  conf.enable_tls ? "hostssl" : "host";
     const auto auth_method = FLAGS_ysql_enable_auth ? (conf.enable_tls ? "md5 clientcert=1" : "md5")
                                                     : "cert";
-
-    for (const auto addr : {"0.0.0.0/0", "::0/0"}) {
-      lines.push_back(Format("$0 all all $1 $2", host_type, addr, auth_method));
-    }
+    lines.push_back(Format("$0 all all all $1", host_type, auth_method));
+    lines.push_back(Format("$0 all all all $1", host_type, auth_method));
```
iSignal, line 236: we don't need both lines now

iSignal (changes requested):

src/yb/yql/pgwrapper/pg_wrapper.cc
pratikSethi: You are right, I have updated my PR.

src/yb/yql/pgwrapper/pg_wrapper.cc
pratikSethi: Change incorporated, thanks!

iSignal (approved):

PR #7061, opened 2021-02-02
Update netlify config

polarweasel:
Update the Netlify configuration as follows:

- Specify current Hugo version to override environment variable. This allows us to update to a new Hugo version without having to share Netlify access credentials with a PR author.
- Add redirects for `legal` and `faq` to always point to `/latest/legal` and `/latest/faq`
- Remove a few redundant redirects

**PREVIEW OF THIS BRANCH** is at https://6018a8945856ba00080f88a6--infallible-bardeen-164bc9.netlify.app/


yb-andrew (approved):
LG. Make sure you remove the HUGO version from the web UI after it lands.

polarweasel:
> LG. Make sure you remove the `HUGO_VERSION` from the web UI after it lands.

Will do. The UI version is superseded by config files, so it won't conflict, but best not to have it anymore, you're right.

yb-andrew (approved):

PR #7074, opened 2021-02-03
Finish missing sentence under schema change in read replica doc (part 2)

polarweasel:
Same as #7072, but for older docs. :)


zhaoalex (approved):
lgtm, sorry for missing these :P

PR #7099, opened 2021-02-04
netlify update: redirect release notes to latest

polarweasel:
Adding a redirect so all /version/releases/* point to /latest/releases/*

More of these to come, as we identify more redundant content that should be version-independent...

(Also a couple of small random fixes)


ameyb (approved):
LGTM!

PR #7138, opened 2021-02-05
docs: updates to v2.2.x release notes

polarweasel:
Adding entry for 2.2.6, and updating links.

While I was in there, fixed some formatting issues.


lizayugabyte (commented):
We should use n instead of x for versions in our docs: v2.2.x should be replaces with v2.2.n

polarweasel:
@lizayugabyte thanks for that note. that 2.2.x is just in the commit message, but i'll adjust in future

polarweasel:
Almost ready to go. Just waiting for 2.2.4 to be re-released on Replicated so we can have correct links and docker image name.

ameyb (approved):
Looks good!

PR #7152, opened 2021-02-06
correct links to download stable packages

polarweasel:
Quick fix to correct version numbers on download links for the stable release.


bllewell (approved):
LGTM

PR #7159, opened 2021-02-08
[Platform] Allow only one cluster per K8s provider with namespace

bhavin192:
This change is applicable to the providers which have AZs with
KUBENAMESPACE set in the config.

Currently, we don't support multiple clusters for these type of
providers. This change makes sure that such universe / cluster
creation fails immediately.

Fixes https://github.com/yugabyte/yugabyte-db/issues/6931

Scenarios tested:
- Created a provider with namespace set for one of the AZs.
- Created universe with that provider.
- Tried creating another universe with same provider, it returned 400.
- Tried adding a read replica to the universe using same provider,
  it failed as expected.
- Created a provider without namespace, tried creating two single zone
  universe with it. Those work as expected.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
@@ -59,11 +59,13 @@ import com.yugabyte.yw.common.PlacementInfoUtil;
 import com.yugabyte.yw.common.Util;
 import com.yugabyte.yw.models.AccessKey;
 import com.yugabyte.yw.models.Audit;
+import com.yugabyte.yw.models.AvailabilityZone;
 import com.yugabyte.yw.models.CertificateInfo;
 import com.yugabyte.yw.models.Customer;
 import com.yugabyte.yw.models.CustomerTask;
 import com.yugabyte.yw.models.HealthCheck;
 import com.yugabyte.yw.models.Provider;
+import com.yugabyte.yw.models.Region;
 import com.yugabyte.yw.models.TaskInfo;
 import com.yugabyte.yw.models.Universe;
 import com.yugabyte.yw.models.helpers.NodeDetails;
@@ -514,6 +516,14 @@ public class UniverseController extends AuthenticatedController {
           if (provider.getConfig().containsKey("USE_HOSTNAME")) {
             c.userIntent.useHostname =
               Boolean.parseBoolean(provider.getConfig().get("USE_HOSTNAME"));
+          }
+        }
+
+        if (c.userIntent.providerType.equals(CloudType.kubernetes)) {
+          try {
+            checkK8sProviderAvailability(provider);
+          } catch (IllegalArgumentException e) {
+            return ApiResponse.error(BAD_REQUEST, e.getMessage());
           }
         }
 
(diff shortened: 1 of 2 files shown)


managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -1762,4 +1781,43 @@ private UniverseDefinitionTaskParams bindFormDataToTaskParams(ObjectNode formDat
...
+    for (Region r : Region.getByProvider(provider.uuid)) {
+      for (AvailabilityZone az : AvailabilityZone.getAZsForRegion(r.uuid)) {
+        if (az.getConfig().containsKey("KUBENAMESPACE")) {
+          isNamespaceSet = true;
+        }
+      }
+    }
+
+    if (isNamespaceSet) {
+      for (Universe uu : Universe.getAllUuids()) {
```
iSignal, line 1804: minor: name these as thisUniv and otherUniv to make it clear instead of uu and u

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -1762,4 +1781,43 @@ private UniverseDefinitionTaskParams bindFormDataToTaskParams(ObjectNode formDat
...
+        Universe u = Universe.get(uu.universeUUID);
+        List<Cluster> clusters = u.getUniverseDetails().getReadOnlyClusters();
+        clusters.add(u.getUniverseDetails().getPrimaryCluster());
+        for (Cluster c : clusters) {
+          UUID pu = UUID.fromString(c.userIntent.provider);
```
iSignal, line 1809: minor: providerUUID

iSignal (approved):
Looks good!

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
bhavin192: Made it `allUniverseUUIDs` and `u`.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -1762,4 +1781,42 @@ private UniverseDefinitionTaskParams bindFormDataToTaskParams(ObjectNode formDat
...
+        List<Cluster> clusters = u.getUniverseDetails().getReadOnlyClusters();
+        clusters.add(u.getUniverseDetails().getPrimaryCluster());
+        for (Cluster c : clusters) {
+          UUID providerUUID = UUID.fromString(c.userIntent.provider);
+          if (providerUUID.equals(providerToCheck.uuid)) {
+            String msg = "Universe " + u.name + " (" + u.universeUUID
+              + ") already exists with provider "
+              + providerToCheck.name + " (" + providerToCheck.uuid
+              + "). Only one universe can be created with providers having KUBENAMESPACE set "
+              + "in the AZ config.";
```
bhavin192, line 1814: @iSignal What do you think about this error message? The resulting message is something like this:
```
Universe my-univ1 (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx) already exists with provider ocp-multi-az (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx). Only one universe can be created with providers having KUBENAMESPACE set in the AZ config.
```

bhavin192: Had a chat with Sanketh offline, we are fine with this message for now.

Issue #6555, opened 2020-12-04
[Platform] Add optional retention duration for manual backups


rao-vasireddy:
We can specify a retention for scheduled backups at the moment but not for backups done manually. It will be good to be able to specify how long the manual should be retained at the time of initiating it. If the retention duration is left blank, then the backup is not deleted by YW.

PR #7162, opened 2021-02-08
[#6555] Platform: Manual Backup Retention

mahendranbhat:
**Description:**
We can specify retention for scheduled backups at the moment but not for backups done manually. It will be good to be able to specify how long the manual should be retained at the time of initiating it. If the retention duration is left blank, then the backup is not deleted by YW.

**Test Plan:**

- Go to the backup section and add manual backup with expiry. Wait till it gets expires or deleted.
- It should delete the backup after the retention time.
- ( For testing I manually edited expiry time in BE so that I do not have to wait for 1 day ) 
- The system did not delete the backup for paused universes.
- The system deleted the backups if it is not paused or even the universe does not exist.
-  Tested manual backup, manual backup with expiry, delete the backup. All are working fine ( NFS ) .

**sbt test:**
[info] Passed: Total 1288, Failed 0, Errors 0, Passed 1284, Skipped 4
[success] Total time: 984 s (16:24), completed 27 Apr, 2021 11:53:07 PM


managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -184,10 +184,10 @@ else if (cronExpression != null) {
             this.runMultiTableBackupsTask(schedule);
           }
         }
-        List<Backup> expiredBackups = Backup.getExpiredBackups(schedule.scheduleUUID);
-        for (Backup backup : expiredBackups) {
-          this.runDeleteBackupTask(backup, schedule);
-        }
+      }
+      List<Backup> expiredBackups = Backup.getExpiredBackups();
```
Arnav15, line 188: We need to not run this per schedule now. So you can move this outside the scope of the for loop of schedules.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -268,28 +268,35 @@ public void runMultiTableBackupsTask(Schedule schedule) {
...
-  public void runDeleteBackupTask(Backup backup, Schedule schedule) {
+  public void runDeleteBackupTask(Backup backup) {
     if (backup.state != Backup.BackupState.Completed) {
       LOG.warn("Cannot delete backup {} since it is not in completed state.",
                 backup.backupUUID);
       return;
     }
     BackupTableParams backupParams = Json.fromJson(backup.backupInfo, BackupTableParams.class);
     Universe universe = null;
+    boolean isScheduled =  (backup.getScheduleUUID()!=null) ? true: false;
```
Arnav15, line 279: Nit: Fix spacing in ternary operation.

managed/src/main/java/com/yugabyte/yw/models/Backup.java
```diff
@@ -190,11 +190,10 @@ public static Backup fetchByTaskUUID(UUID taskUUID) {
       .findOne();
   }
 
-  public static List<Backup> getExpiredBackups(UUID scheduleUUID) {
+  public static List<Backup> getExpiredBackups() {
```
Arnav15, line 193: As mentioned, we should only get expired backups of universes that exist. Otherwise we will keep trying to run delete tasks for backups that cannot be deleted. So after fetching expired backups, you can do a java stream to check all backups if they have universes or not. You can also technically do this in the query itself, but I think a stream to check for existing universe is good enough for now.

Arnav15 (changes requested):

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
mahendranbhat: It is outside of the `for` loop. Guess you got confuse with `try` block `}`

managed/src/main/java/com/yugabyte/yw/models/Backup.java
mahendranbhat: yeah, missed it. Will update the logic.

managed/src/main/java/com/yugabyte/yw/models/Backup.java
```diff
@@ -190,14 +190,17 @@ public static Backup fetchByTaskUUID(UUID taskUUID) {
...
     // Get current timestamp.
     Date now = new Date();
-    return Backup.find.query().where()
-      .eq("schedule_uuid", scheduleUUID)
+    List<Backup> backupList =  Backup.find.query().where()
       .lt("expiry", now)
       .eq("state", BackupState.Completed)
       .findList();
+    return backupList.stream().filter(
```
Arnav15, line 200: Just a thought, why not fetch all universes first, make a set of the UUIDs and then try to see if the universe UUID for the backup exists in the set? The reason I think that would be better is since we wouldn't have to query the DB multiple times to get the universe. Eventually if the backups are a lot, things might pile up.

mahendranbhat: Done

managed/src/main/java/com/yugabyte/yw/models/Backup.java
mahendranbhat: Updated.

SergeyPotachev:
It is unclear why you covered with junit tests only Schedule.java which is actually unchanged in this PR/diff.
And at the same time you changed and didn't cover Backup.java and Scheduler.java.

managed/src/main/java/com/yugabyte/yw/models/Backup.java
Arnav15: @mahendranbhat Fix the getAllUuids function in the Universe class itself. It seems like it isn't referenced from anywhere, and we expect only UUIDs. Also, return a set instead of a list.

SergeyPotachev (changes requested):
Cover Backup.java+ Scheduler.java with junit tests.

Arnav15:
@SergeyPotachev Good find! I totally didn't see the tests were added to Schedule.java and not Backup.java. @mahendranbhat you can just move the test from the Schedule class to the Backup class.

Arnav15:
And @mahendranbhat if you need reference code for mocking the Scheduler.java class, you can take a look at the HealthCheckerTest.java class :)

mahendranbhat:
> And @mahendranbhat if you need reference code for mocking the Scheduler.java class, you can take a look at the HealthCheckerTest.java class :)

Yeah, I will move the test case and the above info is really helpful.

managed/src/main/java/com/yugabyte/yw/models/Backup.java
mahendranbhat: @Arnav15 thought of fixing, but after seeing the comment above the function, thought it would serve a special purpose and written intentionally. I will refactor it as we are not using it anywhere.

Arnav15 (approved):
The backend looks good. For the UI, I hope we have tested the option is available both from the backups tab and the tables tab.

managed/ui/src/components/tables/CreateBackup/CreateBackup.js
```diff
@@ -519,14 +519,12 @@ export default class CreateBackup extends Component {
...
                     />
-                    {isScheduled && (
-                      <Field
-                        name="timeBeforeDelete"
-                        type={'number'}
-                        component={YBFormInput}
-                        label={'Number of Days to Retain Backup'}
-                      />
-                    )}
+                    <Field
```
sshev, line 522: @yb-andrew pls check this change as you most likely more familiar with backups code

managed/src/main/java/com/yugabyte/yw/models/Universe.java
```diff
@@ -275,8 +275,11 @@ public static boolean checkIfUniverseExists(String universeName) {
    * WARNING: Returns partially filled Universe objects!!
    * @return list of UUIDs of all universes
    */
-  public static List<Universe> getAllUuids() {
-    return find.query().select("universeUUID").findList();
+  public static Set<UUID> getAllUuids() {
```
sb-yb, line 278: 1> This function was buggy - it did not support multi-tenant use case. You should look for UUIDs only in current customer.
2> selecting universeUUID and getting list of universes then processing it to get UUID list is very roundabout. Use findIds instead.

I believe I have fixed this at head during another change I was working on. So just sync.

mahendranbhat: Sure.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -249,28 +249,35 @@ private void runMultiTableBackupsTask(Schedule schedule) {
...
-    JsonNode params = schedule.getTaskParams();
     DeleteBackup.Params taskParams = new DeleteBackup.Params();
     taskParams.customerUUID = customerUUID;
     taskParams.backupUUID = backup.backupUUID;
     UUID taskUUID = commissioner.submit(TaskType.DeleteBackup, taskParams);
-    ScheduleTask.create(taskUUID, schedule.getScheduleUUID());
+    if (isScheduled) {
+      Schedule schedule = Schedule.get(backup.getScheduleUUID());
+      JsonNode params = schedule.getTaskParams();
+      ScheduleTask.create(taskUUID, backup.getScheduleUUID());
```
sb-yb, line 279: @Arnav15 Remind me why we need to create a ScheduleTask here for deletion? and create customer task as well?

sb-yb (changes requested):
We need to backport this to 2.4 as this also fixes the bug 
https://github.com/yugabyte/yugabyte-db/issues/7780

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
sb-yb: We do not need this block as discussed.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -249,28 +249,35 @@ private void runMultiTableBackupsTask(Schedule schedule) {
...
     }
     BackupTableParams backupParams = Json.fromJson(backup.backupInfo, BackupTableParams.class);
     Universe universe = null;
+    boolean isScheduled = (backup.getScheduleUUID() != null) ? true : false;
     try {
       universe = Universe.get(backupParams.universeUUID);
     } catch (Exception e) {
-      schedule.stopSchedule();
+      if (isScheduled){
+        Schedule schedule = Schedule.get(backup.getScheduleUUID());
```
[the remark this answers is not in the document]
mahendranbhat: guess ` boolean isScheduled = (backup.getScheduleUUID() != null) ? true : false;` it will raiseNPE because of this condition.
`isScheduled` will be true only if `backup.getScheduleUUID()` is not null.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -249,28 +249,35 @@ private void runMultiTableBackupsTask(Schedule schedule) {
...
+        schedule.stopSchedule();
```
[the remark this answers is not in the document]
mahendranbhat: same as above

managed/src/main/java/com/yugabyte/yw/models/Universe.java
mahendranbhat: I do not think I can reuse the same in this PR as we need to get the expired backups independent of the customer.

sb-yb (changes requested):
I stole part of this change - modified and merged to fix the other bug :)
Pl sync to head. 
Also in getExpiredBkps  add filtering out if on paused universe.

sshev (commented):
UI part looks good

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -80,7 +80,7 @@ private void initialize() {
   /**
    * Iterates through all the schedule entries and runs the tasks that are due to be scheduled.
    */
-  private void scheduleRunner() {
```
mahendranbhat, line 83: `visiblefortesting` not working with private so changed to public.

sb-yb: VisibleForTesting is (mostly) documentation annotation - when you have to make things public just so that tests can see it you have to add that annotation. 
So you have to make it public and *ALSO* add that annotation.

sb-yb: And you should keep it default (package) scope instead of public since your test is in the same package.

sb-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
SergeyPotachev: I'm for the default (package) visibility too.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Test
+  public void scheduleManualBackupWithExpiryTest() {
+    UUID fakeTaskUUID = UUID.randomUUID();
+    when(mockCommissioner.submit(Matchers.any(), Matchers.any())).thenReturn(fakeTaskUUID);
+
+    defaultCustomer = ModelFactory.testCustomer();
+    s3StorageConfig = ModelFactory.createS3StorageConfig(defaultCustomer);
+
+    UUID universeUUID = UUID.randomUUID();
```
SergeyPotachev, line 69: I don't see this variable is used below.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,81 @@
...
+    Universe universe = ModelFactory.createUniverse(defaultCustomer.getCustomerId());
+
+    Backup backup = ModelFactory.createBackupWithExpiry(defaultCustomer.uuid, universe.universeUUID,
+        s3StorageConfig.configUUID);
+    backup.transitionState(Backup.BackupState.Completed);
+
+    scheduler.scheduleRunner();
+ 
+    CustomerTask task = CustomerTask.get(defaultCustomer.uuid, fakeTaskUUID);
+    assertEquals(task.getType(), CustomerTask.TaskType.Delete);
```
SergeyPotachev, line 79: Usually we are using expected value as the first parameter and a value to check as the second one.
  assertEquals(CustomerTask.TaskType.Delete, task.getType());

SergeyPotachev (changes requested):
Small comments.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -251,6 +254,17 @@ private void runMultiTableBackupsTask(Schedule schedule) {
...
+    UUID universeUUID = backup.getBackupInfo().universeUUID;
+    try {
+      Universe universe = Universe.get(universeUUID);
+      if (universe.getUniverseDetails().universePaused) {
+        LOG.warn("Cannot delete the backup since the universe {} is currently in a paused state.",
+            universeUUID.toString());
+        return;
+      }
+    } catch (Exception e) {
+      LOG.warn("Universe not found for the backup {}.", backup.backupUUID.toString());
```
mahendranbhat, line 266: I believe backup deletion does not need the universe. LMK if I have to return here.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -156,7 +159,7 @@ else if (cronExpression != null) {
             }
           }
         }
-
+
```
SergeyPotachev, line 162: Trim the line.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
SergeyPotachev, line 266: "Universe {} not found for the backup {}." - is better as we don't need to search universe uuid inside the backup configuration.

SergeyPotachev (changes requested):
Overall looks quite good. Small comments.

SergeyPotachev:
Add test steps for a case when a universe is paused.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
```diff
@@ -251,6 +254,17 @@ private void runMultiTableBackupsTask(Schedule schedule) {
   }
 
   private void runDeleteBackupTask(Customer customer, Backup backup) {
+    UUID universeUUID = backup.getBackupInfo().universeUUID;
+    try {
+      Universe universe = Universe.get(universeUUID);
+      if (universe.getUniverseDetails().universePaused) {
```
sb-yb, line 260: I was expecting this filtering in `getExpiredBackups` 
That's where rest of the filtering is done.

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
sb-yb: Above suggestion makes it a moot question. Because getExpired will check if there is universe && it is paused skip that. But if there is no universe it will return that backup. 
Most likely that deletion will fail but we are good on that due to the other fix I made earlier this week.

sb-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/scheduler/Scheduler.java
mahendranbhat: yeah, the check is present. Adding paused check there only.

managed/src/main/java/com/yugabyte/yw/models/Backup.java
```diff
@@ -210,7 +210,8 @@ public static Backup fetchByTaskUUID(UUID taskUUID) {
       Customer customer = Customer.get(customerUUID);
       Set<UUID> allUniverseUUIDs = Universe.getAllUUIDs(customer);
       List<Backup> backupsWithValidUniv = backups.stream()
-        .filter(backup -> allUniverseUUIDs.contains(backup.getBackupInfo().universeUUID))
+        .filter(backup -> allUniverseUUIDs.contains(backup.getBackupInfo().universeUUID) && 
+            !Universe.get(backup.getBackupInfo().universeUUID).getUniverseDetails().universePaused)
```
sb-yb, line 214: Given that we are allowing deletion of Universe without deleting backups; as discussed before this check should be:
**Universe not present || Universe not paused**

So remove the allUniverseUUIDs check. 

Also `Universe.get` throws exception when not found. This has been a source of a bug. I am replacing it to Univers.getOrBadRequest and Universe.maybeGet  methods so this does not come as a surprise. 

Meanwhile you should add a static method to Universe:
```
    static boolean isPaused(uuid) {
        return true if universe exists && is paused;
        else return false;
    }
```

Then you can filter on ```!Universe.isPaused(universeUUID)```

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,80 @@
...
+    scheduler = new com.yugabyte.yw.scheduler.Scheduler(mockActorSystem,
+        mockExecutionContext, mockCommissioner);
+  }
+
+  @Test
+  public void scheduleManualBackupWithExpiryTest() {
+    UUID fakeTaskUUID = UUID.randomUUID();
+    when(mockCommissioner.submit(Matchers.any(), Matchers.any())).thenReturn(fakeTaskUUID);
+
+    defaultCustomer = ModelFactory.testCustomer();
```
sb-yb, line 66: move these model setups as part of `@Before` So each test case wont have to repeat it.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,80 @@
...
+import scala.concurrent.duration.Duration;
+import scala.concurrent.duration.FiniteDuration;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+
+@RunWith(MockitoJUnitRunner.class)
+public class SchedulerTest extends FakeDBApplication {
```
sb-yb, line 39: Nice!

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,80 @@
...
+  @Test
+  public void scheduleManualBackupWithExpiryTest() {
+    UUID fakeTaskUUID = UUID.randomUUID();
+    when(mockCommissioner.submit(Matchers.any(), Matchers.any())).thenReturn(fakeTaskUUID);
+
+    defaultCustomer = ModelFactory.testCustomer();
+    s3StorageConfig = ModelFactory.createS3StorageConfig(defaultCustomer);
+
+    Universe universe = ModelFactory.createUniverse(defaultCustomer.getCustomerId());
+
```
sb-yb, line 70: Another test (or update this one)
1. Start as paused universe.
2. Then check that backup is not deleted.
3. unpause the universe and retry this time it deletes.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,80 @@
...
+    Backup backup = ModelFactory.createBackupWithExpiry(defaultCustomer.uuid, universe.universeUUID,
+        s3StorageConfig.configUUID);
+    backup.transitionState(Backup.BackupState.Completed);
+
```
sb-yb, line 74: Another test: 
delete universe for the backup
Now try and make sure that backup gets deleted.

IIUC this test will fail with current code.

sb-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/models/Universe.java
```diff
@@ -895,4 +895,14 @@ public void incrementVersion() {
   public static boolean existsCertificate(UUID certUUID, UUID customerUUID) {
     return universeDetailsIfCertsExists(certUUID, customerUUID).size() != 0;
   }
+
+  static boolean isUniversePaused(UUID uuid) {
+    try {
```
sb-yb, line 900: logic is correct.. But you need to pull the latest. 
Use newly added  `maybeGet` method so you dont have to depend on exception for control flow.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    backup.transitionState(Backup.BackupState.Completed);
+    universe.delete();
+    scheduler.scheduleRunner();
+
+    CustomerTask task = CustomerTask.get(defaultCustomer.uuid, fakeTaskUUID);
+    assertEquals(1, Backup.getExpiredBackups().get(defaultCustomer).size());
+    assertEquals(CustomerTask.TaskType.Delete, task.getType());
+  }
+
+  public static void alterUniversePaused(boolean value, Universe universe) {
```
sb-yb, line 108: rename `setUniversePaused`

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    s3StorageConfig = ModelFactory.createS3StorageConfig(defaultCustomer);
+    
+    when(mockActorSystem.scheduler()).thenReturn(mockScheduler);
+    scheduler = new com.yugabyte.yw.scheduler.Scheduler(mockActorSystem,
+        mockExecutionContext, mockCommissioner);
+    
+  }
+
+  @Test
+  public void scheduleManualBackupWithExpiryTest() {
```
sb-yb, line 68: nit: Remove Test suffix from name

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    Universe universe = ModelFactory.createUniverse(defaultCustomer.getCustomerId());
+    Backup backup = ModelFactory.createBackupWithExpiry(defaultCustomer.uuid,
+        universe.universeUUID, s3StorageConfig.configUUID);
+    backup.transitionState(Backup.BackupState.Completed);
+    alterUniversePaused(true, universe);
+
+    scheduler.scheduleRunner();
+
+    assertEquals(0, Backup.getExpiredBackups().get(defaultCustomer).size());
+
```
sb-yb, line 81: check that CustomerTask.get(defaultCustomer.uuid, fakeTaskUUID); returns null.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+  @Test
+  public void scheduleManualBackupWithExpiryTest() {
+    UUID fakeTaskUUID = UUID.randomUUID();
+    when(mockCommissioner.submit(Matchers.any(), Matchers.any())).thenReturn(fakeTaskUUID);
+
+    Universe universe = ModelFactory.createUniverse(defaultCustomer.getCustomerId());
+    Backup backup = ModelFactory.createBackupWithExpiry(defaultCustomer.uuid,
+        universe.universeUUID, s3StorageConfig.configUUID);
+    backup.transitionState(Backup.BackupState.Completed);
+    alterUniversePaused(true, universe);
```
sb-yb, line 76: Insert newline since we are changing to a different test here.

Then add comment: 
// test that we do not delete paused universe

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    scheduler.scheduleRunner();
+
+    assertEquals(0, Backup.getExpiredBackups().get(defaultCustomer).size());
+
+    alterUniversePaused(false, universe);
```
sb-yb, line 82: newline before this and comment. Unpause the universe and make sure that we will delete it.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    when(mockCommissioner.submit(Matchers.any(), Matchers.any())).thenReturn(fakeTaskUUID);
+
+    Universe universe = ModelFactory.createUniverse(defaultCustomer.getCustomerId());
+    Backup backup = ModelFactory.createBackupWithExpiry(defaultCustomer.uuid,
+        universe.universeUUID, s3StorageConfig.configUUID);
+    backup.transitionState(Backup.BackupState.Completed);
+    alterUniversePaused(true, universe);
+
+    scheduler.scheduleRunner();
+
```
sb-yb, line 79: verify mockCommissioner.submit gets called 0 times.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    assertEquals(0, Backup.getExpiredBackups().get(defaultCustomer).size());
+
+    alterUniversePaused(false, universe);
+    scheduler.scheduleRunner();
+
```
sb-yb, line 84: verify mockCommissioner.submit gets called 1 times.

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    scheduler.scheduleRunner();
+ 
+    CustomerTask task = CustomerTask.get(defaultCustomer.uuid, fakeTaskUUID);
+
+    assertEquals(1, Backup.getExpiredBackups().get(defaultCustomer).size());
+    assertEquals(CustomerTask.TaskType.Delete, task.getType());
+  }
+
+  @Test
+  public void scheduleManualBackupWithExpiryWithDeletedUniverseTest() {
```
sb-yb, line 92: nit: remove Test suffix

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
sb-yb: Based on what this test is testing - rename it to:
`schedulerDeletesExpiredBackups`

managed/src/test/java/com/yugabyte/yw/scheduler/SchedulerTest.java
sb-yb: schedulerDeletesExpiredBackups_universeDeleted

sb-yb (changes requested):
Last bit of comments.

sb-yb (approved):

PR #7168, opened 2021-02-08
[#5594] Platform:Handle invalid certs/keys correctly.

jitendra-12113:
[5594][Platform] Handle invalid certs/keys correctly
**Description:**
We wanted to verify the cert at upload time itself.
**Testing:**
When we try to upload selfSigned cert, we were able to upload any
garbage files. Now with this fix, we have handled this scenrio. User
won't be able to upload garbage file. Incase of self signed cert upload
workflow, We are matching the modulus of
cert RSA Public key with the modulus of its private key if its matched
then it means certs are valid otheriwse Invalid.
For CA Cert, we are just validating the Cert format.
Added the unit Tests as well.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
+++ b/managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
@@ -87,7 +87,7 @@ public class CertificateHelper {
 
   public static UUID createRootCA(String nodePrefix, UUID customerUUID, String storagePath) {
     try {
-      KeyPair clientKeyPair = getKeyPairObject();
+      KeyPair keyPair = getKeyPairObject();
 
       UUID rootCA_UUID = UUID.randomUUID();
       Calendar cal = Calendar.getInstance();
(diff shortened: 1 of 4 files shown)


managed/src/main/java/com/yugabyte/yw/common/CertificateHelper.java
```diff
@@ -390,4 +376,87 @@ public static void createChecksums() {
...
+    File keyFile = new File(keyPath);
+    try (JcaPEMWriter keyWriter = new JcaPEMWriter(new FileWriter(keyFile))) {
+      keyWriter.writeObject(keyContent);
+    } catch (Exception e) {
+      LOG.error(e.getMessage());
+      throw new RuntimeException("Save privateKey failed.");
+    }
+  }
+
+  public static void saveCertContent(X509Certificate cert, String certPath) {
```
Arnav15, line 416: Rename similar to `writeKeyFileContentToKeyPath`

Arnav15 (changes requested):
Looks good overall. Just few minor comments.
1) Add a unit test for a bad cert case as well.
2) Revert the submodule changes. You can do so by running `git submodule update --init --recursive`

Arnav15:
Also, can you update the description a little more. Exactly mention on how we are solving the issue and what all changes are made as part of this PR. And add the output of `sbt test`

jitendra-12113:
Created separated PR for this:
https://github.com/yugabyte/yugabyte-db/pull/7264

PR #7173, opened 2021-02-09
Document queries and joins in YSQL - Doc6849

lizayugabyte:


lizayugabyte:
The doc on queries and joins (for the Explorer section of Yugabyte docs) is ready for review.

lizayugabyte:
This doc is live here: https://docs.yugabyte.com/latest/explore/ysql-language-features/queries/
I'll refresh the PR when I implement comments

docs/content/latest/explore/ysql-language-features/queries.md
```diff
@@ -34,29 +34,28 @@ The following `SELECT` statement clauses provide flexiblity and allow you to fin
 - The `DISTINCT` operator allows you to select distinct rows. 
 - The `ORDER BY` clause lets you sort rows.
 - The `WHERE` clause allows you to apply filters to rows.
-- The `LIMIT` and `FETCH` clauses allow you to select a subset of rows from a table. 
+- The `LIMIT` clauses allow you to select a subset of rows from a table.
```
jaki, line 37: `s/clauses/clause/`

docs/content/latest/explore/ysql-language-features/queries.md
```diff
@@ -253,60 +254,72 @@ INSERT INTO employees (name, department) VALUES
...
-Bette Davis 				
+Bette Davis
+```
+
+### Case Sensitivity
+
+Since YSQL is case-sensitive (that is, it treats the same characters differently depending on whether they are uppercase and lowercase), it is important to use correct cases when forming queries. For example, a table called `Employees` would not be recognized as an `employees` table, even if that is what you mean. That is, if your database contains an `employees` table and you want to query it using the following statement, the query execution will result in an error:
+
+```sql
+SELECT name FROM Employees;
```
m-iancu, line 278: As far as I know this section is not quite correct.
Identifiers are normalized to lowercase unless they are quoted.
So, for example:

```
yugabyte=# create table fOo(iD int PRIMARY KEY);
CREATE TABLE
yugabyte=# select Id from FoO; -- any capitalization works if unquoted
 id
----
(0 rows)

yugabyte=# create table "bAr"("iD" int PRIMARY KEY);
CREATE TABLE
yugabyte=# select iD from bAr; -- no capitalization works because identifiers will let lowercased if unquoted
ERROR:  relation "bar" does not exist
LINE 1: select iD from bAr;
                       ^
yugabyte=# select "iD" from "bAr"; -- quoting with the same as the declared capitalization will work.
 iD
----
(0 rows)

yugabyte=# \d
        List of relations
 Schema | Name | Type  |  Owner
--------+------+-------+----------
 public | bAr  | table | yugabyte
 public | foo  | table | yugabyte
(2 rows)
```

lizayugabyte: Fixed this section

lizayugabyte:
This PR has been "in review" for 7 days. If there are no other comments, then I'm going to merge it today.
The doc is available at https://docs.yugabyte.com/latest/explore/ysql-language-features/queries/ -- if you notice anything that should be changed, please open a ticket, assign it to me, and set Project to "Documentation"

lizayugabyte:
Discussed with Mihnea. Decided to publish the doc and then implement review comments via a new PR.

PR #7254, opened 2021-02-16
update Vitess comparison page

polarweasel:
Quick typo fix and minor frontmatter adjustments across the various versions


rkarthik007 (approved):
LG!

Issue #4580, opened 2020-05-26
[docdb] provide a metric to track WAL files size for each node


rao-vasireddy:

triump2020:
@bmatican  could you  pls describe it more detail ?  or give me a  existing metric reference ?

rahuldesirazu:
Hi @triump2020, I think this would make the most sense as a per tablet metric. We don't have to update the metric every time we write to a WAL file- instead we can simply update every time we close (increment) or GC (decrement) a file. Look at src/yb/consensus/log_metrics.cc for other log metrics. I think it makes the most sense to define a new metric here.

triump2020:
@bmatican  thanks !

triump2020:
@rahuldesirazu  thanks !

triump2020:
@bmatican @rahuldesirazu   I  added some commits to this issue ,  but i 'm not sure  it is [the correct place](https://github.com/triump2020/yugabyte-db/blob/issue-4580/src/yb/consensus/log.cc#L1376) where call Increment() function. 
 Pls help me check  and review all commits,  thanks!

triump2020:
@rahuldesirazu  @bmatican   pls check !  thanks again !

PR #7260, opened 2021-02-17
[#issue 4580] add metric for wal files size 

triump2020:


triump2020:
#4580

rahuldesirazu (changes requested):
One thing to note, it may be more useful to track the size of the WAL file, rather than the number of WAL files. Also, we'll want to decrement the size when we GC files (see log.cc for this logic).

triump2020:
@rahuldesirazu   Thanks for your pointers !   1.  you mean that  we should track the tatal size of all the wal files which belong to a  tablet ?  2.  Sorry !  i would decrement the size when we GC files in the next commit

triump2020:
@rahuldesirazu   Pls check the new commit , thanks !   If anything i missed , pls inform me ,glad to correct it !

src/yb/consensus/log.cc
```diff
@@ -559,6 +559,12 @@ Status Log::Init() {
     RETURN_NOT_OK(reader_->GetSegmentsSnapshot(&segments));
     active_segment_sequence_number_ = segments.back()->header().sequence_number();
     LOG_WITH_PREFIX(INFO) << "Opened existing logs. Last segment is " << segments.back()->path();
+
+    if (metrics_) {
+      std::for_each(segments.begin(), segments.end(), 
+                    [this](auto segment) {
+                    this->metrics_->wal_size->IncrementBy(segment->file_size());});
+    }
```
rahuldesirazu, line 567: I believe this can be removed, since we're already incrementing when we close the last segment.

src/yb/consensus/log.cc
```diff
@@ -597,7 +603,12 @@ Status Log::CloseCurrentSegment() {
                       << ": " << footer_builder_.ShortDebugString();
 
   footer_builder_.set_close_timestamp_micros(GetCurrentTimeMicros());
-  return active_segment_->WriteFooterAndClose(footer_builder_);
+
+  auto status = active_segment_->WriteFooterAndClose(footer_builder_);
+  if (status.ok() && metrics_) {
+      metrics_->wal_size->IncrementBy(active_segment_->written_offset());
```
rahuldesirazu, line 609: Lets use the Size() function.

rahuldesirazu (changes requested):
Appreciate your patience on this! Left a few somewhat small comments.

src/yb/consensus/log_metrics.cc
```diff
@@ -38,6 +38,10 @@ METRIC_DEFINE_counter(tablet, log_bytes_logged, "Bytes Written to WAL",
                       yb::MetricUnit::kBytes,
                       "Number of bytes logged since service start");
 
+METRIC_DEFINE_counter(tablet, log_wal_size, "Size of WAL Files",
+                      yb::MetricUnit::kUnits,
```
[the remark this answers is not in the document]
triump2020: NP

src/yb/consensus/log.cc
triump2020: NP!

src/yb/consensus/log.cc
triump2020: @rahuldesirazu    This code would be called by the  `Log::Init()`  would initialize the `log_wal_size` metric ;  if remove it ,  the `log_wal_size` metric would  represents the number of bytes since server start,  instead of the number of bytes  already existing in the server  ?

triump2020:
@rahuldesirazu   Pls review  , thanks !

rahuldesirazu:
Test suite looks good on jenkins! I'm going to share a patch with, once you commit that, I can go ahead and merge. Great work on this!
[0001-Lint-fixes.log](https://github.com/yugabyte/yugabyte-db/files/6126497/0001-Lint-fixes.log)

triump2020:
@rahuldesirazu   I  have committed the patch for lint fixes, you can merge, thanks !

BTW,  had there been some tools for checking code format ?

triump2020:
Thanks for your patient pointers on this issue,  really appreciate !

rahuldesirazu (approved):

Issue #7002, opened 2021-01-27
[Platform] Script to create a universe from JSON config file


streddy-yb:
Provide a script that creates a universe using these three command line arguments: Platform URL, API token, and path to the file containing the Universe config in JSON format.
cc @chirag-yb

chirag-yb:
User should be able to dump a json payload from an existing universe and use that as an input for the script. 
Need to make sure this works for 2.2.3 and 2.4.0

streddy-yb:
We need scripts for the following actions:
- Dump json payload for a given universe
- Create a universe
- Destroy an existing universe

jaydeepkumara:
PR link :- https://github.com/yugabyte/yugabyte-db/pull/7302/files

PR #7302, opened 2021-02-19
[#7002] Script to create/get and delete universe

jaydeepkumara:
Summary: Provide a script that creates a universe using these three command line arguments: Platform URL, API token, and path to the file containing the Universe config in JSON format.

We need scripts for the following actions:

Dump json payload for a given universe
Create a universe
Destroy an existing universe

The shell script is written in such a way that clients don't require any external library.
Tech:- Shell Script, vanilla python.

Test Plan: <need to add>


managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,415 @@
...
+    ADD_UNIVERSE|CREATE_UNIVERSE)
+        if [[ -z "$input_file" ]]
+        then
+            printf "Input json file required. Use \`-i|--input <file_path>\` to pass json input file\n"
+            exit 255
+        fi
+
+        if [[ $python_3 == "True" ]]
+        then
+            universe_reponse=$(python3 -c "import yb_platform_util_py3; yb_platform_util_py3.create_universe('$base_url', '$customer_uuid', '$auth_uuid', '$input_file', '$universe_name')")
```
mahendranbhat, line 281: check the maximum character limit in each line and format the file.

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,415 @@
...
+            task_id=$(python -c "import yb_platform_util_py2; yb_platform_util_py2.get_key_value('$new_universe_reponse', 'data')")
+        fi
+        if [[ ! $universe_status == "success" ]]
+        then
+            echo $task_id
+            exit
+        fi
+        if [[ ! -z "$skip_wait" ]]
+        then
+            echo "Universe create requested successfull. Use $task_id as task id to get status of universe."
```
mahendranbhat, line 298: `successfully` or  `Universe create request submitted successfully.`

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,443 @@
...
+            else:
+                response = {"data": content, "status": "failed", "error": content}
+                print(response)
+        except Exception as e:
+            response = {"data": str(e), "status": "failed", "error": str(e)}
+            print(response)
+    return inner_function
+
+
+def convert_unicode_json(data):
```
mahendranbhat, line 36: add docstrings

jaydeepkumara: done

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,525 @@
...
+    printf "\n\t export YB_PLATFORM_API_TOKEN=e16d75cf-79af-4c57-8659-2f8c34223551\n"
+    exit 255
+fi
+
+base_url=$(echo "$YB_PLATFORM_URL" | sed 's/\/$//')
+auth_uuid="$YB_PLATFORM_API_TOKEN"
+
+if [[ -z "$customer_uuid" ]]
+then
+    if [[ $python_3 == "True" ]]
```
SergeyPotachev, line 268: See one of my previous comments - if we use two variables instead of one `python_3` then it will be simpler.

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,525 @@
...
+# Operation can be add/create/get/delete/del all are canse insensitive.
+# Add/create will call python api to create universe
+# Get will fetch universe details using name/universe UUID.
+# Delete Will delete universe by name/universe UUID
+case $operation in
+    ADD_UNIVERSE|CREATE_UNIVERSE)
+        if [[ -z "$input_file" ]]
+        then
+            printf "Input json file required. Use \`-i|--input <file_path>\` to "
+            printf "pass json input file\n"
```
SergeyPotachev, line 312: `.` at the end, please.

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,525 @@
...
+                    TaskStatus ${task_id} ${customer_uuid} "Universe Deletion successfully."
+                fi
+                ;;
+            * ) echo "Aborted Delete universe";;
+        esac
+        ;;
+    TASK_STATUS)
+        if [[ -z "$task_id" ]]
+        then
+            echo "Task id required to get task status"
```
SergeyPotachev, line 483: `.`

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+        return dict(map(convert_unicode_json, data.iteritems()))
+    elif isinstance(data, collections.Iterable):
+        return type(data)(map(convert_unicode_json, data))
+    else:
+        return data
+
+
+def call_api(url, auth_uuid, data=None, is_delete=False):
+    """
+    Call the corresponding url with auth token, headers and returns the response.
```
SergeyPotachev, line 53: Calls .. and returns...
or
Call ... and return ...

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+
+    :param url: url to be called.
+    :param auth_uuid: authentication token of the customer.
+    :param data: data for POST request.
+    :param is_delete: to identify the delete call.
+    :return: response of the API call.
+    """
+    if not is_delete:
+        request = urllib2.Request(url)
```
SergeyPotachev, line 62: Why not to put this line before the condition.

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+        cluster = {}
+        for key, val in each_cluster.items():
+            if key not in excluded_keys:
+                user_intent = {}
+                if key == "userIntent":
+                    if universe_name:
+                        val["universeName"] = universe_name
+                    for key1, val1 in val.items():
+                        if key1 not in excluded_keys:
+                            if key1 == "deviceInfo":
```
SergeyPotachev, line 273: Looks slightly strange for me. Please add comments what and why are your doing here.

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+    return data
+
+
+@exception_handling
+def post_universe_config(configure_json, base_url, customer_uuid, auth_uuid):
+    """
+    Call the universe config URL with the updated data.
+
+    :param configure_json: Universe config json.
+    :param base_url: base url of the platform back end.
```
SergeyPotachev, line 311: Please use the same style -
Universe ...
**B**ase ...
UUID ...
**A**uthentication ...

SergeyPotachev: Please update this here and in other functions too.

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+    :return: None
+    """
+    customer_url = base_url + "/api/v1/customers"
+    response = call_api(customer_url, auth_uuid)
+    data = convert_unicode_json(json.load(response))
+    if (len(data) == 1):
+        response = {"data": data[0], "status": "success", "error": ""}
+        print(response)
+    else:
+        response = {"data": data, "status": "failed", "error": "Please provide customer UUID"}
```
SergeyPotachev, line 394: Very strange. Function `get_customer_uuid` returns error `Please provide customer UUID` ???

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+    :return: None
+    """
+    universe = get_universe_by_uuid(base_url, customer_uuid, auth_uuid, universe_uuid)
+    if universe:
+        configure_json = create_universe_config(universe, universe["name"])
+        file = base_dir + '/' + universe["name"] + ".json"
+        with open(file, 'w') as file_obj:
+            json.dump(configure_json, file_obj)
+        response = {
+            "data": "Detail of universe has been saved to " + str(file),
```
SergeyPotachev, line 126: Details ... have been saved

SergeyPotachev: Also from the function name it is not clear that it saves something to a file / returns data through a file.

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+    :return: None
+    """
+    universe = get_universe_by_name(base_url, customer_uuid, auth_uuid, universe_name)
+    if universe:
+        configure_json = create_universe_config(universe, universe["name"])
+        file = base_dir + '/' + universe["name"] + ".json"
+        with open(file, 'w') as file_obj:
+            json.dump(configure_json, file_obj)
+        response = {
+            "data": "Detail of universe has been saved to " + str(file),
```
SergeyPotachev, line 94: Details ... have been saved.
Also from the function name it is not clear that it saves something to a file / returns data through a file.

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+        universe_config_json,
+        base_url,
+        customer_uuid,
+        auth_uuid
+    )
+    if universe_config_json:
+        create_universe_from_config(universe_config_json, base_url, customer_uuid, auth_uuid)
+    else:
+        data = {
+            "data": 'Unable to Configure Universe for Customer %s' % customer_uuid,
```
SergeyPotachev, line 185: Unable to Create Universe ?

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+    :param universe_uuid: UUID of the universe.
+    :return: None
+    """
+    universe_config_url = base_url + "/api/v1/customers/" + customer_uuid + \
+                          "/universes/" + universe_uuid
+    response = call_api(universe_config_url, auth_uuid)
+    return convert_unicode_json(json.load(response))
+
+
+def get_universe_uuid(base_url, customer_uuid, auth_uuid, universe_name):
```
SergeyPotachev, line 358: Maybe `get_universe_uuid_by_name`.

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+    response = call_api(universe_delete_url, auth_uuid, is_delete=True)
+    universe_json = convert_unicode_json(json.loads(response.read()))
+    task_id = universe_json['taskUUID']
+    response = {"data": task_id, "status": "success", "error": ""}
+    print(response)
+
+
+@exception_handling
+def get_provider_data(base_url, customer_uuid, auth_uuid):
+    provider_url = base_url + "/api/v1/customers/" + customer_uuid + "/providers"
```
SergeyPotachev, line 421: Missed docstrings here and in the next three functions.

managed/devops/bin/yb_platform_util_py2.py
```diff
@@ -0,0 +1,498 @@
...
+    Delete the universe by providing UUID.
+
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: UUID of the customer.
+    :param auth_uuid: authentication token for the customer.
+    :param universe_uuid: UUID of the universe to be deleted.
+    :return:
+    """
+    universe_delete_url = base_url + "/api/v1/customers/" + customer_uuid + "/universes/" \
+                          + universe_uuid + "?isForceDelete=true"
```
SergeyPotachev, line 410: Are we intentionally doing Forced delete? I don't like this. Add one more key to the script which forces the universe deletion. Without the key don't use forced delete.

SergeyPotachev (changes requested):
I put some comments for the py2 script but some of them (or maybe all of them) are applicable to the py3 script as well.
Generally looks sloppy...

skorobogatydmitry:
IMO there is no need for bash and Python 2. But it depends on script purpose.
@streddy-yb, do we really need to have bash and Py 2?

svarnau (commented):
Just a few style comments in shell script.

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,525 @@
...
+    printf "\t Example: \n\t\tbash yb_platform_util.sh delete_universe"
+    printf " -n test-universe-by-name -y"
+    printf "\n\n--no-wait \n\t To run command in background and do not wait for task completion task\n"
+    printf "\t Example: \n\t\tbash yb_platform_util.sh create_universe"
+    printf " -f test-universe.json --no-wait"
+    printf "\n\n-t | --task \n\t Task UUID to get task status\n"
+    printf "\t Example: \n\t\tbash yb_platform_util.sh task_status"
+    printf " -t f33e3c9b-75ab-4c30-80ad-cba85646ea39"
+    printf "\n\n-h | --help \n\t Print help message\n\t Example: "
+    printf "\n\t\t bash yb_platform_util.sh -h\n\n"
```
jharveysmith, line 146: heredocs are super useful
https://linuxize.com/post/bash-heredoc/

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+        "awsArnString",
+        "useHostname",
+        "preferredRegion",
+        "regions",
+        "index",
+        "placementInfo"
+    ]
+
+    clusters_list = get_cluster_list(clusters, excluded_keys, universe_name)
+    configure_json["clusters"] = clusters_list
```
skorobogatydmitry, line 231: Don't we need deep copy here ?

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    configure_json["clusterOperation"] = "CREATE"
+    configure_json["userAZSelected"] = user_az_selected
```
skorobogatydmitry, line 233: Ditto deep copy

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,525 @@
...
+            ;;
+    esac
+done
+
+
+# Checking if system has python3 installed.
+# If python3 installed run script using python 3 use 2 otherwise.
+
+python_3="False"
+if [[ "$(python3 -V)" =~ "Python 3" ]]
```
jharveysmith, line 234: Is there really a concern that `python3` would _not_ by python 3?
Wouldn't just a check for `python3` in the $PATH suffice?

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+            if key not in excluded_keys:
+                user_intent = {}
+                if key == "userIntent":
+                    if universe_name:
+                        val["universeName"] = universe_name
+                    for key1, val1 in val.items():
+                        if key1 not in excluded_keys:
+                            if key1 == "deviceInfo":
+                                val1.pop("diskIops")
+                                val1.pop("storageType")
```
skorobogatydmitry, line 261: These methods modifies input dict, please use get instead

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    Get universe details by UUID of the universe.
+
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: UUID of the customer.
+    :param auth_uuid: authentication token for the customer.
+    :param universe_uuid: UUID of the universe.
+    :return: None
+    """
+    universe_config_url = base_url + "/api/v1/customers/" + customer_uuid + \
+                          "/universes/" + universe_uuid
```
skorobogatydmitry, line 338: Please use format string

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+                                val1.pop("storageType")
+                            user_intent[key1] = val1
+                    cluster[key] = user_intent
+                else:
+                    cluster[key] = val
+        clusters_list.append(cluster)
+    return clusters_list
+
+
+def modify_universe_config(file_name, universe_name=""):
```
skorobogatydmitry, line 270: Please rename to `set_universe_name`.

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    """
+    Create the universe from universe config data by calling universe POST API.
+
+    :param universe_config: universe config data.
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: UUID of the customer.
+    :param auth_uuid: authentication token for the customer.
+    :return: None
+    """
+    universe_create_url = base_url + "/api/v1/customers/" + customer_uuid + "/universes"
```
skorobogatydmitry, line 135: Ditto format strings

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: UUID of the customer.
+    :param auth_uuid: authentication token for the customer.
+    :param input_file: Directory of the stored universe json data.
+    :param universe_name: Name of the new universe to be created.
+    :return: None
+    """
+    configure_json = modify_universe_config(input_file, universe_name)
+    universe_config_json = post_universe_config(configure_json, base_url, customer_uuid, auth_uuid)
+    # add the missing fields in the config json.
```
skorobogatydmitry, line 156: So it's just config generating ?

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    """
+    Get details of the ongoing task.
+
+    :param task_id: Task UUID.
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: authentication token for the customer.
+    :param auth_uuid: authentication token for the customer.
+    :return: None
+    """
+    task_url = base_url + "/api/v1/customers/" + customer_uuid + "/tasks/" + task_id
```
skorobogatydmitry, line 188: Ditto format string

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    response = call_api(task_url, auth_uuid)
+    universe_json = json.loads(response.read())
```
skorobogatydmitry, line 190: It's task JSON, not for universe ?

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    Get the UUID of the universe.
+
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: UUID of the customer.
+    :param auth_uuid:  authentication token for the customer.
+    :param universe_name: Name of the universe.
+    :return: None.
+    """
+    universe = get_universe_by_name(base_url, customer_uuid, auth_uuid, universe_name)
+    if universe:
```
skorobogatydmitry, line 354: `if universe and 'universeUUID' in universe`

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    """
+    Delete the universe by providing UUID.
+
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: UUID of the customer.
+    :param auth_uuid: authentication token for the customer.
+    :param universe_uuid: UUID of the universe to be deleted.
+    :return:
+    """
+    universe_delete_url = base_url + "/api/v1/customers/" + customer_uuid + "/universes/" \
```
skorobogatydmitry, line 394: Ditto format string

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    response = call_api(universe_delete_url, auth_uuid, is_delete=True)
+    universe_json = json.loads(response.read())
+    task_id = universe_json['taskUUID']
+    response = {"data": task_id, "status": "success", "error": ""}
+    print(response)
+
+
+@exception_handling
+def get_provider_data(base_url, customer_uuid, auth_uuid):
+    provider_url = base_url + "/api/v1/customers/" + customer_uuid + "/providers"
```
skorobogatydmitry, line 406: Ditto format strings

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+        provider = {}
+        provider['name'] = each["name"]
+        provider['uuid'] = each["uuid"]
+        providers.append(provider)
+    print(json.dumps(providers))
+
+
+@exception_handling
+def get_regions_data(base_url, customer_uuid, auth_uuid):
+    provider_url = base_url + "/api/v1/customers/" + customer_uuid + "/regions"
```
skorobogatydmitry, line 420: Ditto format strings

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+        region['uuid'] = each["uuid"]
+        region['provider'] = each["provider"]["code"]
+        regions.append(region)
+    print(json.dumps(regions))
+
+
+@exception_handling
+def get_universe_list(base_url, customer_uuid, auth_uuid):
+    universe_url = base_url + "/api/v1/customers/" + customer_uuid + \
+                          "/universes"
```
skorobogatydmitry, line 443: Ditto format string

managed/devops/bin/yb_platform_util_py2.py
jaydeepkumara: yes. For the initial version, we thought to do force delete. If you having concern let me discuss with Arnav again.

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,525 @@
...
+    then
+        python_3="True"
+    fi
+
+    while :
+    do
+
+        if [[ $python_3 == "True" ]]
+        then
+            percentage_response=$(python3 -c "import yb_platform_util_py3; \
```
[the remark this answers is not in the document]
jaydeepkumara: Good idea,
Let me check, while fixing those comments I feel it would be better if I can combine python2 and python3 file. i am checking that. That will fix this issue probably.

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,525 @@
...
+    printf "\t Example: \n\t\tbash yb_platform_util.sh create_universe -f test-universe.json"
+    printf " -n test-universe-by-name"
+    printf "\n\ndel_universe |  delete_universe\n\t Delete an existing Universe\n"
+    printf "\t Example: \n\t\tbash yb_platform_util.sh delete_universe -n test-universe\n\n"
+    printf "task_status \n\t To get task status\n"
+    printf "\t Example: \n\t\tbash yb_platform_util.sh task_status"
+    printf " -t f33e3c9b-75ab-4c30-80ad-cba85646ea39"
+    printf "\n\nget_provider \n\t To get list of available providers\n"
+    printf "\t Example: \n\t\tbash yb_platform_util.sh get_provider"
+    printf "\n\nget_region | get_az \n\t list of available region with availability zones.\n"
```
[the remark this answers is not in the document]
jaydeepkumara: yeah, agree. I tried that. 

-f | --file 
	 Json Input/output file for creating universe
	 Example: 
		bash yb_platform_util.sh create_universe -f test-universe.json -n test-universe-by-name

managed/devops/bin/yb_platform_util.sh
jaydeepkumara: Change code a little bit. It might be more clear now. I am trying to give priority to python3. The person with having an old laptop will have default python2 and python3 points to python3 that the reason why checking python3. In case Default python is python3 then Python script is taking care of.

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
+# Copyright (c) YugaByte, Inc.
```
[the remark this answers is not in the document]
jaydeepkumara: It is an executable file. 
-rwxr-xr-x  1 jaahir  staff   19483 Apr 22 19:51 yb_platform_util.py
-rwxr-xr-x  1 jaahir  staff   15799 Apr 22 19:51 yb_platform_util.sh

managed/devops/bin/yb_platform_util_py3.py
jaydeepkumara: Here, We are first creating configuration then and then calling create universe API.

managed/devops/bin/yb_platform_util_py3.py
jaydeepkumara: use case,
Customers will call one API which will save universe JSON to the file and then they will use that JSON file to create a new universe. 

Now the problem here is When we call to get a universe that has different formates and has a lot of extra data which not required for create universe so we are removing that extra feild before saving to the file.

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+        content = {
+            "message": "{0} failed".format(universe_json["title"]),
+            "details": universe_json["details"]
+        }
+        response = {"data": content, "status": "success",
+                    "error": content}
+        print(response)
+
+
+def create_universe_config(universe_data, universe_name):
```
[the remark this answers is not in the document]
jaydeepkumara: Basically, we have one option in the script where the user can pass the custom name of the universe and it will overwrite the universe.  universe_name is the name required to overwrite in current universe data.

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+
+def get_universe_details(base_url, customer_uuid, auth_uuid, universe_name, base_dir):
+    """
+    Get the universe details and store it in a json file after formatting the json.
+
+    :param base_url: base url of the platform back end.
+    :param customer_uuid: UUID of the customer.
+    :param auth_uuid: authentication token for the customer.
+    :param universe_name: Name of the universe to take a json data from.
+    :param base_dir: base directory in which the json file should be stored.
```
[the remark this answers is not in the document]
jaydeepkumara: File name required to be same as universe name. universe name comes as a part of API response. So here we are just asking a base path and then appends universe name(base_dir + '/' + universe["name"] + ".json")

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+                print(response)
+            else:
+                response = {"data": content, "status": "failed", "error": content}
+                print(response)
+        except Exception as e:
+            response = {"data": str(e), "status": "failed", "error": str(e)}
+            print(response)
+    return inner_function
+
+def call_api(url, auth_uuid, data=None, is_delete=False):
```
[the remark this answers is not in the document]
jaydeepkumara: No, it for the delete method.
urllib.request.Request(url, method="DELETE") (line 50)

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+# def exception_handling():
+
+
+def exception_handling(func):
+    def inner_function(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except HTTPError as e:
+            content = e.read().decode("utf-8")
+            print (content)
```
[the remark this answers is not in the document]
jaydeepkumara: This script is a user-friendly script. Print statement return data to script and later script print that response to the Terminal. Because of that reason we had used print.

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+            if "html>" in content:
+                message = "Invalid YB_PLATFORM_URL URL or params, Getting html page in response"
+                response = {"data": message, "status": "failed", "error": message}
```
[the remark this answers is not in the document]
jaydeepkumara: This script is a user-friendly script. Print statement return data to script and later script print that response to the Terminal. Because of that reason we had used print.

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+                print(response)
```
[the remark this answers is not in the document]
jaydeepkumara: This script is a user-friendly script. Print statement return data to script and later script print that response/error to the Terminal. Because of that reason we had used print.

managed/devops/bin/yb_platform_util_py3.py
```diff
@@ -0,0 +1,483 @@
...
+    :param auth_uuid: authentication token for the customer.
+    :param universe_name: Universe name.
+    :return: None
+    """
+    universe_url = base_url + "/api/v1/customers/" + customer_uuid + "/universes"
+    response = call_api(universe_url, auth_uuid)
+    data = json.load(response)
+    for universe in data:
+        if universe["name"] == universe_name:
+            del universe['pricePerHour']
```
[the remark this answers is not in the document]
jaydeepkumara: use case,
Customers will call one API which will save universe JSON to the file and then they will use that JSON file to create a new universe.

Now the problem here is When we call to get a universe that has different formates and has a lot of extra data which not required for create universe so we are removing that extra feild before saving to the file.

managed/devops/bin/yb_platform_util_py3.py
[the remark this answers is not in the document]
skorobogatydmitry: I see, but Python file can't be used as executable (like `./bin/yb_platform_util.py`. Any reasons keep it in `bin/` directory?
Or, if `.py` file is intended to be executable, please add shebang to it. In this case `.sh` file's purpose is unclear.

managed/devops/bin/yb_platform_util_py3.py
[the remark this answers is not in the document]
skorobogatydmitry: Sure, but stderr will be visible to end-user as well, but this allows to filster it out for automation then.

managed/devops/bin/yb_platform_util_py3.py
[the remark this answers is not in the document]
skorobogatydmitry: I meant that you need to add exception's traceback to the "error" field.

managed/devops/bin/yb_platform_util_py3.py
mahendranbhat: @skorobogatydmitry we could use that but we are calling this function here only and only once in the program. also, we are not dealing with the creation of multiple universes in a single shot I believe. Until we use it for creation of single universe it would be fine I guess. But yes we could use deep copy to be on a safer side. Correct me If I am wrong.

managed/devops/bin/yb_platform_util_py2.py
jaydeepkumara: Basically, functionality is to return customers in case only one customer UUID present. If there is multiple UUID then user must be required to pass at least one UUID.

managed/devops/bin/yb_platform_util.sh
```diff
@@ -0,0 +1,467 @@
...
+        echo $result
+        echo
+      ;;
+    *)
+        echo "Invalid opertaion"
+        HelpMessage
+        exit 255
+        ;;
+
+esac
```
mbautin, line 467: Why is there so much new logic written in Bash? Please use Python instead.

mbautin (changes requested):

jaydeepkumara:
Similar PR:- https://github.com/yugabyte/yugabyte-db/pull/8523

PR #7322, opened 2021-02-19
[Docs] remove folders that are now redirected

polarweasel:
We are now redirecting all versions' comparisons, faq, legal, and releases folders to latest.

This PR removes all the files that are no longer needed to build the site.


docs/config.yaml
```diff
@@ -8,7 +8,7 @@ googleAnalytics: UA-104956980-3
 pygmentsCodefences: true
 pygmentsStyle: "pygments"
 enableGitInfo: true
-
+resourceDir: docs/resources
```
yb-andrew, line 11: Is this change needed? I'm assuming it is just to let Hugo know where to look for static assets? I think we use static url to get file locations.

polarweasel:
> Is this change needed? I'm assuming it is just to let Hugo know where to look for static assets? I think we use static url to get file locations. 

I added it to help the Hugo resource-caching plugin find the directory. As it turns out the plugin still isn't caching anything, but I think that's because we're not actually post-processing any images with Hugo. Not harmful, just a no-op at this point.

yb-andrew (approved):
LG

PR #7368, opened 2021-02-24
[doc] Update yugabyte-platform.md

lizayugabyte:


lizayugabyte:
Per Chirag, deleted  A REST API is also exposed by the YugabyteDB Admin Console to the end users in addition to the UI as another means of interacting with the data platform.
from https://docs.yugabyte.com/latest/faq/yugabyte-platform/#how-does-the-yugabytedb-admin-console-interact-with-the-yugabytedb-data-nodes

PR #7476, opened 2021-03-04
[#7448 ][docs] Additional step for macOS Catalina to ensure "npm ci" does not fail

### LINKED ISSUE #7448: [Docs] `npm ci` step for YB Documentation fails and needs additional step for MacOS Catalina
Issue: `npm ci` step for YB Documentation fails and needs additional step for MacOS Catalina

When initially setting up Yugabyte Documentation locally on macOS Catalina, `npm ci` fails even after installing xcode and performing `xcode-select --install` with the same error 
```
No receipt for 'com.apple.pkg.DeveloperToolsCLI'
```
as mentioned in the documentation

**Note**: I have found the fix and I would like to include the additional detail in existing documentation.

aishwarya24:
resolves #7448 
This PR includes a command to ensure "xcode-select" is pointing to the right directory for Catalina.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -74,6 +74,13 @@ Follow these steps if this is the first time you are setting up to work on the d
 >
 > Only now do `git` commands start to work normally. Now, too (but not before) `npm ci` works properly.
 
+Note:
+For Catalina, an additional step maybe necessary to make sure xcode-select is pointing to the Xcode Developer directory using a command like this:
+
+```
+xcode-select -s /Applications/Xcode.app/Contents/Developer
+```
+
 Then:
 
 Fork the yugabyte-db GitHub repository and create a local clone of your fork with a command like this:
--- a/docs/README.md
+++ b/docs/README.md
@@ -75,10 +75,10 @@ Follow these steps if this is the first time you are setting up to work on the d
 > Only now do `git` commands start to work normally. Now, too (but not before) `npm ci` works properly.
 
 Note:
-For Catalina, an additional step maybe necessary to make sure xcode-select is pointing to the Xcode Developer directory using a command like this:
+In some cases, an additional step may be necessary to make sure `xcode-select` is pointing to the correct Xcode or Command-Line Tools installation. For example:
 
-```
-xcode-select -s /Applications/Xcode.app/Contents/Developer
+```sh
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -74,6 +74,13 @@ Follow these steps if this is the first time you are setting up to work on the d
 >
 > Only now do `git` commands start to work normally. Now, too (but not before) `npm ci` works properly.
 
+Note:
+For Catalina, an additional step maybe necessary to make sure xcode-select is pointing to the Xcode Developer directory using a command like this:
```
polarweasel, line 78: Probably don't need to mention Catalina here.

```suggestion
In some cases, an additional step may be necessary to make sure `xcode-select` is pointing to the correct Xcode or Command-Line Tools installation. For example:
```

docs/README.md
```diff
@@ -74,6 +74,13 @@ Follow these steps if this is the first time you are setting up to work on the d
...
+
+```
```
polarweasel, line 80:
```suggestion
```sh
```

docs/README.md
```diff
@@ -74,6 +74,13 @@ Follow these steps if this is the first time you are setting up to work on the d
...
+xcode-select -s /Applications/Xcode.app/Contents/Developer
```
polarweasel, line 81: Per its manpage, `xcode-select --switch` should point just to the version of Xcode you want it to use, and should be run with `sudo`:

> Sets the active developer directory to the given path, for example **/Applications/Xcode-beta.app**. This command must be run with superuser permissions [...]

```suggestion
sudo xcode-select --switch /Applications/Xcode.app
```

polarweasel: And if someone installed the command-line tools standalone, they live in /Library/Developer/CommandLineTools...

I think we need to revisit this whole section with a fresh macOS install, and see what happens. You don't happen to have a fresh machine, do you? I don't, right now. :)

polarweasel (changes requested):

docs/README.md
aishwarya24: I do not have a fresh machine at this point as well. Will review the other changes you suggested.

polarweasel (approved):
LGTM. More work needed in this doc, but this will address certain failures, so it's good to have!

Issue #6945, opened 2021-01-21
[Platform] [2.4] Limit tasks queried in API call by default to 2000


iSignal:
We have proper task pagination and GC in 2.6. For 2.4, however, we could make the tasks UI more functional by adding a `LIMIT 2000` to our tasks db query at https://github.com/yugabyte/yugabyte-db/blob/e98203c9d4cd8180fb3c173c3f43cb78548979f8/managed/src/main/java/com/yugabyte/yw/controllers/CustomerTaskController.java#L69 so we don't end up with a spinning task UI for Platform instances that accumulate many tasks over time.

mahendranbhat:
@VijiYB Please LMK once the testing is done, then will move to 2.6 and 2.4

RP1807:
@mahendranbhat Validated it with the latest master as well as on Portal. Did sanity testing around task API and checked universe task tab and global task page. Working as expected.

mahendranbhat:
2.6 PR https://github.com/yugabyte/yugabyte-db/pull/8624

mahendranbhat:
2.6 backport is done, facing some issues in 2.4 backport. I will update the same once 2.4 backport is done.

mahendranbhat:
2.4 backport is also done.

PR #7477, opened 2021-03-04
[#6945] [Platform] Add limit to customer task API.

mahendranbhat:
**Description**:
We could make the tasks UI more functional by adding a LIMIT 2000 to our tasks DB query, so we don't end up with a spinning task UI for Platform instances that accumulate many tasks over time.
Currently it is set to 2000 in reference.conf.

**Test Plan**:

- Updated reference.conf to 20.
- Verified whether the API returning only 20 entries or not.
-  Also verified the working endpoint of a particular task.
- Verified by creating 10 + tasks and verified pagination in UI too.

SBT TEST:

```	... 18 common frames omitted
[info] Passed: Total 1314, Failed 0, Errors 0, Passed 1310, Skipped 4
[success] Total time: 1026 s (17:06), completed 19 May, 2021 12:55:23 AM
mahendra@mahendra-HP-ZBook-14u-G5:~/yugabyte/yugabyte-db/managed$ 
```


Arnav15 (changes requested):
The change looks good, but this might not actually solve the problem. From what I understand, this doesn't change the actual data that is fetched from the DB, but only parses the first 2000. What that means is we still will have a slow response to the API call when we have a lot of tasks. @daniel-yb can you confirm if my understanding of the ebean setMaxCount is correct? 
Also, I think we can create a mock test for testing this by generating 3000 random entries in the test DB and checking the response only has 2000.

daniel-yb:
> The change looks good, but this might not actually solve the problem. From what I understand, this doesn't change the actual data that is fetched from the DB, but only parses the first 2000. What that means is we still will have a slow response to the API call when we have a lot of tasks. @daniel-yb can you confirm if my understanding of the ebean setMaxCount is correct?
> Also, I think we can create a mock test for testing this by generating 3000 random entries in the test DB and checking the response only has 2000.

Yeah setMaxRows doesn't add a LIMIT clause to the queries AFAIK. It actually loads all results into memory and then only returns x rows. It would probably be better to change this from building the query using ebean methods to using a raw SQL query that includes a LIMIT clause.

yb-andrew:
Yes, the network transfer from API -> UI is the bottleneck. With a lot of tasks the amount of data is large enough to be very noticeable.
@mahendranbhat I'm also refactoring the tasks API so you may want to hold off on the changes.

managed/src/main/java/com/yugabyte/yw/controllers/CustomerTaskController.java
```diff
@@ -61,16 +65,25 @@
...
   private Map<UUID, List<CustomerTaskFormData>> fetchTasks(UUID customerUUID, UUID targetUUID) {
-    Query<CustomerTask> customerTaskQuery = CustomerTask.find.query().where()
-      .eq("customer_uuid", customerUUID)
-      .orderBy("create_time desc");
-
+    Query<CustomerTask> customerTaskQuery;
+    Set<CustomerTask> pendingTasks;
     if (targetUUID != null) {
-      customerTaskQuery.where().eq("target_uuid", targetUUID);
+      String selectQuery = "SELECT * FROM customer_task " +
```
SergeyPotachev, line 71: 1. We can do this without usage of raw sql queries (+ .setMaxRows to finder parameters).
2. Limit should be configurable throw RuntimeConfig (and the initial setting in configuration file).
3. And anyway SQL string preparation looks very ugly.

SergeyPotachev (changes requested):

managed/src/main/java/com/yugabyte/yw/controllers/CustomerTaskController.java
SergeyPotachev: Also step 2 (configurable limit) allows to test this functionality without significant efforts. (Test plan should be updated)

mahendranbhat: @SergeyPotachev https://github.com/yugabyte/yugabyte-db/pull/7477#issuecomment-792540987 The original change was to use setMaxRows but after this suggestion I have used SQL queries.

sb-yb:
```
  List<CustomerTask> pendingTasks = customerTaskQuery
      .setFirstRow(0)
      .setMaxRows(2000)
      .findPagedList()
      .getList();
```
Should do what you want. 

It should add limit clause.

sb-yb (changes requested):
Also make "2000" configurable.

sb-yb:
Also note that query performance will depend on if there is index on 
`create_time` 
So please check and add one if does not exist if the motivation is just to avoid memory hit on yugaware due to fetching all the rows then just limit is sufficient.

managed/src/test/java/com/yugabyte/yw/controllers/CustomerTaskControllerTest.java
```diff
@@ -265,6 +266,23 @@ public void testTaskHistoryUniverseList() {
...
+  @Test
+  public void testTaskHistoryLimit2000() {
+    String authToken = user.createAuthToken();
+    Universe universe1 = createUniverse("Universe 2", customer.getCustomerId());
+    IntStream.range(0, 3000).forEach(i -> createTaskWithStatus(
+        universe.universeUUID, CustomerTask.TargetType.Universe, Create, "Foo", "Running", 50.0));
+    Result result = FakeApiHelper.doRequestWithAuthToken("GET", "/api/customers/" +
+        customer.uuid +  "/universes/" + universe.universeUUID + "/tasks", authToken);
+    assertEquals(OK, result.status());
+    JsonNode json = Json.parse(contentAsString(result));
```
sb-yb, line 278: nit: Can you parse this to a Map with Json.toJson()?
Same with then parsing values to List.

sb-yb (approved):

SergeyPotachev (approved):

managed/src/main/resources/reference.conf
```diff
@@ -88,6 +88,8 @@ yb {
     default_smtp_port_ssl = 465
     debug_email = false
   }
+
+  max_tasks = 2000
```
sb-yb, line 92: You also need to whitelist it in:
```
runtime_config {
  included_paths = [ 
   // here
  ] 
  ...
}

 ```

Thats where runtime config controller looks to determine if a config key can be used as a runtime configuration.

sb-yb (changes requested):

sb-yb (approved):

sb-yb (changes requested):
Better name for config param please.

sb-yb:
I propose
yb.max_task -> yb.customer_task_db_query_limit

But something better than max_tasks which gives very little info.

sb-yb (approved):
Thanks for making the changes!

PR #7483, opened 2021-03-04
[doc] PR for Adding PostgreSQL Range Types tstzrange, tsrange Doc7139

lizayugabyte:


docs/content/latest/explore/ysql-language-features/data-types.md
```diff
@@ -377,3 +377,54 @@ yugabyte=# SELECT (item).name FROM on_hand WHERE (item).price > 0.99;
...
+
+You can create a range type using a constructor function named identically to the range type. The constructor functions typically have two or three arguments, with the former constructing a range in standard form (lower bound inclusive, upper bound exclusive), and the latter constructing a range with bounds specified by the third argument (one of strings "`()`", "`(]`", "`[)`", or "`[]`"). The following example shows the constructor with a lower bound, upper bound, and text argument:
+
+```sql
+SELECT tsrange(2020-01-01 8:30, 2020-02-02 5:30 '(]');
+```
+
+YSQL allows you to impose exclusion constraints, such as non-overlapping constraints, on range types. The following example creates a table containing data on the employee vacation ranges and prevents overlapping dates from being entered in the table at the same time:
```
m-iancu, line 422: we don't support exclusion constraints yet.

docs/content/latest/explore/ysql-language-features/data-types.md
```diff
@@ -377,3 +377,54 @@ yugabyte=# SELECT (item).name FROM on_hand WHERE (item).price > 0.99;
...
 ```
 
+## Range Types
+
+Range data types represent a range of values of an element type. Range types are usually referred to as the subtype of the range. The subtype needs to follow a strict order because it must be well-defined regardless of the position of element values which can be within, before, or after a value range.
+
+YSQL supports the following range types:
+
+- `tsrange`, which corresponds to range of timestamp without time zone.
+- `tstzrange`, which corresponds to range of timestamp with time zone.
```
m-iancu, line 387: I think we support all of:

```
int4range — Range of integer

int8range — Range of bigint

numrange — Range of numeric

tsrange — Range of timestamp without time zone

tstzrange — Range of timestamp with time zone

daterange — Range of date
```
The are tested in this file: https://github.com/yugabyte/yugabyte-db/blob/master/src/postgres/src/test/regress/expected/yb_pg_rangetypes.out

Note that we also support user-defined range types as tested there at the top with:
```
create type textrange as range (subtype=text, collation="C");
```
This is covered in pg docs in the `8.17.8. Defining New Range Types` section: (https://www.postgresql.org/docs/11/rangetypes.html)

Note: As those are more advanced it's ok to omit them and focus either only on tsrange types or only on primitive types here. But I suggest to quickly note them so that users don't get the impression that's _all_ we support.

docs/content/latest/explore/ysql-language-features/data-types.md
lizayugabyte: Deleted

docs/content/latest/explore/ysql-language-features/data-types.md
lizayugabyte: Applied these changes

docs/content/latest/explore/ysql-language-features/data-types.md
```diff
@@ -377,3 +377,59 @@ yugabyte=# SELECT (item).name FROM on_hand WHERE (item).price > 0.99;
...
+```
+
+In addition to using built-in range types, you can define your own custom ones. The following example shows how to define a range type of subtype `text` and use it in a `SELECT` statement:
+
+```sql
+CREATE TYPE textrange 
+  AS RANGE (subtype = text);
+
+SELECT '( " a " " a ", " z " " z " )'::textrange;
+```
```
m-iancu, line 434: Looks like we have already a reference page on range types thanks to @bllewell: https://docs.yugabyte.com/latest/api/ysql/datatypes/type_range/

Does it makes sense to add a note somewhere linking to it along the lines of:
```
For more details on range types see ...
```

lizayugabyte: Yes, Bryn sent me that link last night. I'm definitely including it in the doc  (this was the plan all along)

PR #7498, opened 2021-03-05
[docs] stop redirecting /images

polarweasel:
Yes, it's technically a no-op, butit works to catch the images the `/:version` redirects are breaking!

(Netlify redirects are first-match-wins, so this stops the problem of the broken images on /latest/releases/releases-overview/ as well as others.)

image

### DIFF
--- a/docs/netlify.toml
+++ b/docs/netlify.toml
@@ -22,6 +22,13 @@
 #
 # validator: https://play.netlify.com/redirects
 
+# Make sure we're not redirecting anything under /images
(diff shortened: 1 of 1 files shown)


yb-andrew (approved):
Interesting workaround. The only downside I can think of is it can very slightly affect SEO score with broken links.

polarweasel:
What links are you thinking of that might get broken as a result? This fix is made to address some broken links, not create more. :)

Issue #7472, opened 2021-03-04
[Platform] Non-rolling upgrade should be disabled when the universe is in Pause state


VijiYB:
When the universe is paused, non-rolling upgrade should be disabled. When I try to upgrade with "rolling-upgrade" enabled, getting a warning that the nodes are not live but able to disable the "rolling upgrade" and go ahead with the upgrade. Though it fails as expected, it should not even allow.

nishantSharma459:
![PauseUniverseBug](https://user-images.githubusercontent.com/75416933/109919581-def34180-7cde-11eb-817c-c2ac6dae2fdb.png)
 
Attaching a screenshot. 
PTAL.
@VijiYB  Can you please elaborate on this and add more info?

VijiYB:
In the overview page, you can see the option to upgrade if any new build is available in that por
<img width="562" alt="Screen Shot 2021-03-04 at 8 00 59 AM" src="https://user-images.githubusercontent.com/78048211/109992335-0555b480-7cc0-11eb-9087-4b0cd9eca134.png">
tal.

VijiYB:
<img width="1917" alt="Screen Shot 2021-03-04 at 8 12 16 AM" src="https://user-images.githubusercontent.com/78048211/109993692-69c54380-7cc1-11eb-9dc9-39f851b607a2.png">

On clicking the upgrade on the Overview Info, it opens this pop up to upgrade where by default Rolling-Upgrade is enabled. This actually doesnt allow to continue the upgrade but disable the "Rolling upgrade" its able to proceed and fails.

nishantSharma459:
Here's PR: https://github.com/yugabyte/yugabyte-db/pull/7504

PR #7504, opened 2021-03-05
[#7472] Platform: Hide upgrade action on Pause Universe

nishantSharma459:
When the new upgrade is available then for Pause universe the upgrade button is available which is not supposed to be the case.
If Universe is in a Pause state then no action supposed to be available apart from the Resume and Delete universe.


yb-andrew (approved):
LG.

PR #7517, opened 2021-03-05
Moved query optimization to under Explore top level

lizayugabyte:


lizayugabyte:
@polarweasel  Could you, please, help me with resolving this conflict?

Issue #6889, opened 2021-01-15
Email alerts formatting and missing details


chirag-yb:
There are three different alerts in the platform and all of them with different formatting and there is no consistency. 

- Backup failure alert

```

**Subject** - Yugabyte Platform Alert - <[USERNAME]
**Body** - 
Create Backup failed for aws-test.

Task Info: ybdemo_keyspace:cassandrakeyvalue
```

- Health check failure alert

```

**Subject** - Yugabyte Platform Alert - <[USERNAME]
**Body** -
Common failure for universe 'sp-test':
Health check script got error: Traceback (most recent call last):
 File "bin/cluster_health.py", line 19, in <module>
   error is here
NameError: name 'error' is not defined code (1) [ 636 ms ].
```

- Alert report with various checks for different nodes

 ```

**Subject** - ERROR - <[USERNAME]> tls-test-5

**Body** -

Universe name:tls-test-5 | 2020-12-01 08:54:37
-- | --
Universe version:2.5.1.0-b84

```

Requirements
- All alerts should have the same formatting (body & subject)
- Follow the same guidelines 

1. Include platform IP and hostname details 
2. Include a link to the task page (wherever necessary e.g. - failed backup)

jitendra-12113:
Please review:
https://github.com/yugabyte/yugabyte-db/pull/7580

streddy-yb:
@ymahajan - Can you help elaborate the requirements for this issue? thanks

ymahajan:
Here is the format we should use - 

Subject - ${severity} Alert: ${source} ${condition} 
For eg - 
Critical Alert: Universe puppy-food disk space above 80%

Body [HTML]
{platform_hostname/IP}
{source}/ YB version
Include a link to the task page (wherever necessary e.g. - failed backup)

jitendra-12113:
@ymahajan So the above format should be applicable to all kind of alert right?
1. Backup failure alert ( Include backup task links)
2. Health check Alert
3. Alert report with various checks for different nodes


cc: @SergeyPotachev, @kkg-yb

SergeyPotachev:
@jitendra-12113 
As we are working on alerting mechanic improvements, we need to postpone this issue.
For now we don't have enough information to fill such template (severity, current metric value etc).
cc @streddy-yb @kkg-yb

SergeyPotachev:
cc @ymahajan

PR #7580, opened 2021-03-10
[#6889]Platform: Email alerts formatting and missing details

jitendra-12113:
**Heading:**
[#6889]Platform: Email alerts formatting and missing details
Description:
We need to ensure email alert formatting should be same for all the
alerts. Mail subject and body should be in the same format
Each email should include yugaware hostname and port

Testing:
 1. Check whether each email alerts are in the same format or not
     a. Back up failure email alert
     b. Health check email alert
     c. Alert report with various checks for different nodes

TODO: Unable to test backup email alert - not getting email looking into
this, once testing is done, will Add UTs - This is not final PR

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/commissioner/HealthChecker.java
+++ b/managed/src/main/java/com/yugabyte/yw/commissioner/HealthChecker.java
@@ -211,7 +211,7 @@ public class HealthChecker {
       SmtpData smtpData = emailHelper.getSmtpData(c.uuid);
       if (!StringUtils.isEmpty(emailDestinations) && (smtpData != null)
           && (sendMailAlways || hasErrors)) {
-        String subject = String.format("%s - <%s> %s", hasErrors ? "ERROR" : "OK", c.getTag(),
+        String subject = String.format("%s %s - <%s> %s", "Yugabyte Platform Alert " , hasErrors ? "ERROR" : "OK", c.getTag(),
             u.name);
         String mailError = sendEmailReport(u, c, smtpData, emailDestinations, subject, healthJSON,
             reportOnlyErrors);
--- a/managed/src/main/java/com/yugabyte/yw/common/AlertManager.java
+++ b/managed/src/main/java/com/yugabyte/yw/common/AlertManager.java
@@ -15,6 +15,8 @@ import com.yugabyte.yw.models.*;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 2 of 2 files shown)


managed/src/main/java/com/yugabyte/yw/commissioner/HealthChecker.java
```diff
@@ -211,7 +211,7 @@ private void processResults(Customer c, Universe u, String response, long durati
       SmtpData smtpData = emailHelper.getSmtpData(c.uuid);
       if (!StringUtils.isEmpty(emailDestinations) && (smtpData != null)
           && (sendMailAlways || hasErrors)) {
-        String subject = String.format("%s - <%s> %s", hasErrors ? "ERROR" : "OK", c.getTag(),
+        String subject = String.format("%s %s - <%s> %s", "Yugabyte Platform Alert " , hasErrors ? "ERROR" : "OK", c.getTag(),
```
SergeyPotachev, line 214: - Why "Yugabyte Platform Alert" as a parameter?
- Remove space at the end of "Yugabyte Platform Alert ".
- But overall: I disagree with the new subject. As example, my previous headers for successful health-checks were "OK - <[admin][cloud]> spotachev-test",   in the new format they will be "Yugabyte Platform Alert OK - <[admin][cloud]> spotachev-test". I don't think we need "Yugabyte Platform Alert" here for successful health-checks.

managed/src/main/java/com/yugabyte/yw/common/AlertManager.java
```diff
@@ -87,10 +93,41 @@ public void sendEmail(Alert alert, String state) {
             customer.name, state, alert.message);
       }
     }
+    subject = String.format("Yugabyte Platform Alert ERROR- <%s> %s", customer.getTag(),
+      universe.name);
+    String style = String
+      .format("%s font-size: 14px;background-color:#f7f7f7;padding:25px 30px 5px;", STYLE_FONT);
+    String hostname = "";
+    String ip = "";
+    try {
```
SergeyPotachev, line 102: As we have this code already in several places, let's extract it into com.yugabyte.yw.common.Util (or another utility class if you feel it suits better).

SergeyPotachev (changes requested):
Please see comments.

managed/src/main/java/com/yugabyte/yw/commissioner/HealthChecker.java
jitendra-12113: The intention is to keep the same email alert format for all the email alert

THought to just append in each email subject **Yugabyte Platform Alert**
Yugabyte Platform Alert - <[admin][admin]>
ERROR - <[admin][admin]> jitu-test-gcp-tls
OK - <[admin][admin]> tls-disabled-aws

Issue #7484, opened 2021-03-04
[docdb] Consistently sort IPs of tablet replica nodes in master table view


robertsami:
Currently our UI at `:7000/table?id={table_id}` looks like this:

![image](https://user-images.githubusercontent.com/1457684/110017900-b70ed880-7cf4-11eb-936d-f14784b62a45.png)

If we could order these IPs consistently, and maintain bold formatting on leader, it would make it much easier to discern whether a table is balanced across a three node cluster at a glance.

see: https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/master/master-path-handlers.cc#L1122

PR #7583, opened 2021-03-10
[#7484][docdb]Sort the hosts of tablet replicas

YintongMa:
# Description:
We sort the tablet replica hosts consistently, showed in the UI at :7000/table?id={table_id} 

The host of a replica could be an ip string, a host name or a uuid. Considering both the extensibility and ordering consistency, we sort them by the lexicographical ordering.

Before:
image

After: 
image


src/yb/master/master-path-handlers.cc
```diff
@@ -1011,8 +1011,14 @@ void MasterPathHandlers::HandleCatalogManager(const Webserver::WebRequest& req,
...
 
-bool CompareByRole(const TabletReplica& a, const TabletReplica& b) {
-  return a.role < b.role;
+bool CompareByHost(const TabletReplica& a, const TabletReplica& b) {
+    TSRegistrationPB reg_a = a.ts_desc->GetRegistration();
+    TSRegistrationPB reg_b = b.ts_desc->GetRegistration();
+    if (!reg_a.common().http_addresses().empty() && !reg_b.common().http_addresses().empty()) {
+        return reg_a.common().http_addresses(0).host() < reg_b.common().http_addresses(0).host();
+    } else {
+        return a.ts_desc->permanent_uuid() < b.ts_desc->permanent_uuid();
```
robertsami, line 1020: since permanent_uuid is generally expected to not be empty, and since our goal is only to have /some/ stable sorting of hosts, why dont we simplify this logic by always sorting by permanent_uuid?

YintongMa: The reason is I was trying to keep the hosts' increasing order if they're IPs. That would be more intuitive I think. But it indeed intriduces a more complex logic here. I could modify it to simply sort by permanent_uuid. What do you think?

robertsami: i think lets prefer simplicity in this case, just to enable quick eyeball comparison of replica balance in 3 node clusters

robertsami (approved):

PR #7609, opened 2021-03-11
[docs] Correct prometheus port from 93000 to 9300

dotyjim-work:
Commit 2cc084f introduced some extra ports, and the 93000 had a extra zero on it. This corrects that.


polarweasel (approved):
LGTM. Thanks for the correction @dotyjim-work!

Issue #7280, opened 2021-02-18
[Platform] Refactor gsutil installation


baba230896:
**To-Do:**

1. Add `gsutil.tar.gz` in the third-party dependency image to make the version consistent.
2. Refactor the `gsutil` installation in the YW docker image as per the new `gsutil.tar.gz` in third-party image.

The goal is to avoid pulling in the entire google cloud sdk .tar.gz and include just the gsutil package. @daniel-yb

iSignal:
I verified that the itest testsimpleops-gcp-rf1-gcs is passing on master (2.7.2.0-b20) is passing after this refactoring.

PR #7654, opened 2021-03-15
[Platform #7280]Trim extra spaces in flags.

mahendranbhat:
**Description**: If there's a space in the flag name, YB doesn't seem to accept the flag. Please ltrim and rtrim the flag name prior to saving it in the db as well as updating the yb conf file.

**Testing** : Update the universe with G-flags which has extra spaces at the beginning or end.


managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -413,6 +413,27 @@ public Result configure(UUID customerUUID) {
       //  uuid.
       Cluster c = taskParams.currentClusterType.equals(ClusterType.PRIMARY) ?
         taskParams.getPrimaryCluster() : taskParams.getReadOnlyClusters().get(0);
+      UserIntent primaryIntent = c.userIntent;
+      for (Map.Entry<String, String> intent : primaryIntent.masterGFlags.entrySet()) {
```
SergeyPotachev, line 417: 1. It is better to extract this functionality into a separate function instead of copying it 6 times.
2. Also it is not clear why to put the same item into the map (when key == key.trim and it is already there).
3. Have you checked this when you have more that one item in the map and you are updating a value in the middle of the iteration? I suppose it should fail with an error (not sure about the latest Java versions but some time ago you was unable to modify collections while iterating it).

SergeyPotachev (changes requested):
See comments.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
mahendranbhat: @SergeyPotachev 
Yeah agree, thought of doing the same (1). So moved the PR to draft state. I Will update the heading.
@jaydeepkumara please take care of the changes with test cases.

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
```diff
@@ -1129,6 +1129,41 @@ public void testUniverseGFlagsUpgradeValidParams() {
...
+    JsonNode json = Json.parse(contentAsString(result));
+    assertValue(json, "taskUUID", fakeTaskUUID.toString());
+    verify(mockCommissioner).submit(eq(TaskType.UpgradeUniverse), any(UniverseTaskParams.class));
+
+    CustomerTask th = CustomerTask.find.query().where().eq("task_uuid", fakeTaskUUID).findOne();
+    assertNotNull(th);
+    assertThat(th.getCustomerUUID(), allOf(notNullValue(), equalTo(customer.uuid)));
+    assertThat(th.getTargetName(), allOf(notNullValue(), equalTo("Test Universe")));
+    assertThat(th.getType(), allOf(notNullValue(), equalTo(CustomerTask.TaskType.UpgradeGflags)));
+    assertAuditEntry(1, customer.uuid);
```
SergeyPotachev, line 1164: Can we check that the stored flags are trimmed?

managed/ui/src/components/universes/UniverseForm/UniverseForm.js
```diff
@@ -393,7 +393,7 @@ class UniverseForm extends Component {
             return isNonEmptyString(masterFlag.name) && isNonEmptyString(masterFlag.value);
           })
           .map((masterFlag) => {
-            return { name: masterFlag.name, value: masterFlag.value };
+            return { name: masterFlag.name, value: masterFlag.value.trim() };
```
SergeyPotachev, line 396: Do we need this here? If yes then should we trim the flag name too?

SergeyPotachev (changes requested):
Overall looks good. Need one more test in UniverseControllerTest.java exactly for trimFlags. + Some comments.

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
```diff
@@ -1129,6 +1127,52 @@ public void testUniverseGFlagsUpgradeValidParams() {
...
+  }
+
+  @Test
+  public void testUniverseGFlagsUpgradeWithTrimParams2() {
```
jaydeepkumara, line 1166: rename

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
```diff
@@ -1129,6 +1129,52 @@ public void testUniverseGFlagsUpgradeValidParams() {
...
+    assertNotNull(th);
+    assertThat(th.getCustomerUUID(), allOf(notNullValue(), equalTo(customer.uuid)));
+    assertThat(th.getTargetName(), allOf(notNullValue(), equalTo("Test Universe")));
+    assertThat(th.getType(), allOf(notNullValue(), equalTo(CustomerTask.TaskType.UpgradeGflags)));
+    assertAuditEntry(1, customer.uuid);
+  }
+
+  @Test
+  public void testUniverseTrimFlags() {
+    Map<String, String> data = new HashMap<String, String>();
```
SergeyPotachev, line 1169: Map<String, String> data = new HashMap<>();

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
```diff
@@ -1129,6 +1129,52 @@ public void testUniverseGFlagsUpgradeValidParams() {
...
+    data.put(" Test ", " One ");
+    data.put(" Test 2 ", " Two ");
+
+    Map<String, String> result = UniverseController.trimFlags(data);
+    assertEquals(result.get("Test"), "One");
+    assertEquals(result.get("Test 2"), "Two");
+  }
```
SergeyPotachev, line 1176: Maybe to add one more check that the items count (in the map) wasn't changed.

managed/src/test/java/com/yugabyte/yw/controllers/UniverseControllerTest.java
SergeyPotachev: I'm agree that it is quite hard and the goal is too small. We already have test(s) exactly for this function. So let's omit it.

SergeyPotachev (changes requested):
OK, two very small comments and we are good to go.

SergeyPotachev (approved):
Looks good to me.

PR #7701, opened 2021-03-17
[#6857]Platform: Pre-flight checks for encryption-at-rest keys

jitendra-12113:
**Heading:**
[#6857]Platform: Pre-flight checks for encryption-at-rest keys
**Description:**
We don't have any check to validate the AWS creds and api url while
saving the config.
At the time of creating a new key configuration, need to do a pre-flight check
to verify the key config is valid.

**Testing:**
1. Got to Config->Security->Encryption-at-Rest keys tab
2. Select kms provider as AWS
3. Enter invalid accesskey or secret key - we should get proper error message for the same
4. If we select SmartKey as KMS provider then api url should be one of these
    api.amer.smartkey.io, 
    api.eu.smartkey.io, 
    api.uk.smartkey.io
    
 UT's have been added. SBT tests are getting passed.
    
 **TODO:**  - UI changes are pending. Also not sure whether we want to validate the combination of api url+secret API key incase of smarkey KMS provider if so then needed clarification on how are we going to do that.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
@@ -15,6 +15,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.inject.Inject;
+import com.yugabyte.yw.cloud.CloudAPI;
 import com.yugabyte.yw.commissioner.Commissioner;
 import com.yugabyte.yw.commissioner.tasks.params.KMSConfigTaskParams;
 import com.yugabyte.yw.common.ApiResponse;
@@ -31,7 +32,11 @@ import com.yugabyte.yw.models.helpers.TaskType;
 
 import java.util.Base64;
 import java.util.List;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -44,13 +49,44 @@
 
 public class EncryptionAtRestController extends AuthenticatedController {
     public static final Logger LOG = LoggerFactory.getLogger(EncryptionAtRestController.class);
+    static final String SMARTKEY_KMS_PROVIDER = "SMARTKEY";
```
daniel-yb, line 52: Just use https://github.com/yugabyte/yugabyte-db/blob/master/managed/src/main/java/com/yugabyte/yw/common/kms/util/KeyProvider.java#L22 enums, there's no reason to re-define the different key provider types as strings here.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -44,13 +49,44 @@
...
+    private void validateKMSProviderConfigFormData(ObjectNode formData, String keyProvider) {
+      if(keyProvider.toLowerCase().equals(AWS_KMS_PROVIDER)
+          && (formData.get("AWS_ACCESS_KEY_ID")!=null
+          || formData.get("AWS_SECRET_ACCESS_KEY")!=null)) {
+        CloudAPI cloudAPI = cloudAPIFactory.get(AWS_KMS_PROVIDER);
+        Map<String, String> config = new HashMap<String,String>();
+        config.put("AWS_ACCESS_KEY_ID", formData.get("AWS_ACCESS_KEY_ID").textValue());
+        config.put("AWS_SECRET_ACCESS_KEY", formData.get("AWS_SECRET_ACCESS_KEY")
+            .textValue());
+        if (cloudAPI != null && !cloudAPI.isValidCreds(config, formData.get("AWS_REGION")
```
daniel-yb, line 73: Where is this isValidCreds method defined? I can't seem to find it.

daniel-yb: Also, we should do some pre-flight checks on if the customer has chosen to use IAM instance profile credentials for their AWS KMS config instead of explicitly providing credentials. See AwsEARServiceUtil.java for how we fallback to using those credentials if a user has not explicitly provided any.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -44,13 +49,44 @@
...
+            .textValue())) {
+          throw new RuntimeException("Invalid AWS Credentials.");
+        }
+      }
+      if (keyProvider.equals(SMARTKEY_KMS_PROVIDER)) {
+        Set<String> apiUrl = new HashSet<String>();
+        apiUrl.add("api.amer.smartkey.io");
```
daniel-yb, line 80: Make this set a final class field like: `final Set<String> apiUrl = new HashSet<>(Arrays.asList("api.amer.smartkey.io", "api.eu.smartkey.io", "api.uk.smartkey.io"));` instead of instantiating it each time this method runs.

daniel-yb: Instead of just verifying that the format of provided creds are valid, you can also actually verify that the credentials are valid by running something like SmartKeyEARService::retrieveSessionAuthorization and checking the output.

daniel-yb (changes requested):
Sorry for the delay in looking at this diff. Left a few comments.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
jitendra-12113: **isValidCreds** is present in CloudAPI.java and its implementation is given in AWSCloudImpl.java.

jitendra-12113: This method got recently merged as part of separate PR https://github.com/yugabyte/yugabyte-db/pull/8393

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
jitendra-12113: Sure. Thanks!

jitendra-12113: Would like to merge API error handling change  PR https://github.com/yugabyte/yugabyte-db/pull/8199 before merging this one. 

Will address all the review comments together. This is  OLD PR so may have to create new one. Will update shortly. 

Thanks!

jitendra-12113:
Created separate PR(To avoid merge conflicts): https://github.com/yugabyte/yugabyte-db/pull/8737

@daniel-yb  We will create followup ticket to have check on IAM. I need lil bit more context on the same. 

if (callerIdentity != null) {
      String credentialsArn = callerIdentity.getArn();
      if (credentialsArn.contains(":assumed-role/")) {
        credentialsArn = getRole(configUUID, getResourceNameFromArn(credentialsArn)).getArn();
      } else if (!credentialsArn.contains(":user/")) {
        throw new RuntimeException("Credentials provided are not associated to a user or role");
      }
      String rootArn = String.format("arn:aws:iam::%s:root", callerIdentity.getAccount());
      return bindParamsToPolicyBase(policyBase, credentialsArn, rootArn).toString();
    } else {
      LOG.error("Could not get AWS caller identity from provided credentials");
      return null;
    }

PR #7818, opened 2021-03-25
[Docs] fix broken link to query performance

polarweasel:
We missed one link when we moved the query performance section up a level...


lizayugabyte:
Deleted the table row containing query performance  (because the file is in a different Explore section now)

lizayugabyte:
Closing because it's a duplicate of https://github.com/yugabyte/yugabyte-db/pull/7827

Issue #7421, opened 2021-03-01
[Platform] Enable encryption by default  Client-to-node/node-to-node


adil246:
Comments:

 ("Enable Client-to-node encryption") is off

This can be enabled by allowing encrypted connections

Similar issue: https://github.com/yugabyte/yugabyte-db/issues/6287

adilakram87:
@datamgmt @tylarb

gaurav061:
PR: https://github.com/yugabyte/yugabyte-db/pull/7823

PR #7823, opened 2021-03-26
[#7421]Platform Set node-to-node and client-to-node encryption to true

gaurav061:
Description: Set the default value of node-to-node and client-to-node encryption to true.


nishantSharma459 (commented):
`Client to Node Encryption` depends upon `Node to Node Encryption`. Initially `Client to Node TLS` is disabled once `Node to Node TLS` is enabled then we're allowing the user to enable the `Client to Node TLS`.

So, this change will cause problems with the current behavior.

gaurav061:
`Client to Node Encryption ` and `Node to Node` Encryption will be initially set to true as asked in this issue, once the user sets `Client to Node Encryption `  to false it will force `Node to Node` to be false and your changes will reflect.

nishantSharma459 (approved):

sshev (approved):

PR #7838, opened 2021-03-29
Update OSS helm chart docs to reflect new CPU / memory default values

jvigil-yugabyte:
The default values for CPU / memory have changed. The text on this page
needs to be updated to reflect the new values.


polarweasel (approved):
LGTM. Thanks Jason!

PR #7871, opened 2021-03-30
[docs] tweak prompt-removal logic for code blocks

### LINKED ISSUE #7685: [docs Platform] Add more details on Azure Backup storage doc
Please add more details on the doc like how to get container URL and SAS token on the docs and/or provide some sample values. There are 2 ways to generate the SAS token and the code will not accept if teh SAS token doesnt start with ?sv. So better to provide more information and example.

@polarweasel

polarweasel:
Fixes #7685

* add a couple more labels for SQL languages
* add a couple more common labels for shell prompts
* add a new label called "output" that doesn't get a copy button

### DIFF
--- a/docs/src/index.js
+++ b/docs/src/index.js
@@ -151,16 +151,21 @@ $(document).ready(() => {
       let regExpCopy = /a^/;
       if (languageDescriptor) {
         // Then apply copy button
-        if (['postgresql', 'sql', 'postgres'].includes(languageDescriptor)) {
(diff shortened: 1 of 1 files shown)


polarweasel:
The reason for the new label called `output` is that well-formed Markdown should have a language for every code block. We're going to start enforcing that and other Markdown conventions pretty soon, so being able to label output as, well, output will eliminate a _huge_ number of linter warnings. :)

PR #7897, opened 2021-04-01
YCQL, docs: Changed audit docs for clarity, added example to show how to exclude SELECTs/DMLs

frozenspider:
Also slightly clarified gflag tooltips.

### DIFF
--- a/docs/content/stable/secure/audit-logging/audit-logging-ycql.md
+++ b/docs/content/stable/secure/audit-logging/audit-logging-ycql.md
@@ -66,72 +66,55 @@ The available flags are described in the table below:
 
 <table>
   <tr>
-   <td><strong>Flag</strong>
-   </td>
-   <td><strong>Valid Values</strong>
-   </td>
-   <td><strong>Description</strong>
-   </td>
+   <td><strong>Flag</strong></td>
+   <td><strong>Valid Values</strong></td>
+   <td><strong>Description</strong></td>
+   <td><strong>Default Value</strong></td>
   </tr>
   <tr>
-   <td><code>ycql_enable_audit_log</code>
-   </td>
-   <td><code>false</code> (default) or <code>true</code>.
-   </td>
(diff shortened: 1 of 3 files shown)


docs/content/stable/secure/audit-logging/audit-logging-ycql.md
```diff
@@ -66,72 +66,55 @@ The available flags are described in the table below:
...
+   <td>empty</td>
   </tr>
   <tr>
-   <td><code>ycql_audit_excluded_keyspaces</code>
-   </td>
-   <td>keyspaces to be excluded from auditing.
-   </td>
+   <td><code>ycql_audit_excluded_keyspaces</code></td>
+   <td>keyspaces to be excluded from auditing.</td>
+   <td><code>system,system_schema,system_virtual_schema,system_auth</code></td>
```
kmuthukk, line 111: if this value is set to "ks1"  will "system*" keyspace remaining excluded?

And can these system* keyspaces be audited (by adding them to include list) or are these always special/internal keyspaces?

If not, then we don't really need to add those in the default description (as it is only confusing).

kmuthukk (commented):
Some inline comments, @frozenspider .

docs/content/stable/secure/audit-logging/audit-logging-ycql.md
frozenspider: This is a regular default value with no special properties - if you set this to `ks1`, system keyspaces won't be excluded anymore. Thus, you want to leave them excluded, you have to enumerate them in this filter along with  `ks1`. This design is inspired by Apache Cassandra's audit log configuration.

Thus, the answer to your second question is - as per include/exclude rules, adding them to the include list won't do anything on its own. We audit _included_ entities which _aren't excluded_, as described in `Configure Audit Logging` section.

kmuthukk (approved):
Thx @frozenspider for clarifying.

PR #7943, opened 2021-04-07
docs: fix YSQL backfill client timeout flag

jaki:
Alan pointed out that increasing the timeout flags was not working.  It
turns out that one of the flags was wrong.  Fix that.


m-iancu (approved):
LGTM!

Issue #7959, opened 2021-04-07
[Platform] Disabling Node-to-Node TLS during universe creation causes universe creation to fail


daniel-yb:
Steps to reproduce:
1. Fill out create universe form
2. Disable Node to Node TLS
3. Submit the create request
4. Notice that universe creation fails with `Node-to-node TLS needs to be enabled for client-to-node TLS to be enabled` implying that although client to node TLS showed as disabled in the UI, the actual create universe payload request value was never updated to `false` for node to node TLS.
This bug seems to have been introduced in #7421

gaurav061:
PR: https://github.com/yugabyte/yugabyte-db/pull/7972

PR #7972, opened 2021-04-09
[#7959] Platform: Fixed Universe Failure when node-to-node TLS is disabled. 

gaurav061:
Description: Disabling node-to-node TLS sets the client-to-node TLS to be disabled, but this field was not updating payload form value, it was just done for UI.

Scenario: After this fix, I tried creating Universe for the following scenario.
1. Keeping node-to-node and client-to-node set to true (**Default**)
2. Keeping node-to-node enabled and client-to-node to be disabled.
3. Keeping node-to-node and client-to-node set to false.


gaurav061:
Please Add your comments.

sshev (approved):

PR #7986, opened 2021-04-09
[docs] Update documented instructions on running TPCC

robertsami:
In the interest of simplicity, we'd like to direct folks to run each step independently, rather than running all in one shot. This PR modifies the docs with TPCC instructions accordingly.


polarweasel (approved):
LGTM. Sorry I missed this @robertsami !

Issue #7543, opened 2021-03-08
[docdb] Show ALIVE time for masters in the master home UI


bmatican:
Basically in `master_ip:7000/`, next to the raft information, we can also show the uptime, like we do for the `/tablet-servers`

triump2020:
@bmatican   i'd like to give it a shot !  assign me

kmuthukk:
Thx @triump2020 !

triump2020:
@bmatican   How dose  the leader-master  get the uptime information for the other follower-master  in a raft group ?    the master 
gets the uptime information for the **tserver**  through heartbeating by tserver   , but there seems not be a way  to get the uptime information for the follower-master ?

bmatican:
Hey @triump2020, that's a great question! For TS>master communication, we actually add the uptime in a custom fashion, during the heartbeating code path. However, that won't work for masters, out of the box.

For the master UI, we essentially use the `ListMasters` RPC, which underneath has a list of `ServerEntryPB`, each of which has their corresponding `NodeInstancePB` of the respective masters. This means if the instance data has the startup time, we should be good!

So my suggestion then would be to augment `NodeInstancePB`. Eventually we could refactor the TS heartbeat codepath to also use this. This protobuf is generated at server startup, for both master and tserver, and contains a minimal amount of information:
```
// Uniquely identify a particular instance of a particular server in the
// cluster.
message NodeInstancePB {
  // Unique ID which is created when the server is first started
  // up. This is stored persistently on disk.
  required bytes permanent_uuid = 1;

  // Sequence number incremented on every start-up of the server.
  // This makes it easy to detect when an instance has restarted (and
  // thus can be assumed to have forgotten any soft state it had in
  // memory).
  //
  // On a freshly initialized server, the first sequence number
  // should be 0.
  required int64 instance_seqno = 2;
}
```
We could add a 3rd field above, as optional (so there's no backwards incompatibility issues), where we record the actual start time of the server.

The code that generates this is here:
```
void RpcAndWebServerBase::GenerateInstanceID() {
  instance_pb_.reset(new NodeInstancePB);
  instance_pb_->set_permanent_uuid(fs_manager_->uuid());
  // TODO: maybe actually bump a sequence number on local disk instead of
  // using time.
  instance_pb_->set_instance_seqno(Env::Default()->NowMicros());
}
```
[...]

PR #7992, opened 2021-04-11
Issue 7543

triump2020:
Fix : #7543

### DIFF
--- a/src/yb/common/wire_protocol.proto
+++ b/src/yb/common/wire_protocol.proto
@@ -117,7 +117,7 @@ message NodeInstancePB {
   // On a freshly initialized server, the first sequence number
   // should be 0.
   required int64 instance_seqno = 2;
-  // The time point when Tserver/master start up,it derives from Env::NowMicros()
+  // The time point when Tserver/master start up,it is derived from Env::NowMicros()
   optional uint64 start_time = 3;
 }
 
--- a/src/yb/common/wire_protocol.proto
+++ b/src/yb/common/wire_protocol.proto
@@ -117,6 +117,8 @@ message NodeInstancePB {
   // On a freshly initialized server, the first sequence number
   // should be 0.
   required int64 instance_seqno = 2;
+  // The time point when Tserver/master start up,it derives from Env::NowMicros()
+  optional uint64 start_time = 3;
 }
 
 // RPC and HTTP addresses for each server, as well as cloud related information.
--- a/src/yb/common/wire_protocol.proto
+++ b/src/yb/common/wire_protocol.proto
@@ -117,7 +117,7 @@ message NodeInstancePB {
   // On a freshly initialized server, the first sequence number
   // should be 0.
   required int64 instance_seqno = 2;
-  // The time point when Tserver/master start up,it is derived from Env::NowMicros()
+  // The time point when Tserver/master starts up,it is derived from Env::NowMicros()
   optional uint64 start_time = 3;
 }
 
--- a/src/yb/master/master-path-handlers.cc
+++ b/src/yb/master/master-path-handlers.cc
@@ -1511,7 +1511,8 @@ void MasterPathHandlers::HandleMasters(const Webserver::WebRequest& req,
   (*output) << "<table class='table'>\n";
   (*output) << "  <tr>\n"
             << "    <th>Server</th>\n"
-            << "    <th>RAFT Role</th>"
+            << "    <th>RAFT Role</th>\n"
+            << "    <th>Uptime</th>\n"
             << "    <th>Details</th>\n"
             << "  </tr>\n";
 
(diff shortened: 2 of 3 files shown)


src/yb/common/wire_protocol.proto
```diff
@@ -117,6 +117,8 @@ message NodeInstancePB {
   // On a freshly initialized server, the first sequence number
   // should be 0.
   required int64 instance_seqno = 2;
+  // The time point when Tserver/master start up,it derives from Env::NowMicros()
```
tedyu, line 120: start up -> starts up

it derives -> it is derived

src/yb/server/server_base.cc
```diff
@@ -458,6 +458,7 @@ void RpcAndWebServerBase::GenerateInstanceID() {
   // TODO: maybe actually bump a sequence number on local disk instead of
   // using time.
   instance_pb_->set_instance_seqno(Env::Default()->NowMicros());
+  instance_pb_->set_start_time(Env::Default()->NowMicros());
```
tedyu, line 461: Should the value of Env::Default()->NowMicros() in the previous call be saved and used here ?

tedyu (commented):
Is there any change in thirdparty needed for this pull request ?

src/yb/common/wire_protocol.proto
triump2020: No need .   i have tested on my local cluster .

src/yb/server/server_base.cc
triump2020: @tedyu  ,  As  @bmatican  said :
  `Note, that even though technically, I think we could already use the instance_seqno, because it is implemented as the NowMicros of when the server starts up, that feels like an implementation detail that we should not rely on...`

triump2020: @tedyu  Thanks for your review !

triump2020: @tedyu  Thanks for your review !

src/yb/common/wire_protocol.proto
tedyu: don't quite understand your reply.
I was pointing out grammatical mistake in comment.

src/yb/server/server_base.cc
tedyu: Ack.

src/yb/common/wire_protocol.proto
triump2020: @tedyu  I have already fixed the  grammatical mistake in comment.

triump2020:
> Is there any change in thirdparty needed for this pull request ?

@tedyu   No changes needed in the thirdpary for this issue , but  i don't know why there is a extra commit log  like this :
image

src/yb/common/wire_protocol.proto
tedyu: Now I see the uo-to-date comment in code which is fine.

I don't have other review comment.

src/yb/common/wire_protocol.proto
```diff
@@ -117,6 +117,8 @@ message NodeInstancePB {
   // On a freshly initialized server, the first sequence number
   // should be 0.
   required int64 instance_seqno = 2;
+  // The time point when Tserver/master starts up,it is derived from Env::NowMicros()
```
bmatican, line 120: nit: Tserver -> tserver
nit: space after the ,

also, let's name the field `start_time_us`, so the precision is reflected (makes it easier to reason about in code, directly from the field name)

src/yb/master/master-path-handlers.cc
```diff
@@ -1543,13 +1544,15 @@ void MasterPathHandlers::HandleMasters(const Webserver::WebRequest& req,
       reg_text = Substitute("<b>$0</b>", reg_text);
     }
     string raft_role = master.has_role() ? RaftPeerPB_Role_Name(master.role()) : "N/A";
+    string uptime = UptimeString((Env::Default()->NowMicros() - master.instance_id().start_time()) / 1000000);
```
bmatican, line 1547: to not hardcode this `/ 1000000`, either use something like `MonoDelta::FromMicroseconds` or `MonoTime::GetDeltaSince`

src/yb/server/server_base.cc
bmatican: I think Ted's comment is still fair, just create a new `auto now = ...` and use it in both the spots, that way we don't compute time twice. The comment still remains valid, and if we ever address the TODO, we'd change this.

bmatican (changes requested):
great job!

two high level asks, aside from minor comments:
- can you add a screenshot of the master main page, to the main GH issue, so we can see how this looks like?
- please remove the `thirdparty` change, it was probably unintended...maybe using something like `./bin/undo_submodule_changes.sh`

triump2020:
@bmatican 
image

bmatican:
@triump2020 I guess we can abandon this PR, right?

triump2020:
> @triump2020 I guess we can abandon this PR, right?

Right!

PR #8099, opened 2021-04-20
[docdb] master UI hash_split of last partition should be inclusive

### LINKED ISSUE #6299: [docdb] master UI hash_split of last partition should be inclusive
The master UI table page has tablets with their partition boundaries.  Recent commit dca4cf4ef52d768ea955ffe9a1b0b7ceffd20a9c made the format of the boundaries into hex notation.  It also converted the "start" and "end" words to `0x0000` and `0xffff`, respectively.

Now, it's a lot more obvious that there's an issue with all the boundaries being `[foo, bar)`, `foo` being inclusive and `bar` being exclusive.  Before, it looked like `[foo, <end>)`, which is sort of fine depending on your interpretation of "end"; now, it looks like `[foo, 0xffff)`, which is definitely incorrect.  Change that last partition to be all inclusive so that it looks like `[foo, 0xffff]`.  As of now, this is only applicable to YCQL hash partitioned tables.

imskr:
- Fixes master UI hash_split of the last partition by making it inclusive.
- Changing that last partition to be all-inclusive so that it looks like `[foo, 0xffff]`

Fixes #6299

### DIFF
--- a/src/yb/common/partition.cc
+++ b/src/yb/common/partition.cc
@@ -849,7 +849,7 @@ string PartitionSchema::PartitionDebugString(const Partition& partition,
         const string& pend = partition.partition_key_end();
         uint16_t hash_start = !pstart.empty() ? DecodeMultiColumnHashValue(pstart) : 0;
         uint16_t hash_end = !pend.empty() ? DecodeMultiColumnHashValue(pend) : UINT16_MAX;
-        s.append(Substitute("hash_split: [0x$0, 0x$1)",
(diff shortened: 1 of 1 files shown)


jaki (changes requested):
This looks like it changes all ranges to be inclusive, not just the last partition's.  For example, for 2 tablet table, I expect to see

    [0x0000, 0x7fff), [0x7fff, 0xffff]

but this change will probably do

    [0x0000, 0x7fff], [0x7fff, 0xffff]

PR #8142, opened 2021-04-23
[#7172] Platform: Error handling for create universe failure.

### LINKED ISSUE #7172: [Platform UI] No visual feedback for certain universe creation failures
When a user creates a new universe, it is possible that the creation fails even before creating a new task. These are the cases where some pre-flight validation fails. These failures are not shown in the Tasks section. Usually the API returns 400 bad request along with an error message. The user has no clue about this failure. https://github.com/yugabyte/yugabyte-db/pull/7159 adds one such validation check. 

https://github.com/yugabyte/yugabyte-db/pull/7078 implements a toaster for such cases. It needs to be used at other places like these.

cc @chirag-yb @yb-andrew

gaurav061:
Description: There is no visual feedback for specific failure while creating Universe, i.e If Universe creation fails before the task is generated. This change handles such cases.

Test Plan: Added custom exception to test the flow. (**Need to encounter actual case that may cause this error**)

Fixes #7172


jaydeepkumara (approved):

managed/ui/src/components/universes/UniverseForm/UniverseFormContainer.js
```diff
@@ -79,6 +79,10 @@ const mapDispatchToProps = (dispatch) => {
         dispatch(getTlsCertificates()).then((response) => {
           dispatch(getTlsCertificatesResponse(response.payload));
         });
+        if (response.error) {
+          const errorMessage = response.payload?.response?.data?.error || response.payload.message;
```
nishantSharma459, line 83: 1. `response.payload.response.data` why we are getting JSON like that?
2. And why `data.error` we can check the status code and populate the `data.message` on top of it.

I believe `response.payload.data` is enough. Can we ask BE folks to structure it properly or is there any particular reason behind that ??

sshev: @nishantSharma459 unfortunately, that's a flaw of the current way of making API calls

the `response.payload` part is not an actual network response, but smth what a redux action creator returns:
https://github.com/yugabyte/yugabyte-db/blob/c8a558eca9635c73c4c64650d83d3f6d12ba32c3/managed/ui/src/actions/universe.js#L127-L130
then `response.data` is an actual network response as per axios [response schema](https://axios-http.com/docs/res_schema)

so when API returns `{ error: 'bla-bla-bla' }` - above is the way to extract that message from response

fallback to generic `response.payload.message` is needed for cases when API doesn't return any message, like when it's 500 server error of some network connectivity issues

sshev (approved):

managed/ui/src/components/universes/UniverseForm/UniverseFormContainer.js
nishantSharma459: Understood.

nishantSharma459 (approved):

PR #8151, opened 2021-04-23
[doc] PR for Add link to DDL statements API doc to the Explore section (Doc7578)

lizayugabyte:
Added a row to the table in YSQL vs Postgresql section index file


polarweasel (approved):
Good to go

PR #8164, opened 2021-04-26
[docs][OpenShift] Use different word than IP address and add clarification about admin user

bhavin192:
- Remove note about logging out of admin account
  - The note was there because the whole section of finding Availability Zones was at some other place.
- Use different word than IP address
  - There are setups where you don't have an IP address for a LoadBalancer service, instead you get a hostname. Changing the word IP address with location for now, I wasn't able to find anything better than that. URI can be used but it will add some confusion.
-  Add a note about admin and normal user accounts


lizayugabyte (approved):
LGTM

Issue #3785, opened 2020-02-28
[docdb] Investigate enabling other compression algorithms


bmatican:
Currently, we use snappy for rocksdb files by default. However, rocksdb has support for various compression algorithms.

@psudheer21 recommended considering `LZ4`:
> Compression is about 40% faster and decomprssion is 150% faster.

@ddorian also suggested looking at rocksdb dictionary compression: https://github.com/facebook/rocksdb/wiki/Dictionary-Compression

For all of these, we should be able to do some simple high-pass tests, like:
- create a local cluster
- load a fixed, large amount of data (say 1-10M rows)
- do a compaction to drop down to 1 file, for consistent test results
- time a `select *` or `select count(*)`, to force scanning all the data and going through decompression path

For the actual compression path, it only factors in at the flush/compaction level. Since that's background work, we could probably just use a rocksdb specific benchmark style unit test (if we don't already have one), to get some baseline numbers on speed of compression/decompression.

cc @mbautin @spolitov

ddorian:
> We believe LZ4 is almost always better than Snappy. 

https://github.com/facebook/rocksdb/wiki/Compression

PR #8185, opened 2021-04-27
[#3785][docdb]Add alternative compression type support

YintongMa:
# Description
In this PR, we add another compression type - LZ4 for DocDB. The user now could select compression type from Snappy, LZ4 and NoCompression. 

The following test report shows that LZ4 improves the DocDB performance than Snappy does under the simple high-pass query senario.

### Env
We run the experiments in an MacOS, 2 GHz Quad-Core Intel Core i5 @ 3.40GHz, 16 GB 3733 MHz LPDDR4X, and 766 GB
SSD disk space. 

### Workload
2M rows of CassandraKeyValue generated by tools in https://github.com/yugabyte/yb-sample-apps

### Testing Steps
1. Create a local cluster with 1 replica
2. Load the workload of 2M rows of CassandraKeyValue
3. Do a compaction to drop down to 1 file
4. Restart the cluster to drop cache and time "select count(*)"
5. Execute Step 4 for 10 times and get the average time cost

### Result
LZ4 : 36s
Snappy : 40s
The result shows that LZ4 works better than Snappy with a 10% time cost improvement on average for query 2M records.

#### Detailed CPU monitoring
By TOP cmd, we could see the cpu utilization is higher than 90% during the query for both compression types. They all tend to make full use of the cpu. However, we might need more detailed mornitoring tools to compare the variation of cpu utilization with some resource competition to see which one is more robust.
#### Detailed IO monitoring
By IOSTAT cmd, the iops for the two compression types don't make big difference. We could find more related metrics to conduct a series of detailed IO benchmark.
#### Scalability
Now the tests are done in local env with 1 replica. We should also test it in cluster settings.


src/yb/docdb/docdb_rocksdb_util.cc
```diff
@@ -487,8 +507,16 @@ void InitRocksDBOptions(
...
 
-  options->compression = rocksdb::Snappy_Supported() && FLAGS_enable_ondisk_compression
-      ? rocksdb::kSnappyCompression : rocksdb::kNoCompression;
+  rocksdb::CompressionType compressionType = rocksdb::kNoCompression;
+  if (FLAGS_enable_ondisk_compression){
+      if (FLAGS_compression_type == 0 && rocksdb::LZ4_Supported()){
+          compressionType = rocksdb::kLZ4Compression;
+      }else if (FLAGS_compression_type == 1 && rocksdb::Snappy_Supported()){
+          compressionType = rocksdb::kSnappyCompression;
+      }
```
robertsami, line 516: if we configure a compression type which is not supported, we should probably **_at least_** `LOG(ERROR)`

src/yb/docdb/docdb_rocksdb_util.cc
```diff
@@ -119,6 +119,25 @@ DEFINE_int32(priority_thread_pool_size, -1,
              "If -1 and max_background_compactions is specified - use max_background_compactions. "
              "If -1 and max_background_compactions is not specified - use sqrt(num_cpus).");
 
+DEFINE_int32(compression_type, 1,
+             "compression_type. "
+             "If enable_ondisk_compression is false - no compression. "
+             "If 0 and enable_ondisk_compression is true - use LZ4. "
+             "If 1 and enable_ondisk_compression is true - use Snappy.");
```
robertsami, line 126: can we introduce an enum `YB_DEFINE_ENUM(OnDiskCompression, (kDisabled)(kUseSnappy)(kUseLz4));`

and then remove the `enable_ondisk_compression` flag altogether? and set this flag type to default to `OnDiskCompression::kUseSnappy` by defining it like so:

```
YB_DEFINE_ENUM(OnDiskCompression, (kDisabled)(kUseSnappy)(kUseLz4));

DEFINE_int32(
    compression_type, OnDiskCompression::kUseSnappy, strings::Substitute(
        "On-disk compression to use in RocksDB. Set to $0 to disable compression, $1 to use Snappy, and $2 to use LZ4.",
        OnDiskCompression::kDisabled, OnDiskCompression::kUseSnappy, OnDiskCompression::kUseLz4));
```

src/yb/docdb/docdb_rocksdb_util.cc
```diff
@@ -119,6 +119,25 @@ DEFINE_int32(priority_thread_pool_size, -1,
...
+
+namespace {
+    static bool CompressionTypeValidator(const char* flagname, int value) {
+        if (value > 1 or value < 0) {
+            LOG(INFO) << "Expect " << flagname << " only to be 0,1";
+            return false;
+        }
+        return true;
```
robertsami, line 134: could check against valid enum values here

robertsami (changes requested):

PR #8239, opened 2021-04-30
[docs] fix broken dbeaver page link

polarweasel:
Broken link popped up in SEO audit.

This is a result of splitting the DBeaver page into two, but not adding an alias to the original name. The alias _should_ be redundant now, but I've added it in case old blog posts etc might be linking to it.


lizayugabyte (approved):
LGTM, thanks!

PR #8288, opened 2021-05-04
Reduce time for PgMiniTest.SystemTableTxnTest

### LINKED ISSUE #7657: test: PgMiniTest.SystemTableTxnTest takes too long
C++ test `PgMiniTest.SystemTableTxnTest` takes way too long.  On centos-gcc-debug, I got

```sh
./yb_build.sh --cxx-test pgwrapper_pg_mini-test --gtest_filter PgMiniTest.SystemTableTxnTest
```

```
[       OK ] PgMiniTest.SystemTableTxnTest (957676 ms)
```

which is about 16 minutes.  It has

```cpp
  for (int i = 1; i <= 100; ++i) {
```

and I think this can be reduced down to 50 at least, theoretically cutting time in half.  That's still 8 minutes, though, which is bad, so try to cut it down more without reducing test stability.

Later, it does

```
  ASSERT_GE(commit1_fail_count, 25);
  ASSERT_GE(insert2_fail_count, 25);
```

These should be using variables so that they are relative to the total iterations (namely, 1/4 times the number of iterations).

Use `--test-parallelism 1 --num-repetitions 1000` to check stability.  Test parallelism 1 is usually needed because it gets finicky otherwise.

keywords: PgMiniTest, SystemTableTxnTest

jackstenglein:
Fixes #7657.

Tested with the following command: `./yb_build.sh --cxx-test pgwrapper_pg_mini-test --gtest_filter PgMiniTest.SystemTableTxnTest --test-parallelism 1 --num-repetitions 1000`

This was run on a centos 7 AWS EC2 t2.2xlarge instance and reduced the time to about 9-12 seconds per repetition. All repetitions passed.

### DIFF
--- a/src/yb/yql/pgwrapper/pg_mini-test.cc
+++ b/src/yb/yql/pgwrapper/pg_mini-test.cc
@@ -716,7 +716,8 @@ TEST_F_EX(PgMiniTest, YB_DISABLE_TEST_IN_TSAN(SystemTableTxnTest), PgMiniTestMan
   size_t insert2_fail_count = 0;
 
   const auto kStartTxnStatementStr = "START TRANSACTION ISOLATION LEVEL REPEATABLE READ";
-  for (int i = 1; i <= 100; ++i) {
+  const int iterations = 48;
+  for (int i = 1; i <= iterations; ++i) {
     std::string dictname = Format("contendedkey$0", i);
     const int dictnamespace = i;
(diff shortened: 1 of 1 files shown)


mbautin:
@jackstenglein Thanks for the contribution! What was the running time before the fix, and were there any failures?

jackstenglein:
@mbautin The runtime was originally about a minute per repetition, and there were no failures.

Issue #8225, opened 2021-04-29
Build with make fails on Linux


andrei-mart:
Reproduced on CentOS 7 and CentOS 8
If either ninja-build package is missing or make build is explicitly specified with
`./yb_build.sh --make`
compilation fails with message
```
[...]

PR #8319, opened 2021-05-06
[docs] add missing ninja-build package (#8225)

andrei-mart:


polarweasel:
Hey @svarnau is this PR required, or did you address the missing dependency in c77d489?

svarnau (approved):
Looks good. Technically it is optional, but it is the default build tool.

PR #8357, opened 2021-05-09
[#7571][docdb] Grouping and cleaning up setters to eliminate locking overhead

### LINKED ISSUE #7571: docdb: Group All TSDescriptor Setters Into One Method
On each heartbeat, master calls TSDescriptor methods like 
```
  void set_num_sst_files (uint64_t num_sst_files) {
    std::lock_guard<decltype(lock_)> l(lock_);
    ts_metrics_.num_sst_files = num_sst_files;
  }
```

that takes a write lock, sets a field, and releases the write lock. To avoid doing this many times over for each individual field, create a method that takes a lock once and sets every field before releasing the lock.

bozhiyou:
Fixes #7571. Existing tests suffice.

### DIFF
--- a/src/yb/master/ts_descriptor.cc
+++ b/src/yb/master/ts_descriptor.cc
@@ -164,9 +164,17 @@ std::string TSDescriptor::placement_id() const {
   return placement_id_;
 }
 
-void TSDescriptor::UpdateHeartbeatTime() {
+void TSDescriptor::UpdateHeartbeat(const TSHeartbeatRequestPB* req) {
   std::lock_guard<decltype(lock_)> l(lock_);
   last_heartbeat_ = MonoTime::Now();
+
+  DCHECK_GE(req->num_live_tablets(), 0);
+  DCHECK_GE(req->leader_count(), 0);
+  num_live_replicas_ = req->num_live_tablets();
+  leader_count_ = req->leader_count();
+  physical_time_ = req->ts_physical_time();
+  hybrid_time_ = HybridTime::FromPB(req->ts_hybrid_time());
+  heartbeat_rtt_ = MonoDelta::FromMicroseconds(req->rtt_us());
 }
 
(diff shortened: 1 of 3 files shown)


bmatican:
Oh, sorry @bozhiyou , I was out the past couple of days. Let me get you some reviewers.

src/yb/master/ts_descriptor.cc
```diff
@@ -164,9 +164,17 @@ std::string TSDescriptor::placement_id() const {
   return placement_id_;
 }
 
-void TSDescriptor::UpdateHeartbeatTime() {
+void TSDescriptor::UpdateHeartbeat(const TSHeartbeatRequestPB* req) {
   std::lock_guard<decltype(lock_)> l(lock_);
   last_heartbeat_ = MonoTime::Now();
+
+  DCHECK_GE(req->num_live_tablets(), 0);
```
tedyu, line 171: It seems these two checks can be lifted ahead of taking the lock

bozhiyou: Good catch, thanks! Fix committed.

bozhiyou:
@bmatican Never mind. Could you please also assign some reviewers for #8364? Thanks!

src/yb/master/ts_descriptor.cc
tedyu: nit: you can enclose the lines starting with std::lock_guard to the last line in a pair of curly braces.

You can find many examples of such coding style throughout the YB code base.

I or one of my colleagues would put your patch through internal Jeninks.

Thanks for your contribution.

bozhiyou: Fixed. Thanks for your feedback! It's exciting to contribute to this great project 😆

tedyu: I submitted your patch for YB internal Jenkins.

Once test result comes back green and one of my co-workers confirms, this would be merged.

Issue #7394, opened 2021-02-26
[Platform] [k8s] Simplified k8s provider config for deploying into the same cluster as YW


iSignal:
The plan is to have a simplified flow to deploy into the same k8s cluster as the one YW is already deployed in. From the UI point of view, this could look like a checkbox in the k8s provider config to use the current cluster config or a button to prefill the current cluster config into the provider fields. 

The current k8s provider has the following info:

1. KUBECONFIG/service account - this can be omitted in the same cluster
2. Zone/region info - the k8s node labels have this info
3.  Storage class - current YW volume storage class
4. Pull secret - there is a pull secret in the current NS that was used to pull the YW image (this might be tricky to get).
5. Optional overrides (we could keep this in the simplified UI).
6. Optional image name for the YB image (we could keep this in the simplified UI, prefilled to quay.io as usual)

@chirag-yb @streddy-yb @bhavin192 @baba230896

bhavin192:
The API I'm planning to add will have following JSON. This is similar to what we send to backend when we create a new Kubernetes provider (`POST /api/v1/customers/customer_uuid/providers/kubernetes`)

```json
{
  "code": "kubernetes",
  # "name": "this-cluster", (might not be there).
  "config": {
    "KUBECONFIG_PROVIDER": "custom",
    "KUBECONFIG_SERVICE_ACCOUNT": "yw-test",
  },
  "regionList": [
    {
      # "name": "South Asia", (need to check if there is some mapping for cloud specific name to normal name).
      # "code": "south-asia", (might be different, something like asia-south1).
      # "latitude": 18.4, (won't be there).
      # "longitude": 78.4, (won't be there).
      "zoneList": [
        {
          "code": "asia-south1-a",
          "name": "asia-south1-a",
          "config": {
            "STORAGE_CLASS": "standard",
            "OVERRIDES": "",
            "KUBECONFIG_NAME": "",
            "KUBECONFIG_CONTENT": ""
          }
        },
        {
          "code": "asia-south1-c",
          "name": "asia-south1-c",
          "config": {
            "STORAGE_CLASS": "standard",
            "OVERRIDES": "",
            "KUBECONFIG_NAME": "",
            "KUBECONFIG_CONTENT": ""
          }
        }
      ]
    }
  ]
}
```

bhavin192:
- [ ] https://github.com/yugabyte/yugabyte-db/pull/8856#issuecomment-859492055

[...]

PR #8371, opened 2021-05-10
[Platform] Add API for Kubernetes provider configuration discovery

bhavin192:
-   This API tries to fetch the details which can be used to pre-fill the data during Kubernetes provider creation (should work on OpenShift or Tanzu as well). The returned JSON is similar to what we use for the create call. Please refer this issue for more details: https://github.com/yugabyte/yugabyte-db/issues/7394
-   Use blank string for KUBECONFIG if kubeconfig file is not provided. This allows us to in-cluster credentials (ServiceAccount) when running in the same cluster as of the target cluster. Can be used to simplify the current flow later, we won't need a separate kubeconfig in that case.
-   Finds out the region and AZ based on node annotations.
-   Takes the pull secret name and the storage class name from app config.

Scenarios tested:

-   Ran platform inside Kubernetes, curl the API, it returns the expected JSON.
-   Tested all the failure scenarios like missing config, no permissions to get secret, nodes etc.

A sample output:
```json
{
  "code": "kubernetes",
  "name": null,
  "active": true,
  "config": {
    "KUBECONFIG_IMAGE_PULL_SECRET_NAME": "yugabyte-k8s-pull-secret",
    "KUBECONFIG_PULL_SECRET_NAME": "yugabyte-k8s-pull-secret",
    "KUBECONFIG_PULL_SECRET_CONTENT": "{\"apiVersion\":\"v1\",\"data\":{\".dockerconfigjson\":\"removed==\"},\"kind\":\"Secret\",\"metadata\":{\"annotations\":{},\"name\":\"yugabyte-k8s-pull-secret\"},\"type\":\"kubernetes.io/dockerconfigjson\"}"
  },
  "region": null,
  "regionList": [
    {
      "code": "asia-south1",
      "name": null,
      "latitude": 0.0,
      "longitude": 0.0,
      "zoneList": [
        {
          "code": "asia-south1-a",
          "name": "asia-south1-a",
          "config": {
            "STORAGE_CLASS": "standard"
          }
        },
        {
          "code": "asia-south1-c",
          "name": "asia-south1-c",
          "config": {
            "STORAGE_CLASS": "standard"
          }
        }
      ],
      "config": null
    }
  ]
}
```

Open Questions
- [x] Should the API endpoint as well as the methods be called something like `/customers/:cUUID/providers/kubernetes_discovery`?

This PR is backend part for the  https://github.com/yugabyte/yugabyte-db/issues/7394

### DIFF
--- a/managed/build.sbt
+++ b/managed/build.sbt
@@ -168,7 +168,8 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-core" % "2.10.5",
   "com.jayway.jsonpath" % "json-path" % "2.4.0",
   "commons-io" % "commons-io" % "2.8.0",
-  "commons-codec" % "commons-codec" % "1.15"
+  "commons-codec" % "commons-codec" % "1.15",
+  "org.apache.commons" % "commons-collections4" % "4.4"
 )
 // Clear default resolvers.
 appResolvers := None
--- a/managed/src/main/java/com/yugabyte/yw/common/KubernetesManager.java
+++ b/managed/src/main/java/com/yugabyte/yw/common/KubernetesManager.java
@@ -2,13 +2,17 @@
 
 package com.yugabyte.yw.common;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
+import com.yugabyte.yw.common.Util;
 import com.yugabyte.yw.models.Provider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Singleton;
+
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -133,6 +137,48 @@ public class KubernetesManager {
             namespace,
             "-o",
             "jsonpath=" + SERVICE_INFO_JSONPATH);
+    return execCommand(config, commandList);
+  }
+
+  public JsonNode getNodeInfos(Map<String, String> config) {
+    ShellResponse response = runGetNodeInfos(config);
+    if (response.code != 0) {
+      String msg = "Unable to get node information";
+      if (!response.message.isEmpty()) {
+        msg = String.format("%s: %s", msg, response.message);
+      }
+      throw new RuntimeException(msg);
+    }
+    return Util.convertStringToJson(response.message);
+  }
+
+  public ShellResponse runGetNodeInfos(Map<String, String> config) {
+    List<String> commandList = ImmutableList.of("kubectl", "get", "nodes", "-o", "json");
+    return execCommand(config, commandList);
+  }
+
+  public JsonNode getSecret(Map<String, String> config, String secretName, String namespace) {
+    ShellResponse response = runGetSecret(config, secretName, namespace);
+    if (response.code != 0) {
+      String msg = "Unable to get secret";
+      if (!response.message.isEmpty()) {
+        msg = String.format("%s: %s", msg, response.message);
+      }
+      throw new RuntimeException(msg);
+    }
+    return Util.convertStringToJson(response.message);
+  }
+
+  // TODO: disable the logging of stdout of this command if possibile,
+  // as it just leaks the secret content in the logs at DEBUG level.
+  public ShellResponse runGetSecret(
+      Map<String, String> config, String secretName, String namespace) {
+    List<String> commandList = new ArrayList<String>();
+    commandList.addAll(ImmutableList.of("kubectl", "get", "secret", secretName, "-o", "json"));
+    if (namespace != null) {
+      commandList.add("--namespace");
+      commandList.add(namespace);
+    }
     return execCommand(config, commandList);
   }
 
--- a/managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
+++ b/managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
@@ -4,7 +4,10 @@ package com.yugabyte.yw.controllers;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.inject.Inject;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableList;
 import com.typesafe.config.Config;
 import com.yugabyte.yw.cloud.AWSInitializer;
 import com.yugabyte.yw.cloud.AZUInitializer;
@@ -26,6 +29,8 @@ import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiImplicitParam;
 import io.swagger.annotations.ApiImplicitParams;
 import io.swagger.annotations.ApiOperation;
+import org.apache.commons.collections4.MultiValuedMap;
+import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import play.api.Play;
@@ -83,6 +88,10 @@ public class CloudProviderController extends AuthenticatedController {
   @Inject private play.Environment environment;
 
   @Inject CloudAPI.Factory cloudAPIFactory;
+
+  @Inject KubernetesManager kubernetesManager;
+
+  @Inject play.Configuration appConfig;
 
   /**
    * GET endpoint for listing providers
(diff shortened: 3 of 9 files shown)


managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
```diff
@@ -415,6 +425,115 @@ private void createKubernetesInstanceTypes(Provider provider, UUID customerUUID)
...
+  public Result getKubernetesConfigs(UUID customerUUID) {
+    try {
+      Map<String, Set<String>> regionToAZ = getKubernetesRegionToZoneInfo();
+      String storageClass = appConfig.getString("yb.kubernetes.storageClass");
+      String pullSecretName = appConfig.getString("yb.kubernetes.pullSecretName");
+      if (storageClass == null || pullSecretName == null) {
+        LOG.error("Required configuration keys from yb.kubernetes.* are missing.");
+        return ApiResponse.error(INTERNAL_SERVER_ERROR, "Required configuration is missing.");
+      }
+      String pullSecretJson = getKubernetesPullSecretContent(pullSecretName);
```
iSignal, line 441: minor: would be better to call this just pullSecret here, it isn't a json object any more

managed/src/main/java/com/yugabyte/yw/common/KubernetesManager.java
```diff
@@ -105,6 +107,24 @@ public ShellResponse getServiceIPs(Map<String, String> config,
     return execCommand(config, commandList);
   }
 
+  public ShellResponse getNodeInfos(Map<String, String> config) {
```
iSignal, line 110: would be better to make this command return json directly instead of a ShellResponse object, which isn't very meaningful to the caller.

managed/src/main/java/com/yugabyte/yw/common/KubernetesManager.java
```diff
@@ -105,6 +107,24 @@ public ShellResponse getServiceIPs(Map<String, String> config,
...
+    List<String> commandList = ImmutableList.of("kubectl", "get", "nodes", "-o", "json");
+    return execCommand(config, commandList);
+  }
+
+  // TODO(bhavin192): disable the logging of stdout of this command if
+  // possibile, as it just leaks the secret content in the logs at
+  // debug level.
+  public ShellResponse getSecret(Map<String, String> config, String secretName, String namespace) {
```
iSignal, line 118: same comment: let's return the jsonnode directly to the caller (or String) instead of a ShellResponse object. we can throw runtime exceptions or YWServiceException on failure to let the failure bubble up to the API level directly.

managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
```diff
@@ -415,6 +425,115 @@ private void createKubernetesInstanceTypes(Provider provider, UUID customerUUID)
...
+  // information as a JSON, which is similar to the one which is
+  // passed to the createKubernetes method.
+  public Result getKubernetesConfigs(UUID customerUUID) {
+    try {
+      Map<String, Set<String>> regionToAZ = getKubernetesRegionToZoneInfo();
+      String storageClass = appConfig.getString("yb.kubernetes.storageClass");
+      String pullSecretName = appConfig.getString("yb.kubernetes.pullSecretName");
+      if (storageClass == null || pullSecretName == null) {
+        LOG.error("Required configuration keys from yb.kubernetes.* are missing.");
+        return ApiResponse.error(INTERNAL_SERVER_ERROR, "Required configuration is missing.");
```
iSignal, line 439: minor: throw YWServiceException instead. This is automatically handled by YWErrorHandler for all API requests.

managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
```diff
@@ -415,6 +425,115 @@ private void createKubernetesInstanceTypes(Provider provider, UUID customerUUID)
...
+        LOG.debug("value of the zone or region label is empty for " +
+                 nodeInfo.path("metadata").path("name").asText() + ", skipping.");
+        continue;
+      }
+      Set<String> zones = regionToAZ.get(region) == null ? new HashSet<String>() :
+          regionToAZ.get(region);
+      zones.add(zone);
+      regionToAZ.put(region, zones);
+    }
+    if (regionToAZ.isEmpty()) {
```
iSignal, line 499: minor: would be better to do this check at the caller, that way all the error handling is at that level

managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
```diff
@@ -415,6 +425,115 @@ private void createKubernetesInstanceTypes(Provider provider, UUID customerUUID)
...
+          zoneData.config = ImmutableMap.of("STORAGE_CLASS", storageClass);
+          regionData.zoneList.add(zoneData);
+        }
+        formData.regionList.add(regionData);
+      }
+
+     ObjectMapper mapper = new ObjectMapper();
+     return ApiResponse.success(mapper.valueToTree(formData));
+   } catch (RuntimeException e) {
+      return ApiResponse.error(INTERNAL_SERVER_ERROR, e.getMessage());
```
iSignal, line 467: YWServiceException instead

iSignal (commented):
Looks good! Do update the Test Plan for the PR description with a sample response and the test environment.

Most of the comments are about code organization.

managed/src/main/java/com/yugabyte/yw/common/KubernetesManager.java
bhavin192: The main reason to keep it that way was to make it consistent with rest of the code in this file. That way we can re-use the same function somewhere else if we have to. For example, using the getSecret at some other place which might have different way to handle the errors, than the current caller (which we can still do by catching exceptions). What do you think?

managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
```diff
@@ -415,6 +425,115 @@ private void createKubernetesInstanceTypes(Provider provider, UUID customerUUID)
...
+        nodeLabels.path("failure-domain.beta.kubernetes.io/region").asText() : region;
+      String zone = nodeLabels.path("topology.kubernetes.io/zone").asText();
+      zone = zone.isEmpty() ?
+        nodeLabels.path("failure-domain.beta.kubernetes.io/zone").asText() : zone;
+      if (region.isEmpty() || zone.isEmpty()) {
+        LOG.debug("value of the zone or region label is empty for " +
+                 nodeInfo.path("metadata").path("name").asText() + ", skipping.");
+        continue;
+      }
+      Set<String> zones = regionToAZ.get(region) == null ? new HashSet<String>() :
```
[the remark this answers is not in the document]
bhavin192: This is nice. But I think currently we don't have org.apache.commons.collections4 as a dependency. Should I add it and use MultiValuedMap?

managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
bhavin192: Yes, initially I had it there only. Moved it back to the caller.

managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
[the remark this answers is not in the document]
iSignal: Yes, feel free to add it

managed/src/main/java/com/yugabyte/yw/common/KubernetesManager.java
iSignal: Agree that the rest of the code returns ShellResponse but that is not ideal too. It should be ok to return a JsonNode directly and throw exceptions to communicate this to the caller. One option is to have two methods getNodeInfos -> JsonNode which calls the runGetNodeInfos -> ShellResponse

bhavin192: I have made the changes. Let me know how do you find them.

iSignal (approved):
Looks good! Sorry for the delay, github PRs don't get put back in the "review" list unless you re-request the review.

managed/build.sbt
```diff
@@ -168,7 +168,8 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-core" % "2.10.5",
   "com.jayway.jsonpath" % "json-path" % "2.4.0",
   "commons-io" % "commons-io" % "2.8.0",
-  "commons-codec" % "commons-codec" % "1.15"
+  "commons-codec" % "commons-codec" % "1.15",
+  "org.apache.commons" % "commons-collections4" % "4.4"
```
sb-yb, line 172: We already use guava that has all the necessary collection required by this change. 
Can you please use guava and not add yet another dependency?
@bhavin192

bhavin192: Sure, I wasn't aware of that. I will work on removing that and using guava. I will create a new PR shortly.

Issue #6681, opened 2020-12-16
[PLATFORM UI] Connect button for any pod on pods page doesn't show command to login into the pod


nmantrav2:
- Go to pods page of an universe on k8s portal
- Click on Actions pull down menu and click on connect button

I expected to see a command that can be used to connect to the pod but it didn't seem to be doing anything.

PR #8404, opened 2021-05-12
[Platform UI] Fixed connect button for K8s-based universes.

baba230896:
### Summary

**Issues:** 
- The Connect button wasn't working for the Kubernetes-based universe due to the unavailability of access keys. It works fine if the YW has one k8s based universe & a VM-based cloud-config provider (GCP, AWS). The condition checks for the following parameters - access keys & node IPs.
- YW appends the zone name with pod name in multi-zone deployment so the kubectl command created that has appended name. If the user copies it & runs then it won't work.

**Fix:** 
- Conditional check has been updated & now it has one more condition to check the universe type.
- Extract pod name from the appended nodeName.

### Test Plan

- Spin up the YW UI locally & connected the local UI to the YW container deployed on the GKE cluster.
image

fixes: https://github.com/yugabyte/yugabyte-db/issues/6681

### DIFF
--- a/managed/ui/src/components/universes/NodeDetails/NodeConnectModal.js
+++ b/managed/ui/src/components/universes/NodeDetails/NodeConnectModal.js
@@ -42,7 +42,8 @@ class NodeConnectModal extends Component {
 
     if (
       (isEmptyObject(nodeIPs) || currentRow.cloudInfo.cloud !== 'kubernetes') &&
-      !getPromiseState(accessKeys).isSuccess()) {
+      !getPromiseState(accessKeys).isSuccess()
+    ) {
       return <MenuItem>{label}</MenuItem>;
     }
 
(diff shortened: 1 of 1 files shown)


managed/ui/src/components/universes/NodeDetails/NodeConnectModal.js
```diff
@@ -40,7 +40,9 @@ class NodeConnectModal extends Component {
     let accessCommand = null;
     let accessTitle = null;
 
-    if (isEmptyObject(nodeIPs) || !getPromiseState(accessKeys).isSuccess()) {
+    if (
+      (isEmptyObject(nodeIPs) || currentRow.cloudInfo.cloud !== 'kubernetes') &&
+      !getPromiseState(accessKeys).isSuccess()) {
```
sshev, line 45:
```suggestion
      !getPromiseState(accessKeys).isSuccess()
    ) {
```

managed/ui/src/components/universes/NodeDetails/NodeConnectModal.js
```diff
@@ -52,7 +54,7 @@ class NodeConnectModal extends Component {
     if (currentRow.cloudInfo.cloud === 'kubernetes') {
       accessTitle = 'Access your pod';
       const podNamespace = currentRow.privateIP.split(".")[2];
-      accessCommand = `kubectl exec -it -n ${podNamespace} ${currentRow.name} -- sh`;
+      accessCommand = `kubectl exec -it -n ${podNamespace} ${currentRow.name.split("_")[0]} -- sh`;
```
sshev, line 57:
```suggestion
      accessCommand = `kubectl exec -it -n ${podNamespace} ${currentRow.name.split('_')[0]} -- sh`;
```

sshev (approved):

PR #8425, opened 2021-05-12
[#6586][Platform] Fix resource leaks

### LINKED ISSUE #6586: [Platform] Leaks of resources
1. UniverseTaskBase.java, isServerAlive(), absent ybService.closeClient();
2. Universe.java, getMasterLeader(), it is better to put closeClient() into the `finally` block;
3. SetFlagInMemory.java, run(), closeClient() could be missed in case of any exceptions raised after getClient() and before closeClient();
4. WaitForServerReady.java, run(), it is better to put closeClient() into the `finally` block. Yes, it is almost impossible that ServerSubTaskBase.getHostPort() will throw an exception but actually this probability is not 0;
5. SessionController.java, getLogs(), Resource leak: 'reader' is never closed;
6. SwamperHelper.java, writeTargetJsonFile(), the file may stay opened (in case of file.write of file.flush throws an exception);
7. CertificateHelper.java, createRootCA(), Resource leak: 'certWriter' is never closed, Resource leak: 'keyWriter' is never 
 closed.

jackstenglein:
Fixes #6586.

Closes leaked resources listed in the issue.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UniverseTaskBase.java
+++ b/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UniverseTaskBase.java
@@ -1272,7 +1272,9 @@ public abstract class UniverseTaskBase extends AbstractTaskBase {
 
     HostAndPort hp = HostAndPort.fromParts(node.cloudInfo.private_ip,
         server == ServerType.MASTER ? node.masterRpcPort : node.tserverRpcPort);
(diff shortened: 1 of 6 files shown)


SergeyPotachev:
@jackstenglein 
Sorry, missed this PR. Some of these places are already fixed.
For other places + a few more findings @nkhogen has prepared another pull request partly based on your fix.
Thank you for your participation!

Issue #7405, opened 2021-02-26
Add GFlag validators for ensuring correct values for dependent flags.


amitanandaiyer:
* leader_lease_interval_ms needs to be at least raft_heartbeat_interval_ms

* For backfill we need  grace_margin_ms < 1000 * batch_size / rate_per_sec

will be good to enforce these in code using gflag validators so that customers do not set these incorrectly.

amitanandaiyer:
Not needed after #7547

PR #8428, opened 2021-05-12
[yugabyte#1348] Add validator for Leader lease to be larger than heartbeat interval

### LINKED ISSUE #1348: Set Leader lease to be larger than the heartbeat interval
Jira Link: [DB-1493](https://yugabyte.atlassian.net/browse/DB-1493)
Ran into  a perf issue on a cluster where the heartbeat interval was increased to be larger than the leader lease. This caused a few tserver rpc's to fail because the leader did not have a lease, and caused unnecessary retrys, leading to a higher latency.

For the heartbeat mechanism to successfully renew the lease and ensure that the leader has a continous lease : whenever we change one of `raft_heartbeat_interval_ms` or `leader_lease_duration_ms`; we should ensure that 
```
 raft_heartbeat_interval_ms < leader_lease_duration_ms
```

Couple of options here:
 

1. Add a gflag validator method https://gflags.github.io/gflags/#validate to ensure that the requirement is satisfied whenever one of the flags changes.
2. Consider computing one as a function of other. So, if one changes, the other is automatically updated.

Preferred approach: Use the approach taken by https://github.com/yugabyte/yugabyte-db/commit/9c0f84ec8fa386b86a4adc39471489677e892e66

[DB-1493]: https://yugabyte.atlassian.net/browse/DB-1493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

saumyaprakash:
Uses gflag validators to ensure that `leader_lease_duration_ms` will be greater than `raft_heartbeat_interval_ms` 
Fixes #1348

### DIFF
--- a/src/yb/consensus/raft_consensus.cc
+++ b/src/yb/consensus/raft_consensus.cc
@@ -81,11 +81,24 @@
 using namespace std::literals;
 using namespace std::placeholders;
 
+namespace{
+
+static bool HeartbeatIntervalLeaderLeaseDurationValidator(const char* flagname, int32_t value) {
+	if(value > FLAGS_leader_lease_duration_ms) {
+		LOG(INFO) << "Expect " << flagname << " to be less than leader_lease_duration_ms";
+		return false;
+	}
+	return true;
+}
+
+} // namespace
+
 DEFINE_int32(raft_heartbeat_interval_ms, yb::NonTsanVsTsan(500, 1000),
              "The heartbeat interval for Raft replication. The leader produces heartbeats "
              "to followers at this interval. The followers expect a heartbeat at this interval "
              "and consider a leader to have failed if it misses several in a row.");
 TAG_FLAG(raft_heartbeat_interval_ms, advanced);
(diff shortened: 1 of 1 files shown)


amitanandaiyer (changes requested):
Hi Saumyaprakash,
   Thanks a lot for jumping onto this issue!!

I think it would be better if we try to automatically set the value for a dependent flag, rather than enforce a validator across 2 flags.

   We tried a similar approach recently (https://github.com/yugabyte/yugabyte-db/issues/7405) and it leads to unexpected crashes; depending on the order in which the two inter-dependent flags get updated. So `--A=a --B=b` may work; but `--B=b --A=a` will not work when using the validators. 
   
   Probably a better approach is to default one of them to -1 and automatically the value for it, based on the other flag. See https://github.com/yugabyte/yugabyte-db/commit/a30c1f203ea70f823b362263b4e374216d79b349 for example.
   
   Thanks,
   -Amit

Issue #8481, opened 2021-05-17
[Platform] Disable Bulk delete backup option for "In-Progress" backups.


gaurav061:
Description: Try taking a backup on any universe, hit browser reload soon after you take the backup, Try selecting the backup with the status `InProgress`. Deleting such backups will cause failure. 
![Screenshot from 2021-05-17 12-42-23](https://user-images.githubusercontent.com/60691495/118488017-f025e900-b738-11eb-8cfe-1e438039e3a7.png)

Backup select should be disabled for `InProgress` backups

gaurav061:
PR: https://github.com/yugabyte/yugabyte-db/pull/8482

PR #8482, opened 2021-05-17
[#8481] [Platform] Added check to disable backups if it is in progress.

gaurav061:
Description: The user is able to mark the `in-progress backups for bulk delete, which will eventually fail. Added a check to keep the select in the disabled state if the backup status in `in-progress.
Screenshot from 2021-05-17 18-08-03


nishantSharma459 (approved):

managed/ui/src/components/tables/ListBackups/ListBackups.js
```diff
@@ -385,6 +392,13 @@ export default class ListBackups extends Component {
       })
       .filter(Boolean);
 
+    // Check for in-progress backups and mark them as unselctable.
+    selectRowProp.unselectable = backupInfos
+      .filter((backup) => {
+        return backup.status === 'InProgress';
+      })
+      .map((data) => data.backupUUID);
```
sshev, line 400: just for consistency this would be better
```suggestion
      .filter((backup) => backup.status === 'InProgress')
      .map((data) => data.backupUUID);
```

sshev (approved):

managed/ui/src/components/tables/ListBackups/ListBackups.js
gaurav061: Done

Issue #8461, opened 2021-05-14
[Platform][UI] User not redirected to the original URL after signing in


streddy-yb:
Platform power user feedback: "_When I click on a platform url, but I have to sign in first, I’m then redirected to the home page, rather than to the URL I initially clicked. Any cheap way we can fix that with some referral magic?_"

streddy-yb:
hi @gaurav061 - can you please back port this to 2.6? thanks

gaurav061:
Done @streddy-yb. Thanks.

PR #8520, opened 2021-05-19
[#8461] [Platform] Added fix for redirecting user to intended url.

gaurav061:
Description: Platform power user feedback: "When I click on a platform URL, but I have to sign in first, I’m then redirected to the home page, rather than to the URL I initially clicked. Any cheap way we can fix that with some referral magic?"

Fix uses search param to pass the redirect page URL.

Test Plan: 
1. Navigate to a valid URL, copy it from the URL bar, and log out from the active session. Now hit the copied URL you will be redirected to the login page and once the valid credential is entered you will land on the original URL instead of the default landing page.
2. Navigate to a valid URL other than `/` and clear the browser cookies and local storage details, so it expires the session, now try to navigate to some other URL, you will be redirected to the login page and once the valid credential is entered you will land on the original URL instead of the default landing page.


managed/ui/src/routes.js
```diff
@@ -121,6 +121,10 @@ function validateSession(store, replacePath, callback) {
   const userId = Cookies.get('userId') || localStorage.getItem('userId');
   const customerId = Cookies.get('customerId') || localStorage.getItem('customerId');
   if (_.isEmpty(customerId) || _.isEmpty(userId)) {
+    var location = window.location.pathname;
```
nishantSharma459, line 124: let.

managed/ui/src/routes.js
```diff
@@ -121,6 +121,10 @@ function validateSession(store, replacePath, callback) {
...
+    if (location !== '/') {
+      localStorage.setItem('pathToRedirect', location);
```
nishantSharma459, line 126: can we use `initialPath` instead of  `pathToRedirect`?

managed/ui/src/routes.js
gaurav061: Done

managed/ui/src/routes.js
gaurav061: For now, I am leaving this as is, let's hear from others.

nishantSharma459: Cool. Just keep it in mind.

nishantSharma459 (approved):

jaydeepkumara (approved):

managed/ui/src/routes.js
sshev: actually, that is the last visited page before possible the redirect to the login page, so smth like `lastVisitedPage` better reflects its value

managed/ui/src/routes.js
```diff
@@ -121,6 +121,10 @@ function validateSession(store, replacePath, callback) {
   const userId = Cookies.get('userId') || localStorage.getItem('userId');
   const customerId = Cookies.get('customerId') || localStorage.getItem('customerId');
   if (_.isEmpty(customerId) || _.isEmpty(userId)) {
+    let location = window.location.pathname;
+    if (location !== '/') {
```
sshev, line 125: I'd add check for `/login` to avoid possible circular redirects from `/login` to `/login`

managed/ui/src/routes.js
sshev: One more note - I'd suggest using `sessionStorage` to avoid redirecting to stale URL as localStorage persists value indefinitely

sshev (changes requested):
Add detailed test plan pls

managed/ui/src/routes.js
gaurav061: Yeah, I will change that, but for the same reason I am clearing it out once it is redirected to the intended page.

managed/ui/src/routes.js
```diff
@@ -121,6 +121,10 @@ function validateSession(store, replacePath, callback) {
   const userId = Cookies.get('userId') || localStorage.getItem('userId');
   const customerId = Cookies.get('customerId') || localStorage.getItem('customerId');
   if (_.isEmpty(customerId) || _.isEmpty(userId)) {
+    let location = window.location.pathname;
```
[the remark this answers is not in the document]
gaurav061: Done

managed/ui/src/routes.js
gaurav061: I didn't see that while testing, but there is no harm in having that check so, done

gaurav061:
I'll add a detailed test plan in a while.

managed/ui/src/routes.js
```diff
@@ -121,6 +121,10 @@ function validateSession(store, replacePath, callback) {
   const userId = Cookies.get('userId') || localStorage.getItem('userId');
   const customerId = Cookies.get('customerId') || localStorage.getItem('customerId');
   if (_.isEmpty(customerId) || _.isEmpty(userId)) {
+    const location = window.location.pathname;
```
mjoshi0923, line 124: We should use queryParam to define redirect URL post login. When multiple tabs open local storage doesn't give you consistent behaviour and it creates buggy interaction.

managed/ui/src/routes.js
sunilsutarDCE: My suggestion is to put intented url in querystring , for e.g.  http://localhost:3000/login?redirectUrl=<intendedUrl>

mjoshi0923: +1 to @sunilsutarDCE @nishantSharma459 Look at my previous comment, we should use redirectURL or something similar in the query parameter.

gaurav061: Sure @sunilsutarDCE @mjoshi0923 I will do the needful.

sunilsutarDCE:
I am considering one more use case here that should be checked :
Suppose user is on universe detail page and accessToken expired then it should redirect login with redirect Url of that universe detail page and after login it should redirect to universe detail page . 
I have managed find angular e.g. check [here](https://stackblitz.com/edit/route-session-angular-atmm8q)

gaurav061:
@sunilsutarDCE, this flow will cover the mentioned use case, please refer to point 2 on the test plan.

managed/ui/src/routes.js
```diff
@@ -143,7 +149,10 @@ function validateSession(store, replacePath, callback) {
       }
     });
     store.dispatch(customerTokenError());
-    browserHistory.push('/login');
+    location && location !== '/'
+      ? browserHistory.push(`/login?redirectUrl=${location}`)
+      : browserHistory.push('/login');
+    // location ? browserHistory.push(`/login?redirectUrl=${location}`) : browserHistory.push('/login');
```
sunilsutarDCE, line 155: Please remove commented code

gaurav061: Sorry, Done

managed/ui/src/routes.js
gaurav061: Done.

managed/ui/src/routes.js
```diff
@@ -63,7 +65,9 @@ export const clearCredentials = () => {
   Cookies.remove('authToken');
   Cookies.remove('customerId');
   Cookies.remove('userId');
-  browserHistory.push('/');
+  pathToRedirect
```
mjoshi0923, line 68: clearCredential is not right place to clear queryParam, minor comment. If you can please either rename the function or pull this snippet in another function.

mjoshi0923 (approved):

managed/ui/src/routes.js
gaurav061: I am not clearing the `queryParams` in `clearCredential` function, but it's a check that we should load the page with params intact if we have the query param set previously. Actual clear logic is in `validateSession` function once it redirects.

gaurav061: The reason for this check is because the login component is calling clear credential and thus this will preserve the `queryParam`

mjoshi0923: Cool

mjoshi0923 (approved):

mjoshi0923 (approved):

Issue #8083, opened 2021-04-19
[Platform] UI: Onprem number of nodes check is not happening when configuring read replica


hkandala:
Steps to reproduce:
1. Configure onprem provider and add instances (in this case only 1 instance is added)
2. Go to create universe page and select the onprem provider
3. Because only 1 instance is configured, selected number of nodes as 1 <br/><br/><img width="1788" alt="Screenshot 2021-04-19 at 3 12 48 PM" src="https://user-images.githubusercontent.com/9005092/115218034-bfe03000-a123-11eb-92c7-d1973c6001fc.png">
4. Click on Configure Read Replica. Ideally UI should block here when selecting nodes because all onprem nodes are used in the Primary cluster. But UI allows to add 1 node for read replica too <br/><br/><img width="1787" alt="Screenshot 2021-04-19 at 3 13 22 PM" src="https://user-images.githubusercontent.com/9005092/115218524-4432b300-a124-11eb-82e0-7974c6f3e287.png">

Even though create universe fails in the backend with this configuration, it would be more intuitive if UI limits the number of nodes in read replica based on selected number of nodes in primary

nishantSharma459:
Putting this on hold. Before this, I need to resolve the ticket: [#8407]

jaydeepkumara:
As per discussion. We are planning to revamp create the universe flow. We need to take care of this bug while revamping.

PR #8522, opened 2021-05-19
[#8083] [Platform] Added nodes per region and set the error message accordingly.

nishantSharma459:
Description: 
While creating a universe for the on-prem provider if you select a region then the error message `Not Enough Nodes Configured` is not showing up when the selected number of nodes surpasses the limit of the available number of nodes.

Test Plan:

1. Configure the on-prem provider with 2 regions.
    Region1 -> 3 nodes
    Region2 -> 2 nodes
    Total nodes -> 5 nodes

2. While creating the universe we select the provider as on-prem and select Region1 which has 3 nodes available.

3. If you try increasing the node and exceeds the total available nodes for the respective region, it is expected to show the error message `Not Enough Nodes Configured`, which is not happening.


managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -492,16 +493,24 @@ export default class ClusterFields extends Component {
       this.props.fetchUniverseResources(nextProps.universe.universeConfigTemplate.data);
     }
     // If nodeInstanceList changes, fetch number of available nodes
-    if (
-      getPromiseState(nodeInstanceList).isSuccess() &&
-      getPromiseState(this.props.cloud.nodeInstanceList).isLoading()
-    ) {
-      let numNodesAvailable = nodeInstanceList.data.reduce((acc, val) => {
+    if (getPromiseState(nodeInstanceList).isSuccess()) {
+      let nodesPerRegion = [];
```
sshev, line 497: should be `const` as it doesn't reassign anywhere

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -492,16 +493,24 @@ export default class ClusterFields extends Component {
       this.props.fetchUniverseResources(nextProps.universe.universeConfigTemplate.data);
     }
     // If nodeInstanceList changes, fetch number of available nodes
-    if (
-      getPromiseState(nodeInstanceList).isSuccess() &&
-      getPromiseState(this.props.cloud.nodeInstanceList).isLoading()
```
sshev, line 497: is this condition removed intentionally?

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -492,16 +493,24 @@ export default class ClusterFields extends Component {
...
-    ) {
-      let numNodesAvailable = nodeInstanceList.data.reduce((acc, val) => {
+    if (getPromiseState(nodeInstanceList).isSuccess()) {
+      let nodesPerRegion = [];
+      nodeInstanceList.data.forEach((node) => {
+        isNonEmptyArray(this.state.regionList) &&
+          this.state.regionList.forEach((region) => {
```
sshev, line 500: seems `isNonEmptyArray()` is exessive as for an empty array `forEach()` won't perform an iteration anyway

sshev (changes requested):
I'd really like to see a unit tests for this change as it's unobvious how to validate if it fixes the issue and doesn't introduce a regression

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
nishantSharma459: Done.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
nishantSharma459: once you select a region for the primary cluster and hit the `configure read replica` button then the region list is undefined so to avoid that I've used `isNonEmptyArray()`.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
nishantSharma459: Yeah. Actually, I think we don't need this `isLoading()` here in `willRecieveProps`. This condition will get fail if you come back to the `primary` cluster from `read replica`.  And now we're setting up `maxNodes` based on region selection so I believe it's fine. 

Please lemme know if this needs to be corrected.

nishantSharma459:
Putting this on hold. As per the discussion, the new UX discussion is going on and I believe in that we gonna take care of issues related to the AZ options.

PR #8544, opened 2021-05-20
[Docs] fix links

ddhodge:
Fixed broken links

### DIFF
--- a/docs/content/latest/explore/security/security.md
+++ b/docs/content/latest/explore/security/security.md
@@ -41,13 +41,13 @@ You can choose the method to use to authenticate a particular client connection 
 
 YugabyteDB stores authentication credentials internally in the YB-Master system tables. The authentication mechanisms available to clients depend on what is supported and exposed by the YSQL, YCQL, and YEDIS APIs.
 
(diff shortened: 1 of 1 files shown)


docs/content/latest/explore/security/security.md
```diff
@@ -41,13 +41,13 @@ You can choose the method to use to authenticate a particular client connection
 
 YugabyteDB stores authentication credentials internally in the YB-Master system tables. The authentication mechanisms available to clients depend on what is supported and exposed by the YSQL, YCQL, and YEDIS APIs.
 
-Read more about [how to enable authentication in YugabyteDB](../../secure/authentication).
+Read more about [how to enable authentication in YugabyteDB](/latest/secure/authentication).
```
polarweasel, line 44: Avoid absolute links whenever you can. They'll break later when this becomes a numbered version.

polarweasel (changes requested):
No absolute links!

docs/content/latest/explore/security/security.md
polarweasel: (Each link just needs one more `../` in it and they should all work)

ddhodge: fixed

PR #8545, opened 2021-05-20
[docs] Update airgapped.md

kerneljake:
fix non sequitur


polarweasel:
@kerneljake in this case, this isn't a non-sequitur, but a style choice. Saying "as follows" makes it very clear that what comes next is what you need to do.

Issue #8324, opened 2021-05-06
[Platform] when the universe is paused, disable query monitoring /reset stats option on slow queries


VijiYB:
When the universe is paused, disable "Query Monitoring" and "Reset Stats" option from Slow Queries tab. Though the functionality will not work as the nodes are down, still its better to disable as it doesn't makes sense to keep in enabled.

PR #8576, opened 2021-05-24
[#8324] [Platform] Disabled the query monitoring under Queries tab.

nishantSharma459:
**Description:** 
When the Universe is in a `paused state` the users will not perform any actions for that respective universe. But for now, when the users go under the `Queries tab` they can perform the actions like `query monitoring and reset stats`. This is not supposed to happen.

**Test Plan:**

1. Go to the universe page.
2. Hit the Action button next to connect and hit Pause universe.
3. Once the universe is paused then hit the Queries tab for that universe.
4. Click on Show queries. 
5. Check whether Reset stats and Query Monitoring is present or not.

If the above-mentioned buttons are not present then it's working fine. Check for both the Universes:
1. For Pasued Universe - Buttons are not supposed to be there on the UI.
2. For Normal Universe(Ready State) - Buttons are supposed to be there.


yb-andrew (approved):
LG

PR #8582, opened 2021-05-24
[docs] Query tuning intro

ddhodge:
Intro to query tuning section
Preview https://deploy-preview-8582--infallible-bardeen-164bc9.netlify.app/latest/explore/query-1-performance/query-tuning-intro/


docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+    Seq Scan on contacts (cost=0.00..105.00 rows=1000 width=0)
+        Filter: ((NOT is_paid) AND (updated_at > '2021-04-12 12:00:00'::timestamp without time zone) AND (account_id =1234))
+(3 rows)
+```
+
+In each case, the queries do a `Seq Scan` (sequential scan) on the tables. This operation requires scanning the entire table to retrieve the desired columns. Even though we are using the partition keys to do the look up, it still needs to do a lot of scanning. 
+
+Generally, you want to avoid `SELECT COUNT(*)` queries as they can require a full scan of the table to get the results. This can cause query degradation and in some cases cause the query to not return at all.
+
+Because most of the queries above use `account_id` as the main qualifier, we can avaoid a sequential scan by creating a direct index on that column, and then use the `INCLUDE` feature to cover the other columns that we also want in the index. Indexing is a powerful tool that can speed up queries with higher latencies. When creating an index, we want to consider the column cardinality, as well as the different index types.
```
polarweasel, line 177:
```suggestion
Because most of the queries above use `account_id` as the main qualifier, you can avoid a sequential scan by creating a direct index on that column, and then using the `INCLUDE` feature to cover the other columns that you also want in the index. Indexing is a powerful tool that can speed up queries with higher latencies. When creating an index, consider the column cardinality, as well as the different index types.
```

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+
+We create the index as follows:
```
polarweasel, line 179:
```suggestion
Create the index as follows:
```

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+----------------------------------------------
+Limit (cost=128.22..128.28 rows=25 width=1642)
+    Sort (cost=128.22..130.72 rows=1000 width=1642)
+        Sort Key: id DESC
+            Seq Scan on accounts (cost=0.00..100.00 rows=1000 width=1642)
+```
+
+In this case, a sort is run first, which adds extra time to the query, before running a sequential scan of the table, which, as we have seen, also degrades performance.
+
+To optimize this query, we can adjust the sorting of the table by the primary key to be `DESC` rather than `HASH`.
```
polarweasel, line 295:
```suggestion
To optimize this query, adjust the sorting of the table by the primary key to be `DESC` rather than `HASH`.
```

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+
+```sql
+CREATE TABLE public.accounts ( id bigint NOT NULL, ... ,PRIMARY KEY(id desc);
+```
+
+The result is the query no longer does a sequential scan; instead it uses an index scan, cutting execution time from 460ms to 3ms.
```
polarweasel, line 301:
```suggestion
Now, the query no longer does a sequential scan; instead it uses an index scan, cutting execution time from 430ms to 3ms.
```

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+QUERY PLAN
+------------------------------------------
+Limit (cost=0.00..2.85 rows=25 width=1642)
+    Index Scan using accounts_pkey on accounts (cost=0.00..114.00 rows=1000 width=1642)
+
+Time: 2.994 ms
+Previous run time: 426.627 ms
+```
+
+### Optimizing `SELECT` using an index
```
polarweasel, line 318: All of these headings could be imperative. It's a decision we should make across the board sometime, but it works well in this doc.

```suggestion
### Optimize `SELECT` using an index
```

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+
+The following example shows another query running a sequential scan, which we can fix by adding an index.
```
polarweasel, line 320:
```suggestion
The query in the following example runs a sequential scan, which you can fix by adding an index.
```

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+```output
+QUERY PLAN
+----------------------------------------------
+Sort (cost=152.33..154.83 rows=1000 width=237)
+    Sort Key: url
+        Seq Scan on account_type (cost=0.00..102.50 rows=1000 width=237)
+            Filter: (account_id = 6873)
+```
+
+The query runs a sequential scan on the account_type table, so by adding an index on `account_id`, we can prevent the full scan of the table, as follows:
```
polarweasel, line 357:
```suggestion
The query runs a sequential scan on the account_type table, so adding an index on `account_id` prevents the full scan of the table, as follows:
```

polarweasel (changes requested):
This is a _really_ useful addition to the docs! A few tweaks and it's good to go.

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+* I/O intensity
+* time consumption
+* response time
+* memory consumption
+* temporary space consumption
+
+The [`pg_stat_statements`](https://www.postgresql.org/docs/11/pgstatstatements.html) extension module is installed by default, but must be enabled before the `pg_stat_statements` view can be queried.
+
+```sql
+CREATE EXTENSION pg_stat_statements;
```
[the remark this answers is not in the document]
ddhodge: Fixed

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,380 @@
...
+menu:
+  latest:
+    identifier: query-tuning-intro
+    parent: query-tuning
+    weight: 150
+isTocNested: true
+showAsideToc: true
+---
+
+Improving the performance of queries in YugabyteDB typically involves some combination of the following general steps:
```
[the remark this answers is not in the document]
ddhodge: fixed

docs/content/latest/explore/query-1-performance/query-tuning-intro.md
```diff
@@ -0,0 +1,379 @@
...
+```output
+QUERY PLAN
+----------------------------------------------
+Limit (cost=128.22..128.28 rows=25 width=1642)
+    Sort (cost=128.22..130.72 rows=1000 width=1642)
+        Sort Key: id DESC
+            Seq Scan on accounts (cost=0.00..100.00 rows=1000 width=1642)
+```
+
+In this case, a sort is run first, which adds extra time to the query, before running a sequential scan of the table, which, as we have seen, also degrades performance.
```
polarweasel, line 292: One missed "we", gonna just commit this for you :) 

```suggestion
In this case, a sort is run first, which adds extra time to the query, before running a sequential scan of the table, which also degrades performance.
```

polarweasel (approved):

epratt-yb (approved):

polarweasel:
That rebase turned out to be a disaster... I think you're gonna need to close this one and make a new PR with just the files you need in it ☹️

PR #8598, opened 2021-05-25
Add format check for Java code under managed

iSignal:
Run an `sbt javafmtCheck` on PRs and pushes.

Also disable the C++ build for managed sub directory.

Test Plan: https://github.com/iSignal/yugabyte-db/actions/runs/875999659

### DIFF
--- a/.github/workflows/managed-java-checks.yaml
+++ b/.github/workflows/managed-java-checks.yaml
@@ -1,0 +1,38 @@
+name: Lint java code under managed
+
+on:
+  push:
+    branches:
+      - master
+      - '2.6'
+    paths:
+      - .github/workflows/managed-java-checks.yaml
+      - managed/**/*.java
+
+  pull_request:
+    branches:
+      - master
+      - '2.6'
+    paths:
+      - .github/workflows/managed-java-checks.yaml
+      - managed/**/*.java
+
+
+jobs:
+  lint-checks:
+    name: Lint checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install Java
+        uses: actions/setup-java@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/managed-format-java.yaml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/checkout@v2
+
+      - name: Install Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: 8
+
+      - name: Check formatting
+        run: sbt javafmtCheck
```
Arnav15, line 34: `sbt javafmtCheck` just checks the files in the src directory. We need to run `sbt javafmtCheckAll` for including test files.
Also, what are your thoughts on running the formatter explicitly during the push to master?
Also, as a follow up, can we make doing an `sbt compile` as part of our build process? We currently have no checks for our managed code build.

Arnav15 (changes requested):

Arnav15 (approved):

.github/workflows/managed-format-java.yaml
iSignal: It isn't easy to add a compile step actually, because of the client jar. To prepare the client jar, we would need to run ./yb_build.sh --java-only (just mvn isn't sufficient because we need protobuf declarations) and that needs more setup. Let's start with this for now.

.github/workflows/managed-java-checks.yaml
```diff
@@ -0,0 +1,38 @@
+name: Lint java code under managed
+
+on:
+  push:
+    branches:
+      - master
+      - "2.6"
```
Arnav15, line 7: Don't know if double quote or single quote matters, but for consistency, maybe use single quotes?

sb-yb (approved):

PR #8617, opened 2021-05-25
[docs] PR for Document Indexes and Constraints in Explore (Doc8451)

lizayugabyte:
Build preview: https://deploy-preview-8617--infallible-bardeen-164bc9.netlify.app/latest/explore/indexes-constraints/


lizayugabyte:
Asked (via slack) @m-iancu to review this doc on May 25, 2021

docs/content/latest/explore/indexes-constraints/constraints.md
```diff
@@ -0,0 +1,298 @@
...
+- [Table with Primary Key](https://docs.yugabyte.com/latest/api/ysql/the-sql-language/statements/ddl_create_table/#table-with-primary-key)
+- [Primary Keys in PostgreSQL documentation](https://www.postgresql.org/docs/12/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS)
+
+## Foreign Key
+
+A foreign key represents one or more columns in a table referencing the primary key of another table.
+
+Tables can have multiple foreign keys.
+
+To define a foreign key, you use the foreign key constraint which enables you to maintain the referential integrity of data between two tables: values in columns in one table equal the values in columns in another table.
```
m-iancu, line 110: > To define a foreign key, you use the foreign key

This reads a bit off, can we rephrase. Maybe along the lines of
`Foreign key constraints are used to maintain referential integrity ...`

docs/content/latest/explore/indexes-constraints/constraints.md
```diff
@@ -0,0 +1,298 @@
...
+
+For more information and examples, see the following: 
+
+- [Primary Key](https://docs.yugabyte.com/latest/api/ysql/the-sql-language/statements/ddl_create_table/#primary-key)
+- [Table with Primary Key](https://docs.yugabyte.com/latest/api/ysql/the-sql-language/statements/ddl_create_table/#table-with-primary-key)
+- [Primary Keys in PostgreSQL documentation](https://www.postgresql.org/docs/12/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS)
+
+## Foreign Key
+
+A foreign key represents one or more columns in a table referencing the primary key of another table.
```
m-iancu, line 106: I think it can reference either a primary key or unique index (i.e. columns that are restricted with unique index or unique constraint) but let's confirm that by testing.

docs/content/latest/explore/indexes-constraints/constraints.md
```diff
@@ -0,0 +1,298 @@
...
+
+```
+[CONSTRAINT fk_name] 
+  FOREIGN KEY(fk_columns) 
+    REFERENCES parent_table(parent_key_columns)
+    [ON DELETE delete_action]
+    [ON UPDATE update_action]
+```
+
+Defining the `CONSTRAINT` clause and naming the foreign key is optional. If you omit it, an auto-generated name is provided by YSQL. The `REFERENCES` clause specifies the parent table and its columns referenced by the *fk_columns*. Defining actions is also optional; if defined, they determine the behaviors when the primary key in the parent table is deleted or, rarely, updated. YSQL allows you to perform the following actions:
```
m-iancu, line 122: >is deleted or, rarely, updated

I think "rarely" is un-necessary here.

docs/content/latest/explore/indexes-constraints/indexes-1.md
```diff
@@ -0,0 +1,172 @@
...
+
+You create indexes in YSQL using the `CREATE INDEX` statement that has the following syntax:
+
+```
+CREATE INDEX index_name ON table_name(column_list);
+```
+
+*column_list* represents a column or a comma-separated list of several columns to be stored in the index. An index created for more than one column is called a composite index.
+
+You can also create a functional-based index, in which case you would replace *column_list* with an expression. For more information, see [Using Indexes on Expressions](#using-indexes-on-expressions).
```
m-iancu, line 30: >in which case you would replace *column_list* with an expression

I think any of the elements in the column list can be an expression, not just the whole list.
For instance, this is perfectly fine (though definitely contrived).
```
yugabyte=# create index on t((v * 2), (v + 1), (v + 5));
CREATE INDEX
```

Postgres defines it simply as:

```
{ column_name | ( expression ) }
```
which means a repeated sequence (list) of either column_name or (parenthesis-separated) expression.
But I don't know how they summarize it in text.

docs/content/latest/explore/indexes-constraints/indexes-1.md
```diff
@@ -0,0 +1,172 @@
...
+
+If you execute the same `SELECT` query with the `EXPLAIN` statement as in [Creating Indexes](#creating-indexes), the query plan will not include any information about the index. 
+
+## Listing Indexes
+
+YSQL inherits all the functionality of the PostgeSQL `pg_indexes` view that allows you to retrieve a list of all indexes in the database as well as detailed information about every index.
+
+For details, see [pg_indexes](https://www.postgresql.org/docs/12/view-pg-indexes.html) in the PostgreSQL documentation.
+
+## Using the UNIQUE Index
```
m-iancu, line 109: I think "using *a* unique .."  or simply "Using unique indexes" matching the style for the other sections is better.
"the" might suggest there is only one unique index which is not the case.

docs/content/latest/explore/indexes-constraints/constraints.md
lizayugabyte: Rephrased as: 
You use a foreign key constraint to maintain the referential integrity ...

docs/content/latest/explore/indexes-constraints/constraints.md
lizayugabyte: Removed "rarely"

docs/content/latest/explore/indexes-constraints/indexes-1.md
lizayugabyte: Fixed

docs/content/latest/explore/indexes-constraints/indexes-1.md
lizayugabyte: Changed as follows:

You can also create a functional-based index, in which case you would replace any element of *column_list* with an expression. For more information, see [Using Indexes on Expressions](#using-indexes-on-expressions).

docs/content/latest/explore/indexes-constraints/constraints.md
lizayugabyte: Yes, you are right, I tested both the unique index and unique constraint, and the both scenarios work. I'm adding this info to the document.

lizayugabyte:
Thanks for the review, @m-iancu , I implemented your suggestions. Would you like to re-review? Or should I just merge the PR (because the changes are minor)

m-iancu (approved):

PR #8623, opened 2021-05-26
Add python wheel package to requirements files.

### LINKED ISSUE #8622: Build error message with missing python wheel package
Building on ubuntu 20.04 LTS with the command `./yb_build.sh --clean release` or `./yb_build.sh --clean debug` gives the output similar to:

`error: invalid command 'bdist_wheel'` and `command not found, in setup.py`

Likely needs the "wheel" package to be explicitly mentioned in requirements.txt and requirements_frozen.txt

vjr:
Fixes #8622


requirements_frozen.txt
```diff
@@ -1,3 +1,4 @@
+wheel==0.36.2
```
svarnau, line 1: It would be nice to sort these two files and keep them alphabetized, but it isn't currently, so I guess it is not required for this change.

svarnau (commented):
Looks good to me. We should check the wheel install when the build actions finish.

vjr:
> Looks good to me. We should check the wheel install when the build actions finish.

Did you mean I need to check the wheel install after the build actions are done in this PR? If so, could you point me to where this needs to done?

svarnau (approved):

svarnau:
> > Looks good to me. We should check the wheel install when the build actions finish.
> 
> Did you mean I need to check the wheel install after the build actions are done in this PR? If so, could you point me to where this needs to done?

I just meant I was not going to approve until I saw the build logs from the checks tab. They look good to me.
Thanks for the fix.

Issue #8620, opened 2021-05-26
[Platform] Clicking cancel on 'Edit configuration' on backup shows the edited values 


VijiYB:
Backup -> Edit configuration. I edited something in AWS key but I clicked cancel, though the value is not saved, the values I edited are still showing up. Once I reload the page, the edited values are gone and original (saved) values are showed, which is correct. Its good to reload the original values if cancel clicked rather than showing the edited values.

cc @nishantSharma459

shashwat-yb:
Tested the cancel button in Backup -> AWS -> Edit configuration by changing some values in the respective fields and then checking whether it restores the previous credentials when the cancel button is clicked.
The backup page was able to handle this scenario and the previous details were visible once the cancel button is clicked.

PR #8627, opened 2021-05-26
[#8620] [Platform] [Backport-2.4] Reset the values on cancel.

nishantSharma459:
**Description:**
Backup -> Edit configuration. I edited something in the AWS key but I clicked cancel, though the value is not saved, the values I edited are still showing up. Once I reload the page, the edited values are gone and original (saved) values are shown, which is correct. It's good to reload the original values if cancel clicked rather than showing the edited values.

**Test Plan:**

1. Click on config from the side nav.
2. Go to Backup-> AWS.
3. Create a new Backup config.
4. Hit the Edit configuration button.
5. Change some values and hit the Cancel button.

Check if it's restoring the original values or not.


sshev (approved):
LGTM

Issue #8611, opened 2021-05-25
[Platform] Clicking "Edit Configuration" on one backup applies to all Backup tabs


VijiYB:
I clicked "Edit Configuration" on AWS but did not change anything, I navigated to another tab like GCS/Azure, and those storage are automatically went to "edit configuration" mode even though I did not click "Edit Configuration" explicitly.

Repro:
1. Backup -> AWS
2. Click "Edit Configuration"
3. Without canceling/update, click on Azure/GCS
4. You will see that Azure/GCS is in Edit Configuration mode even without me explicitly clicking "Edit configuration"

Expected - By default, the backup config should be in non-editable mode unless click on "Edit Configuration" on that specific tab

cc @nishantSharma459

VijiYB:
Once we click on Edit and move to different menu and come back, the edit state is retained currently. Once we move out of the tab, the state of "edit" should not be retained.

shashwat-yb:
Validated that Backup -> AWS -> Edit Configuration only allows the active tab to change and update values and edit mode for other tabs are disabled. Also verified that this holds correct for other tabs as well.
cc @VijiYB

VijiYB:
@shaharuk-yb - Can you create a new issue for this? - _

> But in case we return back to the same tab it still remains in active edit mode which should me disabled as we moved out of that tab earlier

_

PR #8673, opened 2021-05-27
[#8611] [Platform] [Backport-2.6] Enable edit backup configuration for the active tab only.

nishantSharma459:
**Description:**
When the users will hit the `Edit configuration` button from a particular backup config then the other tabs also enables the editing even though you clicked it or not. So it shouldn't happen. Edit config will only allow editing the backup config of the respective tab. 

**Test Plan:**
1. Click config from the side nav.
2. Hit the backup tab and create backup configs for the respective tabs.
3. Then go to one of the tabs and hit the edit-config button.
4. Without saving it or cancelling it go to another tab and see if edit-config is available or not. If yes then it's a bug otherwise we're good to go.

### DIFF
--- a/managed/ui/src/components/config/Storage/StorageConfiguration.js
+++ b/managed/ui/src/components/config/Storage/StorageConfiguration.js
@@ -92,12 +92,7 @@ class StorageConfiguration extends Component {
     super(props);
 
     this.state = {
-      enableEdit: {
-        s3: false,
-        nfs: false,
-        gcs: false,
-        az: false
-      },
+      editingTab: '',
       iamRoleEnabled: false
     };
   }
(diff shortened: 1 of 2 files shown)


managed/ui/src/components/config/Storage/StorageConfiguration.js
```diff
@@ -92,7 +92,12 @@ class StorageConfiguration extends Component {
     super(props);
 
     this.state = {
-      enableEdit: false,
+      enableEdit: {
+        s3: false,
+        nfs: false,
+        gcs: false,
+        az: false
```
yb-andrew, line 99: Isn't only one of these values ever true at a time? I think it makes more sense to have an `activeTab: string` as the state.

yb-andrew (changes requested):
Seems a little overkill to track every tab. See comment.

managed/ui/src/components/config/Storage/StorageConfiguration.js
nishantSharma459: Yes, only 1 value true at a time. This is just an initial state I defined but I'm using `activeTab` only when I'm setting up the state for the respective tab. You can check the `onEditConfig` method call.

yb-andrew (approved):
LG

PR #8674, opened 2021-05-27
[#8611] [Platform] [Backport-2.4] Enable edit backup configuration for the active tab only. 

nishantSharma459:
**Description:**
When the users will hit the Edit configuration button from a particular backup config then the other tabs also enables the editing even though you clicked it or not. So it shouldn't happen. Edit config will only allow editing the backup config of the respective tab.

**Test Plan:**

1. Click config from the side nav.
2. Hit the backup tab and create backup configs for the respective tabs.
3. Then go to one of the tabs and hit the edit-config button.
4. Without saving it or cancelling it go to another tab and see if edit-config is available or not. If yes then it's a bug otherwise we're good to go.


yb-andrew (approved):
LG

Issue #8619, opened 2021-05-25
Backfill Table Rpc should show index names instead of ids in master:7000/tasks


acube123:
![Screen Shot 2021-05-25 at 3 11 33 PM](https://user-images.githubusercontent.com/6658556/119581518-9300e780-bd77-11eb-9aab-3d8700b8091e.png)

Backfill Table Rpc should use index names instead of index ids. similar to GetSafeTime Api.

Also, will be good to have the index names/ids sorted in the order of creation.

ConcurrencyPractitioner:
Hi @acube123 Do you mind if I try this issue as my first newbie task? I wanted to try my hand at some Yugabyte issues, and felt that this was a good one.

amitanandaiyer:
Sure @ConcurrencyPractitioner

amitanandaiyer:
Note that for internal accounting/rpc purposes we may still want to keep track of the ids. But the ToString functions and the BackfillJob (used for the UI rendering above) can be a more user-friendly if they show the index names instead. 

Feel free to ping me if you need any clarification.

ConcurrencyPractitioner:
Cool, thanks for the opportunity! I've created a pull request. So, from my understanding, I thought that I should log both the index table name plus the associated table id (since I think you wanted to keep the table id in the log as well). Hope this is what you meant. :)

PR #8690, opened 2021-05-27
[#8619] Backfilling rpc should log index name

ConcurrencyPractitioner:
This is a short fix for issue #8619. I decided to change the current field requested_index_names_ in BackfillTable so that it includes both index name plus the associated id, and then logged that instead.


src/yb/master/backfill_index.cc
```diff
@@ -653,7 +653,8 @@ BackfillTable::BackfillTable(
     if (!first) {
       out << ", ";
     }
-    out << master_->catalog_manager()->GetTableInfo(index_id)->name();
```
amitanandaiyer, line 656: I think we can revert this part. Having just the names should be fine for displaying information.

(ids may make it too cluttered.)

ConcurrencyPractitioner:
Closing this pull request as I'm migrating to a new branch.

ConcurrencyPractitioner:
@amitanandaiyer Just wanted to let you know that the new PR can be found here: #8693

PR #8693, opened 2021-05-28
[#8619] Log index table name instead of index id

ConcurrencyPractitioner:
I went over the changes, and I observed that their might've been some inconsistencies with my approach. I added a helper method instead which would retrieve the index table names instead (as the field ```requested_index_names_``` might not necessarily contain only the indexes we are currently backfilling).

### DIFF
--- a/src/yb/master/backfill_index.cc
+++ b/src/yb/master/backfill_index.cc
@@ -635,7 +635,8 @@ std::unordered_set<TableId> IndexIdsFromInfos(const std::vector<IndexInfoPB>& in
   return idx_ids;
 }
 
-std::string RetrieveIndexNames(CatalogManager* mgr, const std::unordered_set<std::string>& index_ids) {
+std::string RetrieveIndexNames(CatalogManager* mgr,
+                               const std::unordered_set<std::string>& index_ids) {
   std::ostringstream out;
   out << "{ ";
   bool first = true;
--- a/src/yb/master/backfill_index.cc
+++ b/src/yb/master/backfill_index.cc
@@ -635,6 +635,27 @@ std::unordered_set<TableId> IndexIdsFromInfos(const std::vector<IndexInfoPB>& in
   return idx_ids;
 }
 
+std::string RetrieveIndexNames(CatalogManager* mgr, const std::unordered_set<std::string>& index_ids) {
+  std::ostringstream out;
+  out << "{ ";
+  bool first = true;
+  for (const auto& index_id : index_ids) {
+    const auto table_info = mgr->GetTableInfo(index_id);
+    if (!table_info) {
+      LOG(WARNING) << "No table info can be found with index table id " << index_id;
+      continue;
+    }
+    if (!first) {
+      out << ", ";
+    }
+    first = false;
+
+    out << table_info->name();
+  }
+  out << " }";
+  return out.str();
+}
+
 }  // namespace
 
 BackfillTable::BackfillTable(
@@ -645,20 +666,8 @@ BackfillTable::BackfillTable(
       indexed_table_(indexed_table),
       index_infos_(indexes),
       requested_index_ids_(IndexIdsFromInfos(indexes)),
+      requested_index_names_(RetrieveIndexNames(master->catalog_manager(), requested_index_ids_)),
       ns_info_(ns_info) {
-  std::ostringstream out;
-  out << "{ ";
-  bool first = true;
-  for (const auto& index_id : requested_index_ids_) {
-    if (!first) {
-      out << ", ";
-    }
-    out << master_->catalog_manager()->GetTableInfo(index_id)->name();
-    first = false;
-  }
-  out << " }";
-  requested_index_names_ = out.str();
-
   auto l = indexed_table_->LockForRead();
   schema_version_ = indexed_table_->metadata().state().pb.version();
   leader_term_ = master_->catalog_manager()->leader_ready_term();
@@ -1243,6 +1252,21 @@ void GetSafeTimeForTablet::UnregisterAsyncTaskCallback() {
   }
   WARN_NOT_OK(backfill_table_->UpdateSafeTime(status, safe_time),
     "Could not UpdateSafeTime");
+}
+
+BackfillChunk::BackfillChunk(std::shared_ptr<BackfillTablet> backfill_tablet,
+                             const std::string& start_key) : 
+                             RetryingTSRpcTask(backfill_tablet->master(),
+                                              backfill_tablet->threadpool(),
+                                              gscoped_ptr<TSPicker>(new PickLeaderReplica(
+                                              backfill_tablet->tablet())),
+                            backfill_tablet->tablet()->table().get()),
+                            indexes_being_backfilled_(backfill_tablet->indexes_to_build()),
+                            backfill_tablet_(backfill_tablet),
+                            start_key_(start_key), 
+                            requested_index_names_(RetrieveIndexNames(backfill_tablet->master()->catalog_manager(),
+                                                                      indexes_being_backfilled_)) {
+  deadline_ = MonoTime::Max(); // Never time out.
 }
 
 // -----------------------------------------------------------------------------------------------
--- a/src/yb/master/backfill_index.cc
+++ b/src/yb/master/backfill_index.cc
@@ -1255,17 +1255,16 @@ void GetSafeTimeForTablet::UnregisterAsyncTaskCallback() {
 }
 
 BackfillChunk::BackfillChunk(std::shared_ptr<BackfillTablet> backfill_tablet,
-                             const std::string& start_key) : 
-                             RetryingTSRpcTask(backfill_tablet->master(),
-                                              backfill_tablet->threadpool(),
-                                              gscoped_ptr<TSPicker>(new PickLeaderReplica(
-                                              backfill_tablet->tablet())),
-                            backfill_tablet->tablet()->table().get()),
-                            indexes_being_backfilled_(backfill_tablet->indexes_to_build()),
-                            backfill_tablet_(backfill_tablet),
-                            start_key_(start_key), 
-                            requested_index_names_(RetrieveIndexNames(backfill_tablet->master()->catalog_manager(),
-                                                                      indexes_being_backfilled_)) {
+                             const std::string& start_key)
+    : RetryingTSRpcTask(backfill_tablet->master(),
+                        backfill_tablet->threadpool(),
+                        gscoped_ptr<TSPicker>(new PickLeaderReplica(backfill_tablet->tablet())),
+                        backfill_tablet->tablet()->table().get()),
+      indexes_being_backfilled_(backfill_tablet->indexes_to_build()),
+      backfill_tablet_(backfill_tablet),
+      start_key_(start_key),
+      requested_index_names_(RetrieveIndexNames(backfill_tablet->master()->catalog_manager(),
+                                                indexes_being_backfilled_)) {
   deadline_ = MonoTime::Max(); // Never time out.
 }
 
--- a/src/yb/master/backfill_index.h
+++ b/src/yb/master/backfill_index.h
@@ -172,7 +172,6 @@ class BackfillTable : public std::enable_shared_from_this<BackfillTable> {
   int32_t schema_version_;
   int64_t leader_term_;
 
-  std::string requested_index_names_;
   std::atomic_bool done_{false};
   std::atomic_bool timestamp_chosen_{false};
   std::atomic<size_t> tablets_pending_;
@@ -181,6 +180,8 @@ class BackfillTable : public std::enable_shared_from_this<BackfillTable> {
   mutable simple_spinlock mutex_;
   HybridTime read_time_for_backfill_ GUARDED_BY(mutex_){HybridTime::kMin};
   const std::unordered_set<TableId> requested_index_ids_;
+  const std::string requested_index_names_;
+
   const scoped_refptr<NamespaceInfo> ns_info_;
 };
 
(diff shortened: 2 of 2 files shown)


src/yb/master/backfill_index.cc
```diff
@@ -678,6 +667,21 @@ BackfillTable::BackfillTable(
   done_.store(false, std::memory_order_release);
 }
 
+void BackfillTable::retrieve_index_names(const std::unordered_set<std::string>& requested_index_ids, std::string& requested_id_names) {
```
amitanandaiyer, line 670: Our internal coding style is largely based on https://google.github.io/styleguide/cppguide.html#Function_Names 

I'd suggest using CamelCase to name the function RetrieveIndexNames or IndexNamesForIndexIds(...)

Also, it would be preferable to return the string; instead of using an output parameter

src/yb/master/backfill_index.cc
```diff
@@ -678,6 +667,21 @@ BackfillTable::BackfillTable(
...
+  std::ostringstream out;
+  out << "{ ";
+  bool first = true;
+  for (const auto& index_id : requested_index_ids) {
+    if (!first) {
+      out << ", ";
+    }
+    out << master_->catalog_manager()->GetTableInfo(index_id)->name();
```
amitanandaiyer, line 678: Probably a pre-existing bug here, but it may be better to check/skip in case `master_->catalog_manager()->GetTableInfo(index_id)` returns nullptr

src/yb/master/backfill_index.h
```diff
@@ -101,6 +101,8 @@ class BackfillTable : public std::enable_shared_from_this<BackfillTable> {
 
   const std::string& requested_index_names() const { return requested_index_names_; }
 
+  void retrieve_index_names(const std::unordered_set<std::string>& requested_index_ids, std::string& requested_index_names);
```
amitanandaiyer, line 104: Pls see comment above.

src/yb/master/backfill_index.h
```diff
@@ -253,6 +255,9 @@ class BackfillTablet : public std::enable_shared_from_this<BackfillTablet> {
   const std::vector<IndexInfoPB>& index_infos() const { return backfill_table_->index_infos(); }
 
   const std::string& requested_index_names() { return backfill_table_->requested_index_names(); }
+  void retrieve_index_names(const std::unordered_set<std::string> indexes, std::string& requested_index_names) {
```
amitanandaiyer, line 258: ditto

src/yb/master/backfill_index.h
```diff
@@ -350,8 +355,10 @@ class BackfillChunk : public RetryingTSRpcTask {
   std::string type_name() const override { return "Backfill Index Table"; }
 
   std::string description() const override {
-    return yb::Format("Backfilling index_ids $0 for tablet $1 from key '$2'",
-                      indexes_being_backfilled_, tablet_id(),
+    std::string requested_index_names = "";
```
amitanandaiyer, line 358: you could store this as a member variable in BackfillChunk, instead of recomputing it each time we call description.

amitanandaiyer (commented):
Thanks Richard!!

I like this approach.  It handles the case of partial-failures better, the index-names will show up for just the currently backfilling indexes; instead of all the indexes that we began backfilling with. 

Some minor stylistic comments inline on the naming.

For testing purposes, I'd suggest you can try to compile the code; and launch a local cluster with 
bin/yb-ctl and create a cql table/index on it.

master:7000/tasks should show you a view that I posted on the github-issue. You should be able to see the descriptions as desired.

src/yb/master/backfill_index.h
```diff
@@ -381,6 +387,7 @@ class BackfillChunk : public RetryingTSRpcTask {
   tserver::BackfillIndexResponsePB resp_;
   std::shared_ptr<BackfillTablet> backfill_tablet_;
   std::string start_key_;
+  std::string requested_index_names_;
```
amitanandaiyer, line 390: could make this const and initialize it as a member-initialization

src/yb/master/backfill_index.cc
```diff
@@ -678,6 +667,21 @@ BackfillTable::BackfillTable(
   done_.store(false, std::memory_order_release);
 }
 
+const std::string BackfillTable::RetrieveIndexNames(const std::unordered_set<std::string>& requested_index_ids) {
```
amitanandaiyer, line 670: const is probably not relevant for the return type.

ConcurrencyPractitioner:
Hi @amitanandaiyer,

Thanks for the tips, they were incredibly helpful!

I ran my changes as you suggested and found the following log:
image

It looks like the log statements are working as expected.
If there is anything else, please let me know!

src/yb/master/backfill_index.cc
```diff
@@ -678,6 +667,26 @@ BackfillTable::BackfillTable(
...
+    if (!first) {
+      out << ", ";
+    }
+    first = false;
+
+    out << table_info->name();
+  }
+  out << " }";
+  return out.str();
+}
```
jaki, line 688:
```suggestion
}
```

src/yb/master/backfill_index.h
```diff
@@ -253,6 +255,9 @@ class BackfillTablet : public std::enable_shared_from_this<BackfillTablet> {
   const std::vector<IndexInfoPB>& index_infos() const { return backfill_table_->index_infos(); }
 
   const std::string& requested_index_names() { return backfill_table_->requested_index_names(); }
+  std::string RetrieveIndexNames(const std::unordered_set<std::string>& indexes) {
```
jaki, line 258:
```suggestion
  std::string RetrieveIndexNames(const std::unordered_set<std::string>& indexes) {
```

src/yb/master/backfill_index.cc
```diff
@@ -678,6 +667,26 @@ BackfillTable::BackfillTable(
...
 
+std::string BackfillTable::RetrieveIndexNames(const std::unordered_set<std::string>& requested_index_ids) {
+  std::ostringstream out;
+  out << "{ ";
+  bool first = true;
+  for (const auto& index_id : requested_index_ids) {
+    const auto table_info = master_->catalog_manager()->GetTableInfo(index_id);
+    if (!table_info) {
+       continue;
+    }
```
jaki, line 678: Should this missing table info log warning or do something more severe?  For example, `LOG(WARNING) << "foo";`

src/yb/master/backfill_index.cc
```diff
@@ -678,6 +667,26 @@ BackfillTable::BackfillTable(
   done_.store(false, std::memory_order_release);
 }
 
+std::string BackfillTable::RetrieveIndexNames(const std::unordered_set<std::string>& requested_index_ids) {
```
jaki, line 670: line too long.  Make it <=100

jaki (commented):
Thanks for this.

ConcurrencyPractitioner:
@amitanandaiyer @jaki I'm ready for review, I think.

jaki (approved):
Ran internal builds: look fine.

The current state looks fine, but it would be great if you could address the additional comments.

src/yb/master/backfill_index.cc
```diff
@@ -678,6 +667,27 @@ BackfillTable::BackfillTable(
   done_.store(false, std::memory_order_release);
 }
 
+std::string BackfillTable::RetrieveIndexNames(const std::unordered_set<std::string>& index_ids) {
```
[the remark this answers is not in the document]
ConcurrencyPractitioner: @jaki I don't know if this is workable, mainly because when trying to call ```Master::catalog__manager()``` in ```backfill_index.h```, the following error was obtained:
```
In file included from src/yb/master/backfill_index.cc:13:0:
src/yb/master/backfill_index.h: In constructor 'yb::master::BackfillChunk::BackfillChunk(std::shared_ptr<yb::master::BackfillTablet>, const string&)':
src/yb/master/backfill_index.h:345:57: error: invalid use of incomplete type 'class yb::master::Master'
         requested_index_names_(backfill_tablet->master()->catalog_manager()
                                                         ^
In file included from ent/src/yb/master/async_rpc_tasks.h:16:0,
                 from src/yb/master/backfill_index.h:22,
                 from src/yb/master/backfill_index.cc:13:
src/yb/master/async_rpc_tasks.h:59:7: note: forward declaration of 'class yb::master::Master'
 class Master;
       ^
```
I can call the ```catalog_manager.cc``` class in ```backfill_index.cc``` method, not in the .h file. The main problem is that the statement where the error is thrown is in BackfillChunk which is located in .h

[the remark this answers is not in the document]
jaki: @ConcurrencyPractitioner, okay.  How about anonymous namespace, like for `IndexIdsFromInfos`?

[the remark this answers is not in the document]
jaki: Actually...try moving the constructor definition to the .cc, then.  Both suggestions don't work without that.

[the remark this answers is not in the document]
ConcurrencyPractitioner: @jaki Cool, sounds good. I've made the change.

src/yb/master/backfill_index.cc
```diff
@@ -646,18 +646,7 @@ BackfillTable::BackfillTable(
...
-  for (const auto& index_id : requested_index_ids_) {
-    if (!first) {
-      out << ", ";
-    }
-    out << master_->catalog_manager()->GetTableInfo(index_id)->name();
-    first = false;
-  }
-  out << " }";
-  requested_index_names_ = out.str();
+  requested_index_names_ = RetrieveIndexNames(requested_index_ids_);
```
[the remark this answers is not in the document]
ConcurrencyPractitioner: Acknowledged.

src/yb/master/backfill_index.cc
```diff
@@ -1255,6 +1233,21 @@ void GetSafeTimeForTablet::UnregisterAsyncTaskCallback() {
...
+                             RetryingTSRpcTask(backfill_tablet->master(),
+                                              backfill_tablet->threadpool(),
+                                              gscoped_ptr<TSPicker>(new PickLeaderReplica(
+                                              backfill_tablet->tablet())),
+                            backfill_tablet->tablet()->table().get()),
+                            indexes_being_backfilled_(backfill_tablet->indexes_to_build()),
+                            backfill_tablet_(backfill_tablet),
+                            start_key_(start_key), 
+                            requested_index_names_(backfill_tablet->master()->catalog_manager()
+                                                   ->RetrieveIndexNames(indexes_being_backfilled_)) {
```
jaki, line 1247:
```suggestion
                             const std::string& start_key)
    : RetryingTSRpcTask(backfill_tablet->master(),
                        backfill_tablet->threadpool(),
                        gscoped_ptr<TSPicker>(new PickLeaderReplica(backfill_tablet->tablet())),
      backfill_tablet->tablet()->table().get()),
      indexes_being_backfilled_(backfill_tablet->indexes_to_build()),
      backfill_tablet_(backfill_tablet),
      start_key_(start_key),
      requested_index_names_(backfill_tablet->master()->catalog_manager()
                             ->RetrieveIndexNames(indexes_being_backfilled_)) {
```

jaki (commented):
I'll run another round of tests on this.

ConcurrencyPractitioner:
Cool, so I accepted your suggestion. @jaki 

@amitanandaiyer If you think there is anything else, please let me know. :)

src/yb/master/backfill_index.cc
```diff
@@ -645,20 +645,8 @@ BackfillTable::BackfillTable(
       indexed_table_(indexed_table),
       index_infos_(indexes),
       requested_index_ids_(IndexIdsFromInfos(indexes)),
+      requested_index_names_(master->catalog_manager()->RetrieveIndexNames(requested_index_ids_)),
```
amitanandaiyer, line 648: I'm not too sure if I'd want this function in catalog_manager.cc

Seems like this is something we'd only use for backfill_indexes and may be polluting catalog manager (which is already doing too many things).

have you considered,  adding it in this file in an anonymous namespace, as 

```
namespace {

std::string RetrieveIndexNames(CatalogManager *mgr, const std::unordered_set<std::string>& requested_index_ids) {
...
}

};

Comment (ConcurrencyPractitioner) [reply src/yb/master/backfill_index.cc]: @amitanandaiyer Sure, sounds fair. I've made the change accordingly. It looks neater now, actually.

File: src/yb/master/backfill_index.cc [hunk 16]
```diff
@@ -1245,6 +1254,21 @@ void GetSafeTimeForTablet::UnregisterAsyncTaskCallback() {
...
+                             RetryingTSRpcTask(backfill_tablet->master(),
+                                              backfill_tablet->threadpool(),
+                                              gscoped_ptr<TSPicker>(new PickLeaderReplica(
+                                              backfill_tablet->tablet())),
+                            backfill_tablet->tablet()->table().get()),
+                            indexes_being_backfilled_(backfill_tablet->indexes_to_build()),
+                            backfill_tablet_(backfill_tablet),
+                            start_key_(start_key), 
+                            requested_index_names_(RetrieveIndexNames(backfill_tablet->master()->catalog_manager(),
+                                                                      indexes_being_backfilled_)) {
```
jaki, line 1268:
```suggestion
                             const std::string& start_key)
    : RetryingTSRpcTask(backfill_tablet->master(),
                        backfill_tablet->threadpool(),
                        gscoped_ptr<TSPicker>(new PickLeaderReplica(backfill_tablet->tablet())),
                        backfill_tablet->tablet()->table().get()),
      indexes_being_backfilled_(backfill_tablet->indexes_to_build()),
      backfill_tablet_(backfill_tablet),
      start_key_(start_key),
      requested_index_names_(RetrieveIndexNames(backfill_tablet->master()->catalog_manager(),
                                                indexes_being_backfilled_)) {
```

jaki (commented):
Thanks for the quick adaption.  The same style issue seems to have come back.  I'll run another round of tests/builds.

src/yb/master/backfill_index.cc
```diff
@@ -635,6 +635,27 @@ std::unordered_set<TableId> IndexIdsFromInfos(const std::vector<IndexInfoPB>& in
   return idx_ids;
 }
 
+std::string RetrieveIndexNames(CatalogManager* mgr, const std::unordered_set<std::string>& index_ids) {
```
jaki, line 638: line too long

ConcurrencyPractitioner: ack

amitanandaiyer (approved):
lgtm!! thanks @ConcurrencyPractitioner

src/yb/master/backfill_index.cc
amitanandaiyer: btw,  fyi: there is this script https://github.com/yugabyte/yugabyte-db/blob/master/arcanist_util/pre_commit_hook.sh

that you can run locally to reformat your changes in a yb style guide compliant way. 

it's not perfect but gets most of the stylistic guidelines right.

jaki:
Thanks, @ConcurrencyPractitioner.  Public and internal build results look fine.

PR #8737, opened 2021-06-03
[#6857]Platform: Pre-flight checks for encryption-at-rest keys

jitendra-12113:
**Heading**
[#6857]Platform: Pre-flight checks for encryption-at-rest keys
**Description:**
We don't have any check to validate the AWS creds and api url while
saving the config.
At the time of creating a new key configuration, need to do a pre-flight check
to verify the key config is valid.

**Test Plan:**

Got to Config->Security->Encryption-at-Rest keys tab
Select kms provider as AWS
Enter invalid accesskey or secret key - we should get proper error message for the same
If we select SmartKey as KMS provider then api url should be one of these
api.amer.smartkey.io,
api.eu.smartkey.io,
api.uk.smartkey.io
UT's have been added. SBT tests are getting passed.

[info] Test com.yugabyte.yw.commissioner.TaskGarbageCollectorTest.testPurge_invalidData started
[info] Test run finished: 0 failed, 0 ignored, 7 total, 0.149s
[info] Passed: Total 1363, Failed 0, Errors 0, Passed 1359, Skipped 4
[success] Total time: 1992 s (33:12), completed 3 Jun, 2021 10:28:55 AM

Ref Diff: https://github.com/yugabyte/yugabyte-db/pull/7701


nishantSharma459 (approved):

nishantSharma459:
FE looks good.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -42,6 +48,32 @@
...
   @Inject Commissioner commissioner;
 
+  @Inject CloudAPI.Factory cloudAPIFactory;
+
+  private void validateKMSProviderConfigFormData(ObjectNode formData, String keyProvider) {
+    if (keyProvider.toUpperCase().equals(KeyProvider.AWS.toString())
+        && (formData.get("AWS_ACCESS_KEY_ID") != null
+            || formData.get("AWS_SECRET_ACCESS_KEY") != null)) {
+      CloudAPI cloudAPI = cloudAPIFactory.get(KeyProvider.AWS.toString().toLowerCase());
+      Map<String, String> config = new HashMap<String, String>();
```
SergeyPotachev, line 58: Map<String, String> config = new HashMap<>();

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -42,6 +48,32 @@
...
+      if (cloudAPI != null
+          && !cloudAPI.isValidCreds(config, formData.get("AWS_REGION").textValue())) {
+        throw new YWServiceException(BAD_REQUEST, "Invalid AWS Credentials.");
+      }
+    }
+    if (keyProvider.toUpperCase().equals(KeyProvider.SMARTKEY.toString())) {
+      final Set<String> apiUrl =
+          new HashSet<>(
+              Arrays.asList("api.amer.smartkey.io", "api.eu.smartkey.io", "api.uk.smartkey.io"));
+      if (!(formData.get("base_url") != null
```
SergeyPotachev, line 70: nit: Looks too complex. For me it is better to use smtp like this:
 
`if (formData.get("base_url") == null  || !apiUrl.contains(formData.get("base_url").textValue()))
`
or even:
`
JsonNode baseUrl=formData.get("base_url");`
`if (baseUrl == null  || !apiUrl.contains(baseUrl.textValue()))
`

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -42,6 +48,32 @@
...
+      Map<String, String> config = new HashMap<String, String>();
+      config.put("AWS_ACCESS_KEY_ID", formData.get("AWS_ACCESS_KEY_ID").textValue());
+      config.put("AWS_SECRET_ACCESS_KEY", formData.get("AWS_SECRET_ACCESS_KEY").textValue());
+      if (cloudAPI != null
+          && !cloudAPI.isValidCreds(config, formData.get("AWS_REGION").textValue())) {
+        throw new YWServiceException(BAD_REQUEST, "Invalid AWS Credentials.");
+      }
+    }
+    if (keyProvider.toUpperCase().equals(KeyProvider.SMARTKEY.toString())) {
+      final Set<String> apiUrl =
```
SergeyPotachev, line 67: Why final + why not simply:
`
List<String> apiUrl = Arrays.asList(...);
`

SergeyPotachev: Also why are we limited by these three domains only? Maybe it is better to use a mask like api.*.smartkey.io?

SergeyPotachev (changes requested):

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
daniel-yb: @SergeyPotachev those are the only valid APIs that smartkey supports. From their documentation: `Note that the U.S. API endpoint is api.amer.smartkey.io, the E.U API endpoint is api.eu.smartkey.io, and the U.K. API endpoint is api.uk.smartkey.io`

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -42,6 +48,32 @@
...
+          new HashSet<>(
+              Arrays.asList("api.amer.smartkey.io", "api.eu.smartkey.io", "api.uk.smartkey.io"));
```
daniel-yb, line 69: Would be good for this list to be a static class field or even an enum as we don't expect it to ever change and re-building the list/set over and over is unnecessary work

managed/src/test/java/com/yugabyte/yw/controllers/EncryptionAtRestControllerTest.java
```diff
@@ -190,6 +191,74 @@ public void testCreateAndRecoverKey() {
     assertAuditEntry(2, customer.uuid);
   }
 
+  @Test
+  public void testCreateSMARTKEYKmsProviderWithInValidAPIUrl() {
```
daniel-yb, line 195: nit: InValid -> Invalid

managed/src/test/java/com/yugabyte/yw/controllers/EncryptionAtRestControllerTest.java
```diff
@@ -190,6 +191,74 @@ public void testCreateAndRecoverKey() {
...
+    UUID fakeTaskUUID = UUID.randomUUID();
+    when(mockCommissioner.submit(any(TaskType.class), any(KMSConfigTaskParams.class)))
+        .thenReturn(fakeTaskUUID);
+    Result createKMSResult =
+        doRequestWithAuthTokenAndBody("POST", kmsConfigUrl, authToken, kmsConfigReq);
+    assertOk(createKMSResult);
+  }
+
+  @Test
+  public void testCreateAwsKmsProviderWithInValidCreds() {
```
daniel-yb, line 224: Nit: InValid -> Invalid

daniel-yb:
@jitendra-12113 we can validate the combination of API key + URL by calling SmartkeyEARService::retrieveSessionAuthorization and seeing if it throws an exception or not.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
jitendra-12113: Done.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
jitendra-12113: Fixed. Thanks for pointing out.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
jitendra-12113: yeah make sense. Declared static class field var. Thanks!

managed/src/test/java/com/yugabyte/yw/controllers/EncryptionAtRestControllerTest.java
jitendra-12113: Done.

managed/src/test/java/com/yugabyte/yw/controllers/EncryptionAtRestControllerTest.java
jitendra-12113: Done.

jitendra-12113:
> @jitendra-12113 we can validate the combination of API key + URL by calling SmartkeyEARService::retrieveSessionAuthorization and seeing if it throws an exception or not.

Hi @daniel-yb  , could you please tell from where to get valid secret API key, I don't have any so couldn't test this change. Appreciate your help on this! Thanks!

We can have this check in followup PR along with IAM.

managed/ui/src/components/config/Security/KeyManagementConfigurationContainer.js
```diff
@@ -41,6 +41,13 @@ const mapDispatchToProps = (dispatch) => {
     setKMSConfig: (provider, body) => {
       return dispatch(createKMSProviderConfig(provider, body))
         .then?.((response) => {
+          if (response.error) {
+            const errorMessage =
+              response.payload?.response?.data?.error || response.payload.message;
+            toast.error(errorMessage);
+          } else {
+            toast.success('Successfully added the configuration');
```
sunilsutarDCE, line 49: Success message should also come from backend , instead of static message .

managed/ui/src/components/config/Security/KeyManagementConfigurationContainer.js
```diff
@@ -41,6 +41,13 @@ const mapDispatchToProps = (dispatch) => {
     setKMSConfig: (provider, body) => {
       return dispatch(createKMSProviderConfig(provider, body))
         .then?.((response) => {
+          if (response.error) {
+            const errorMessage =
+              response.payload?.response?.data?.error || response.payload.message;
```
sunilsutarDCE, line 46: Any reason for OR here , does error message come in 2 different source ?

gaurav061: Earlier implementation of toaster had a comment from Sergey, that this field may not contain a message in case of 500 or some scenarios like that, so we should also check for payload messages, However, I will check with @jitendra-12113 on this.

managed/ui/src/components/config/Security/KeyManagementConfigurationContainer.js
gaurav061: Sure, I will check with @jitendra-12113  on this.

gaurav061: As per discussion with @jitendra-12113, W need to check with Sashank on this, For now `response.payload.message` is for handling generic exception.

managed/ui/src/components/config/Security/KeyManagementConfigurationContainer.js
gaurav061: `response.payload.message` holds a generic message in case of exception like 500.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
SergeyPotachev: @daniel-yb OK, I see.

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
```diff
@@ -30,18 +33,57 @@
...
+
   @Inject EncryptionAtRestManager keyManager;
 
   @Inject Commissioner commissioner;
 
+  @Inject CloudAPI.Factory cloudAPIFactory;
+
+  private void validateKMSProviderConfigFormData(ObjectNode formData, String keyProvider) {
+    if (keyProvider.toUpperCase().equals(KeyProvider.AWS.toString())
+        && (formData.get("AWS_ACCESS_KEY_ID") != null
```
SergeyPotachev, line 59: I see these constants again and again. We have them in CustomerConfigValidator.java, as example. If you/we don't like them being there, we can extract them to a proper class (in another package, maybe something like AwsConstants). But don't use simple string constants, please.

SergeyPotachev (changes requested):

managed/src/main/java/com/yugabyte/yw/controllers/EncryptionAtRestController.java
jitendra-12113: Good point: Used consts for now. Will create AWSConstants in the follow up diff. Thanks!

SergeyPotachev (approved):
Please do constants and code cleanup later to not have same constants in different classes.

PR #8745, opened 2021-06-03
github action tests: fix YB download path

svarnau:
Recent YugabyteDB releases are not posted on github.

### DIFF
--- a/.github/workflows/yugabyted-test.yml
+++ b/.github/workflows/yugabyted-test.yml
@@ -6,12 +6,16 @@ on:
       - master
     paths:
       - '**/yugabyted*'
+    paths-ignore:
+        - 'docs/**'
 
   pull_request:
     branches:
       - master
     paths:
       - '**/yugabyted*'
+    paths-ignore:
+        - 'docs/**'
 
 jobs:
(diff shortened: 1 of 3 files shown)


polarweasel (approved):
LGTM! Tests pass now.

The output of the tests points to two issues that may be worth addressing in the near future.

1. Making downloads quieter would be very good. 9846 lines in each log are just downloading the tarball.
2. Making `yugabyted` aware that it's running in a non-interactive environment would also be very good. 2439 lines in one log (2427 in the other) are the "Running system checks..." cursor animation.

svarnau:
> LGTM! Tests pass now.
> 
> The output of the tests points to two issues that may be worth addressing in the near future.
> 
> 1. Making downloads quieter would be very good. 9846 lines in each log are just downloading the tarball.
> 2. Making `yugabyted` aware that it's running in a non-interactive environment would also be very good. 2439 lines in one log (2427 in the other) are the "Running system checks..." cursor animation.

1. Okay, adding the quiet option to the wget command.
2. I don't enough about yugabyted to turn that off, or to know if that might hide info on a failure.

Issue #8311, opened 2021-05-06
Fix the download url for Replicated airgap bundle in Platform docs


streddy-yb:
Change the replicated download URL in `https://docs.yugabyte.com/latest/yugabyte-platform/install-yugabyte-platform/install-software/airgapped/#install-replicated` from  `https://downloads.yugabyte.com/replicated.tar.gz `to 
`https://s3.amazonaws.com/replicated-airgap-work/replicated.tar.gz`

polarweasel:
Can we use one of Replicated's own URLs instead of pointing to their s3 bucket?

```sh
$ curl -I https://get.replicated.com/airgap
HTTP/2 302 
   [...]
location: https://s3.amazonaws.com/replicated-airgap-work/replicated.tar.gz
```

ddhodge:
The download for Yugaware is https://downloads.yugabyte.com/yugaware-2.1.2.0-b10.airgap, should that also be updated?

polarweasel:
This `wget` command will work to get past the redirect and naming problem. I can't get `curl` to do it right so far.

```sh
wget --trust-server-names https://get.replicated.com/airgap
```

PR #8751, opened 2021-06-03
[docs] Update airgapped install

ddhodge:
#8311 updated URLs

### DIFF
--- a/docs/content/latest/yugabyte-platform/install-yugabyte-platform/install-software/airgapped.md
+++ b/docs/content/latest/yugabyte-platform/install-yugabyte-platform/install-software/airgapped.md
@@ -65,13 +65,13 @@ $ cd /opt/downloads
 Download the `replicated.tar.gz` file by executing the following command:
 
 ```sh
-$ wget https://s3.amazonaws.com/replicated-airgap-work/replicated.tar.gz
+$ wget --trust-server-names https://get.replicated.com/airgap
 ```
 
 Download the `yugaware` binary and change the following number, as required:
 
 ```sh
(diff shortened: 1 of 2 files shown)


docs/content/latest/yugabyte-platform/install-yugabyte-platform/install-software/airgapped.md
```diff
@@ -65,13 +65,13 @@ $ cd /opt/downloads
...
 ```sh
-$ wget https://downloads.yugabyte.com/replicated.tar.gz
+$ wget --trust-server-names https://get.replicated.com/airgap
 ```
 
 Download the `yugaware` binary and change the following number, as required:
 
 ```sh
-$ wget https://downloads.yugabyte.com/yugaware-2.1.2.0-b10.airgap
+$ wget https://downloads.yugabyte.com/yugaware-2.4.4.0-b7.airgap
```
polarweasel, line 74: Seems odd to be providing a `stable` release download in the `latest` docs. Should it be this instead?
```suggestion
$ wget https://downloads.yugabyte.com/yugaware-2.7.1.1-b1.airgap
```

polarweasel: (And if so, you'll need to also update the same file in the stable folder)

ddhodge: fixed

polarweasel (approved):
LGTM. Is there a good way for us to tell the reader _where_ they might find the version and build number they need, or not really? (download.yugabyte.com may be the easiest place to point them, if we do want to point them anywhere)

Issue #8754, opened 2021-06-03
[Docs] Docker fault tolerance example does not provide working ycqlsh connection command


jvigil-yugabyte:
https://docs.yugabyte.com/latest/explore/fault-tolerance/docker/#1-create-universe

Fails with error:
```
docker exec -it yb-tserver-n1 /home/yugabyte/bin/ycqlsh
Connection error: ('Unable to connect to any servers', {'127.0.0.1': error(111, "Tried connecting to [('127.0.0.1', 9042)]. Last error: Connection refused")})
```

It should use the correct address bound to by the server running in the docker container.

PR #8755, opened 2021-06-03
[#8754] Docs: Fix docker fault tolerance ycqlsh commands to use address

jvigil-yugabyte:


polarweasel (approved):
LGTM. I've run through the whole scenario on my local machine, and it all works as advertised.

Issue #8760, opened 2021-06-04
[Platform][API] Metric, Certificate info, Provider, Region and Yugaware property  Raw Json Cleanup


mahendranbhat:
subtask of https://github.com/yugabyte/yugabyte-db/issues/7877

Description :
There are places where we just create raw json node and return that. Ebeans already has built in serialization and deserialization for DBJson fields.
In many cases these entities are also part of out API request responses and they end up polluting the API definition with raw opaque json fields instead of first class objects. Hence making this part of our API cleanup.
As a side effect this leads to much better code redability.

Fields included:

1. Metrcs model :  layout, filters
2. Yugaware property : value
3. Provider : config
4. Region : detail, config
5. Certificat info: customCertInfo

test plan:

All UI operations should work with the mentioned models API.
basic compile test, run platform should work

PR #8797, opened 2021-06-07
[#8760][Platform][API] Metric,  Region Raw Json Cleanup

mahendranbhat:
Description :
There are places where we just create raw json node and return that. Ebeans already has built in serialization and deserialization for DBJson fields.
In many cases these entities are also part of out API request responses and they end up polluting the API definition with raw opaque json fields instead of first class objects. Hence making this part of our API cleanup.
As a side effect this leads to much better code redability.

Fields included:

Metrcs model : layout, filters
Region : detail, config
test plan:

All UI operations should work with the mentioned models API.
basic compile test, run platform should work


jaydeepkumara (approved):

managed/src/main/java/com/yugabyte/yw/models/YugawareProperty.java
```diff
@@ -24,7 +26,7 @@
   public static final Logger LOG = LoggerFactory.getLogger(YugawareProperty.class);
 
   // The name of the property.
-  @Id private String name;
+  @Id public String name;
```
sb-yb, line 29: Why public?

sb-yb (commented):
Looks good. Are you still having problem with the test?

managed/src/main/java/com/yugabyte/yw/models/YugawareProperty.java
mahendranbhat: not sure how it got changes, let me change.

sb-yb:
AccessManagerTest has poor use of generics. It should do:

```
  ArgumentCaptor<List<String>> command;
  ArgumentCaptor<Map<String, String>> cloudCredentials;
```
And iniitialize as:
```
    command = ArgumentCaptor.forClass(List.class);
    cloudCredentials = ArgumentCaptor.forClass(Map.class);
```

mahendranbhat:
> AccessManagerTest has poor use of generics. It should do:
> 
> ```
>   ArgumentCaptor<List<String>> command;
>   ArgumentCaptor<Map<String, String>> cloudCredentials;
> ```
> 
> And iniitialize as:
> 
> ```
>     command = ArgumentCaptor.forClass(List.class);
>     cloudCredentials = ArgumentCaptor.forClass(Map.class);
> ```

Thanks! worked and updated the PR.

sb-yb (approved):

managed/src/main/java/com/yugabyte/yw/models/YugawareProperty.java
```diff
@@ -44,9 +46,10 @@
   // The property config.
   @Constraints.Required
   @Column(columnDefinition = "TEXT")
-  private JsonNode value;
+  @DbJson
+  private Map<String, Object> value;
```
sb-yb, line 50: This is a true Json Node. So no need to fix this.

managed/src/main/java/com/yugabyte/yw/models/CertificateInfo.java
```diff
@@ -92,19 +83,19 @@ public void setChecksum() throws IOException, NoSuchAlgorithmException {
 
   @Column(columnDefinition = "TEXT", nullable = true)
   @DbJson
-  public JsonNode customCertInfo;
+  public CertificateParams.CustomCertInfo customCertInfo;
 
   public CertificateParams.CustomCertInfo getCustomCertInfo() {
     if (this.customCertInfo != null) {
-      return Json.fromJson(this.customCertInfo, CertificateParams.CustomCertInfo.class);
+      return this.customCertInfo;
```
sb-yb, line 90: why not just always `return this.customCertInfo;` ?

managed/src/main/java/com/yugabyte/yw/models/CertificateInfo.java
```diff
@@ -92,19 +83,19 @@ public void setChecksum() throws IOException, NoSuchAlgorithmException {
 
   @Column(columnDefinition = "TEXT", nullable = true)
   @DbJson
-  public JsonNode customCertInfo;
+  public CertificateParams.CustomCertInfo customCertInfo;
```
sb-yb, line 86: make it private so no one can skip checkEditable validation.

managed/src/main/java/com/yugabyte/yw/models/MetricConfig.java
```diff
@@ -70,14 +70,14 @@ public MetricConfig getConfig() {
...
-      return Json.fromJson(this.getConfig().filters, Map.class);
+      return this.getConfig().filters;
     }
     return new HashMap<>();
   }
 
   public Layout getLayout() {
     if (this.getConfig().layout != null) {
-      return Json.fromJson(this.getConfig().layout, Layout.class);
+      return this.getConfig().layout;
```
sb-yb, line 80: same

managed/src/main/java/com/yugabyte/yw/models/Region.java
```diff
@@ -88,50 +88,50 @@ public void setActiveFlag(Boolean active) {
 
   @DbJson
   @Column(columnDefinition = "TEXT")
-  public JsonNode details;
+  public Map<String, String> details;
```
sb-yb, line 91: Create a POJO  `RegionDetails` with two properties `sg_id` and `vnet`. (Just public static inner class is fine)  And use that class here.
Fix the below getters and setters to work with that POJO.

sb-yb: This is low priority btw becaue I am hiding this property and only exposing the setters and getters below. But It is a simple fix so you may want to do it in this PR.

sb-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/models/YugawareProperty.java
mahendranbhat: how to identify this?

sb-yb: Follow the trail for where it is getting set.  it is a json comming from some external tool. For example health check script or in this case some python script that is storing too much variety.
Of course you can follow the trail into that script and deduce that it has some fixed format that can be captured by a POJO. in that case we can use that POJO. 
Or you see that it is being updated as a map by that tool.

Map<String, String> is better than ObjectNode as value wont be another objectNode or arrayNode.
But Map<String, Object> does not buy us any additional type safety or contract beyond ObjectNode. 
It is just an equivalent representation that is equially opaque.

But in API having Json node is useless and Map<String Object> will be better. To achieve that we can simply use 
`@ApiModelProperty(dataType="Map")`
or 
`@ApiModelProperty(dataType="Object")`

And we will get to enter a JsonObject equivalent.

managed/src/main/java/com/yugabyte/yw/models/Region.java
mahendranbhat: if it only can contains above two fields, i will make the change. My local has only one field for the region i have. So made it map.

managed/src/main/java/com/yugabyte/yw/models/MetricConfig.java
mahendranbhat: we have isEmpty() checks

managed/src/main/java/com/yugabyte/yw/models/Region.java
```diff
@@ -86,52 +85,57 @@ public void setActiveFlag(Boolean active) {
     this.active = active;
   }
 
+  class RegionDetails {
+    public String sg_id; // Security group ID.
+    public String vet; // Vnet key.
```
sb-yb, line 90: typo? vnet I think.

managed/src/main/java/com/yugabyte/yw/models/Region.java
```diff
@@ -86,52 +85,57 @@ public void setActiveFlag(Boolean active) {
...
+      details = new RegionDetails();
     }
-    ((ObjectNode) details).put(VNET_KEY, vnetName);
+    details.vet = vnetName;
     save();
   }
 
   public String getVnetName() {
     if (details != null) {
-      JsonNode vnetNode = details.get(VNET_KEY);
```
sb-yb, line 119: delete the static constants

managed/src/main/java/com/yugabyte/yw/models/Region.java
```diff
@@ -86,52 +85,57 @@ public void setActiveFlag(Boolean active) {
...
+      String vnetNode = details.vet;
+      return vnetNode == null || vnetNode.isEmpty() ? null : vnetNode;
     }
     return null;
   }
 
   @DbJson
   @Column(columnDefinition = "TEXT")
-  public JsonNode config;
+  public Map<String, String> config;
```
sb-yb, line 131: I think this can be private

managed/src/main/java/com/yugabyte/yw/models/Region.java
```diff
@@ -86,52 +85,57 @@ public void setActiveFlag(Boolean active) {
...
 
   public void setConfig(Map<String, String> configMap) {
```
sb-yb, line 133: Tyr to capture as many keys of this map as possible in static consts. 
For example I see "OVERRIDES" for kubernetes. There may be more.

sb-yb (changes requested):
Mostly looks good. Just few small comments.

managed/src/main/java/com/yugabyte/yw/models/Region.java
mahendranbhat: are you suggesting to use POJO here?

sb-yb: If you can figure out all the fields yes. But I suspect that it will merge in random things like some gcp service account creds etc. i.e. it is inherantly open-ended.

I am just saying that capture in static constants at least the keys that we know of. You can do that in follow up. This is similar to `VNET_KEY` and `SECURITY_GROUP_KEY` in case of details. It at least give some sanity to code as we can track references and most keys get tracked at one place.

sb-yb (approved):

sb-yb (approved):

PR #8840, opened 2021-06-10
[Platform][API] Health check  API json cleanups

mahendranbhat:
Description :
There are places where we just create raw json node and return that. Ebeans already has built in serialization and deserialization for DBJson fields.
In many cases these entities are also part of out API request responses and they end up polluting the API definition with raw opaque json fields instead of first class objects. Hence making this part of our API cleanup.
As a side effect this leads to much better code redability.

Fields included:

Healthcheck: detailsjson

test plan:
All UI operations should work with the mentioned models API.
basic compile test, run platform should work.

	at play.db.ebean.DefaultEbeanConfig$EbeanConfigParser$WrappingDatasource.getConnection(DefaultEbeanConfig.java:153)
	at io.ebeaninternal.server.transaction.TransactionFactoryBasic.createQueryTransaction(TransactionFactoryBasic.java:28)
	... 16 common frames omitted
[error] Failed: Total 1517, Failed 3, Errors 0, Passed 1510, Skipped 4
[error] Failed tests:
[error] 	com.yugabyte.yw.controllers.ApiDiscoveryControllerTest
[error] 	com.yugabyte.yw.commissioner.tasks.UpgradeUniverseTest
[error] 	com.yugabyte.yw.models.helpers.CustomerConfigValidatorTest
[error] (Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 1510 s (25:10), completed 23 Jun, 2021 12:00:43 PM


managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
```diff
@@ -36,15 +40,18 @@
 
   // The Json serialized version of the details. This is used only in read from and writing to the
   // DB.
+  @DbJson
   @Constraints.Required
   @Column(columnDefinition = "TEXT", nullable = false)
-  public String detailsJson;
+  public  Map<String, Object>  detailsJson;
```
sb-yb, line 46: I checked and this is not a Map. It is setting from JsonData returned by healthCheck script. 
The way to deal with this is to create a POJO that captures the format of what the healthcheck script can return.  
i.e. Its going to be something like 
```
class HealthCheckDetails {
    boolean has_error;
    ... // other things 
}
```

May be all the properties are known and we can come up with such pojo sinc ethe health check script is in our control. So just check what format it generates.

If it is unbounded i.e. can contain any fields then we can just leave it as is.

managed/src/main/java/com/yugabyte/yw/models/TaskInfo.java
```diff
@@ -91,7 +93,7 @@
   @Constraints.Required
   @Column(columnDefinition = "TEXT default '{}'", nullable = false)
   @DbJson
-  private JsonNode details;
+  private Map<String, Object>  details;
```
sb-yb, line 96: This is not a Map. I followed the trail and this is of type ITaskParams.
Let me know if you have questions.

sb-yb: Customers need this API so prioritize this and address HelthCheck separately.

sb-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/models/TaskInfo.java
sb-yb: So what you really need is so that ITaskParam subtype can be deserialized back correctly:
https://stackoverflow.com/questions/30362446/deserialize-json-with-jackson-into-polymorphic-types-a-complete-example-is-giv

Thats too much to do and I do not think we need to expose the task params anyway.
Leave it as is. i.e. revert this part.

managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
```diff
@@ -23,6 +27,23 @@
...
+      public String node_name;
+      public Boolean has_error;
+      public List<String> details;
+      public String message;
+    }
+
+    public Date timestamp;
+    public List<nodeData> data = new ArrayList<>();
+    public String yb_version;
+    public Boolean error;
```
mahendranbhat, line 44: @sb-yb 
```{"error": false, "data": [ {"node":"n", "has_error": true, "message":"c" } ] }
2021-06-22 00:23:03.788 [info] HikariCPModule.scala:79 [application-akka.actor.default-dispatcher-5] Shutting down connection pool.
[error] Test com.yugabyte.yw.commissioner.HealthCheckerTest.testSingleUniverseNoEmail failed: java.lang.IllegalArgumentException: Unrecognized field "error" (class com.yugabyte.yw.models.HealthCheck$details), not marked as ignorable (4 known properties: "data", "has_error", "timestamp", "yb_version"])
[error]  at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: com.yugabyte.yw.models.HealthCheck$details["error"]), took 0.767```
it should have been has_error ( checked with jitendra's DB entry ) I will setup healthcheck and validate but test case is only passing `error`  , not sure how it was working before if the field which is passed is error.

File: managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java [hunk 4]
```diff
@@ -23,6 +28,25 @@
 public class HealthCheck extends Model {
   public static final Logger LOG = LoggerFactory.getLogger(HealthCheck.class);
 
+  public static class details {
```
sb-yb, line 31: Details

managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
```diff
@@ -23,6 +28,25 @@
...
+    public static class nodeData {
```
sb-yb, line 32: NodeData

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -1746,7 +1746,7 @@ public Result healthCheck(UUID customerUUID, UUID universeUUID) {
     try {
       List<HealthCheck> checks = HealthCheck.getAll(universeUUID);
       for (HealthCheck check : checks) {
-        detailsList.add(Json.stringify(Json.parse(check.detailsJson)));
+        detailsList.add(check.detailsJson.toString());
```
sb-yb, line 1749: Is there a unit test that reads this response so that we know its not broken.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -1746,7 +1746,7 @@ public Result healthCheck(UUID customerUUID, UUID universeUUID) {
     try {
       List<HealthCheck> checks = HealthCheck.getAll(universeUUID);
```
sb-yb, line 1747: Should this be getAllOrBadRequest? Will that get rid of the runtime exception catching below?

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
```diff
@@ -1746,7 +1746,7 @@ public Result healthCheck(UUID customerUUID, UUID universeUUID) {
...
       for (HealthCheck check : checks) {
-        detailsList.add(Json.stringify(Json.parse(check.detailsJson)));
+        detailsList.add(check.detailsJson.toString());
       }
     } catch (RuntimeException e) {
       // TODO(API) dig deeper and find root cause of RuntimeException
```
sb-yb, line 1752: See if there is unit test for bad request that gives any clue

managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
```diff
@@ -23,6 +28,25 @@
...
+      public Boolean has_error;
+      public List<String> details;
+      public String message;
+    }
+
+    public Date timestamp;
+    public List<nodeData> data = new ArrayList<>();
+    public String yb_version;
+
+    @JsonAlias({"error", "has_error"})
```
sb-yb, line 46: Add a // TODO: This was done to make <testName> happy. Reconcile this.

managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
```diff
@@ -36,15 +60,18 @@
...
 
   public boolean hasError() {
-    JsonNode details = Json.parse(detailsJson);
-    JsonNode hasErrorField = details.get(FIELD_HAS_ERROR);
-    // Only return true if we have the top-level has_error field with a value of true.
-    return hasErrorField != null && hasErrorField.asBoolean();
+    try {
+      // Only return true if we have the top-level has_error field with a value of true.
+      return detailsJson.has_error;
+    } catch (NullPointerException e) {
```
sb-yb, line 72: why npe catch? was that temporary to debug failing test?
If so remove. Otherwise use if condition to get rid of NPE.

managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
```diff
@@ -36,15 +60,18 @@
...
   // DB.
+  @DbJson
   @Constraints.Required
   @Column(columnDefinition = "TEXT", nullable = false)
-  public String detailsJson;
+  public details detailsJson;
 
   public boolean hasError() {
-    JsonNode details = Json.parse(detailsJson);
-    JsonNode hasErrorField = details.get(FIELD_HAS_ERROR);
```
sb-yb, line 45: Get rid of static const if not too much of work.

managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
```diff
@@ -64,7 +91,9 @@ public static HealthCheck addAndPrune(UUID universeUUID, Long customerId, String
     check.idKey = HealthCheckKey.create(universeUUID);
     check.customerId = customerId;
     // Validate it is correct JSON.
-    check.detailsJson = Json.stringify(Json.parse(details));
+    ObjectMapper objMapper = new ObjectMapper();
```
sb-yb, line 94: Just use `Json.fromJson`

sb-yb: Avoid creating new ObjectMapper if you are doing that in any of the diffs in flight.

sb-yb (changes requested):

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
mahendranbhat: No

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
mahendranbhat: Not sure, thereis no unit tests also.

sb-yb (approved):
Just add the todo mentioning which test was the issue if this is test only issue.
Otherwise looks good.

managed/src/main/java/com/yugabyte/yw/controllers/UniverseController.java
sb-yb: Either add unitest tests or run the platform create universe and check if health checking functionality works

managed/src/main/java/com/yugabyte/yw/models/HealthCheck.java
SergeyPotachev: Controversial point: Static constants are preferable before string constants used in several places.

sb-yb: Been long time gone ... but see changed code. Basically use first class object.

PR #8856, opened 2021-06-11
[Platform] Remove commons-collections4 and use guava instead

bhavin192:
This is a follow up change from https://github.com/yugabyte/yugabyte-db/pull/8371

Scenarios tested:
- Ran platform inside Kubernetes, curl the API, it returns the expected JSON.

### DIFF
--- a/managed/build.sbt
+++ b/managed/build.sbt
@@ -169,8 +169,7 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-core" % "2.10.5",
   "com.jayway.jsonpath" % "json-path" % "2.4.0",
   "commons-io" % "commons-io" % "2.8.0",
-  "commons-codec" % "commons-codec" % "1.15",
-  "org.apache.commons" % "commons-collections4" % "4.4"
+  "commons-codec" % "commons-codec" % "1.15"
 )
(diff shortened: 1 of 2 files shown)


sb-yb (approved):
Thanks! Unrelated rant about your comment:
```
// Performs discovery of region, zones, pull secret, storageClass  
// when running inside a Kubernetes cluster. Returns the discovered  
// information as a JSON, which is similar to the one which is  
// passed to the createKubernetes method.
```
How much effort to change the other method to accept this form data? You say it is "similar". If its not much difference then lets bridge the gap. 
You can leave it in different diff and do when time permits but please track it in a bug.

bhavin192:
@sb-yb good question. The thing is, some of the fields need to come from the frontend. Frontend is going to consume this API, add or modify some of the field (user might also modify). These are things like
- The provider type (in Kubernetes we have OpenShift, Tanzu, GKE etc). 
- Lat Lon of the region (the region name might also change, as the frontend has a different way to name the regions).

Frontend -> Discovery API -> show pre-filled data to user -> User modifies or just accepts -> frontend calls the provider creation API.

The main issue where we are tracking this feature is still open, I will add the details there, as few things might change when the UI part of this is implemented.

Issue #6859, opened 2021-01-13
[Platform] Preflight checks for GCS backup storage configuration


streddy-yb:
At the time of creating GCS backup configuration, do a pre-flight check whether the backup configuration is valid before saving the config.

When user initiates a manual backup -or- a scheduled backup, check whether GCS bucket is still accessible. This check can be performed on the platform side.

PR #8870, opened 2021-06-12
[#6859] [Platform] Preflight checks for GCS backup storage configuration 

jitendra-12113:
**Heading:**
[#6859]Platform: Pre-flight checks for GCS backup storage configuration
**Description:**
We don't have any check to validate the GCP creds and GCP bucket name.
At the time of creating new gcp backup storage configuration, need to do a pre-flight check
to verify whether config details are valid or not.

**Test Plan:**
 Go to config -> try to configure GCP backup storage configurations with invalid gcp bucket name or gcp creds json
 We should get appropriate message. 
 With valid config details, we should able to proceed.


managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
```diff
@@ -229,6 +238,56 @@ protected void doValidate(String value, ObjectNode errorsCollector) {
...
+    protected final String fieldName;
+
+    public ConfigGCSPreflightCheckValidator(String type, String name, String fieldName) {
+      super(type, name);
+      this.fieldName = fieldName;
+    }
+
+    @Override
+    public void doValidate(JsonNode data, ObjectNode errorJson) {
+      if (this.name.equals("GCS") && data.get(GCS_CREDENTIALS_JSON_FIELDNAME) != null) {
```
SergeyPotachev, line 252: NAME_GCS

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
```diff
@@ -229,6 +238,56 @@ protected void doValidate(String value, ObjectNode errorsCollector) {
...
+                GoogleCredentials.fromStream(
+                    new ByteArrayInputStream(gcpCredentials.getBytes("UTF-8")));
+            Storage storage =
+                StorageOptions.newBuilder().setCredentials(credentials).build().getService();
+            Page<Blob> blobs =
+                storage.list(
+                    bucketName,
+                    Storage.BlobListOption.prefix(prefix),
+                    Storage.BlobListOption.currentDirectory());
+            final List<Blob> blobItems = new ArrayList<>();
```
SergeyPotachev, line 275: Is it the complete logic and we only need to check that "blobs" has any data? If yes, then we don't need the `blobItems` list and don't need the loop.

SergeyPotachev (changes requested):
Looks very good but need to clarify a small moment.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
jitendra-12113: Yeah but I don't see other way to check whether **blobs** has some **blob** or not.
https://googleapis.dev/java/gax/latest/com/google/api/gax/paging/Page.html
Please suggest! Thanks!

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
jitendra-12113: Done.

managed/src/main/java/com/yugabyte/yw/models/helpers/CustomerConfigValidator.java
SergeyPotachev: Which type has blob.iterateAll()? Does it have `size()`, `hasNext()` or something like this?

jitendra-12113: blob.iterateAll() has java.lang.Iterable<ResourceT> type.


It has getNextPage(), hasNextPage(), getValues(), getNextPageToken() but that don't seem to be working in this case. Already tried them.

SergeyPotachev (approved):
Good work!

Issue #8871, opened 2021-06-13
[ybase] Pass timeout to the Session object in FlushIndexBatchIfRequired()


tedyu:
Currently the Session instance used by FlushIndexBatchIfRequired() doesn't set timeout.

This change sets the client timeout to the Session instance.

PR #8925, opened 2021-06-16
[#8871] Adding mechanism to push number of rows processed to master

ConcurrencyPractitioner:


ConcurrencyPractitioner:
Note: Still in progress, so this is not done yet.

Issue #8809, opened 2021-06-08
Editing fields is only available alternatively when we click 'Edit Configuration'


Arpit-yb:
**Description**
On the 2.6 portal , the edit configuration button under cloud provider config makes editing available alternatively.

**Steps to repro**
1. log in 2.6 portal
2. Go to Configs -> Backup -> amazon S3
3. Click on edit Configuration 
4. Click on Update without making any changes
5. Click on Edit configuration again
6. The field are not editable
<img width="1792" alt="Screenshot 2021-06-08 at 10 37 17 AM" src="https://user-images.githubusercontent.com/54769894/121126665-29d4a600-c846-11eb-90a9-f8fb28c78f5a.png">

VijiYB:
cc @streddy-yb @ymahajan

hsiaosu-yb:
Seens to work fine on portal right now.  Please double check the latest version to see if this bug reproduces.

Arpit-yb:
I can still reproduce this issue in 2.6 portal

nishantSharma459:
The same needs to be done in **`2.4`** as well.

streddy-yb:
hi @Arpit-yb - does this repro in 2.4 as well?

Arpit-yb:
Yes , i just checked and i am able to repro this bug in 2.4 as well
Link - http://portal2.4.dev.yugabyte.com/releases

nishantSharma459:
@streddy-yb  Will take a look at 2.4 as well and raise a PR by EOD.

Arpit-yb:
**Validated in the edit backup configuration in following scenarios**

1. Created a backup
2. Edited the backup 
3. Clicked Update(backup gets updated successfully)
4. Clicked edit config
5. Clicked Cancel(Previous state is retained)
6. Clicked Edit again
7.  Fields are editable

Works correctly

PR #8975, opened 2021-06-18
[#8809] [Platfrom] [Backport-2.4] Editing fields is only available alternatively when we click 'Edit Configuration'.

nishantSharma459:
On the 2.4 portal, the edit configuration button under cloud provider config makes editing available alternatively.

Test Plan:

1. login 2.4 portal
2. Go to Configs -> Backup -> Amazon S3
3. Click on edit Configuration
4. Click on Cancel without making any changes
5. Click on Edit configuration again
6. The fields are editable. If yes then we're good to go.


yb-andrew (approved):
LG

Issue #9010, opened 2021-06-22
[DOC] cosmetic updates in sharding, replication, and explicit-locking flows


srinivasa-vasu:
Updates in the following links,
https://docs.yugabyte.com/latest/explore/transactions/explicit-locking/
```
<<<<<<< UPSTREAM
Commit fails
======= LOCAL_UPDATE
Commit succeeds
>>>>>>>
```
https://docs.yugabyte.com/latest/architecture/docdb-sharding/sharding/
```
<<<<<<< UPSTREAM
This can be mitigated in some cases where the distribution is keys is known ahead of time by presplitting the table into multiple shards, however this is hard in practice
======= LOCAL_UPDATE
This can be mitigated in some cases where the distribution of keys is known ahead of time by pre-splitting the table into multiple shards. However, this is hard in practice
>>>>>>>
```

https://docs.yugabyte.com/latest/architecture/docdb-replication/replication/
```
<<<<<<< UPSTREAM
A fault domains are comprised of a group of nodes that are prone to correlated failures
======= LOCAL_UPDATE
A fault domain comprises a group of nodes that are prone to correlated failures
>>>>>>>

```

PR #9011, opened 2021-06-22
[DOC][#9010] PR - cosmetic updates in sharding, replication, and explicit-locking flows

srinivasa-vasu:
PR - cosmetic updates in sharding, replication, and explicit-locking flows #9010


docs/content/latest/architecture/docdb-replication/replication.md
```diff
@@ -22,7 +22,7 @@ DocDB automatically replicates data synchronously in order to survive failures w
 
 ### Fault domains
 
-A fault domains are comprised of a group of nodes that are prone to correlated failures. Examples of fault domains are:
+A fault domain comprises a group of nodes that are prone to correlated failures. Examples of fault domains are:
```
polarweasel, line 25:
```suggestion
A fault domain is a group of nodes that are prone to correlated failures. Examples of fault domains are:
```

docs/content/latest/architecture/docdb-sharding/sharding.md
```diff
@@ -118,7 +118,7 @@ This type of sharding allows efficiently querying a range of rows by the primary
...
+Firstly, when starting out with a single shard implies only a single node is taking all the user queries. This often results in a database “warming” problem, where all queries are handled by a single node even if there are multiple nodes in the cluster. The user would have to wait for enough splits to happen and these shards to get redistributed before all nodes in the cluster are being utilized. This can be a big issue in production workloads. This can be mitigated in some cases where the distribution of keys is known ahead of time by pre-splitting the table into multiple shards. However, this is hard in practice.
```
polarweasel, line 121:
```suggestion
Firstly, when starting out with a single shard implies only a single node is taking all the user queries. This often results in a database “warming” problem, where all queries are handled by a single node even if there are multiple nodes in the cluster. The user would have to wait for enough splits to happen and these shards to get redistributed before all nodes in the cluster are being utilized. This can be a big issue in production workloads. This can be mitigated in some cases where the distribution of keys is known ahead of time by pre-splitting the table into multiple shards, but this is hard in practice.
```

polarweasel (changes requested):
Few small changes and it's good to go

docs/content/latest/architecture/docdb-sharding/sharding.md
srinivasa-vasu: @polarweasel - these changes are addressed in #9194. I'll close this request.

PR #9097, opened 2021-06-25
[Backport 2.6][#8461][Platform] Added fix for redirecting user to intended url.

gaurav061:
Description: Platform power user feedback: "When I click on a platform URL, but I have to sign in first, I’m then redirected to the home page, rather than to the URL I initially clicked. Any cheap way we can fix that with some referral magic?"

Fix uses search param to pass the redirect page URL.

Test Plan:
Navigate to a valid URL other than / and clear the browser cookies and local storage details, so it expires the session, now try to navigate to some other URL, you will be redirected to the login page and once the valid credential is entered you will land on the original URL instead of the default landing page.


nishantSharma459 (approved):
LGTM.

sunilsutarDCE (approved):
LGTM

Issue #9089, opened 2021-06-24
[docs] yb-admin list_cdc_streams not documented


RunningJon:
When setting up xCluster replication, you need to check to see if there are any tables already configured for replication or not. If not, you run `yb-admin setup_universe_replication` but if there already tables, you add more tables with `yb-admin alter_universe_replication`.
You can check if there is xCluster replication already configured with `yb-admin list_cdc_streams`
This information is missing from https://docs.yugabyte.com/latest/admin/yb-admin/#setup-universe-replication

PR #9100, opened 2021-06-25
[docs] add list_cdc_streams

ddhodge:
Add list_cdc_streams YB-Admin command with notes on use with setup and apply replication (#9089)

@RunningJon in Slack thread: "This looks good."


lizayugabyte (commented):
LGTM

Issue #9103, opened 2021-06-25
[Platform] During universe deployment, pre-flight network check fails when ICMP is blocked


cigoldstein:
When attempting to deploy a universe in 2.4.5, the network check will fail if ICMP is blocked on the network. We can see below that it fails stating that `"Internet Connection": false`:

```
Failed to execute task java.util.concurrent.FutureTask@373cec36, hit error java.lang.RuntimeException: 
-----
Node <redacted nodename>failed preflight checks:
{
  "(Prometheus) /opt/prometheus is writable": true, 
  "(Prometheus) No Pre-existing Node Exporter Running": true, 
  "Internet Connection": false,
  "SSH Connection": true, 
  "(Prometheus) /var/run/prometheus is writable": true, 
  "(Prometheus) /etc/prometheus is writable": true, 
  "Sudo Access to Python": true, 
  "(Prometheus) /lib/systemd/system/node_exporter.service is writable": true, 
  "Yugabyte User in Yugabyte Group": true, 
  "(PAM Limits) /etc/security/limits.conf is writable": true, 
  "Try Ansible Command": true, 
  "(Mount Point) /app is writable": true, 
  "(Prometheus) /var/log/prometheus is writable": true, 
  "(Prometheus) /var/lib/prometheus is writable": true
}
-----
```

Ping test to `yugabyte.com` fails on this network (expected, since ICMP is blocked):

```
[root@<redacted> ~]# ping www.yugabyte.com
PING www.yugabyte.com (104.26.4.123) 56(84) bytes of data.
^C
--- www.yugabyte.com ping statistics ---
4 packets transmitted, 0 received, 100% packet loss, time 2999ms
```

Network connectivity works on 80/443:
```
[root@<redacted> ~]# nc -v www.yugabyte.com 80
Ncat: Version 7.50 ( https://nmap.org/ncat )
Ncat: Connected to 104.26.5.123:80.
^C

[root@<redacted> ~]# nc -v www.yugabyte.com 443
Ncat: Version 7.50 ( https://nmap.org/ncat )
Ncat: Connected to 172.67.73.218:443.
```

RunningJon:
Maybe a more appropriate check would be a curl which uses TCP rather than ping which uses ICMP.

tylarb:
Linked PR - https://github.com/yugabyte/yugabyte-db/pull/9133

ashu0008:
I see internet check is implemented using tcp check. 

[...]

PR #9133, opened 2021-06-29
[Platform][Issue #9103] Update "preflight_checks.sh" from ping to tcp check.

cigoldstein:
- The existing check for internet connectivity runs a "ping" check to "yugabyte.com".

- This can be problematic since enterprise customers often block icmp, even when there is internet access.

- The new check uses "/dev/tcp" to try to reach "yugabyte.com" on port 443. This was chosen in favor of netcat since there's no assurance that netcat will be present on the system.

### DIFF
--- a/managed/devops/opscli/ybops/data/preflight_checks.sh
+++ b/managed/devops/opscli/ybops/data/preflight_checks.sh
@@ -31,7 +31,7 @@ preflight_provision_check() {
     PORT=443
 
     for i in 1 2 3; do
-      timeout 3 bash -c "cat < /dev/null > /dev/tcp/$HOST/$PORT" --preserve-status && break
+      timeout 3 bash -c "cat < /dev/null > /dev/tcp/${HOST}/${PORT}" --preserve-status && break
     done
     update_result_json_with_rc "Internet Connection" "$?"
   fi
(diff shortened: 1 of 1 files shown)


managed/devops/opscli/ybops/data/preflight_checks.sh
```diff
@@ -25,10 +25,13 @@ preflight_provision_check() {
...
-    # Send 3 packets with 3 second timeout and return success if any succeed. Do not send
-    # multiple packets at once since ping will return an error if any packet fails.
+    # Attempt to run "/dev/tcp" 3 times with a 3 second timeout and return success if any succeed.
+    # --preserve-status flag will maintain the exit code of the "/dev/tcp" command.
+    HOST="yugabyte.com"
+    PORT=443
+
     for i in 1 2 3; do
-      ping -c 1 -W 3 www.yugabyte.com && break
+      timeout 3 bash -c "cat < /dev/null > /dev/tcp/$HOST/$PORT" --preserve-status && break
```
WesleyW, line 34: This syntax would be better:
`/dev/tcp/${HOST}/${PORT}`

WesleyW (approved):
Thanks for taking a look at this!

managed/devops/opscli/ybops/data/preflight_checks.sh
cigoldstein: Happy to help! I've applied the change above.

WesleyW (approved):

PR #9105, opened 2021-06-25
[docs] PR for Restructure doc site wrt Ecosystem Integration (Doc8860)

lizayugabyte:

PR #9143, opened 2021-06-29
[docs] Minor changes based on review comments in the now-closed PR 9105

lizayugabyte:
See https://github.com/yugabyte/yugabyte-db/pull/9105

### DIFF
--- a/docs/content/latest/integrations/janusgraph.md
+++ b/docs/content/latest/integrations/janusgraph.md
@@ -26,8 +26,13 @@ $ ycqlsh
 Connected to local cluster at 127.0.0.1:9042.
 [ycqlsh 5.0.1 | Cassandra 3.9-SNAPSHOT | CQL spec 3.4.2 | Native protocol v4]
 Use HELP for help.
-ycqlsh> DESCRIBE KEYSPACES;
+```
 
+```shell
+ycqlsh> DESCRIBE KEYSPACES;
+```
+
+```
 system_schema  system_auth  system
 
 ycqlsh>
--- a/docs/content/latest/integrations/janusgraph.md
+++ b/docs/content/latest/integrations/janusgraph.md
@@ -28,11 +28,11 @@ Connected to local cluster at 127.0.0.1:9042.
 Use HELP for help.
 ```
 
-```shell
+```sql
 ycqlsh> DESCRIBE KEYSPACES;
 ```
 
-```
+```output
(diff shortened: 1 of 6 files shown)


docs/content/latest/integrations/janusgraph.md
```diff
@@ -26,8 +26,13 @@ $ ycqlsh
 Connected to local cluster at 127.0.0.1:9042.
 [ycqlsh 5.0.1 | Cassandra 3.9-SNAPSHOT | CQL spec 3.4.2 | Native protocol v4]
 Use HELP for help.
+```
+
+```shell
```
polarweasel, line 31: All ysqlsh and ycqlsh commands should be tagged as sql...

````suggestion
```sql
````

File: docs/content/latest/integrations/janusgraph.md [hunk 2]
```diff
@@ -26,8 +26,13 @@ $ ycqlsh
...
 ycqlsh> DESCRIBE KEYSPACES;
+```
 
+```
```
polarweasel, line 35:
````suggestion
```output
````

File: docs/content/latest/integrations/kairosdb.md [hunk 3]
```diff
@@ -73,7 +73,7 @@ Use HELP for help.
 ycqlsh>
 ```
 
-- Run a cql command to verify it is working:
+- Run the following YCQL commands to verify it is working:
 
 ```sql
 ycqlsh> describe keyspaces;
```
polarweasel, line 79: Our docs are so inconsistent with uppercase vs lowercase statements. And I know the interpreters are case-insensitive, but it's definitely something of a religious war on the order of vi vs emacs. Bryn hates uppercase, so I suppose that's as good a reason as any to standardize on lowercase. ;-)

(I think the main argument for uppercase is making the statements stand out, but isn't that what syntax highlighting is for in modern docs?)

polarweasel (changes requested):
Thanks for doing this! I've left a couple of suggestions and comments.

docs/content/latest/integrations/janusgraph.md
lizayugabyte: Thanks. Changed.

docs/content/latest/integrations/janusgraph.md
lizayugabyte: fixed

Issue #8969, opened 2021-06-17
[Platform] [API] [UI] API portal whitelabeling


sb-yb:
This is for whitelabeling and any other look and feel changes for swagger-ui available on `/docs/`
The assets are currently checked in under `/managed/ui/public/swagger-ui/`
Please discuss with @ymahajan for ideas/products requirements.

sb-yb:
Some background:
So far I have made only minor tweaks like  changed favicon that shows up on chrome tab to yb and changed the default swagger.json to /swagger.json instead of petstore example.

Some thoughts 
Looks like there are various way to change layouts etc. 
There are logos that we should fix.
FWIW there is also 
https://www.npmjs.com/package/swagger-ui-express
in case you want to completely swap it out for node module (You are on your own to get it working :))

jaydeepkumara:
On hold for now, Need to wait to merge all swagger annotation. 
Reason:- New UI uses URL hash(http://localhost:9000/docs/#/operations, operations is hash for this API). new UI swagger uses the hash key as API function name. There are multiple places where we have used the same function name in API(for example list and index have used a lot of places as a function name).  To overwrite we required to use the nickname at the API swagger annotation key. Few API annotations are in review once that get merge I need to figure out all places where nicknames required and implement it.

PR #9189, opened 2021-07-06
[#8969][Platform] [API] [UI] API portal whitelabeling 

jaydeepkumara:
Summary:- This is for whitelabeling and any other look and feel changes for swagger-ui available on /docs/
The assets are currently checked in under /managed/ui/public/swagger-ui/
New UI look and feel:- https://stoplight.io/open-source/elements/
Implementation :- https://meta.stoplight.io/docs/elements/docs/getting-started/usage/web-component.md

Test Plan:- 
1. Open Swagger UI. try to create, update, list etc.
2. Validate operationId for all API are differ as I use operationId to disguise API.

**New Swagger UI screenshot**

Screenshot 2021-07-09 at 10 06 26 PM

**Delete User API**

Screenshot 2021-07-10 at 1 22 17 AM

**Add User API**
Screenshot 2021-07-10 at 1 21 34 AM


managed/src/main/java/com/yugabyte/yw/controllers/CloudProviderController.java
```diff
@@ -42,6 +42,7 @@
    *
    * @return JSON response of newly created provider
    */
+  @ApiOperation(value = "createCloudProvider", nickname = "createCloudProvider")
```
sb-yb, line 45: I think someday I can change this to generate operationId automatically. For now lets live with this.

sb-yb:
This looks much better that swagger-UI. 
High level question: 
  Is there a way to add this as a library dependency and configure it?
For example with swagger-UI I was going to start using the library and only add swagger-config.yaml. See https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md

Do we have equivalent so we don't have to add ton of third party code to our repo.

managed/ui/public/swagger-ui/index.html
```diff
@@ -4,74 +4,23 @@
...
-        deepLinking: true,
-        presets: [
-          SwaggerUIBundle.presets.apis,
-          SwaggerUIStandalonePreset
-        ],
-        plugins: [
-          SwaggerUIBundle.plugins.DownloadUrl
-        ],
-        layout: "StandaloneLayout",
-        requestInterceptor: function(req) {
```
sb-yb, line 66: See this. I had fixed the CSRF issue so swagger-ui could be used to try out post and put requests as well. 
Does stoplight take care of it internally?

sb-yb (changes requested):
Just one comment about applying csrf fix.

managed/ui/public/swagger-ui/index.html
jaydeepkumara: Yes. I have tested delete and post API.

jaydeepkumara: Worked Added screenshot.

sb-yb (approved):

Issue #8152, opened 2021-04-23
Enforce configured password policy on Platform UI


anmalysh-yb:
Configurable password policy was implemented here on backend side: https://github.com/yugabyte/yugabyte-db/issues/7939
However, on UI side it's still hardcoded.
Need to get configured password policy from backend and apply it on UI side.
Password policy can be retrieved from CustomerConfigController as CustomerConfig object with type==PASSWORD_POLICY.

rajmaddy89:
@streddy-yb or @iSignal : Can you please assign this issue to my name

anmalysh-yb:
Actually, we have at least 3 issues with CustomerConfigController:
1. It have different methods to create and update config. Not transient for the client.
2. It has no method to get configured values OR default. Without that - UI code will need to have defaults hardcoded.
3. It’s raw json - the only way to find out the exact object which need to be passed to the API - is to dig into backend code.

I’d say that at least #2 should be addressed on the backend side
These are common issues for all types of customer config - so probably there should be some common fix.

rajmaddy89:
@anmalysh-yb : I am Raj and I just started working on this issue, the below is my plan for this story, do you have any thoughts on the steps:

1. Create a method in `PasswordPolicyService.java` to return(expose) PolicyPasswordFormData
2. Create a new endpoint in `CustomerConfigController.java `which should call the above function 
3. The above end point should return a plain old Java Object which will contain the policy information of password
4. Now client side REGISTER function in `register.js` should call the above API which in turn will give us the policy and we will show validation based on that

rajmaddy89:
PR: https://github.com/yugabyte/yugabyte-db/pull/9210

rajmaddy89:
Can this issue be closed now ?

anmalysh-yb:
I took a quick look into code review.
It seems like UI code is still using old regexp for password policy check:

https://github.com/yugabyte/yugabyte-db/pull/9210/files#diff-62db5af8d2a399fe2d5ebe276571f87df6d5c5c962b02265556aee399ef1a541L47

Basically you just changed the error message.

Also, One more place which applies password policy on UI is UserProfileForm.js - which is not covered in your PR.

One more note - your getPasswordPolicyData() method in PasswordPolicyService seems weird.
It should get policy, configured for customer first (so at least it should get customer ID as an input). And only if it's missing - should use default values from config object.
[...]

PR #9210, opened 2021-07-08
[#8152] [Platform] Enforce configured password policy

rajmaddy89:
Description
--
However, on UI side it's still hardcoded.
Need to get configured password policy from backend and apply it on UI side.
Password policy can be retrieved from CustomerConfigController as CustomerConfig object with type==PASSWORD_POLICY.
This should apply for `Register form` and `Add User Modal`

Changes
--
On a high level
1. Create a method in PasswordPolicyService.java to return(expose) PolicyPasswordFormData
2. Create a new endpoint in CustomerConfigController.java which should call the above function
3. The above end point should return a plain old Java Object which will contain the policy information of password
4. Now client side REGISTER function in register.js should call the above API which in turn will give us the policy and we will show validation based on that


rajmaddy89:
I have covered mostly here: However we need to get Password Policy from CustomerConfig object with type==PASSWORD_POLICY, which needs CustomerId as input.

My question is when a customer registers through `RegisterForm` or `AddModal` form, CustomerId is always `null` as it is a new customer in this case, how do we call the CustomerConfig in this case.

Please refer to my work and suggest the input there

managed/src/main/resources/v1.routes
```diff
@@ -17,6 +17,7 @@ PUT     /customers/:cUUID/api_token                                            c
 GET     /logout                                                                com.yugabyte.yw.controllers.SessionController.logout()
 + nocsrf
 POST    /register                                                              com.yugabyte.yw.controllers.SessionController.register()
+GET     /validate                                                              com.yugabyte.yw.controllers.SessionController.validate()
```
yb-andrew, line 20: Can we update the naming to be more specific? Should be something like `password_policy` not `validate`, which implies a POST request.

managed/ui/src/components/common/forms/RegisterForm/RegisterForm.js
```diff
@@ -43,9 +50,12 @@ class RegisterForm extends Component {
...
       password: Yup.string()
         .required('Enter password')
-        .min(8, 'Password is too short - must be 8 characters minimum.')
+        .min(minPasswordLength, `Password is too short - must be ${minPasswordLength} characters minimum.`)
         .matches(/^(?=.*[0-9])(?=.*[!@#$%^&*])(?=.*[a-z])(?=.*[A-Z])[a-zA-Z0-9!@#$%^&*]{8,256}$/,
-          'Password must contain at least 1 digit, 1 capital, 1 lowercase and one of the !@#$%^&* (special) characters.'),
+          `Password must contain at least ${passwordValidationInfo?.minDigits} digit
+          , ${passwordValidationInfo?.minUppercase} capital
+          , ${passwordValidationInfo?.minLowercase} lowercase
+          and ${passwordValidationInfo?.minSpecialCharacters} of the !@#$%^&* (special) characters.`),
```
yb-andrew, line 58: Nice 👍

yb-andrew (changes requested):
Minor comments, everything else looks good.

managed/src/main/resources/v1.routes
rajmaddy89: @yb-andrew : I have resolved the issue, changed the naming part and tested it locally. Please re-review, thanks !

managed/ui/src/actions/customers.js
```diff
@@ -13,6 +13,10 @@ export const VALIDATE_FROM_TOKEN_RESPONSE = 'VALIDATE_FROM_TOKEN_RESPONSE';
 export const REGISTER = 'REGISTER';
 export const REGISTER_RESPONSE = 'REGISTER_RESPONSE';
 
+// Validate Customer registration
+export const VALIDATE = 'VALIDATE';
+export const VALIDATE_RESPONSE = 'VALIDATE_RESPONSE';
```
yb-andrew, line 18: Can you also change these action types to reflect what you are doing: `FETCH_PASSWORD_POLICY` seems like a better choice of wording.

yb-andrew (approved):
LG. Can you change the action types and action creation function before merging?
`VALIDATE` -> `FETCH_PASSWORD_POLICY`
`function validate()` -> `function fetchPasswordPolicy()`

rajmaddy89:
@yb-andrew : I agree with you, I have done this change yesterday :-) I will push the change shortly

yb-andrew (approved):
LG! Thanks for making the changes.

rajmaddy89:
@yb-andrew : Can I know what is the merge blocker here, do I need to do anything here ?

yb-andrew:
@rajmaddy89 I approved running the Github workflows. Can you run `sbt javafmt` in the `/managed` directory in order to format the Java files for our linter?

rajmaddy89:
Sure @yb-andrew , will do it by EOD

managed/src/main/resources/reference.conf
```diff
@@ -10,7 +10,7 @@ db {
     host="localhost"
     port=5432
     dbname="yugaware"
-    username="postgres"
+    username="root"
```
yb-andrew, line 13:
```suggestion
    username="postgres"
```

managed/ui/src/components/common/forms/RegisterForm/RegisterForm.js
```diff
@@ -23,14 +29,27 @@ class RegisterForm extends Component {
...
-    const { registerCustomer } = this.props;
+    const config = {
+      type: 'PASSWORD_POLICY',
+      name: 'password policy',
+      data: {
+        minLength: 8,
+        minUppercase: 1,
+        minLowercase: 1,
+        minDigits: 1,
+        minSpecialCharacters: 1
```
yb-andrew, line 40: We should be storing the password configuration that the backend returns and not hard-coding it here.

managed/ui/src/components/profile/modals/AddUserModal.js
```diff
@@ -17,7 +18,19 @@ export class AddUserModal extends Component {
...
     try {
+      const config = {
+        type: 'PASSWORD_POLICY',
+        name: 'password policy',
+        data: {
+          minLength: 8,
+          minUppercase: 1,
+          minLowercase: 1,
+          minDigits: 1,
+          minSpecialCharacters: 1
```
yb-andrew, line 29: Same comment as above.

project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.4.6
```
yb-andrew, line 1: Revert?

yb-andrew (changes requested):
Please address the review comments and remove the changes to the configuration files: `build.properties` and `reference.conf`

managed/ui/src/components/common/forms/RegisterForm/RegisterForm.js
rajmaddy89: @yb-andrew : Just a question, I agree we should not hardcode the config data here, but we do need to pass the config data that needs to be stored in backend. Do I need to pass the above config data in the `[create](https://github.com/yugabyte/yugabyte-db/blob/master/managed/src/main/java/com/yugabyte/yw/controllers/CustomerConfigController.java#L46)` function in `CustomerConfigController.java` when the formData type is `PASSWORD_CONFIG` ?

rajmaddy89: Background of why I add the config above:

So, while customer is not yet registered - it can't have custom password policy configured - so default one is used at this point.
We should use customer-specific policy only when customer is already created.
So we can create customer policy via the following API once customer is registered:

POST /api/v1/customers/<customer_uuid>/configs

with the body like that:

{
  "name": "password policy",
  "type": "PASSWORD_POLICY",
  "data": {
     "minLength": 8,
     "minUppercase": 1,
     "minLowercase": 1,
     "minDigits": 1,
     "minSpecialCharacters": 1
  }
}

This is what I have done, please let me know

yb-andrew: @rajmaddy89 When the customer is not registered, there is no customerUUID, but the password policy can be retrieved from the server's configuration (kind of like the environment variables). This code should look similar to the original code in `PasswordPolicyService.java` file. Here is pseudo-code
```
PasswordPolicyFormData effectivePolicy;
if (configuredPolicy == null) { // If customer is new, there will not be a configured policy yet
      effectivePolicy = new PasswordPolicyFormData();
      effectivePolicy.setMinLength(config.getInt(DEFAULT_MIN_LENGTH_PARAM));
      effectivePolicy.setMinUppercase(config.getInt(DEFAULT_MIN_UPPERCASE_PARAM));
      effectivePolicy.setMinLowercase(config.getInt(DEFAULT_MIN_LOWERCASE_PARAM));
      effectivePolicy.setMinDigits(config.getInt(DEFAULT_MIN_DIGITS_PARAM));
      effectivePolicy.setMinSpecialCharacters(config.getInt(DEFAULT_MIN_SPECIAL_CHAR_PARAM));
}
```

yb-andrew (approved):
Changes LG, thanks!

Issue #8565, opened 2021-05-22
[Platform] Support deploying multiple k8s universes in the same namespace


iSignal:
Currently, Platform allows specifying the namespace in the k8s provider to deploy the k8s universe in but doesn't allow more than one universe to be deployed per namespace. This is because of the older naming convention of YB helm charts which doesn't include the release name in some of the k8s object names.

https://github.com/yugabyte/charts/commit/1b0a823f2b5abe0c64c15251ce6511c61e7ec108 introduced a newer naming strategy for helm that allows two YB chart deployments in the same namespace. Using that would allow us to deploy multiple universes in the same namespace.

PR #9273, opened 2021-07-13
[Platform] Changes to allow multiple Helm releases in one namespace 

bhavin192:
Adds a new universe configuration key HELM_NAMING_STYLE, the new naming style is enabled for all the YugabyteDB installations which are newer than 2.7.2.0-b113, as this version has Helm chart related changes required for this naming to work.

Make changes to Create, Update, Edit flows, backup and healthchecks scripts, metrics accordingly.

When we try to create multiple Helm releases in one namespace, we try to apply the same pull secret in same namespace in parallel. There is a race condition with `kubectl apply` on the client side (kubectl side), it fails with an error saying the secret already exist. This is fixed by using server side apply, the flag is experimental in 1.15 but from Kubernetes 1.16 the feature is marked as beta and hence available everywhere.

I will update the commit message, add test scenarios, once an initial review is done. 

Comments marked with `TODO(bhavin192): REVIEW:` are the ones where I would like to have opinion or comment from reviewer.

Fixes https://github.com/yugabyte/yugabyte-db/issues/8565

### DIFF
--- a/managed/devops/bin/cluster_health.py
+++ b/managed/devops/bin/cluster_health.py
@@ -180,14 +180,10 @@ class KubernetesDetails():
     def __init__(self, node_fqdn, config_map):
         self.namespace = node_fqdn.split('.')[2]
         self.pod_name = node_fqdn.split('.')[0]
-        # The pod names are yb-master-n/yb-tserver-n where n is the pod number
-        # and yb-master/yb-tserver are the container names.
-
-        # TODO(bhavin192): need to change in case of multiple releases
-        # in one namespace. Something like find the word 'master' in
-        # the name.
-
-        self.container = self.pod_name.rsplit('-', 1)[0]
+        # The pod names are <helm fullname>yb-<master|tserver>-n where
+        # n is the pod number. <helm fullname> can be blank. And
+        # yb-master/yb-tserver are the container names.
+        self.container = "yb-master" if self.pod_name.find("master") > 0 else "yb-tserver"
         self.config = config_map[self.namespace]
 
 
--- a/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/EditKubernetesUniverse.java
+++ b/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/EditKubernetesUniverse.java
@@ -89,22 +89,46 @@ public class EditKubernetesUniverse extends KubernetesTaskBase {
 
       currPlacement = new KubernetesPlacement(currPI);
 
+      // TODO(bhavin192): FIX ME: should this be a taskParam?
+      boolean newNamingStyle = universe.usesHelmNewNamingStyle();
+
+      String nodePrefix = taskParams().nodePrefix;
+
       Set<NodeDetails> mastersToAdd =
-          getPodsToAdd(newPlacement.masters, currPlacement.masters, ServerType.MASTER, isMultiAz);
+          getPodsToAdd(
+              newPlacement.masters,
+              currPlacement.masters,
+              ServerType.MASTER,
+              nodePrefix,
+              isMultiAz,
+              newNamingStyle);
       Set<NodeDetails> mastersToRemove =
           getPodsToRemove(
-              newPlacement.masters, currPlacement.masters, ServerType.MASTER, universe, isMultiAz);
+              newPlacement.masters,
+              currPlacement.masters,
+              ServerType.MASTER,
+              universe,
+              nodePrefix,
+              isMultiAz,
+              newNamingStyle);
 
       Set<NodeDetails> tserversToAdd =
           getPodsToAdd(
-              newPlacement.tservers, currPlacement.tservers, ServerType.TSERVER, isMultiAz);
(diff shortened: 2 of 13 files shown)


managed/src/main/java/com/yugabyte/yw/forms/UniverseDefinitionTaskParams.java
```diff
@@ -139,6 +139,11 @@ public ClusterType getCurrentClusterType() {
   // EDIT mode: Set to true if nodes could be resized without full move.
   @ApiModelProperty public boolean nodesResizeAvailable = false;
 
+  // This flag indicates whether the Kubernetes universe will use new
+  // naming style of the Helm chart. Default is set to false for
+  // backward compatability.
+  @ApiModelProperty public boolean useHelmNewNamingStyle = false;
```
iSignal, line 145: minor: useNewHelmNamingStyle

managed/src/main/resources/reference.conf
```diff
@@ -56,6 +56,7 @@ db {
 yb {
   mode="PLATFORM"
   use_kubectl=true
+  use_helm_new_naming=false
```
iSignal, line 59: I would also add this to the runtime config section below, which allows setting it without a restart. Setting it would change the behavior for new univs created from the non.

managed/src/main/java/com/yugabyte/yw/common/PlacementInfoUtil.java
```diff
@@ -2300,14 +2307,46 @@ public static Map<String, String> getConfigPerNamespace(
     return namespaceToConfig;
   }
 
+  /**
+   * Returns kubeconfig for each Pod's FQDN from the nodeDetailsSet. This method is useful for both
```
iSignal, line 2311: minor: Returns a map `pod_fqdn -> KUBECONFIG string` for all pods in nodeDetailSet

managed/src/main/java/com/yugabyte/yw/common/PlacementInfoUtil.java
```diff
@@ -2300,14 +2307,46 @@ public static Map<String, String> getConfigPerNamespace(
...
+      if (kubeconfig == null) {
+        throw new NullPointerException("Couldn't find a kubeconfig");
+      }
+      azToKubeconfig.put(entry.getKey(), kubeconfig);
+    }
+
+    for (NodeDetails nd : nodeDetailsSet) {
+      String kubeconfig = azToKubeconfig.get(nd.azUuid);
+      if (kubeconfig == null) {
+        throw new NullPointerException("Couldn't find a kubeconfig");
```
iSignal, line 2334: minor: `"Couldn't find a kubeconfig for AZ " + azUUID

managed/src/main/java/com/yugabyte/yw/commissioner/tasks/EditKubernetesUniverse.java
```diff
@@ -54,6 +54,20 @@ public void run() {
 
       Universe universe = lockUniverseForUpdate(taskParams().expectedUniverseVersion);
 
+      // TODO(bhavin192): FIX ME: is this a better way? There is no
+      // way that one who writes code knows subsequent calls where
+      // this is actually needed i.e. while creating the
+      // KubernetesCommandExecutor tasks. If we go by this route, this
+      // must be verified somewhere down the call tree, so that the
```
iSignal, line 61: The code below which sets it from taskParams is ok. In our case, this is not a userIntent setting which can be changed by the user, so getting it from UI shouldn't matter, user can never edit it.

So we don't need to set it in taskParams here too, we could just pick it up from universeDetails directly.

managed/src/main/java/com/yugabyte/yw/commissioner/tasks/UpgradeKubernetesUniverse.java
```diff
@@ -51,6 +55,20 @@ public void run() {
 
       taskParams().rootCA = universe.getUniverseDetails().rootCA;
 
+      // TODO(bhavin192): FIX ME: is this a better way? There is no
+      // way that one who writes code knows subsequent calls where
+      // this is actually needed i.e. while creating the
+      // KubernetesCommandExecutor tasks. If we go by this route, this
```
iSignal, line 61: see similar comment in EditUniverse.

managed/devops/bin/cluster_health.py
```diff
@@ -210,21 +210,17 @@ class KubernetesDetails():
     def __init__(self, node_fqdn, config_map):
         self.namespace = node_fqdn.split('.')[2]
         self.pod_name = node_fqdn.split('.')[0]
```
iSignal, line 212: minor: I guess even this part will become problematic for multicluster. We need to ideally pass in some kind of a "node-name" -> (ns, pod name, kubeconfig) map in the future

iSignal (approved):
Awesome, thanks for all the iterations on it. Let's enable this by default in a succeeding diff after verifying any remaining itests.

PR #9299, opened 2021-07-14
Update migrate-application.md

kerneljake:
consistency between examples and python3 syntax


rkarthik007 (approved):
Nice catch! Thanks for the fix!

rkarthik007:
@polarweasel - will let you merge this in, fix looks good.

PR #9348, opened 2021-07-18
Release lock before program exits

jenny-cheung:
A fix patch for issue#9324 to avoid the unreleased lock.

### DIFF
--- a/src/postgres/src/test/thread/thread_test.c
+++ b/src/postgres/src/test/thread/thread_test.c
@@ -147,7 +147,8 @@ main(int argc, char *argv[])
 	{
 		fprintf(stderr, "Failed to create thread 1: %s **\nexiting\n",
 				strerror(errno));
-		exit(1);
+        pthread_mutex_unlock(&init_mutex);
+        exit(1);
 	}
 	rc = pthread_create(&thread2, NULL, (void *(*) (void *)) func_call_2, NULL);
 	if (rc != 0)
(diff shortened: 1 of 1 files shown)


src/postgres/src/test/thread/thread_test.c
```diff
@@ -157,6 +158,7 @@ main(int argc, char *argv[])
 		 * 1 that uses it, so avoid using it.
 		 */
 		fprintf(stderr, "Failed to create thread 2 **\nexiting\n");
+        pthread_mutex_unlock(&init_mutex);
```
tedyu, line 161: Please align pthread_mutex_unlock() with the rest of the code (left indent 2 chars).

tedyu:
Please sign CLA.

After indentation is fixed, I will run your patch through test suite.

jenny-cheung:
The indentation has been fixed.

d-uspenskiy (commented):
Hello @jenny-cheung could you please describe the impact of the issue? What is going wrong without your fix?
As you know YugaByte uses postgres fork and this code belong to postgres native code. Latest version of postgres sources has the same code now. There is no mutex unlocking before exit https://github.com/postgres/postgres/blob/5865e064abfbbe11ebfc09881be009c0f69b4dc2/config/thread_test.c#L129

In case it is real problem it will be good to create issue in postgres project as well.

jenny-cheung:
@d-uspenskiy Hi, the impact is that the lock is not correctly released before exit. It is not a very harmful bug( e.g., resource leak, deadlock) since the program exits and all of the program resources will be cleaned up by some common OS. However, other OS systems (e.g., embedded systems) do not automatically free some resources at the exit. It is a good manner for resource management. Adding the unlock statement adds symmetry, so the code looks better. Also, the debugger would not warn this case : )

d-uspenskiy:
Issue was closed by the originator. No fix planned.

PR #9474, opened 2021-07-27
[docs] fix a few 404s

polarweasel:
(I've also started testing a link checker, hence the new gitignore entry)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,6 +113,3 @@ submodules/
 
 # Local Netlify folder
 .netlify/
-
-# HTML link checker temp files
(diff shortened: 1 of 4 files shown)


jaki:
@polarweasel, if you're testing something that only relates to your setup, not anyone else's, consider moving the ignore entry from `.gitignore` to `.git/info/exclude`.

polarweasel:
Thanks for the tip, @jaki! I think this is likely to become part of the docs workflow, but I'll take it out for now.

ddhodge (approved):
LGTM

PR #9486, opened 2021-07-27
Replace YugaByte -> Yugabyte across the board

mbautin:


jaki:
Regress tests should be run because some test output show the github links.

PR #9540, opened 2021-07-30
[docdb] fix issue-4262

### LINKED ISSUE #4262: num_tablet_servers_live metric does not reset when a master's status changes from leader to follower
Impact appears to be simply cosmetic...this can be reproduced by triggering a master leader that does not involve restarting the active leader's yb-master process. Observed in v2.1.3.0.

SGZW:
fix #4262

### DIFF
--- a/src/yb/master/catalog_manager_bg_tasks.cc
+++ b/src/yb/master/catalog_manager_bg_tasks.cc
@@ -175,6 +175,9 @@ void CatalogManagerBgTasks::Run() {
 
       // Start the tablespace background task.
       catalog_manager_->StartTablespaceBgTaskIfStopped();
+    } else {
+      // Reset Metrics when leader_status is not ok.
+      catalog_manager_->ResetMetrics();
     }
     WARN_NOT_OK(catalog_manager_->encryption_manager_->
                 GetUniverseKeyRegistry(&catalog_manager_->master_->proxy_cache()),
--- a/src/yb/master/catalog_manager.cc
+++ b/src/yb/master/catalog_manager.cc
@@ -9150,6 +9150,12 @@ void CatalogManager::ReportMetrics() {
   metric_num_tablet_servers_dead_->set_value(ts_descs.size() - num_live_servers);
 }
 
+void CatalogManager::ResetMetrics() {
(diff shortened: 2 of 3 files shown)


SGZW:
@rahuldesirazu please review this when you are free, thanks

rahuldesirazu:
@SGZW Code looks good! I'll go ahead and run it through our test suite.

src/yb/master/catalog_manager_bg_tasks.cc
```diff
@@ -175,6 +175,9 @@ void CatalogManagerBgTasks::Run() {
 
       // Start the tablespace background task.
       catalog_manager_->StartTablespaceBgTaskIfStopped();
+    } else {
```
rahuldesirazu, line 178: I think there should be a `if (!l.catalog_status().ok())` clause above this, I think its best to put this there.

rahuldesirazu (changes requested):

rahuldesirazu:
@SGZW Actually sorry just had one small comment, will put it through the test suite once its addressed.

src/yb/master/catalog_manager_bg_tasks.cc
SGZW: when (!l.catalog_status().ok()) is true,  catalog-manager status is either kConstructed or kStarting, it means catalog manager is init that metrics already reset, we should do nothing. we only need to clear metrics when raft role change from leader to other state so that reset metrics in branch "else"

SGZW:
> @SGZW Actually sorry just had one small comment, will put it through the test suite once its addressed.

SGZW:
> @SGZW Actually sorry just had one small comment, will put it through the test suite once its addressed.

please recheck my comment, thanks

src/yb/master/catalog_manager_bg_tasks.cc
rahuldesirazu: Sorry for the late reply, that's a fair point. The code looks good and I will put it through our jenkins pipeline.

SGZW: thanks

PR #9576, opened 2021-08-03
[YSQL] Merge user provided shared_preload_libraries to enable custom PSQL extensions

radekg:
This PR enables the `--ysql_pg_conf[_csv]=shared_preload_libraries=...` functionality by merging a user provided list of extensions with the predefined YugabyteDB list of metrics related extensions.

The background to this is the following: I was trying to use a custom PSQL extension with YugabyteDB. I was able to install the extension in YugabyteDB using the method described [Extensions requiring manual installation](https://docs.yugabyte.com/latest/api/ysql/extensions/#extensions-requiring-installation) but regardless of the settings provided, even though `create extension ...;` would result in `CREATE EXTENSION` status, the actual extension would never get invoked.

I've managed to narrow it down to the hard coded list of `shared_preload_libraries`. It seems that Postgres will not invoke an extension if it is not specified in that list, even though it will actually allow installing it with `create extension ...` SQL command.

## Caveats

- I'm not working with C/C++ so this might not be the best idiomatic code.
- I am not sure how to write a proper test for this.

In both cases, guidance is welcomed.

### Other

- This should be pretty straightforward to back port to previous releases. This patch [applies directly to 2.7.2.0, example](https://github.com/radekg/yugabyte-db/tree/2.7.2-merge-user-shared-preload-libraries).

## Testing the PR

Using a toolkit to build YugabyteDB, prepare a Docker image from the PR:

```sh
mkdir -p /tmp/pr-test && cd /tmp/pr-test
git clone https://github.com/radekg/yugabytedb-postgres-extensions-build-infrastructure.git .

## Build the YugabyteDB build infrastructure.
## This will take a bit because it has to build gcc.
## Maybe the gcc make -j has to be adjusted by setting GCC_MAKE_PARALLELISM to value lower than default 32.
make GCC_MAKE_PARALLELISM=32 ybdb-build-infrastructure

## Put the example extension in the third-party postgres location
## so it gets compiled together with the PR.
mkdir -p .tmp/extensions/example
cd .tmp/extensions/example
git clone https://github.com/radekg/yugabytedb-example-extension.git .
cd -

## Build YugabyteDB from the PR:
make YB_REPOSITORY=https://github.com/radekg/yugabyte-db.git \
    YB_SOURCE_VERSION=ysql-merge-user-shared_preload_libraries \
    ybdb-build-first-pass

## Create the distribution:
make YB_RELEASE_VERSION=master-pr-9576 ybdb-distribution

## Create the Docker image:
make YB_RELEASE_VERSION=master-pr-9576 \
    YB_REPOSITORY=https://github.com/radekg/yugabyte-db.git \
    YB_SOURCE_VERSION=ysql-merge-user-shared_preload_libraries \
    YB_RELEASE_DOCKER_VERSION=master-pr-9576 \
    ybdb-build-docker
```

## Validating

Run the container from the resulting Docker image. First, without the `--tserver_flags="ysql_pg_conf=shared_preload_libraries=example"` flag:

```sh
docker run --rm -ti \
    -p 7000:7000 \
    -p 5433:5433 \
    local/yugabytedb:master-pr-9576 yugabyted start --daemon=false --ui=false
```

In another terminal:

```
docker exec -ti $(docker ps | grep 'master-pr-9576' | awk '{print $1}') /bin/bash
bash-4.2$ ysqlsh
ysqlsh (11.2-YB-2.7.3.0-b0)
Type "help" for help.

yugabyte=# create extension example;
CREATE EXTENSION
yugabyte=# \dx
                                     List of installed extensions
        Name        | Version |   Schema   |                        Description
--------------------+---------+------------+-----------------------------------------------------------
 example            | 0.1.0   | public     | Example library
 pg_stat_statements | 1.6     | pg_catalog | track execution statistics of all SQL statements executed
 plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language
(3 rows)

yugabyte=# create role tenant1 with login password 'tenant1' connection limit 10;
CREATE ROLE
yugabyte=# \q
bash-4.2$ ysqlsh -U tenant1
ysqlsh (11.2-YB-2.7.3.0-b0)
Type "help" for help.

yugabyte=> set default_tablespace='pg_default';
SET
yugabyte=> \q
bash-4.2$ exit
exit
```

The `SET` command was confirmed - extension wasn't invoked. Stop the YugabyteDB container.

## Revalidate with user defined `shared_preload_libraries`

```sh
docker run --rm -ti \
    -p 7000:7000 \
    -p 5433:5433 \
    local/yugabytedb:master-pr-9576 yugabyted start --daemon=false --ui=false \
        --tserver_flags="ysql_pg_conf=shared_preload_libraries=example"
```

In another terminal:

```
docker exec -ti $(docker ps | grep 'master-pr-9576' | awk '{print $1}') /bin/bash
bash-4.2$ ysqlsh
ysqlsh (11.2-YB-2.7.3.0-b0)
Type "help" for help.

yugabyte=# create extension example;
CREATE EXTENSION
yugabyte=# \dx
                                     List of installed extensions
        Name        | Version |   Schema   |                        Description
--------------------+---------+------------+-----------------------------------------------------------
 example            | 0.1.0   | public     | Example library
 pg_stat_statements | 1.6     | pg_catalog | track execution statistics of all SQL statements executed
 plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language
(3 rows)

yugabyte=# create role tenant1 with login password 'tenant1' connection limit 10;
CREATE ROLE
yugabyte=# \q
bash-4.2$ ysqlsh -U tenant1
ysqlsh (11.2-YB-2.7.3.0-b0)
Type "help" for help.

yugabyte=> set default_tablespace='pg_default';
ERROR:  operation not allowed: "SET ... default_tablespace"
yugabyte=> create table test (rowval varchar(100)) tablespace pg_default;
ERROR:  operation not allowed: "CREATE TABLE ... TABLESPACE pg_default"
yugabyte=> \q
bash-4.2$ exit
exit
```

The `example` **extension prevented setting** the `default_tablespace` and creating a table in the `pg_default` tablespace. This is what the example extension does.

**This proves that the extension was invoked**. The **user provided preload libraries have been successfully merged with hard coded values** from the `pgwrapper`.

## What should be the test

The premise of the code from this PR can be tested by executing:

```cpp
#include <iostream>
#include <string>
#include <vector>
#include <boost/algorithm/string.hpp>

using namespace std;

  void MergeSharedPreloadLibraries(const string& src, vector<string>* defaults) {
    string copy = boost::erase_first_copy(boost::replace_all_copy(src, " ", ""), "shared_preload_libraries=");
    copy = boost::trim_copy_if(copy, boost::is_any_of("'\""));
    vector<string> new_items;
    boost::split(new_items, copy, boost::is_any_of(","));
    defaults->insert(defaults->end(), new_items.begin(), new_items.end());
  }
 
int main () {

  vector<string> pgConfLines {
    "shared_preload_libraries='given1,given2'",
    "some_other_config_option=aaa"
  };

  vector<string> metricsLibs {
  	"test1",
    "test2"
  };
  
  for ( string &value : pgConfLines ) {
    if (boost::starts_with(value, "shared_preload_libraries")) {
      MergeSharedPreloadLibraries(value, &metricsLibs);
    }
  }
  
  cout << string("meh=").append(boost::join(metricsLibs, ","));
  
  return 0;
}
```

in this online cpp runner: http://cpp.sh/5k3cx. The above code should be folded into a test.


src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -120,6 +120,14 @@ void ReadCommaSeparatedValues(const string& src, vector<string>* lines) {
   lines->insert(lines->end(), new_lines.begin(), new_lines.end());
 }
 
+void MergeSharedPreloadLibraries(const string& src, vector<string>* defaults) {
+  string copy = boost::erase_first_copy(boost::replace_all_copy(src, " ", ""), "shared_preload_libraries=");
```
tedyu, line 124: nit: long line.
Please wrap so that each line is 100 chars long or shorter.

src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -349,14 +357,37 @@ Status PgWrapper::Start() {
...
-  // TODO: we should probably load the metrics library in a different way once we let
-  // users change the shared_preload_libraries conf parameter.
+
+  // Gather the default extensions:
+  vector<string> metricsLibs;
   if (FLAGS_pg_stat_statements_enabled) {
-    argv.push_back("shared_preload_libraries=pg_stat_statements,yb_pg_metrics,pgaudit,"
-      "pg_hint_plan");
+    metricsLibs.push_back("pg_stat_statements");
+    metricsLibs.push_back("yb_pg_metrics");
```
tedyu, line 365: It seems this is the difference between the two branches.
You can lift the statements from else block up. And leave only this line in if block (to reduce code dup).

src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -349,14 +357,37 @@ Status PgWrapper::Start() {
...
+  // Once again parse the pg conf:
+  vector<string> pgConfLines;
+  if (!FLAGS_ysql_pg_conf_csv.empty()) {
+    RETURN_NOT_OK(ReadCSVValues(FLAGS_ysql_pg_conf_csv, &pgConfLines));
+  } else if (!FLAGS_ysql_pg_conf.empty()) {
+    ReadCommaSeparatedValues(FLAGS_ysql_pg_conf, &pgConfLines);
+  }
+  
+  // If the user has given any shared_preload_libraries -> merge them in
+  for ( string &value : pgConfLines ) {
```
tedyu, line 383: nit: no space after (, no space before )

src/yb/yql/pgwrapper/pg_wrapper.cc
radekg: Makes sense, done in https://github.com/yugabyte/yugabyte-db/pull/9576/commits/dc53f3c3b497f605e6dd7a97a2a03b12ed581d37 but not sure if the order is significant so left it with the if branch in the middle.

src/yb/yql/pgwrapper/pg_wrapper.cc
radekg: Done in https://github.com/yugabyte/yugabyte-db/pull/9576/commits/dc53f3c3b497f605e6dd7a97a2a03b12ed581d37.

src/yb/yql/pgwrapper/pg_wrapper.cc
radekg: Done in https://github.com/yugabyte/yugabyte-db/pull/9576/commits/dc53f3c3b497f605e6dd7a97a2a03b12ed581d37.

radekg:
Updated the test case:

```cpp
#include <iostream>
#include <string>
#include <vector>
#include <boost/algorithm/string.hpp>

using namespace std;

  void MergeSharedPreloadLibraries(const string& src, vector<string>* defaults) {
    string copy = boost::erase_first_copy(boost::replace_all_copy(src, " ", ""), "shared_preload_libraries=");
    copy = boost::trim_copy_if(copy, boost::is_any_of("'\""));
    vector<string> new_items;
    boost::split(new_items, copy, boost::is_any_of(","));
    new_items.erase(std::remove_if(new_items.begin(),
        new_items.end(),
        [](const std::string& s){return s.empty();}),
        new_items.end());
    defaults->insert(defaults->end(), new_items.begin(), new_items.end());
  }
 
int main () {

  vector<string> pgConfLines {
    "shared_preload_libraries='given1,,given2'",
    "some_other_config_option=aaa"
  };

  vector<string> metricsLibs {
  	"test1",
    "test2"
  };
  
  for ( string &value : pgConfLines ) {
    if (boost::starts_with(value, "shared_preload_libraries")) {
      MergeSharedPreloadLibraries(value, &metricsLibs);
    }
  }
  
  cout << string("shared_preload_libraries=").append(boost::join(metricsLibs, ","));
  
  return 0;
}
```

Link to a working example: http://cpp.sh/4bupcz.

src/yb/yql/pgwrapper/pg_wrapper.cc
tedyu: I agree to keep the order.

tedyu:
I don't have further comment.
A member from YQL team would take another look.

radekg:
Significant tidbit for this story is is using the configuration switch when handling multiple extensions. Originally referenced as `--tserver_flags="ysql_pg_conf=shared_preload_libraries=example"`, however, because of how the `--ysql_pg_conf` handles commas (and the fact that it is deprecated), it is better to use `--ysql_pg_conf_csv`. A comma delimited value list must be single quoted in the in the ysql_pg.conf file so the actual reference switch is:

```
--tserver_flags="ysql_pg_conf_csv=\"shared_preload_libraries='example'\""
```

shaunstoltz:
@rkarthik007 Hi Karthik, who would need to review this next to get this moving forward?

src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -120,6 +120,33 @@ void ReadCommaSeparatedValues(const string& src, vector<string>* lines) {
   lines->insert(lines->end(), new_lines.begin(), new_lines.end());
 }
 
+void MergeSharedPreloadLibraries(const string& src, vector<string>* defaults) {
+  string copy = boost::replace_all_copy(src, " ", "");
+  copy = boost::erase_first_copy(copy, "shared_preload_libraries");
+  // According to the documentation in postgresql.conf file,
+  // the '=' is optional hence it needs to be handled separate.
```
m-iancu, line 127: nit: separately.

src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -349,14 +376,33 @@ Status PgWrapper::Start() {
...
+  metricsLibs.push_back("pg_hint_plan");
+
+  // Once again parse the pg conf:
+  vector<string> pgConfLines;
+  if (!FLAGS_ysql_pg_conf_csv.empty()) {
+    RETURN_NOT_OK(ReadCSVValues(FLAGS_ysql_pg_conf_csv, &pgConfLines));
+  } else if (!FLAGS_ysql_pg_conf.empty()) {
+    ReadCommaSeparatedValues(FLAGS_ysql_pg_conf, &pgConfLines);
+  }
+
```
m-iancu, line 396: nit: looks like this line contains trailing whitespace (i.e. up to the indent level) -- it should just be an empty line.

src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -349,14 +376,33 @@ Status PgWrapper::Start() {
...
+  metricsLibs.push_back("pg_stat_statements");
   if (FLAGS_pg_stat_statements_enabled) {
-    argv.push_back("shared_preload_libraries=pg_stat_statements,yb_pg_metrics,pgaudit,"
-      "pg_hint_plan");
-  } else {
-    argv.push_back("shared_preload_libraries=yb_pg_metrics,pgaudit,pg_hint_plan");
+    metricsLibs.push_back("yb_pg_metrics");
   }
+  metricsLibs.push_back("pgaudit");
+  metricsLibs.push_back("pg_hint_plan");
```
m-iancu, line 387: The logic here seems changed. Previously `FLAGS_pg_stat_statements_enabled` determined if `pg_stat_statements` extensions is loaded. Now it toggles `yb_pg_metrics` instead.
I think this could be:
```
if (FLAGS_pg_stat_statements_enabled) {
  metricsLibs.push_back("pg_stat_statements");
}
metricsLibs.push_back("yb_pg_metrics");
metricsLibs.push_back("pgaudit");
metricsLibs.push_back("pg_hint_plan");
```

src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -349,14 +376,33 @@ Status PgWrapper::Start() {
...
+
+  // Once again parse the pg conf:
+  vector<string> pgConfLines;
+  if (!FLAGS_ysql_pg_conf_csv.empty()) {
+    RETURN_NOT_OK(ReadCSVValues(FLAGS_ysql_pg_conf_csv, &pgConfLines));
```
m-iancu, line 392: We already process `FLAGS_ysql_pg_conf_csv`  in `WritePostgresConfig` so this now gets processed twice.
I guess the code works because of the override rules between the two, but I think it might be cleaner to do it just once.

Basically, we now have two ways to set config options:
1. Add them to the auto-generated `ysql_pg.conf` configuration file.
2. Setting them in the `postgres` process shell arguments using the `-c` option.

I think it might not be too complicated to simplify that by just using the conf file for the shared preloaded extensions part (we could do it for the other `-c` options also as a follow-up, but not needed in this diff).
So this is a bit of a less trivial change but we could:
1. Move the logic starting around line `380` (from `// Gather the default extensions:`) to the `WritePostgresConfig(` function.
2. In that function, we can reuse the existing `ReadCSVValues` call instead of re-parsing the pg_config flags.
3. Instead of setting the `-c` options, we just iterate over the `lines` from the config, and, similarly to current code, if they match the `shared_preload_libraries` prefix we merge the pre-computed metricsLibs list before proceeding to write them to the final config.

Note: I might have missed some details/complications here so let me know if this change seems a bit involved. If so I can double-check and provide more details or maybe a better suggestion.

m-iancu (commented):
Awesome @radekg, thanks for fixing this, and sorry for the late review.
I left a few inline comments.

src/yb/yql/pgwrapper/pg_wrapper.cc
radekg: Thank you for catching this. Addressed: https://github.com/yugabyte/yugabyte-db/pull/9576/commits/ab1d047dee772c55ffd9c4dba4f11470ec81f455.

src/yb/yql/pgwrapper/pg_wrapper.cc
radekg: Addressed: https://github.com/yugabyte/yugabyte-db/pull/9576/commits/ab1d047dee772c55ffd9c4dba4f11470ec81f455.

src/yb/yql/pgwrapper/pg_wrapper.cc
radekg: Addressed: https://github.com/yugabyte/yugabyte-db/pull/9576/commits/ab1d047dee772c55ffd9c4dba4f11470ec81f455.

src/yb/yql/pgwrapper/pg_wrapper.cc
radekg: As in, completely skip the `shared_preload_libraries` from the `-c` switches and move it to the config file? That would work. If you confirm this is the way to go, I can do that.

m-iancu: @radekg Correct -- I think moving `shared_preload_libraries` to just the config file will simplify and consolidate our logic/code there.

radekg: Hey @m-iancu, implemented as requested: https://github.com/yugabyte/yugabyte-db/pull/9576/commits/34c9330fbb0667d7193051596816e0d219486bfc.

m-iancu: Thanks @radekg, looks great.  Another small follow-up to this. The user config lines will be few, but there are a lot of lines in the config sample file (all commented out) so it'd be better if we skip iterating over those during processing (to search-for any `shared_preload_libraries` entries and later to remove any duplicates).

I think this patch might work for that but I did not test thoroughly.
```
+  vector<string> user_configs;
   if (!FLAGS_ysql_pg_conf_csv.empty()) {
-    RETURN_NOT_OK(ReadCSVValues(FLAGS_ysql_pg_conf_csv, &lines));
+    RETURN_NOT_OK(ReadCSVValues(FLAGS_ysql_pg_conf_csv, &user_configs));
   } else if (!FLAGS_ysql_pg_conf.empty()) {
-    ReadCommaSeparatedValues(FLAGS_ysql_pg_conf, &lines);
+    ReadCommaSeparatedValues(FLAGS_ysql_pg_conf, &user_configs);
   }

   // If the user has given any shared_preload_libraries, merge them in.
-  for (string &value : lines) {
+  for (string &value : user_configs) {
     if (boost::starts_with(value, "shared_preload_libraries")) {
       MergeSharedPreloadLibraries(value, &metricsLibs);
+    } else {
+      lines.push_back(value);
     }
   }

-  // Ensure we do not have the line duplicated in the config.
-  lines.erase(std::remove_if(lines.begin(), lines.end(), [](const std::string& s) {
-    return boost::starts_with(s, "shared_preload_libraries");
-  }));
-
```
Note:  I (re)started a full run of our tests based on your current patch plus this change, so testing should be done by tomorrow (hopefully everything is good there).

radekg: Added your diff: https://github.com/yugabyte/yugabyte-db/pull/9576/commits/6ba7647eec280ecaa35871ba707e205822347dc0.

m-iancu (approved):
Thanks @radekg, LGTM!

radekg:
\o/ looking forward to this being merged, great news @m-iancu. When done, will bring these changes into the https://github.com/yugabyte/yugabyte-db/pull/9650.

m-iancu:
This is now merged into latest: https://github.com/yugabyte/yugabyte-db/commit/dfc7c94bbfb5a0fef8f7a4371f10e0fbcd74c343.
Thanks again @radekg !

PR #9686, opened 2021-08-11
[docs] adjust copy button behaviour on code blocks

polarweasel:
A couple of code-block handling changes.

1. Strip prompts from `cassandra` and `cql` blocks. (Yes, these are valid source-highlighter languages, and distinct from the sql ones...)

2. Adjust how the `output` language works, and add a `nocopy` one. Specify either one alone and they both work as output did before. Or, append a language name to have the code block highlighted but _without_ a copy button.

    \
    All of these are valid: `output`, `nocopy`, `nocopy.java`, and `output.json`. (The language name needs to come last!)

**Discussion**

First thing, you can keep on going with blocks tagged as `output` just like you could before. This change adds options, but doesn't change that base behaviour.

We could drop the "output" and just go with "nocopy", but:

- there's a lot of output already tagged in the docs
- `output` makes sense, semantically, as a (non-)highlight language for _output_
- as does `nocopy`, for other stuff the reader shouldn't need to copy :)

**One potential downside**: you need a bit more complex search pattern to find all code blocks of a given type. Instead of a simple search for this:

````markdown
```xml
````

you'd need to do a regular-expression search along the lines of this, instead:

````markdown
```.*xml
````

### DIFF
--- a/docs/src/index.js
+++ b/docs/src/index.js
@@ -151,8 +151,8 @@ $(document).ready(() => {
       let regExpCopy = /a^/;
       if (languageDescriptor) {
         // Then apply copy button
-        // Strip the prompt from SQL languages
-        if (['pgsql', 'plpgsql', 'postgres', 'postgresql', 'sql'].includes(languageDescriptor)) {
+        // Strip the prompt from CQL/SQL languages
+        if (['cassandra', 'cql', 'pgsql', 'plpgsql', 'postgres', 'postgresql', 'sql'].includes(languageDescriptor)) {
           if (element.textContent.match(/^[0-9a-z_.:@=^]{1,30}[>|#]\s/gm)) {
             regExpCopy = /^[0-9a-z_.:@=^]{1,30}[>|#]\s/gm;
           }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by lizayugabyte** [comment]
My editor (Typora) doesn't give me an option for "output". I don't set anything for things like output and syntax. When published, they look fine, just don't have a "copy" button. No difference, as far as I an tell, between setting nothing and setting "output".

**Comment by polarweasel** [comment]
As far as you can tell, yes. But when I turn on the markdown linter, you'll get errors on your PRs if you don't tag every code block with a language. :)

**Comment by polarweasel** [comment]
Also, this gives you the option of actually syntax-highlighting output. Handy in particular for stuff like JSON output that's an unintelligible wall of characters otherwise.

**Review by ddhodge — APPROVED** [review]
LGTM

File: docs/src/index.js [hunk 1]
```diff
@@ -163,8 +163,11 @@ $(document).ready(() => {
           } else if (element.textContent.match(/^[0-9a-z_.:@=^]{1,30}[>|#]\s/gm)) {
             regExpCopy = /^[0-9a-z_.:@=^]{1,30}[>|#]\s/gm;
           }
-        // Don't add a copy button to blocks labeled "output"
-        } else if (['output'].includes(languageDescriptor)) {
```
aishwarya24, line 167: Do you think this can be rewritten as 
```
else if (['output', 'nocopy'].includes(languageDescriptor)) 
```

aishwarya24:
LGTM

aishwarya24 (approved):

docs/src/index.js
polarweasel: The ordering is significant, so we needed to flip the order in order to accommodate the `output|nocopy <languageName>` construction. If you'd like to combine lines 168 and 170, though, go for it...

aishwarya24: Understood! This is good then.

PR #9857, opened 2021-08-27
[docs] yb-ctl explain how to quote flags

jaki:
yb-ctl --master_flags and --tserver_flags each take a comma-separated
list of key-value pairs.  It gets tricky when the value has commas or
equals.  Explain how to deal with that in the docs and script docstring.

### DIFF
--- a/scripts/installation/bin/yb-ctl
+++ b/scripts/installation/bin/yb-ctl
@@ -25,12 +25,13 @@ Creating a cluster with replication factor 5
   yb-ctl --rf 5 start
 
 Creating a cluster with placement_info
-  yb-ctl start
-  --placement_info 'cloud1.region1.zone1,cloud2.region2.zone2,cloud3.region3.zone3'
+  yb-ctl start \\
+    --placement_info 'cloud1.region1.zone1,cloud2.region2.zone2,cloud3.region3.zone3'
 
 Creating a cluster with custom flags
-  yb-ctl start --master_flags 'flag1=value,flag2=value,flag3=value'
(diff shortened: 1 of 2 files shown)


scripts/installation/bin/yb-ctl
```diff
@@ -25,12 +25,12 @@ Creating a cluster with replication factor 5
...
 Creating a cluster with placement_info
-  yb-ctl start
-  --placement_info "cloud1.region1.zone1,cloud2.region2.zone2,cloud3.region3.zone3"
+  yb-ctl start --placement_info "cloud1.region1.zone1,cloud2.region2.zone2,cloud3.region3.zone3"
 
-Creating a cluster with custom_flags
-  yb-ctl start --master_flags "flag1=value,flag2=value,flag3=value"
-  --tserver_flags "flag1=value,flag2=value,flag3=value"
+Creating a cluster with custom flags
+  yb-ctl start \\
```
polarweasel, line 31: Why the double-backslash on each line? Is it a python thing? (But this is a comment block, so it shouldn't be parsing those...)

polarweasel (approved):
LGTM, provided the backslashes in the script are correct...

scripts/installation/bin/yb-ctl
jaki: It's a docstring, so still a string:

```py
print("""abc \
123""")
print("""abc \\
123""")
print(r"""abc \
123""")
```

My intent is for there to be backslashes.

polarweasel: Thanks for the clarification! This PR is good to go. Merge when you're ready!

PR #9910, opened 2021-09-02
[docs] WIP: playing with a proxy redirect

polarweasel:
Playing around with a Netlify feature that could allow us to break the docs up into a few microsites and hopefully get our build times down to just a few minutes.

The build preview should point to "List availability zones".

@netlify /platform-api-docs/b3A6MTg5NDc2ODQ-list-availability-zones

### DIFF
--- a/docs/netlify.toml
+++ b/docs/netlify.toml
@@ -28,6 +28,17 @@
 #
 # validator: https://play.netlify.com/redirects
 
+### ALEX TESTING
+# This style of redirect will pull content from the remote URL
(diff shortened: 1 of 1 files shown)


polarweasel:
It appears that things start breaking in interesting ways if paths don't match up between the "main" site and the proxied site. I might need to make a bit more realistic proof of concept for this one. (Also, this is probably not a good way to embed the API docs into the main doc site, which I'm still not convinced is a good approach in any case!)

polarweasel:
Not required.

PR #9962, opened 2021-09-09
[docs] JSON index expressions

kerneljake:
range queries on JSON int with index expressions


polarweasel (approved):
LGTM. Thank you @kerneljake!

PR #9985, opened 2021-09-14
[docs] update explore data-types: typo+linting

polarweasel:
Fix a typo in "To insert a row, use the RUN keyword" (should be ROW, and now is). While I was at it, I fixed all the markdownlint warnings and cleaned up the code snippet tagging.

@netlify /latest/explore/ysql-language-features/data-types


lizayugabyte (approved):
LGTM 
(btw, it wasn't my file; I only added the range types section at the end)

PR #9993, opened 2021-09-14
 [Doc] PR for Clarify OS requirements for Platform vs DB (Doc9847) 

lizayugabyte:
I added a note.
(however, since the document is clearly only applicable to Platform, I don't think this was necessary)

### DIFF
--- a/docs/content/latest/faq/yugabyte-platform.md
+++ b/docs/content/latest/faq/yugabyte-platform.md
@@ -26,14 +26,20 @@ YugaWare first needs to be installed on any machine. The next step is to configu
 
 **OS requirements**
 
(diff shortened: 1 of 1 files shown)


ddhodge:
@hsiaosu-yb There is a FAQ item directly below "What are the OS requirements and permissions to run the YugabyteDB data nodes?" that also sort of addresses this. However the link is to a page that no longer exists (https://docs.yugabyte.com/latest/deploy/multi-node-cluster/#prerequisites) and redirects to https://docs.yugabyte.com/latest/deploy/manual-deployment/, which doesn't actually list any OS requirements for YugabyteDB, but possibly should link to https://docs.yugabyte.com/latest/deploy/checklist/ instead; however this page has vaguer and different OS requirements than the suggested page: "YugabyteDB works on a variety of operating systems. For production workloads, the recommended operating systems are CentOS 7.x and RHEL 7.x." Is there a distinction between YugabyteDB requirements and YugabyteDB on cluster node requirements (is this what they mean by production environment for example)? Can you clarify so that we can then link to a common location for both, or at least make this clearer

ddhodge (approved):
LGTM

Issue #6302, opened 2020-11-06
[docdb] master UI remove parentheses around range partition


jaki:
The master UI table page shows tablets and their partition boundaries.  For YSQL range partitioned tables, this looks like `range: [(DocKey([], [1]), DocKey([], [3])))`.  There is an extra pair of parentheses around all of these.  Change `[(foo, bar))` to `[foo, bar)`.

YSQL range partitioned tables can be created as follows:

```
CREATE TABLE t (i int, PRIMARY KEY (i ASC)) SPLIT AT VALUES ((1), (3), (5));
```

Master UI is at 127.0.0.1:7000 by default.

bansaljas:
Hey!
I'd like to work on the given issue. Could tell me which files need this change if possible?

Thank you.

jaki:
Hi @bansaljas!  My guess is that it's around `PartitionSchema::RangePartitionDebugString` in `src/yb/common/partition.cc`.  You may also want to see related PR #6223.

bansaljas:
Great, I'll go through that!
Thanks :)

rajmaddy89:
Can you please assign this issue to me ?

jaki:
Hi, @rajmaddy89.  Anyone can work on it, so the issue is left unassigned.  If you would like to work on it, it shouldn't take too long to do, then you can create a PR.

rajmaddy89:
Reference to PR: https://github.com/yugabyte/yugabyte-db/pull/10028

Issue #8517, opened 2021-05-19
[Platform][UI] Create Universe should auto-populate dropdown for provider and region


kc-yb:
The Create Universe screen needs enhancement when showing dropdown list for provider and region. For example, if there is only one provider with one region, the drop down list is still unfilled - it should be auto populated with some provider/region - based on some criteria - for ex: alphabetically.

Or an advanced feature - like default config provider and default region (like shopping apps - where you can have multiple addresses and one can be your default) - it helps reduce the number of clicks required to create the universe.

kc-yb:
[
![Region-Screenshot](https://user-images.githubusercontent.com/18676881/118761687-76a20e00-b892-11eb-8e14-9573202d6ef3.png)
![Provider-screenshot](https://user-images.githubusercontent.com/18676881/118761692-786bd180-b892-11eb-8dd3-a28ad710796b.png)

](url)

rajmaddy89:
Can you please assign this issue to me ?

rajmaddy89:
In case of In Region Dropdown for issue 8517, its a little bit tricky to auto-populate. Here is the reason

It is multi select type and uses React-Select component,

when you assign the value prop it points to a default value but it will not change at all. This is because when we try to hardcode it say like universeRegionList[0], we do an anti-pattern in React

This is because you've tied the value prop directly to universeRegionList[0], which doesn't change. Better to tie it to a state variable that you default to the option, and have your onChange update the state variable.

React is intended to have an unmount happen from a parent to child relationship. Now if you want a child to unmount itself, you can simulate this with a state change in the parent that is triggered by the child. let me show you in code. So the value must be tied to state variable and can be re-rendered by the parent component and this is how React default value works.

This is the reason why I have tied the value prop to state variable

rajmaddy89:
Reference to PR: https://github.com/yugabyte/yugabyte-db/pull/10028

PR #10028, opened 2021-09-17
[#8517, #6302, #5355][Platform] - Auto-populate dropdown, remove paranthesis around range partition and adjust DB version margin

rajmaddy89:
Description
--
1. Fix issue [8517](https://github.com/yugabyte/yugabyte-db/issues/8517)
2. Fix issue [6302](https://github.com/yugabyte/yugabyte-db/issues/6302)
3. Fix issue [5355](https://github.com/yugabyte/yugabyte-db/issues/5355)

### DIFF
--- a/managed/ui/src/components/universes/UniverseForm/ClusterFields.js
+++ b/managed/ui/src/components/universes/UniverseForm/ClusterFields.js
@@ -97,6 +97,7 @@ const GP3_MAX_THROUGHPUT = 1000;
 const GP3_IOPS_TO_MAX_DISK_THROUGHPUT = 4;
 
 const initialState = {
+  defaultRegion: {},
   universeName: '',
   instanceTypeSelected: '',
   azCheckState: true,
@@ -376,7 +377,7 @@ export default class ClusterFields extends Component {
   UNSAFE_componentWillReceiveProps(nextProps) {
     const {
       universe: { currentUniverse },
-      cloud: { nodeInstanceList, instanceTypes },
+      cloud: { nodeInstanceList, instanceTypes, regions },
       clusterType,
       formValues
     } = nextProps;
@@ -385,6 +386,15 @@ export default class ClusterFields extends Component {
     let providerSelected = this.state.providerSelected;
     if (isNonEmptyObject(currentFormValues) && isNonEmptyString(currentFormValues.provider)) {
       providerSelected = currentFormValues.provider;
+      if (providerSelected && regions.data.length > 0) {
+        const providerRegions = regions.data &&
+          regions.data.map(function (regionItem) {
+          return { value: regionItem.uuid, label: regionItem.name };
+        });
+        this.setState({
+          defaultRegion: providerRegions[0]
+        });
+      }
     }
 
     if (
@@ -1629,7 +1639,9 @@ export default class ClusterFields extends Component {
       );
     }
 
-    universeProviderList.unshift(<option key="" value=""></option>);
+    if (universeProviderList.length == 0) {
+      universeProviderList.unshift(<option key="" value=""></option>);
+    }
 
     let universeRegionList = [];
     if (self.state.providerSelected) {
@@ -1894,6 +1906,7 @@ export default class ClusterFields extends Component {
                   isMulti={true}
                   selectValChanged={this.regionListChanged}
                   providerSelected={currentProviderUUID}
+                  defaultValue={this.state.defaultRegion}
                 />
                 {clusterType === 'async'
                   ? [
--- a/src/yb/common/partition.cc
+++ b/src/yb/common/partition.cc
@@ -817,7 +817,7 @@ string PartitionSchema::RangePartitionDebugString(const Partition& partition,
                                                   const Schema& schema) const {
   CHECK(!schema.num_hash_key_columns());
   std::string s;
-  s.append("range: [(");
+  s.append("range: [");
   if (partition.partition_key_start().empty()) {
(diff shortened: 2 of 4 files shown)


rajmaddy89:
In case of In Region Dropdown for issue [8517](https://github.com/yugabyte/yugabyte-db/issues/8517), its a little bit tricky to auto-populate. Here is the reason

1. It is multi select type and uses React-Select component, 

2. when you assign the `value` prop it points to a default value but it will not change at all. This is because when we try to hardcode it say like `universeRegionList[0]`, we do an anti-pattern in React

3. This is because you've tied the value prop directly to `universeRegionList[0]`, which doesn't change. Better to tie it to a state variable that you default to the option, and have your onChange update the state variable.

4. React is intended to have an unmount happen from a parent to child relationship. Now if you want a child to unmount itself, you can simulate this with a state change in the parent that is triggered by the child. let me show you in code. So the `value` must be tied to state variable and can be re-rendered by the parent component and this is how React default value works.

This is the reason why I have tied the value prop to state variable

jaki:
This appears to fix #6302.  There is another similar code path, but it seems dead, so what you have should be good for #6302.

I will not review the other items.

rajmaddy89:
Thanks Jason

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -1629,7 +1639,9 @@ export default class ClusterFields extends Component {
       );
     }
 
-    universeProviderList.unshift(<option key="" value=""></option>);
+    if (universeProviderList.length == 0) {
+      universeProviderList.unshift(<option key="" value=""></option>);
+    }
```
yb-andrew, line 1644: I think we should remove this entirely. Don't insert an empty option under any circumstances. If there is no provider, the form should fail.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
```diff
@@ -385,6 +386,15 @@ export default class ClusterFields extends Component {
...
     if (isNonEmptyObject(currentFormValues) && isNonEmptyString(currentFormValues.provider)) {
       providerSelected = currentFormValues.provider;
+      if (providerSelected && regions.data.length > 0) {
+        const providerRegions = regions.data &&
+          regions.data.map(function (regionItem) {
+          return { value: regionItem.uuid, label: regionItem.name };
+        });
+        this.setState({
+          defaultRegion: providerRegions[0]
+        });
```
yb-andrew, line 396: No need to put the regions into another array. Since you already checked the length in the if-statement, just access the first value:
```
this.setState({
  defaultRegion: { value: regions.data[0].uuid, label: regions.data[0].name }
})
```

yb-andrew (changes requested):
Looks mostly good. Just some minor points to fix.

managed/ui/src/components/universes/UniverseForm/ClusterFields.js
rajmaddy89: I initially thought about that option, but it won't work due to the reason I mentioned here:

https://github.com/yugabyte/yugabyte-db/pull/10028#issuecomment-921414760

@yb-andrew : Please let me know about your thoughts

jaki:
Turns out that the other code path is not dead, though it is very difficult to come by.  Please fix the parentheses in that code path (`PartitionSchema::PartitionDebugString`) as well.

You can see this code path hit using

```diff
diff --git i/src/yb/tserver/tablet_server-test.cc w/src/yb/tserver/tablet_server-test.cc
index 41aa0943f..373f8430a 100644
--- i/src/yb/tserver/tablet_server-test.cc
+++ w/src/yb/tserver/tablet_server-test.cc
@@ -194,6 +194,7 @@ TEST_F(TabletServerTest, TestSetFlagsAndCheckWebPages) {
   // Tablets page should list tablet.
   ASSERT_OK(c.FetchURL(Substitute("http://$0/tablets", addr),
                        &buf));
+  LOG(WARNING) << "buf: " << buf.ToString();
   ASSERT_STR_CONTAINS(buf.ToString(), kTabletId);
 
   // Tablet page should include the schema.
```

and running `./yb_build.sh release --cxx-test tserver_tablet_server-test --gtest_filter TabletServerTest.TestSetFlagsAndCheckWebPages`.  One other test is `TabletServerTest.TestCreateTablet_TabletExists`.

I suggest you also add to the WebPages test to `ASSERT_STR_CONTAINS` the proper range string without the parentheses.

yb-andrew (approved):
LG!

src/yb/common/partition.cc
```diff
@@ -900,7 +900,7 @@ string PartitionSchema::PartitionDebugString(const Partition& partition,
     } else {
```
jaki, line 900: Don't forget to change `s.append("), (")` above to `s.append(", ")`.

jaki (commented):
Please do this small adjustment as well.

src/yb/common/partition.cc
rajmaddy89: May be I missed this, sorry about that @jaki , fixed now

mbautin:
I have not done a detailed review, but are we taking into account the fact that the same version of Platform can interact with multiple different versions of YugabyteDB, and potentially deal with both old and new versions of partition formatting that is being updated?

jaki:
@rajmaddy89, I'm not a good person to answer about whether Platform uses the partition string.

PR #10032, opened 2021-09-17
Correcting DDL statements for YSQL Table Partitioning document.

shubh1608:
Executing Range and Hash partitioning DDL statement with primary key constraints won't allow creating a table. We should remove the primary key constraints from the Range and Hash partitioning 'CREATE TABLE' statements as the primary key has to be a part of partition keys.


deeps1991 (approved):
Thanks for fixing this!

PR #10035, opened 2021-09-17
[docs] vpc peering info

ddhodge:
@netlify /latest/yugabyte-cloud/cloud-network/vpc-peers/


juanman2 (approved):
Good first cut, people like Taylor and or Chirag may want to review.  Do note that this will be temporary, and we will be changing things again, next week!

PR #10056, opened 2021-09-20
Run yugabyte test every couple of days

iSignal:
We have a github action that tests yugabyted. We only run it when yugabyted file is changed but it would be good to also run it periodically.


svarnau (approved):
Looks okay. Where do these tests show up?

iSignal:
https://github.com/yugabyte/yugabyte-db/actions/workflows/yugabyted-test.yml

iSignal:
The download url detection was broken for 2.9.0 because it has only 3 digits while the download path has 4  - 2.9.0.0. Fixed it so the test runs again.

PR #9346, opened 2021-07-17
(YUGA-9304) - Create issue templates with pre-populated suggestions

aishwarya24:
Description
--
Created different templates for
1. Docs 
2. DocDB
3. YSQL
4. YCQL
5. UI
6. Feature requests

When a user tries create a new issue, they will be redirected to a page to select one of the above template types and then the issue will be also assigned to the right labels.

NOTE
--
We can add PROJECT tag too, but GITHUB does not currently support it. The workaround is to include a URL in the `config.yaml` file something like this

```
contact_links:
  - name: <Name>
    about: <About>
    url: https://github.com/aishwarya24/HelloWorld/issues/new?assignees=&labels=area%2Fdocumentation&template=docs.md&title=%5BDocs%5D+Title&projects=aishwarya24%2FHelloWorld%2F1
```
However the challenge is to get the **user name** dynamically added to the url which is something that needs to be investigated

Issue #9304, opened 2021-07-15
Create issue templates with pre-populated suggestions.


aishwarya24:
Currently , while creating new issues , there needs to be a template helping users across the community to sort the issue by its type at the time of creation.
Issue types to be included:
1.docs
2.core (ycql, ysql, docdb)
3.platform
4.UI

polarweasel:
To clarify: we'll need to create several issue templates, one per issue type.

aishwarya24:
Reference PR: https://github.com/yugabyte/yugabyte-db/pull/9346

PR #10261, opened 2021-10-12
[Docs] Create issue templates with pre-populated suggestions

aishwarya24:
Created a new PR for #9304 since the old one was having some issue fetching my updated commit.
#9346 is the old PR. I added the requested changes and also included an issue template for Yugabyte Cloud.

Attaching the summary again:
Description 
--
Created different templates for
1. Docs 
2. DocDB
3. YSQL
4. YCQL
5. UI
6. Feature requests
7. Cloud

When a user tries create a new issue, they will be redirected to a page to select one of the above template types and then the issue will be also assigned to the right labels.

NOTE
--
We can add PROJECT tag too, but GITHUB does not currently support it. The workaround is to include a URL in the `config.yaml` file something like this

```
contact_links:
  - name: <Name>
    about: <About>
    url: https://github.com/aishwarya24/HelloWorld/issues/new?assignees=&labels=area%2Fdocumentation&template=docs.md&title=%5BDocs%5D+Title&projects=aishwarya24%2FHelloWorld%2F1
```
However the challenge is to get the **user name** dynamically added to the url which is something that needs to be investigated.
 
Additional note from @polarweasel : Issue forms do not currently support adding projects, though (I've contacted their team to ask about adding the feature, fwiw) :(
So hopefully we can have them in the future.


polarweasel:
Quick update for watchers: Waiting on some research on redirects for cloud (and maybe platform) issues. If we can get those reports going straight to Zendesk, that'd be ideal.

PR #10412, opened 2021-10-26
[Docs] need to modify the command to run the Java application

tichimura:
Change the command to run the application with `mvn` command, otherwise, the original `java` cmd will return error as `no main manifest attribute`. 
If still using `java`, 
- it can be resolved with adding the main class `com.yugabyte.HelloSqlSslApp` as follows the sample code.
- Also, the postgres driver will not found with the classpath, the external library will be found in `$HOME/.m2/ directory` if no additional configurations are there in maven. so may need to add the classpath to that, or configure maven to copy the dependencies to `target/lib`


polarweasel (approved):
LGTM

aishwarya24 (approved):

aishwarya24:
LGTM

tichimura:
@aishwarya24 may need to add a comment for our recommendation of ssl configured cluster.

Issue #10094, opened 2021-09-23
optimize rocksdb's delta encoding for YugabyteDB key format


kmuthukk:
The delta encoding scheme in RocksDB does simple prefix compression. This works well if when the common part of the two adjacent keys is at the prefix, and the non-common part is at the end.

The YugabyteDB key format is something like:

```
<hashcode>:<hash-key-column(s)>:<range-key-column(s)>:<column-id>:<hybrid-TS>:<uniq-write-id>
```

For a row with say N non-primary key columns... the N such keys are written that only differ in the column-id (and perhaps the uniq-write-id).  Also note that in cases where the column values themselves are small - such as integers or small text fields, the key portion can dominate the space usage.

A simple prefix-compression based delta encoding scheme is only able to save the bytes to the left of the <column-id>.. and therefore not as effective than a more "custom" delta encoding scheme that we can implement. 

For example, if we are able to save 90% of space used for the keys instead of 70%, it might seem like a small incremental benefit but in the overall equation (especially for cases where the value portion is small) the space savings can be significant (2-3x even). 

In one workload, we noticed that 220MB of an SSTable was for keys and 20MB was for values, and prefix compression was given an effective savings of 70%.

So logical space used was with simple prefix compression based delta-encoding is:

`(100-70%) of 220MB (for keys) + 20MB  (for values) = 86MB`

We can easily implement a more custom scheme that can give nearly 90% savings on keys. So the math will then be:

`(100-90)% of 220MB (for keys) + 20MB (for values) = 42MB`

Nearly 2x wins! Some early experimentation shows that in many cases we can get 2-3x wins here.

The nice thing is that the change can easily be done in upgrade safe manner where newer SSTable files written by newer versions of the software start writing data in this new delta encoding format... while supporting both formats on the read-path. A metadata bit in the SSTable file can indicate which delta encoding format is in use.

tylarb:
Backport to 2.6 branch is preferred.

PR #10424, opened 2021-10-27
[#10094] docs: advanced delta encoding format description

ttyusupov:
Preview: https://github.com/ttyusupov/yugabyte-db/blob/advanced-delta-encoding-design/architecture/design/advanced-delta-encoding.md


mbautin (approved):

architecture/design/advanced-delta-encoding.md
```diff
@@ -0,0 +1,168 @@
...
+
+Standard rocksdb key delta encoding saves some space by extracting common shared prefix and encoding those records using the following scheme: `<shared_prefix_size><non_shared_size><value_size><non_shared><value>`. For our example this will be encoded as:
+```
+471210488000000121488000000121 4A80 23800185F0027D73BA804A   0114000000000004 -> null => <0><24><0><0x4712104880000001214880000001214A8023800185F0027D73BA804A0114000000000004><>
+471210488000000121488000000121 4B8C 23800185F0027D73BA803FAB 0115000000000004 -> 12 => <15><32><4><0x4B8C23800185F0027D73BA803FAB0115000000000004><12>
+471210488000000121488000000121 4B8D 23800185F0027D73BA803F8B 0116000000000004 -> 24 => <16><31><4><0x8D23800185F0027D73BA803FAB0115000000000004><24>
+```
+
+But the only difference between keys of each of these docdb records and the next one is only `ColumnId` and `w` (write_id) component of hybrid time. In the binary form, we also have the last internal rocksdb component which is usually just got incremented for the next record. So if we can only store this difference, but not the whole key part after shared_prefix - we can save more space.
```
mbautin, line 28: just got incremented -> just gets incremented

mbautin (approved):

Issue #10480, opened 2021-11-04
[UI] [docs] Yugabyte docs version dropdown fails to re-render when there is a value change for the version.


aishwarya24:
### Description

If the Yugabyte docs version is changed from the dropdown list , it still displays the first index from the dropdown.

rajmaddy89:
can I work on this issue ?

rajmaddy89:
Can you assign this to me ?

rajmaddy89:
https://github.com/yugabyte/yugabyte-db/pull/10481

rajmaddy89:
<img width="508" alt="Screen Shot 2021-11-03 at 9 13 57 PM" src="https://user-images.githubusercontent.com/10095310/140242337-1197e5f0-45e4-44de-89b2-5fc4c72c4f37.png">

If you see the above image, even though the dropdown value was changed to `stable` version , its still pointing to `latest` version in the dropdown but the content on the right side displays the `stable` docs content accordingly

polarweasel:
Closed by #10481

PR #10481, opened 2021-11-04
[UI] [docs] Fix re-rendering of docs page version dropdown

rajmaddy89:
Issue: https://github.com/yugabyte/yugabyte-db/issues/10480

Description
--
The issue currently today is that when a user goes to Yugabyte document page when they change the version of docs on the left side menu, the dropdown value always sticks to the first index instead of reacting to the dropDownValue change

Fix
--
This happens because the `version_switcher.html` component is cached and is not allowed to re-render in `drawer.html`.

### DIFF
--- a/docs/layouts/partials/drawer.html
+++ b/docs/layouts/partials/drawer.html
@@ -23,7 +23,7 @@
       <div class="toc">
 
         {{ partial "detect_version" . }}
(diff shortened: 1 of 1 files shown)


rajmaddy89:
Please refer to issue: #10480

polarweasel (approved):
Great catch. Thank you so much!

PR #10485, opened 2021-11-04
[docs] redirect updates

polarweasel:
Redirects:

* all older versions of /_version_/yugabyte-cloud point to /latest/yugabyte-cloud
* all older versions of /_version_/integrations and /_version_/develop/ecosystem-integrations point to /latest/integrations

Also bumping Hugo to [current 89.1](https://github.com/gohugoio/hugo/releases) from 88.1. (This required adding the build lock to .gitignore.)

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -15,3 +15,5 @@ static/js/site.js
 .session.vim*
 
 *.sw?
+
+.hugo_build.lock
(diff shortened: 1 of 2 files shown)


polarweasel:
All looks good, and builds are working. Hugo 89.1 fixed a lock-related issue, so bumped up that today. All still builds fine.

PR #10514, opened 2021-11-06
[docs] network compression docs

### LINKED ISSUE #10003: [docs] add network compression options for yb-tserver config
Add a new section called "Network compression" at the same level as ["Performance flags"](https://docs.yugabyte.com/latest/reference/configuration/yb-tserver/#performance-flags).

Content should be a well-edited version of the following:

> In order to reduce network bandwidth consumption, YugabyteDB can compress network traffic between nodes of a cluster in order to reduce the amount of bandwidth used between peers. With the RPC compression feature, you can significantly reduce costs associated with cross-AZ or cross-region network traffic with minimal CPU overhead.
>
> Three different traffic compression algorithms are available: Gzip, Snappy, and LZ4. It supports easy enable or disable compression with rolling upgrade support and provides the ability to select the compression algorithm of your choice.
> * `enable_stream_compression` is designed to be rolling upgrade safe — as in you can do a rolling upgrade to enable or disable compression overall
> * `stream_compression_algo` can only be set after the compression is enabled though!
>   * you cannot change this in the same gflag change as enabling compression
>   * you cannot change this if only a subset of all nodes (eg masters vs tservers) have compression enabled
>   * However, if compression is already enabled, then this flag is safe to change in a rolling way.

polarweasel:
Adding docs for RPC compression. Latest only right now, but once we're happy, I'll also add to 2.6 and 2.8.

Fixes #10003

@netlify /latest/reference/configuration/yb-tserver/#network-compression

### DIFF
--- a/docs/content/latest/reference/configuration/yb-tserver.md
+++ b/docs/content/latest/reference/configuration/yb-tserver.md
@@ -680,16 +680,16 @@ Use the following two gflags to configure RPC compression:
 
 ##### --enable_stream_compression
 
-Controls whether YugabyteDB uses RPC compression. If you set this flag to `true`, you must also specify a stream compression algorithm.
+Controls whether YugabyteDB uses RPC compression. Valid values are `true` or `false`.
 
 ##### --stream_compression_algo
 
-Specifies which compression algorithm to use, if stream compression is enabled. Valid values are:
+Specifies which compression algorithm to use. Requires `enable_stream_compression` to be set to true. Valid values are:
 
-- 0: no compression
-- 1: gzip
-- 2: snappy
-- 3: lz4
+- 0: No compression (default value)
+- 1: Gzip
+- 2: Snappy
+- 3: LZ4
 
 In most cases, LZ4 (`--stream_compression_algo=3`) offers the best compromise of compression performance versus CPU overhead.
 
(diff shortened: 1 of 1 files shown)


docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -674,6 +674,35 @@ Default: `256MB`
 
 ---
 
+### Network compression
+
+Use the following two gflags to configure RPC compression:
+
+##### --enable_stream_compression
+
+Controls whether YugabyteDB uses RPC compression. If you set this flag to `true`, you must also specify a stream compression algorithm.
```
bmatican, line 683: This part was probably unclearly described. You do not *have to* also specify the algo. But the other way around, you cannot specify an algo, until you have enabled compression support.

docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -674,6 +674,35 @@ Default: `256MB`
...
+- 2: snappy
+- 3: lz4
+
+In most cases, LZ4 (`--stream_compression_algo=3`) offers the best compromise of compression performance versus CPU overhead.
+
+{{< note title="Upgrade notes" >}}
+
+To upgrade from an older version that doesn't support RPC compression (such as 2.4), to a newer version that does (such as 2.6), you need to do the following:
+
+1. Rolling restart to upgrade YugabyteDB, such as from 2.4 to 2.6.
```
bmatican, line 700: maybe expand on this to make it clear that the role of the upgrade is to get you to a version of the software that is aware of compression, as a feature

docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -674,6 +674,35 @@ Default: `256MB`
...
+1. Rolling restart to enable compression, on both master and tserver, by setting `enable_stream_compression=true`.
```
bmatican, line 701: We should mark somehow that this step is optional, if the version you're upgrading to already has this enabled by default, eg: if you're upgrading to the next stable release of 2.6 >= 2.6.3, or any 2.8 version, you can skip this step

bmatican (changes requested):

bmatican (approved):
thank you!

PR #10567, opened 2021-11-11
[Docs] update hugo config for proper dates in RSS

polarweasel:
This change tells Hugo to pull each file's last-committed date from Git to populate the `<pubDate>` elements in the RSS.

The feed we most care about currently is /latest/releases/whats-new/index.xml — verify that the two pages in that file actually have non-zero pub dates!


polarweasel:
Based on wrong branch, closing.

PR #10568, opened 2021-11-11
[docs] update hugo config to provide dates in RSS

polarweasel:
This change tells Hugo to pull each file's last-committed date from Git to populate the `<pubDate>` elements in the RSS.

The feed we most care about currently is `/latest/releases/whats-new/index.xml` — verify that the two pages in that file actually have non-zero pub dates!

Compare the new output to the old output...

```sh
# new output:
$ curl https://deploy-preview-10568--infallible-bardeen-164bc9.netlify.app/latest/releases/whats-new/index.xml

# old output:
$ curl https://docs.yugabyte.com/latest/releases/whats-new/index.xml

# or, for less noise in the output:

$ curl https://deploy-preview-10568--infallible-bardeen-164bc9.netlify.app/latest/releases/whats-new/index.xml | grep pubDate
$ curl https://docs.yugabyte.com/latest/releases/whats-new/index.xml | grep pubDate
```

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -10,6 +10,11 @@ pygmentsStyle: "pygments"
 enableGitInfo: true
 resourceDir: docs/resources
 
+frontmatter:
+  date:
+    - :git
+    - :default
+
 params:
   author: Team Yugabyte
   description: 'YugabyteDB, the transactional, high-performance database for building internet-scale, globally-distributed apps.'


tylarb:
@polarweasel  I think I am a little lost... this commit seems to just add the timestamp of the commit, but I'm not seeing a way to subscribe to an RSS feed still. 

I've got an RSS feed subscription extension, and it doesn't show the preview page as having a valid rss.

Is this the first commit in a multi step process?

polarweasel:
@tylarb I think we're all good now, after our slack chat. Can you confirm with an approval here if the RSS feed actually works correctly in your reader?

PR #10608, opened 2021-11-16
[docs] alias to redirect to the new jdbc page

aishwarya24:
Added an alias to redirect the smart driver to the new JDBC page.

@netlify /latest/integrations/jdbc-driver/


polarweasel (approved):
LGTM. Thanks for the quick fix!

PR #10612, opened 2021-11-16
[docs] redirect smart driver to jdbc driver 

aishwarya24:
Redirected smart driver page under integrations to Yugabyte JDBC driver page.


polarweasel (approved):
LGTM. Double-check on the build preview to make sure the redirect works as you expect, then merge!

docs/netlify.toml
```diff
@@ -94,6 +94,11 @@
   to = "/latest/integrations/"
   force = true
 
+[[redirects]]
```
polarweasel, line 97: I was wrong! This needs to go above the more general redirect on line 87, because the first match wins in this file.

polarweasel (changes requested):

docs/netlify.toml
aishwarya24: Oh thank you and in that case , I added it right above the first redirect that goes to /latest/integrations.

polarweasel (approved):

PR #10613, opened 2021-11-16
[docs] yb_hash_code docs

tanujnay112:
Doc page for yb_hash_code

### DIFF
--- a/docs/content/latest/api/ysql/exprs/func_currval.md
+++ b/docs/content/latest/api/ysql/exprs/func_currval.md
@@ -4,7 +4,7 @@ headerTitle: currval()
 linkTitle: currval()
 description: Returns the last value returned by the nextval() function for the specified sequence in the current session.
 menu:
-  latest:
+  stable:
     identifier: api-ysql-exprs-currval
     parent: api-ysql-exprs
 aliases:
--- a/docs/content/latest/api/ysql/exprs/func_currval.md
+++ b/docs/content/latest/api/ysql/exprs/func_currval.md
@@ -4,7 +4,7 @@ headerTitle: currval()
 linkTitle: currval()
 description: Returns the last value returned by the nextval() function for the specified sequence in the current session.
 menu:
-  stable:
+  latest:
     identifier: api-ysql-exprs-currval
     parent: api-ysql-exprs
 aliases:
--- a/docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
+++ b/docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
@@ -29,10 +29,11 @@ Where `a1, a2, a3...` are expressions of type `t1, t2, t3...`, respectively. `t1
 
 ## Function Pushdown
 
-This function can be either evaluated at the YSQL layer after resolving each of its argument expressions, or certain invocations of it may be pushed down to be evaluated at the DocDB layer. When pushdown happens, the restrictions implied by the yb_hash_code conditions are used to ensure we only send requests to the tablets that definitely contain values in the requested ranges, as well as only scan the relevant ranges within each tablet. This section discusses the situations where they are pushed down. `yb_hash_code` invocations are pushed down when the YSQL optimizer determines that the return values of the function will directly match the values of the hash value column in a requested base table or index table. For example, if you create a table as follows
+This function can be either evaluated at the YSQL layer after resolving each of its argument expressions, or certain invocations of it may be pushed down to be evaluated at the DocDB layer. When pushdown happens, the restrictions implied by the yb_hash_code conditions are used to ensure we only send requests to the tablets that definitely contain values in the requested ranges, as well as only scan the relevant ranges within each tablet. This section discusses the situations where they are pushed down. `yb_hash_code` invocations are pushed down when the YSQL optimizer determines that the return values of the function will directly match the values of the hash value column in a requested base table or index table. For example, if you create a table and insert 10000 rows as follows
 
 ```sql
 CREATE TABLE sample_table (x INTEGER, y INTEGER, z INTEGER, PRIMARY KEY((x,y) HASH, z ASC));
+INSERT INTO sample_table SELECT i,i,i FROM generate_series(1,10000) i;
 ```
 
 Then the following query will evaluate the `yb_hash_code` calls at the DocDB layer:
@@ -41,7 +42,10 @@ Then the following query will evaluate the `yb_hash_code` calls at the DocDB lay
 SELECT * FROM sample_table WHERE yb_hash_code(x,y) <= 128 AND yb_hash_code(x,y) >= 0;
 ```
 
-You can expect to verify this with the `EXPLAIN ANALYZE` result of this statement
+You can verify this with the `EXPLAIN ANALYZE` result of this statement
+```sql
+EXPLAIN ANALYZE SELECT * FROM sample_table WHERE yb_hash_code(x,y) <= 128 AND yb_hash_code(x,y) >= 0;
+```
 
 ```output
                                                             QUERY PLAN
@@ -70,6 +74,10 @@ SELECT * FROM sample_table WHERE yb_hash_code(x,z) <= 128 AND yb_hash_code(x,z) 
 ```
 
 Here, the `yb_hash_code` calls are on `x` and `z`. `x` and `z` do not form the full hash key of the base sample_table table but they do form the full hash key of the index table, `sample_idx`. Hence, the optimizer will consider pushing down the calls to the DocDB layer if an index scan using `sample_idx` is chosen. Note that the optimizer may choose not to go with a secondary scan if it deems the requested hash range to be large enough to warrant doing a simple full table scan instead. The `EXPLAIN ANALYZE` result for this could be as follows
+
+```sql
+EXPLAIN ANALYZE SELECT * FROM sample_table WHERE yb_hash_code(x,z) <= 128 AND yb_hash_code(x,z) >= 0;
+```
 
 ```output
                                                          QUERY PLAN
@@ -140,9 +148,9 @@ SELECT yb_hash_code(1::int, 2::int, 'sample string'::text);
 ```
 
 ```output
-   yb_hash_code
----------------------
-      5067
+ yb_hash_code 
+--------------
+        23017
 ```
 
 ### Querying rows from a specific tablet
--- a/docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
+++ b/docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
@@ -21,7 +21,7 @@ showAsideToc: true
 
 `yb_hash_code` can be used anywhere a normal function can be used in a YSQL expression. There is no limitation to where it can be used. It takes in a variable number of parameters whose types must be allowed in YB primary keys. More formally,
 
-```sql
+```output
 yb_hash_code(a1:t1, a2:t2, a3:t3, a4:t4...) → int4 (32 bit integer)
 ```
 
@@ -43,7 +43,7 @@ SELECT * FROM sample_table WHERE yb_hash_code(x,y) <= 128 AND yb_hash_code(x,y) 
 
 You can expect to verify this with the `EXPLAIN ANALYZE` result of this statement
 
-```sql
+```output
                                                             QUERY PLAN
 -----------------------------------------------------------------------------------------------------------------------------------
  Index Scan using sample_table_pkey on sample_table  (cost=0.00..5.88 rows=16 width=12) (actual time=2.867..2.919 rows=18 loops=1)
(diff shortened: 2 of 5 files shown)


docs/content/latest/api/ysql/exprs/func_currval.md
```diff
@@ -4,7 +4,7 @@ headerTitle: currval()
 linkTitle: currval()
 description: Returns the last value returned by the nextval() function for the specified sequence in the current session.
 menu:
-  latest:
```
polarweasel, line 7: This should be latest. You need to edit the corresponding stable file:  `/docs/content/stable/api/ysql/exprs/func_currval.md`

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+title: yb_hash_code() function [YSQL]
+headerTitle: yb_hash_code()
+linkTitle: yb_hash_code()
+description: Returns the partition hash code for a given set of expressions. 
+menu:
+  latest:
+    identifier: api-ysql-exprs-yb_hash_code
+    parent: api-ysql-exprs
+aliases:
+  - /latest/api/ysql/exprs/func_yb_hash_code
```
polarweasel, line 11: Not required. That's already the file's URL. :)
```suggestion
```

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+`yb_hash_code` is a function that returns the hash of a set of given input values using the hash function DocDB uses to shard its data. In effect, it provides direct access to the hash value of any given row of a YSQL table, allowing one to infer a row’s physical location. This enables an application to specify queries based on the physical location of a row or set of rows. A user’s newfound access to the physical location of a row should open up possibilities for more efficient queries.
+
+## Instructions
+`yb_hash_code` can be used anywhere a normal function can be used in a YSQL expression. There is no limitation to where it can be used. It takes in a variable number of parameters whose types must be allowed in YB primary keys. More formally,
+
+```
+yb_hash_code(a1:t1, a2:t2, a3:t3, a4:t4…..) → int4 (32 bit integer)
+```
+
+Where `a1, a2, a3…` are expressions of type `t1, t2, t3…`, respectively. `t1, t2, t3…` must be types that are currently allowed in a primary key.
```
polarweasel, line 26: Careful on the auto-replace
```suggestion
Where `a1, a2, a3...` are expressions of type `t1, t2, t3...`, respectively. `t1, t2, t3...` must be types that are currently allowed in a primary key. 
```

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+showAsideToc: true
+---
+
+## Synopsis
+`yb_hash_code` is a function that returns the hash of a set of given input values using the hash function DocDB uses to shard its data. In effect, it provides direct access to the hash value of any given row of a YSQL table, allowing one to infer a row’s physical location. This enables an application to specify queries based on the physical location of a row or set of rows. A user’s newfound access to the physical location of a row should open up possibilities for more efficient queries.
+
+## Instructions
+`yb_hash_code` can be used anywhere a normal function can be used in a YSQL expression. There is no limitation to where it can be used. It takes in a variable number of parameters whose types must be allowed in YB primary keys. More formally,
+
+```
```
polarweasel, line 22: Every code block needs a language. (`output` removes the copy button and doesn't source-highlight.)
````suggestion
```output
````

File: docs/content/latest/api/ysql/exprs/func_yb_hash_code.md [hunk 5]
```diff
@@ -0,0 +1,195 @@
...
+
+Where `a1, a2, a3…` are expressions of type `t1, t2, t3…`, respectively. `t1, t2, t3…` must be types that are currently allowed in a primary key. 
+
+## Function Pushdown
+
+
+This function can be either evaluated at the Postgres layer after resolving each of its argument expressions, or certain invocations of it may be pushed down to be evaluated at the DocDB layer. We discuss the situations where they are pushed down. `yb_hash_code` invocations are pushed down when the Postgres optimizer determines that the return values of the function will directly match the values of the hash value column in a requested base table or index table. For example, if we create a table as follows
+
+```
```
polarweasel, line 33: (Also do this for the rest)
````suggestion
```sql
````

File: docs/content/latest/api/ysql/exprs/func_yb_hash_code.md [hunk 6]
```diff
@@ -0,0 +1,195 @@
...
+
+Then the following query will evaluate the `yb_hash_code` calls at the DocDB layer:
+
+```
+SELECT * FROM sample_table WHERE yb_hash_code(x,y) <= 128 AND yb_hash_code(x,y) >= 0;
+```
+
+You can expect to verify this with the `EXPLAIN ANALYZE` result of this statement
+```
```
polarweasel, line 44: Always surround code blocks with blank lines
````suggestion

```output
````

File: docs/content/latest/api/ysql/exprs/func_yb_hash_code.md [hunk 7]
```diff
@@ -0,0 +1,195 @@
...
+                                                            QUERY PLAN
```
polarweasel, line 45: Kill trailing whitespace
```suggestion
                                                            QUERY PLAN
```

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+                                                            QUERY PLAN                                                             
+-----------------------------------------------------------------------------------------------------------------------------------
+ Index Scan using sample_table_pkey on sample_table  (cost=0.00..5.88 rows=16 width=12) (actual time=2.867..2.919 rows=18 loops=1)
+   Index Cond: ((yb_hash_code(x, y) <= 128) AND (yb_hash_code(x, y) >= 0))
+ Planning Time: 0.052 ms
+ Execution Time: 2.953 ms
+(4 rows)
+```
+
+Here, we see that the primary key index was used and no row was rechecked at the postgres layer.
```
polarweasel, line 54: PostgreSQL is a trademark, so we should use it rather than nicknames. ("Postgres" with a capital P is also trademarked, but for consistency, I'm trying to get us using the full name.)
```suggestion
Here, we see that the primary key index was used and no row was rechecked at the PostgreSQL layer.
```

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+CREATE INDEX sample_idx_dup ON sample_table ((x,z) HASH);
+```
+
+And then we execute the same query with a pg_hint_plan directive:
+
+```
+SET pg_hint_plan.enable_hint=ON;
+/*+IndexScan(sample_table sample_idx_dup) */
+EXPLAIN ANALYZE SELECT * FROM sample_table WHERE yb_hash_code(x,z) <= 128 AND yb_hash_code(x,z) >= 0;
+                                                          QUERY PLAN
```
polarweasel, line 95: Split the inputs and outputs
````suggestion
```sql
SET pg_hint_plan.enable_hint=ON;
/*+IndexScan(sample_table sample_idx_dup) */
EXPLAIN ANALYZE SELECT * FROM sample_table WHERE yb_hash_code(x,z) <= 128 AND yb_hash_code(x,z) >= 0;
```

```output
                                                          QUERY PLAN
````

File: docs/content/latest/api/ysql/exprs/func_yb_hash_code.md [hunk 10]
```diff
@@ -0,0 +1,195 @@
...
+ Index Scan using sample_idx_dup on sample_table  (cost=0.00..6.04 rows=16 width=12) (actual time=15.551..15.608 rows=18 loops=1)
+   Index Cond: ((yb_hash_code(x, z) <= 128) AND (yb_hash_code(x, z) >= 0))
+ Planning Time: 14.775 ms
+ Execution Time: 15.709 ms
+(4 rows)
+```
+
+We can also mix calls that can be pushed down and calls that cannot in a single statement as such 
+
+```
```
polarweasel, line 106: Same, split input/output

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+  latest:
+    identifier: api-ysql-exprs-yb_hash_code
+    parent: api-ysql-exprs
+aliases:
+  - /latest/api/ysql/exprs/func_yb_hash_code
+isTocNested: true
+showAsideToc: true
+---
+
+## Synopsis
```
polarweasel, line 16: Single blank line before/after headings (lots of these throughout)
```suggestion
## Synopsis

```

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+`yb_hash_code` is a function that returns the hash of a set of given input values using the hash function DocDB uses to shard its data. In effect, it provides direct access to the hash value of any given row of a YSQL table, allowing one to infer a row’s physical location. This enables an application to specify queries based on the physical location of a row or set of rows. A user’s newfound access to the physical location of a row should open up possibilities for more efficient queries.
+
+## Instructions
```
polarweasel, line 19:
```suggestion
## Instructions

```

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+
+We can consider a modified version of the above `SELECT` query as follows:
+
+```
+SELECT * FROM sample_table WHERE yb_hash_code(x,z) <= 128 AND yb_hash_code(x,z) >= 0;
+```
+
+Here, the `yb_hash_code` calls are on `x` and `z`. `x` and `z` do not form the full hash key of the base sample_table table but they do form the full hash key of the index table, `sample_idx`. Hence, the optimizer will consider pushing down the calls to the DocDB layer if an index scan using `sample_idx` is chosen. Note that the optimizer may choose not to go with a secondary scan if it deems the requested hash range to be large enough to warrant doing a simple full table scan instead. The `EXPLAIN ANALYZE` result for this could be as follows
+
+```
```
polarweasel, line 72: Split inputs and outputs as line 91

docs/content/stable/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
+---
```
polarweasel, line 1: Same comments as the version in /latest :)

docs/content/latest/api/ysql/exprs/func_currval.md
```diff
@@ -4,7 +4,7 @@ headerTitle: currval()
 linkTitle: currval()
 description: Returns the last value returned by the nextval() function for the specified sequence in the current session.
```
polarweasel, line 5: This description doesn't agree with the page title
```suggestion
description: Returns the last value returned by the currval() function for the specified sequence in the current session.
```

polarweasel (changes requested):
This is looking really good. I've left some comments on particular lines. One overall comment: we need to adjust this from first-person plural to second-person singular. In other words, use "you" instead of "we". If you'd like me to do that adjustment as part of an editing pass, just let me know.

docs/content/latest/api/ysql/exprs/func_currval.md
m-iancu: @polarweasel I think this is actually correct just ambiguously written. `nextval` advances the sequence and returns the new value. `currval` just returns the current value. So it ends up returning the value of the latest `nextval` call.
Postgres docs: https://www.postgresql.org/docs/11/functions-sequence.html phrase it as:
```
Return value most recently obtained with nextval for specified sequence
```
cc @tanujnay112

docs/content/latest/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+SELECT COUNT(*) FROM sample_table WHERE yb_hash_code(x,y) >= 16384 and yb_hash_code(x,y) < 24576;
+.
+.
+.
+.
+
+SELECT COUNT(*) FROM sample_table WHERE yb_hash_code(x,y) >= 57344 and yb_hash_code(x,y) < 65536;
+```
+
+Summing up all the results of these queries gives us the equivalent of a `COUNT(*)` over all of `sample_table`.
```
m-iancu, line 191: nit: once your script lands to yb-tools maybe we can reference that somewhere in this section?

docs/content/stable/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -0,0 +1,195 @@
...
+title: yb_hash_code() function [YSQL]
```
m-iancu, line 2: ditto about the comments -- make sure once you land you verify the diff between latest and stable versions are only the expected macro stuff (e.g. the path being stable vs latest).

m-iancu (approved):
LGTM, thanks @tanujnay112

polarweasel (approved):
LGTM

PR #10647, opened 2021-11-18
[docs] update markdownlint configuration

polarweasel:
Removing a couple of redundant entries, and adding some more specificity to a couple of rules.

(Also a tiny update to docs package-lock)


ddhodge (approved):
LGTM

PR #10657, opened 2021-11-19
[docs] Add to CREATE INDEX api/ysql

jaki:
In the api/ysql section of docs, the CREATE INDEX grammar is missing some components:

- NONCONCURRENTLY
- USING

Add those.


lizayugabyte:
The page TOC should be removed
Screen Shot 2021-11-19 at 4 25 32 PM

jaki:
@lizayugabyte, not sure what page you're showing, but I see the same issue on other pages like https://deploy-preview-10657--infallible-bardeen-164bc9.netlify.app/latest/api/ysql/keywords/ and https://deploy-preview-10657--infallible-bardeen-164bc9.netlify.app/latest/api/ysql/reserved_names/.  This should be out of scope and done separately.

bllewell:
The grammar has this:

```
[ USING access_method_name ]
```

But the "Semantics section" starts with this:

```
... USING method ...
```

Should be the same word `access_method_name` as in the grammar.

— — — — — — — — — — — — — — — — — — — — — — — — — 

Look for this:

«
access_method_name
The name of the index access method. By default, lsm is used for YugabyteDB tables and btree is used otherwise (for example, temporary tables). GIN indexes can be created in YugabyteDB by using the ybgin access method.
»

This is add odds with:

«
... USING ... options are not yet supported.
»

aelier on the page.

polarweasel (approved):
LGTM

(Aside: the issue @lizayugabyte commented on appears to be an artifact of the responsive design. Can't easily do anything about it.)

bllewell:
Presumably the grammar will eventually be this:

```
INDEX [ NONCONCURRENTLY | CONCURRENTLY ] 
```

It would be more consistent with the rest of this section if you showed this now and said in the semantics that CONCURRENTLY isn't supported.

Yes? So one of these must be the default when you use neither. You must state which one it is.

The wording:

«
To disable online schema migration for YSQL CREATE INDEX, set the flag ysql_disable_index_backfill=true on all nodes and both master and tserver. To disable online schema migration for one CREATE INDEX, use CREATE INDEX NONCONCURRENTLY.
»

suggests that CONCURRENTLY (currently not supported) must be the default.

This is so confusing that you can't publish before you fix it.

jaki:
@bllewell, addressed your comments.

PR #10670, opened 2021-11-23
[docs] Follower reads in YSQL

### LINKED ISSUE #10506: [Docs] Add docs for YSQL follower reads
### Description

internal functional spec: https://docs.google.com/document/d/1Pb-ZxegHKwFxvMBfcRoOIB2ldZluKAYkCiqM0uT4o2A/edit#heading=h.mior1do4w7bm

cc @amitanandaiyer for any content asks

aishwarya24:
Resolves #10506

Updated Build preview : https://deploy-preview-10670--infallible-bardeen-164bc9.netlify.app/latest/explore/multi-region-deployments/follower-reads-ysql/ 
- Added the document to both latest and stable. 
- The document currently includes an introduction, details about the session variables and various examples that demonstrates follower reads.
- The default for follower reads now points to YSQL.
 
@netlify /latest/explore/multi-region-deployments/follower-reads-ysql/


bmatican:
Added Amit, if you wanted a technical review of the content as well.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
       <i class="icon-cassandra" aria-hidden="true"></i>YCQL</a>
-  </li>  
-
+  </li>
 </ul>
 
-YSQL follower reads support is currently in Beta. 
+## Introduction
+
+YugabyteDB requires reading from the **leader** to read the latest data. However, for applications that do not require the latest data, and/or are working with unchanging data; the cost of contacting a potentially remote leader to fetch the data may be wasteful. The application could benefit with better latency by reading from a **replica** that may be closer to the client.
```
amitanandaiyer, line 31: `benefit with` sounds off. Perhaps a native speaker may check if this is the right usage.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+```
+
+This next two examples use follower reads since **pg_hint** can be used during PREPARE and CREATE FUNCTION to do follower reads.
+
+```sql
+set session characteristics as transaction read write;
+set yb_read_from_followers = true;
+
+PREPARE select_stmt(text) AS
+/*+ Set(transaction_read_only on)*/
```
amitanandaiyer, line 114: Do we need a space before */

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+
+- *yb_follower_read_staleness_ms* : Defaults to 30 seconds.
+
+### Expected behavior
+
+The table describes what the expected behavior is when a read happens from a follower.
+
+| Conditions | Expected behavior |
+| --- | --- |
+| yb_read_from_followers is true AND <br> (Transaction is marked read-only <br>OR<br> Single-statement-select can be inferred to be read-only) <br> | Read will happen from the closest follower|
```
amitanandaiyer, line 55: single statement select inferrence isn't landed yet.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+|yb_read_from_followers is false OR <br> Transaction/statement is not read only | Read will happen from the leader |
+
+### Read from follower conditions
+
+- If the follower’s safe-time is larger than the read-time thus chosen, the follower may serve the read without any delay.
```
kmuthukk, line 60: "thus chosen" -- chosen how? This hasn't been explained before on this page.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+
+- If the follower is not yet caught up to the specified timestamp, the read will be redirected to a different replica transparently from the end-user. The end user may see a slight increase in latency depending on the location of the replica which satisfies the read.
```
kmuthukk, line 62: "the specified timestamp" -- which timestamp? Again, this concept has not been explained.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+  - The current behavior is that a follower/replica would wait for the requested timestamp to become safe. This behavior shall be changed to respond back to the `ybclient/pggate` immediately, so that the request may be routed to a different server that has the data.
```
kmuthukk, line 63: This discussion is not relevant to external documentation, @amitanandaiyer .

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+  - The current redirection policy is to directly go to the leader if the follower read is rejected by a follower. This behavior could be also be changed going forward, to explore other replicas before going to the leader.
+  - If the request reaches the leader, it will no longer reject/redirect the request.
+
+### Read-only transaction conditions
+
+Regarding marking the transaction as read only, a user can do one of the following:
+
+- `SET TRANSACTION READ ONLY` : Applies only to the current transaction block.
```
kmuthukk, line 71: Minor: No space before the : 

Ditto on the next two lines also.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+
+- If the follower is not yet caught up to the specified timestamp, the read will be redirected to a different replica transparently from the end-user. The end user may see a slight increase in latency depending on the location of the replica which satisfies the read.
+  - The current behavior is that a follower/replica would wait for the requested timestamp to become safe. This behavior shall be changed to respond back to the `ybclient/pggate` immediately, so that the request may be routed to a different server that has the data.
+  - The current redirection policy is to directly go to the leader if the follower read is rejected by a follower. This behavior could be also be changed going forward, to explore other replicas before going to the leader.
+  - If the request reaches the leader, it will no longer reject/redirect the request.
+
+### Read-only transaction conditions
+
+Regarding marking the transaction as read only, a user can do one of the following:
+
```
kmuthukk, line 70: How come there's no mention of pg_hint syntax in this section as another way of achieving same thing?

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+```
+
+```output
+ k  | v
+----+----
+ k1 | v1
+(1 row)
+```
+
+This next two examples use follower reads since **pg_hint** can be used during PREPARE and CREATE FUNCTION to do follower reads.
```
kmuthukk, line 107: Can't pg_hint method be used for a top-level/standalone SELECT statement? Does it have to be only via PREPARE or CREATE FUNCTIONA?

kmuthukk (commented):
Left some comments/questions on this.

Thx. CC: @amitanandaiyer

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+```
+
+```output
+ k  |  v
+----+------
+ k1 | v1
+(1 row)
+```
+
+The final example demonstrates that SELECTS error out in read-only transactions.
```
kmuthukk, line 242: The wording needs to be improved here. 

It is not that SELECTs error out in a read-only transaction. It has more do with the fact that SELECT INTO is not a read-only transaction.

And this example has nothing to do with follower reads. Because we would get the same error even if follower reads was off.

So what's the point  of this example?

CC: @amitanandaiyer , @aishwarya24 ?

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+ k  |  v
+----+------
+ k1 | v1
+(1 row)
+```
+
+```sql
+set yb_follower_read_staleness_ms = 5000;
+
+select * from t where k = 'k1';   /* 5s old value */
```
kmuthukk, line 207: Is this 5s old value or up to 5s old value?

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+
+- *yb_read_from_followers* : This session variable controls whether read from followers is enabled.
+
+- *yb_follower_read_staleness_ms* : This session variable defines the maximum staleness that the user is willing to tolerate.
+
+### Default behavior
+
+- *yb_read_from_followers* : Defaults to false.
+
+- *yb_follower_read_staleness_ms* : Defaults to 30 seconds.
```
polarweasel, line 47:
```suggestion
Two session variables control the behavior of follower reads:

- `yb_read_from_followers` controls whether reading from followers is enabled. Default is false.

- `yb_follower_read_staleness_ms` sets the maximum allowable staleness. Default is 30 seconds.
```

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
polarweasel, line 55:
```suggestion
| Conditions | Expected behavior |
| :--------- | :---------------- |
| yb_read_from_followers is true AND transaction is marked read-only | Read happens from the closest follower |
```

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+
+### Expected behavior
+
+The table describes what the expected behavior is when a read happens from a follower.
+
+| Conditions | Expected behavior |
+| --- | --- |
+| yb_read_from_followers is true AND <br> (Transaction is marked read-only <br>OR<br> Single-statement-select can be inferred to be read-only) <br> | Read will happen from the closest follower|
+|yb_read_from_followers is false OR <br> Transaction/statement is not read only | Read will happen from the leader |
```
polarweasel, line 56:
```suggestion
| yb_read_from_followers is false OR transaction/statement is not read-only | Read happens from the leader |
```

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
polarweasel: My suggestion below addresses this, and we should add a separate row to the table when it does land. There's no need for complex logic in a reference table. :)

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
polarweasel: Do we actually mean "If the follower's data is fresher than the client's yb_follower_read_staleness_ms value" here?

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
polarweasel: Even better: no colon, lowercase the word that follows: "`SET TRANSACTION READ ONLY` applies only to the..."

polarweasel, line 71:
```suggestion
- `SET TRANSACTION READ ONLY` applies only to the current transaction block.
```

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+  - The current redirection policy is to directly go to the leader if the follower read is rejected by a follower. This behavior could be also be changed going forward, to explore other replicas before going to the leader.
+  - If the request reaches the leader, it will no longer reject/redirect the request.
+
+### Read-only transaction conditions
+
+Regarding marking the transaction as read only, a user can do one of the following:
+
+- `SET TRANSACTION READ ONLY` : Applies only to the current transaction block.
+- `SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY` : Applies the read-only setting for all statements and transaction blocks that follow.
+- `SET default_transaction_read_only = TRUE` : Applies the read-only setting for all statements and transaction blocks that follow.
```
polarweasel, line 73:
```suggestion
- `SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY` applies the read-only setting to all statements and transaction blocks that follow.
- `SET default_transaction_read_only = TRUE` applies the read-only setting to all statements and transaction blocks that follow.
```

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
aishwarya24: Will change this to "up to 5s old value" wherever applicable.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,245 @@ menu:
...
+```output
+ k  |  v
+----+------
+ k1 | v1+1
+(1 row)
+```
+
+```sql
+postgres=# set yb_follower_read_staleness_ms = 25000;
+postgres=# select * from t where k = 'k1';   /* 25s old value */
```
polarweasel, line 232: Why are these Postgres prompts and not ysqlsh?

polarweasel (changes requested):

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
polarweasel: (Same for the 15 and 25 versions :) )

polarweasel:
The FR page(s) clearly need more contextual information, but in the interests of getting _something_ documented since 2.11 is already out, I would suggest we address the errors and other things we can do reasonably quickly, and then follow up with more thorough explanations soon, like next week. @kmuthukk and @amitanandaiyer does that work for you?

kmuthukk:
Sure @polarweasel .

@aishwarya24 -- when the round of comments have been incorporated, please let me know the netlify preview url, and will take a look again,

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: I'd say rewite this as:

If the follower's safe-time is at least <current_time> - <staleness>, 

If the follower is not yet caught up to <current_time> - <staleness> ...

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: If the follower is not yet caught up to `<current_time> - <staleness> `...

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: Yes. The 3 sub bullets:
Screen Shot 2021-11-30 at 10 32 08 AM

may be removed from the external documentation.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: updated the FS.

Can you pls also add

* SET transaction_read_only = TRUE (applies to the current transaction block)
* /*+ Set(transaction_read_only true) */ SELECT (applies only to the current transaction/select statement)
       
here.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: Yes. the pg_hint method can also be used for stand_alone statements.

The important point to highlight there is that for Prepare/Functions, the pg_hint needs to be applied at the prepare/function-definition stage. Not at the execute stage.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: Yes. That's true. We could just remove this example.

It was added to the FS to clarify the discussion regarding how we'd prevent/safegurard against a select statement  having non-read-only side-effects. 

Doesn't need to be part of the user documentation.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
aishwarya24: I will include it as a note.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+```
+
+```output
+ k  | v
+----+----
+ k1 | v1
+(1 row)
+```
+
+The following examples use follower reads since **pg_hint** can be used during PREPARE, CREATE FUNCTION, and SELECT to do follower reads.
```
kmuthukk, line 100: replace all occurrences of pg_hint with pg_hint_plan, as I believe that's the correct word.

since **pg_hint**
-->
since the **pg_hint_plan** mechanism

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+Replicas may not be completely up to date with all updates, so this design may respond with stale data. You can specify how much staleness the application can tolerate. When enabled, read-only operations may be handled by the closest replica, instead of having to go to the leader.
+
+## Surface area and usage
+
+### Surface area
+
+Two session variables control the behavior of follower reads:
+
+- `yb_read_from_followers` controls whether reading from followers is enabled. Default is false.
+- `yb_follower_read_staleness_ms` sets the maximum allowable staleness. Default is 30 seconds.
```
kmuthukk, line 42: is 30 seconds.
-->
is 30000 (i.e. 30 secs).

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+
+### Read from follower conditions
+
+- If the follower’s safe-time is at least `<current_time> - <staleness>`, then the follower may serve the read without any delay.
+
+- If the follower is not yet caught up to `<current_time> - <staleness>`, then the read will be redirected to a different replica transparently from the end-user. The end user may see a slight increase in latency depending on the location of the replica which satisfies the read.
+
+### Read-only transaction conditions
+
+Regarding marking the transaction as read only, a user can do one of the following:
```
kmuthukk, line 61: Regarding marking the transaction
-->
To mark a transaction

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+
+- `SET TRANSACTION READ ONLY` applies only to the current transaction block.
+- `SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY` applies the read-only setting to all statements and transaction blocks that follow.
+- `SET default_transaction_read_only = TRUE` applies the read-only setting to all statements and transaction blocks that follow.
+- `/*+ Set(transaction_read_only true) */ SELECT` applies only to the current transaction/select statement.
```
kmuthukk, line 66: Use [pg_hint_plan](https://docs.yugabyte.com/latest/explore/query-1-performance/pg-hint-plan/) mechanism	to embed the hint along with the SELECT statement. For example,` /*+ Set(transaction_read_only true) */ SELECT ...`  applies only	to the current SELECT  statement.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+
+```output
+ k  | v
+----+----
+ k1 | v1
+(1 row)
+```
+
+```sql
+set session characteristics as transaction read write;
```
kmuthukk, line 143: Why is this line needed?

And overall the simple SELECT with the embedded hint should be the first example for pg_hint_plan, instead of the more uncommon cases of PREPARE and CREATE FUNCTION. The flow needs to be simple to advanced rather than reverse.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+ k  | v
+----+----
+ k1 | v1
+(1 row)
+```
+
+The following examples use follower reads since **pg_hint** can be used during PREPARE, CREATE FUNCTION, and SELECT to do follower reads.
+
+{{< note title="Note" >}}
+The pg_hint needs to be applied at the prepare/function-definition stage and not at the `execute` stage.
```
kmuthukk, line 103: The pg_hint needs to 
-->
The pg_hint_plan hint needs to

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+ k | v | k | v
+---+---+---+---
+ 3 | 6 | 2 | 6
+(1 row)
+```
+
+The following examples demonstrate **staleness** after enabling the `yb_follower_read_staleness_ms` property.
+
+```sql
+set session characteristics as transaction read write;
```
kmuthukk, line 177: @amitanandaiyer , @aishwarya24 :

between 'v1' becoming 'v1+1' don't you need a 10sec sleep to illustrate the point. I don't see that being explicitly mentioned. Would be good to show the INSERT of 'v1' and a sleep 10s after that also, in my opinion.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -9,25 +9,247 @@ menu:
...
+UPDATE t SET  v = 'v1+1' where k = 'k1';
+/* sleep 10s */
+
+UPDATE t SET  v = 'v1+2' where k = 'k1';
+/* sleep 10s */
+
+select * from t where k = 'k1';
+```
+
+```output
```
kmuthukk, line 198: Might be worth adding some commentary along with..

For example, here I would say:

This selects the latest version of the row because the transaction setting for the session is `read write`.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: It is just to highlight that pg_hint can be used in a read_write session to override the read only property.

We don't need this line per se. The example would work even without this line. So we can remove this, and otehr places in line 125 and 107.

Also, agree with moving this example above line 107. We can show this example, then the PREPARE example and then the CREATE FUNCTION example.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
amitanandaiyer: So the assumption here was that the initial state of the row was `v1` (as depicted by the select). With the write being made (at least 5+ seconds) in the past.

we could add it here, to make it more explicitly, if the default assumption is that the table/row is empty.

So replace the

```
insert into t values ('k1', 'v1')
/* sleep 10s */
```

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
aishwarya24: Thank you ! I'll make the change accordingly.

docs/content/latest/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -39,7 +39,7 @@ Replicas may not be completely up to date with all updates, so this design may r
 Two session variables control the behavior of follower reads:
 
 - `yb_read_from_followers` controls whether reading from followers is enabled. Default is false.
-- `yb_follower_read_staleness_ms` sets the maximum allowable staleness. Default is 30 seconds.
+- `yb_follower_read_staleness_ms` sets the maximum allowable staleness. Default is 30000(30 seconds).
```
kmuthukk, line 42: there should be a space before (

docs/content/stable/explore/multi-region-deployments/follower-reads-ysql.md
```diff
@@ -39,7 +39,7 @@ Replicas may not be completely up to date with all updates, so this design may r
 Two session variables control the behavior of follower reads:
 
 - `yb_read_from_followers` controls whether reading from followers is enabled. Default is false.
-- `yb_follower_read_staleness_ms` sets the maximum allowable staleness. Default is 30 seconds.
+- `yb_follower_read_staleness_ms` sets the maximum allowable staleness. Default is 30000(30 seconds).
```
kmuthukk, line 42: space before (

kmuthukk (commented):
Why do we see both latest/explore and stable/explore version of changes here?

My understanding is that follower reads is right now only available in 2.11 and not yet in a stable release.

CC: @ymahajan

aishwarya24:
@kmuthukk , Confirmed with @polarweasel and have applied the changes only to latest/explore. Thank you.

kmuthukk (approved):
Thx!

polarweasel:
Sent you a few slack comments. Probably GTG for now, but will need some followup adjustments.

PR #10782, opened 2021-12-04
[docs] Update extensions in GIN explore doc

jaki:
- change hstore to supported
- add intarray to the list as unsupported


polarweasel (approved):
LGTM

PR #10817, opened 2021-12-09
[docs] remove conflicting redirects from /develop

polarweasel:
Removing a pair of conflicting (and very old) aliases. The deploy preview should be a 404.

@netlify /develop


ddhodge (approved):
LGTM

PR #10823, opened 2021-12-09
[docs] Make sure references to firewall are accurately titled

tylarb:
Our docs reference firewalld commands as SElinux, but the two are
different concepts. The docs we have in this troubleshooting section are
specifically for firewall, so this commit just corrects the titles of
the sections.

I also added a proper link to our support portal

Signed-off-by: Tyler Ramer <tramer@yugabyte.com>


polarweasel (approved):
LGTM. Thanks, Tyler!

PR #10892, opened 2021-12-16
[docs] 2.11.1.0 release

polarweasel:
@netlify /latest/releases/whats-new/latest-release/#v2.11.1.0

@VijiYB: if you get workaround info for the platform known issue, please add it!


VijiYB:
Sure. Can you add this one as known issue too? While adding read replicas, sometimes the instance type and volume doesn't show any values. Work around is to reselect the provider so that these values populate correctly as primary cluster. Bug # https://yugabyte.atlassian.net/browse/PLAT-2487

VijiYB (approved):

Issue #10860, opened 2021-12-13
[Platform][Kubernetes] Zone configurations are overwritten when kubeconfig is not provided


bhavin192:
### Description

The commit https://github.com/yugabyte/yugabyte-db/commit/d29f73f818de303531034abb4b6599666d6d57b0#diff-fc957da0b60dcc2d3632c191ca98d5b5d765d560800f4a34b0e224fc3232d1acR262 added support for in-cluster credentials. It basically sets KUBECONFIG variable to a blank string. 

It uses `az.setConfig` whose behavior has been changed in https://github.com/yugabyte/yugabyte-db/commit/c0b9786ec2eec41d5773eef07f88d87ed77d189e#diff-440e6f700b5643beb4942f556420450439b718acf8faf493cd5c7311be538274R66

So, when we don't upload kubeconfig, it replaces all the AZ level configs such as Helm overrides with `{"KUBECONFIG": ""}`. It should be changed to use the new method `az.updateConfig` instead.

PR #10928, opened 2021-12-22
[Platform][Kubernetes] Use az.updateConfig for incluster credentials

bhavin192:
c0b9786ec2eec41d5773eef07f88d87ed77d189e changed the behavior of az.setConfig and added az.updateConfig. This change uses the correct method to add `{"KUBECONFIG": ""}` to the existing configuration instead of overwriting the whole AZ configuration.

This was missed when I rebased https://github.com/yugabyte/yugabyte-db/pull/8371 against latest changes.

Fixes https://github.com/yugabyte/yugabyte-db/issues/10860


sb-yb (approved):
Thanks for the change!
FWIW we have some notebooks here for demoing and testing k8s API:
https://github.com/yugabyte/yugabyte-db/blob/master/managed/api-examples/python-simple/k8s.ipynb

Feel free to add to it.

PR #11058, opened 2022-01-11
[Docs] Fix tree typo in docs

soccerties:
Simple typo fix in scaling ysql queries files


polarweasel (approved):
Good catch! And thanks for fixing it all the way back to 2.0 👍

PR #11083, opened 2022-01-13
[Docs] [manage] yb-admin page updated with upgrade_ysql syntax

aishwarya24:
Summary:
- Added a note to **Upgrade a deployment** page for `upgrade_ysql` about Yugabyte version support (2.8 or higher).
- Updated **yb-admin** cli page with the details.

@netlify /latest/admin/yb-admin/


docs/content/latest/manage/upgrade-deployment.md
```diff
@@ -110,7 +114,7 @@ For a successful YSQL upgrade, a message will be displayed as follows:
 YSQL successfully upgraded to the latest version
 ```
 
-In certain scenarios, YSQL upgrade can take longer than 60 seconds, which is the default timeout value for `yb-admin`. To account for that, run the command with a higher timeout value:
+In certain scenarios, a YSQL upgrade can take longer than 60 seconds, which is the default timeout value for `yb-admin`. To account for that, run the command with a higher timeout value:
```
polarweasel, line 117: What happens if yb-admin times out while the upgrade is running?

sushantrmishra: yeah, that is a good suggestions..

aishwarya24: Thank you @polarweasel and @sushantrmishra. Let me add it to the `yb-admin` page as well.

polarweasel (approved):
LGTM! Thank you @aishwarya24 and @sushantrmishra

sushantrmishra (approved):

Issue #9160, opened 2021-07-01
[docs] Describe the process for using xCluster replication with TLS


bmatican:
These Scenarios should be documented publicly

1. Setting up xCluster with both Producer & Consumer using the same TLS cert
2. Setting up xCluster with Producer & Consumer using different TLS certs
3. Setting up xCluster with Producer using TLS cert but not Consumer
4. Setting up xCluster with Consumer using TLS cert but not Producer

We support #1 & #2 for sure.  Steps are listed below.  #3 & #4 need some investigation to verify that they are supported use cases.

cc @nspiegelberg @rahuldesirazu

nspiegelberg:
I know that we have FLAGS_certs_for_cdc_dir  to handle a situation where the producer and consumer have different certificates.  I don’t think we’ve tested this lately or mixed mode.  This above flag also requires FLAGS_use_node_to_node_encryption, which tells me we were only coding for both on / both off situations.

PR #11337, opened 2022-02-02
[docs] [#9160] Describe the process for using xCluster replication with TLS

hulien22:
Adding section to the xcluster setup docs for how to create xCluster replications setups with TLS (where both universes have either the same certs or different certs)

[Alex edit to add: you can set the entry point for the PR preview by tagging the Netlify bot. It only reads these tags at initial PR creation time and again after any subsequent commits.]

@netlify /latest/deploy/multi-dc/async-replication/

### DIFF
--- a/docs/content/latest/deploy/multi-dc/async-replication.md
+++ b/docs/content/latest/deploy/multi-dc/async-replication.md
@@ -130,19 +130,19 @@ When both universes use different certificates, you need to store the certificat
 
 1. Copy this file to each node on the target. It needs to be copied to a directory named: `<certs_for_cdc_dir>/<source_universe_uuid>`.
 
-  For example, if you previously set `certs_for_cdc_dir=/home/yugabyte/yugabyte_producer_certs`, and the source universe's ID is `00000000-1111-2222-3333-444444444444`, then you would need to copy the cert file to `/home/yugabyte/yugabyte_producer_certs/00000000-1111-2222-3333-444444444444/ca.crt`.
+    For example, if you previously set `certs_for_cdc_dir=/home/yugabyte/yugabyte_producer_certs`, and the source universe's ID is `00000000-1111-2222-3333-444444444444`, then you would need to copy the cert file to `/home/yugabyte/yugabyte_producer_certs/00000000-1111-2222-3333-444444444444/ca.crt`.
 
 1. Finally, set up replication using `yb-admin setup_universe_replication`, making sure to also set the `-certs_dir_name` flag to the directory with the *target universe's* certificates (this should be different from the directory used in the previous steps).
 
-  For example, if you have the target's certificates in `/home/yugabyte/yugabyte-tls-config`, then you would run:
+    For example, if you have the target's certificates in `/home/yugabyte/yugabyte-tls-config`, then you would run:
 
-  ```sh
-  ./bin/yb-admin -master_addresses 127.0.0.11:7100,127.0.0.12:7100,127.0.0.13:7100 \
-    -certs_dir_name /home/yugabyte/yugabyte-tls-config \
-    setup_universe_replication 00000000-1111-2222-3333-444444444444 \
-    127.0.0.1:7100,127.0.0.2:7100,127.0.0.3:7100 \
-    000030a5000030008000000000004000,000030a5000030008000000000004005,dfef757c415c4b2cacc9315b8acb539a
-  ```
+    ```sh
+    ./bin/yb-admin -master_addresses 127.0.0.11:7100,127.0.0.12:7100,127.0.0.13:7100 \
+      -certs_dir_name /home/yugabyte/yugabyte-tls-config \
+      setup_universe_replication 00000000-1111-2222-3333-444444444444 \
+      127.0.0.1:7100,127.0.0.2:7100,127.0.0.3:7100 \
+      000030a5000030008000000000004000,000030a5000030008000000000004005,dfef757c415c4b2cacc9315b8acb539a
+    ```
 
 ## Verifying Replication
 
--- a/docs/content/latest/deploy/multi-dc/async-replication.md
+++ b/docs/content/latest/deploy/multi-dc/async-replication.md
@@ -64,7 +64,7 @@ After you created the required tables, you can set up asynchronous replication a
     ```
 
 
-The preceding command contains three table IDs: the first two are YSQL for the base table and index, and the third is the YCQL table. 
+The preceding command contains three table IDs: the first two are YSQL for the base table and index, and the third is the YCQL table.
 
 Also, be sure to specify all master addresses for source and target universes in the command.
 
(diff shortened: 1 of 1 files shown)


hulien22:
This is my first docs push so do let me know if I'm missing any steps..

docs/content/latest/deploy/multi-dc/async-replication.md
```diff
@@ -102,6 +102,23 @@ Once you have set up replication, load data into the source universe as follows:
...
+
+1. Ensure that `use_node_to_node_encryption` is set to `true` on all nodes (source and target).
+2. For each master and tserver on the target universe, set the gflag `certs_for_cdc_dir` to the parent directory where you will store all the producer certs for replication.
+3. Fetch all of the certificates that the source universe uses.
+4. Copy these certificates to each node on the target. They need to be copied to a directory named: `<certs_for_cdc_dir>/<producer_cluster_id>`.
+  - For example, if we previously set `certs_for_cdc_dir=/home/yugabyte/yugabyte_producer_certs`, and the source universe's id is `00000000-1111-2222-3333-444444444444`, then we would need to save all the source universe's certificates to `/home/yugabyte/yugabyte_producer_certs/00000000-1111-2222-3333-444444444444/*`.
+5. Finally, setup replication using `yb-admin setup_universe_replication`, making sure to also set the `-certs_dir_name` flag to the directory with the *target universe's* certificates (this should be different from the directory used in the previous steps).
```
polarweasel, line 120: Using 1. for each makes maintaining lists a lot easier. Also, are you able you add sample command lines for each of these?

````suggestion
1. Ensure that `use_node_to_node_encryption` is set to `true` on all nodes (source and target).

    ```sh
    $ command_here
    ```

1. For each master and tserver on the target universe, set the gflag `certs_for_cdc_dir` to the parent directory where you will store all the producer certs for replication.

1. Fetch all of the certificates that the source universe uses.

1. Copy these certificates to each node on the target, into a directory named `<certs_for_cdc_dir>/<producer_cluster_id>`.

  - For example, if you previously set `certs_for_cdc_dir=/home/yugabyte/yugabyte_producer_certs`, and the source universe's id is `00000000-1111-2222-3333-444444444444`, then you would save all the source universe's certificates to `/home/yugabyte/yugabyte_producer_certs/00000000-1111-2222-3333-444444444444/*`.

1. Finally, set up replication using `yb-admin setup_universe_replication`, making sure to also set the `-certs_dir_name` flag to the directory with the _target universe's_ certificates (this should be different from the directory used in the previous steps).
````

File: docs/content/latest/deploy/multi-dc/async-replication.md [hunk 2]
```diff
@@ -102,6 +102,23 @@ Once you have set up replication, load data into the source universe as follows:
 
 When completed, proceed to [Verifying Replication](#verifying-replication).
 
+## Setting Up Replication with TLS
+
+### Source and Target Universes have the Same Certificates
+
+If both universes use the same certificates, run the same `yb-admin` [`setup_universe_replication`](../../../admin/yb-admin/#setup-universe-replication) command as above, but also include the [`-certs_dir_name`](../../../admin/yb-admin#syntax) flag. Setting that to the target universe's certificate directory will make replication use those certificates for connecting to both universes.
```
polarweasel, line 109:
````suggestion
If both universes use the same certificates, run the same `yb-admin` [`setup_universe_replication`](../../../admin/yb-admin/#setup-universe-replication) command as above, but also include the [`-certs_dir_name`](../../../admin/yb-admin#syntax) flag. Setting that to the target universe's certificate directory will make replication use those certificates for connecting to both universes.

For example:

```sh
$ ./bin/yb-admin do_the_things_here --with_the_arguments
````

**Review by polarweasel — CHANGES_REQUESTED** [review]
Thanks, Julien! Great start. Small things, and this'll be good to go. 👍

File: docs/content/latest/deploy/multi-dc/async-replication.md (hunk 2 above)
Comment (lizayugabyte) [inline docs/content/latest/deploy/multi-dc/async-replication.md:109]: I recommend not using "as above" but instead repeating the command.

File: docs/content/latest/deploy/multi-dc/async-replication.md [hunk 3]
```diff
@@ -102,6 +102,23 @@ Once you have set up replication, load data into the source universe as follows:
...
+
+### Source and Target Universes have Different Certificates
+
+When both universes use different certificates, we need to store the certificates for the producer universe on the target universe. Each node on the target universe will have a parent directory (determined by the `certs_for_cdc_dir` gflag), under which there will be sub-directories for each source universe, which each contain their respective source universe's certificates.
```
lizayugabyte, line 113: "We" is not used in our documentation ("you" to indicate the reader).

lizayugabyte, line 113: "Each node on the target universe will have a parent directory (determined by the `certs_for_cdc_dir` gflag), under which there will be sub-directories for each source universe, which each contain their respective source universe's certificates." sentence should be broken down into bite-size chunks.

docs/content/latest/deploy/multi-dc/async-replication.md
```diff
@@ -102,6 +102,23 @@ Once you have set up replication, load data into the source universe as follows:
...
+
+1. Ensure that `use_node_to_node_encryption` is set to `true` on all nodes (source and target).
+2. For each master and tserver on the target universe, set the gflag `certs_for_cdc_dir` to the parent directory where you will store all the producer certs for replication.
+3. Fetch all of the certificates that the source universe uses.
+4. Copy these certificates to each node on the target. They need to be copied to a directory named: `<certs_for_cdc_dir>/<producer_cluster_id>`.
+  - For example, if we previously set `certs_for_cdc_dir=/home/yugabyte/yugabyte_producer_certs`, and the source universe's id is `00000000-1111-2222-3333-444444444444`, then we would need to save all the source universe's certificates to `/home/yugabyte/yugabyte_producer_certs/00000000-1111-2222-3333-444444444444/*`.
```
lizayugabyte, line 119: "We" is not used in our documentation; should be replaced with "you".

lizayugabyte, line 119: "universe's id" should be "universe's ID"

docs/content/latest/deploy/multi-dc/async-replication.md
lizayugabyte, line 120: "setup" should be "set up"

lizayugabyte (commented):
Left a few comments

docs/content/latest/deploy/multi-dc/async-replication.md
polarweasel: Covered in my suggested change above :)

polarweasel: got it above

docs/content/latest/deploy/multi-dc/async-replication.md
polarweasel: Got it above

hulien22: Ah cool, didn't know about the "1." trick. 

As for adding sample command lines, I don't really know what I could add.. For setting the flags, there are various ways of doing that (set as part of the yb-master command, set in the config files, set via platform, set via yb-ts-cli, etc). And then for copying files, there again multiple ways to do that.

polarweasel: Looking at the other sections of this file, there's not a lot in the way of example commands. I think you're right that we're good without adding them here.

polarweasel (approved):
I think we're good to go now...

Issue #11439, opened 2022-02-10
[docs] Build a Node.js application that uses YCQL


nijova:
Jira Link: [DB-4655](https://yugabyte.atlassian.net/browse/DB-4655)
https://docs.yugabyte.com/latest/quick-start/build-apps/nodejs/ycql/

## Issue description

In the sample code, in the select function:
```
const city = JSON.parse(row.location).city;
```
JSON.parse is not correct here, as retrieved result is already parsed.

## Suggested resolution

```
const city = row.location.city;
```

PR #11451, opened 2022-02-11
fix(docs): avoid error retrieving json data #11439

nijova:
fix issue #11439


nijova:
@aishwarya24 I am using node v16.14.0

docs/content/latest/quick-start/build-apps/nodejs/ycql.md
```diff
@@ -107,7 +107,7 @@ async.series([
     client.execute(select, function (err, result) {
       if (err) return next(err);
       var row = result.first();
-      const city = JSON.parse(row.location).city;
+      const city = row.location.city;
```
polarweasel, line 110: This change is correct, but we also need line 97 to change, as the INSERT statement is missing `location`:

```js
const insert = "INSERT INTO ybdemo.employee (id, name, age, language, location) " +
```

polarweasel (changes requested):
One further change required to get the code to actually run... If you can make it on each file, this is good to go. Thank you!

docs/content/latest/quick-start/build-apps/nodejs/ycql.md
nijova: @polarweasel I made the changes as suggested. But it looks to me like I messed up with git. So I'll see if I can fix that.

nijova:
Cant figure this out, so I think it's easiest to close this and make a new pr. (please see #11545)

PR #11524, opened 2022-02-18
[docs] remove restrictions from verify-migration

polarweasel:
Removing the million-row restriction now that @andrei-mart has fixed the underlying bug.

Right now the change is in latest only, but we'll need to backport this to 2.8 and 2.6 docs as those code updates happen.

@netlify /latest/migrate/migrate-from-postgresql/verify-migration/


andrei-mart (approved):
LGTM

Issue #11543, opened 2022-02-19
[docs] Comparison matrix for DynamoDB has a mistake


zyake:
https://docs.yugabyte.com/latest/comparisons/

## Issue description
Comparison matrix for DynamoDB has a mistake.

## Suggested improvement
On the matrix, "Consensus-driven, strongly-consistent replication" is not supported for DynamoDB.
But, in reality, DynamoDB is a consensus-driven database and it supports strongly-consistent replication.
Despite similarity of their product names(Dynamo and DynamoDB), DynamoDB adapts Paxos consensus protocol and leader-based approach.

The difference between Dynamo and DynamoDB is explained in the following re:invent video.
https://youtu.be/yvBR71D0nAQ?t=332

In addition, leader election is described as the fundamental design approach behind Amazon DynamoDB (DynamoDB) on the following article.
https://aws.amazon.com/builders-library/leader-election-in-distributed-systems/

PR #11549, opened 2022-02-21
Comparison matrix for DynamoDB has a mistake[#11543]

zyake:
Change DynamoDB's "Consensus-driven, strongly-consistent replication" category to supported as DynamoDB is a Consensus-driven database and supports strongly-consistent replication.


polarweasel:
Already addressed in https://github.com/yugabyte/yugabyte-db/pull/11577. Thank you for catching it, as well!

PR #11550, opened 2022-02-21
[PLAT-3171, PLAT-3169] Use provider KUBECONFIG in NodeUniverseManager

bhavin192:
Currently we don't use kubeconfig given by user at the time of provider creation, this in-turn means that any kubectl commands run by NodeUniverseManager uses in-cluster ServiceAccount credentials. In case of OpenShift the platform pod's ServiceAccount doesn't have permissions to run `kubectl exec` and hence operations like software upgrade fail.

- Addresses the scenario where node_name is pod-hostname_azname (multiaz universes): this fixes issues with operations like software upgrade, download logs etc.

Scenarios tested:

- Created one multi AZ Kubernetes universe, then ran universe upgrade. It worked as expected.
- Download logs works as expected as well.
- Ran same steps on OpenShift where platform's ServiceAccount doesn't have permission to run `kubectl exec`

### DIFF
--- a/managed/devops/bin/node_client_utils.py
+++ b/managed/devops/bin/node_client_utils.py
@@ -1,3 +1,4 @@
+import os
 import paramiko
 import subprocess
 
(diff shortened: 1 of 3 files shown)


sb-yb (approved):
Looks good. 
I was wondering why this code went through trouble of invoking python executable and then that in turn invoked kube exec. 
Recently we changed to use kubectl java library instead of invoking kubectl shell processes. 
I think e missed this case where we are doing the same thru python.

bhavin192:
@sb-yb yes, there are at least two more such scripts which I know, cluster_health.py and yb_backup.py.

PR #11567, opened 2022-02-23
[Backport 2.12][PLAT-3171, PLAT-3169] Use provider KUBECONFIG in NodeUniverseManager

bhavin192:
Original commit: 0a3c893cf25ef19544d84b2c21d5b156554d342e

Currently we don't use kubeconfig given by user at the time of
provider creation, this in-turn means that any kubectl commands run by
NodeUniverseManager uses in-cluster ServiceAccount credentials. In
case of OpenShift the platform pod's ServiceAccount doesn't have
permissions to run `kubectl exec` and hence operations like software
upgrade fail.

- Addresses the scenario where node_name is
  pod-hostname_azname (multiaz universes): this fixes issues with
  operations like software upgrade, download logs etc.

Scenarios tested:

- Created OpenShift universe where platform's ServiceAccount doesn't
  have permission to run `kubectl exec`
- Then ran universe upgrade. It worked as expected.
- Download logs works as expected as well.

### DIFF
--- a/managed/devops/bin/node_client_utils.py
+++ b/managed/devops/bin/node_client_utils.py
@@ -1,3 +1,4 @@
+import os
 import paramiko
 import subprocess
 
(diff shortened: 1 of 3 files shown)


Sahith02 (approved):
Looks good!

PR #11585, opened 2022-02-24
[docs][cloud] YSQL Ops known issue

ddhodge:
spikes in the YSQL Operations/sec chart due to cluster health checks

@netlify /latest/yugabyte-cloud/release-notes/#known-issues


VijiYB (approved):
LGTM

juanman2 (approved):
I think there is probably a better way to say that, but I cannot think of what that would be at this time.

PR #11595, opened 2022-02-25
[Backport 2.8][PLAT-3171, PLAT-3169] Use provider KUBECONFIG in NodeUniverseManager

bhavin192:
Currently we don't use kubeconfig given by user at the time of
provider creation, this in-turn means that any kubectl commands run by
NodeUniverseManager uses in-cluster ServiceAccount credentials. In
case of OpenShift the platform pod's ServiceAccount doesn't have
permissions to run `kubectl exec` and hence operations like software
upgrade fail.

- Addresses the scenario where node_name is
  pod-hostname_azname (multiaz universes): this fixes issues with
  operations like software upgrade, download logs etc.

### DIFF
--- a/managed/devops/bin/node_client_utils.py
+++ b/managed/devops/bin/node_client_utils.py
@@ -1,3 +1,4 @@
+import os
 import paramiko
 import subprocess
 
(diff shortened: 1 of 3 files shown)


Sahith02 (approved):
Looks good to me! 
Btw, have you tested if the download logs feature is working here without any problems?

bhavin192:
> Btw, have you tested if the download logs feature is working here without any problem?
Yes, I did. It is working fine with single AZ and multi AZ universes.

PR #11609, opened 2022-02-28
[docs] PR for Documentation for multiple subscription support with azure

lizayugabyte:
Addresses https://yugabyte.atlassian.net/browse/PLAT-3115

Preview: https://deploy-preview-11609--infallible-bardeen-164bc9.netlify.app/latest/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/azure/#how-to-use-a-shared-image-gallery


lizayugabyte:
@SergeyPotachev Could you, please, review and approve doc changes?

docs/content/latest/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/azure.md
```diff
@@ -180,15 +180,15 @@ You set up a shared gallery image on Azure as follows:
...
 
-    Ensure that the images are replicated to each region in which you are planning to use them. For example, configuration shown in the following illustration would only work for US East:
+    <br><br>Ensure that the images are replicated to each region in which you are planning to use them. For example, configuration shown in the following illustration would only work for US East:
 
     ![description](/images/yb-platform/install/azure/shared-gallery-replication.png)
 
 1. To use the image in Yugabyte Platform, enter the image version's resource ID into the **Marketplace Image URN/Shared Gallery Image ID** field of the **Specify Region Info** dialog.
 
-    
+   <br>Note that if you provide the resource ID, Yugabyte Platform infers  the resource group from it. If you provide only the name, then Yugabyte  Platform assumes that the resource group is the same as the group used  in the cloud provider.
```
SergeyPotachev, line 194: 1. I haven't heard that we can provide only name. I think it should be a full URL otherwise it will be not picked up.
2. This URL contains not only the resource group but subscription ID as well.

lizayugabyte: @SergeyPotachev I guess I misunderstood your comment in the doc ticket: "_The part about Private DNS Zone looks good enough. Also we need to add similar comments to a part related to Shared gallery image._"  I thought you suggested to add the Note from "How to define a private DNS zone" section to "How to use a shared image gallery" section, which I did.

Could you let me know what comments you'd like to be added to the Shared gallery image section?

SergeyPotachev: It looks like I missed some things while was checking the Private DNS section. My apologies.
What we need to say:
1. `Private DNS zone` could be `id` only and in such case it will be used together with default subscription id and resource group; If it has subscription id and resource group then they will be used instead of default values;
2. `Gallery Image ID` could be a full URL only which should contain a subscription id, a resource group name and the resource name itself. If subscription id/resource group are different from the default ones, they will be used instead.

lizayugabyte: Thanks, @SergeyPotachev. 
Just to make sure I understand correctly, I have to do this:
1) Delete "Note that if you provide the Resource ID, Yugabyte Platform infers the resource group from it. If you provide only the name, then Yugabyte Platform assumes that the resource group is the same as the group used in the cloud provider." from both sections.
2) Add "Private DNS zone could be id only and in such case it will be used together with default subscription id and resource group; If it has subscription id and resource group then they will be used instead of default values." to the  "How to define a private DNS zone" section.
3) Add "Gallery Image ID could be a full URL only which should contain a subscription id, a resource group name and the resource name itself. If subscription id/resource group are different from the default ones, they will be used instead." to the "How to use a shared image gallery" section.

Could you, please, confirm?

SergeyPotachev: For item 2. "if it has subscription ..." should be something like "if it is a full URL which contains subscription ID and resource group..."
All other things look good to me.

SergeyPotachev (approved):
LGTM!

PR #11651, opened 2022-03-04
Add about persistent connection setting in django doc

Sfurti-yb:
Added parameter to turn on persistent connections with Django for now. Let me know if I need to add a more detailed description on the performance of persistent connections. Although I am planning to add that in the blog.


ddorian (approved):
LGTM

PR #11719, opened 2022-03-10
[docs] [infra] update font awesome to version 5.14.4

aishwarya24:
Part 1 - Update the version to latest in the 5 series and then look at [Upgrade to version 6](https://fontawesome.com/docs/web/setup/upgrade/)
@netlify /latest/quick-start/build-apps/


polarweasel (approved):
This should be safe to merge, shouldn't it?

PR #11739, opened 2022-03-12
[docs] shortcodes for version numbers

polarweasel:
Adding a shortcode to insert versions and build numbers, which are stored in a JSON file (`currentVersions.json`). For an example, look at the change in `versions.md` that uses the new shortcode to add the current versions of latest and stable.

**Note**: The Hugo --renderStaticToDisk option that I've added into package.json requires a forthcoming version of Hugo. It was removed from 0.94.1 and will be re-added later. It promises to speed up the local live-reload server, but Hugo 0.94.x has a bad markdown rendering bug, so we have to wait for 0.95.x.

@netlify /latest/releases/versioning/


docs/content/latest/releases/_index.md
```diff
@@ -47,7 +47,7 @@ Support for YugabyteDB stable release series includes:
   * Yugabyte may direct Customers to Upgrade to a current release if a workaround does not exist.
```
ddhodge, line 47: missing bracket in bullet under On this page

ddhodge (approved):
LGTM

ddhodge (approved):
LGTM

docs/content/latest/releases/_index.md
polarweasel: I noticed that in some other PR and fixed it :)

PR #11748, opened 2022-03-14
[Docs] PR for Missing Information about add node

lizayugabyte:


docs/content/latest/yugabyte-platform/manage-deployments/remove-nodes.md
```diff
@@ -66,7 +66,7 @@ The action to release a node instance is available from the Removed internal sta
 Taking this action transfers the node to a BeingDecommissioned and then Decommissioned internal state, as follows:
 
 1. There is a wait for Master leader.
-2. This server is removed from blacklist on Master leader (blacklisting is described in [Remove node](#remove-node)).
+2. This server is removed from blacklist on Master leader (blacklisting is described in [Remove node](#remove-node/)).
```
ddhodge, line 69: Link broken. Pretty sure you don't add trailing slash on in-page links

docs/content/latest/yugabyte-platform/manage-deployments/remove-nodes.md
```diff
@@ -87,3 +87,5 @@ Taking this action completely eliminates the node, as follows:
 
 1. Removes the node record from the universe metadata.
 2. Updates metadata in the database only.
+
+You can recover a deleted node by following instructions provided in [Use Yugabyte Platform to add a node](../add-nodes/).
```
ddhodge, line 91: Reading this page, not clear to me that you can recover a 'deleted' node; in Add Node it says you can add back a node in 'decommissioned' status, which comes after Release Node Instance steps. Deleted node may be gone for good? Needs clarification.

ddhodge (approved):
2 comments

ddhodge:
The way these topics read, it appears you can only add a node back, after removing it

1. should Add a Node be moved after Eliminate in the left nav? 
2. should Add a Node be renamed Revive a node? 
3. should the Revive node page should be incorporated into Eliminate if it can only be done in that context?

Can you spontaneously add a brand new node using this process? If this is only about reviving nodes, then I would do 3.

docs/content/latest/yugabyte-platform/manage-deployments/remove-nodes.md
lizayugabyte: fixed

docs/content/latest/yugabyte-platform/manage-deployments/remove-nodes.md
lizayugabyte: Good catch. I moved that sentence to under "Release node instance".
(this whole article was really carefully crafted based on the input from dev, as internal node states and displayed statuses weren't always a match)

PR #11795, opened 2022-03-18
[docs][cloud] Notes on maintenance

ddhodge:
Add info on effect of rolling maintenance/upgrade on cluster

@netlify /latest/yugabyte-cloud/cloud-clusters/cloud-maintenance/#what-to-expect-during-maintenance

### DIFF
--- a/docs/content/latest/yugabyte-cloud/cloud-clusters/cloud-maintenance.md
+++ b/docs/content/latest/yugabyte-cloud/cloud-clusters/cloud-maintenance.md
@@ -55,5 +55,5 @@ To set the maintenance exclusion period for a cluster:
 Yugabyte performs rolling maintenance and upgrades on multi-node clusters with zero downtime. However, the cluster is still subject to the following:
 
 - Dropped connections - Connections to the stopped node are dropped. Verify your connection pool, driver, and application to ensure they handle dropped connections correctly. Any failures need to be retried.
-- No high availability - During maintenance, one node is always offline. In a cluster with a replication factor (RF) of 3, if one of the remaining nodes goes down, you can no longer access to the database. For clusters with a higher RF, there is less risk.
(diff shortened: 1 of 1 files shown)


docs/content/latest/yugabyte-cloud/cloud-clusters/cloud-maintenance.md
```diff
@@ -49,3 +49,11 @@ To set the maintenance exclusion period for a cluster:
 1. On the **Maintenance** tab, click **Edit Maintenance Preferences** to display the **Maintenance Preferences** dialog.
 1. Set a start date and end date. The exclusion period includes the day of the start date, and every day up to, but not including, the end date.
 1. Click **Save**.
+
+## What to expect during maintenance
+
+Yugabyte performs rolling maintenance and upgrades on multi-node clusters with zero downtime. However, the cluster is still subject to the following:
+
+- Dropped connections - Connections to the stopped node are dropped. Verify your connection pool, driver, and application to ensure they handle dropped connections correctly. Any failures need to be retried.
+- No high availability - During maintenance, one node is always offline. If one of the remaining 2 nodes goes down in a 3 node cluster, you lose access to the database. For clusters with more nodes, there is less risk.
```
lizayugabyte, line 58: Suggestion:
one of the remaining two nodes goes down in a 3-node cluster

lizayugabyte (approved):
One minor comment. Otherwise LGTM

mrajcevic01 (changes requested):
For the HA piece, it is more dependent on RF than number of nodes. This specific example should be calling out a RF of 3 rather than a 3 node cluster. Since RF of 3 gives you a fault tolerance of 1, that causes issues regardless of the number of nodes because whatever shards are on the node going down will not have HA.

PR #11802, opened 2022-03-18
[Doc] PR for Update (outdated) Platform IAM role example 

lizayugabyte:
Addresses https://yugabyte.atlassian.net/browse/PLAT-3341


lizayugabyte:
@chrisckchang Could you review the doc and approve the changes, please?

kevincox:
@lizayugabyte I don't have any way to verify this change. It appears to just add the permissions that we noticed with our feature set and our production use. I think it is good to add this as a start but to really get a reliable set of permissions we need to run a full QA eval using the listed permissions and ensure that all features are working correctly.

chidmuthu (approved):
Definitely need these, but work may not be finished. We can run itests to verify.

PR #11813, opened 2022-03-21
xCluster with GeoPartitioning doc

Arjun-yb:
Added a note in xCluster setup with GP section

### DIFF
--- a/docs/content/latest/deploy/multi-dc/async-replication.md
+++ b/docs/content/latest/deploy/multi-dc/async-replication.md
@@ -249,9 +249,13 @@ CREATE TABLE transactions_us
 ```
 
 ### Steps to create Replication(Unidirectional):
-1. Collect child table UUIDs from source universe (partition tables, here transactions_eu, transactions_in and transactions_us) - these will be used while setting up replication. To collect table UUIDs, go to the `Tables` section in the Admin UI (127.0.0.1:7000).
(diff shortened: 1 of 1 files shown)


docs/content/latest/deploy/multi-dc/async-replication.md
```diff
@@ -249,10 +249,14 @@ CREATE TABLE transactions_us
 ```
 
 ### Steps to create Replication(Unidirectional):
-1. Collect child table UUIDs from source universe (partition tables, here transactions_eu, transactions_in and transactions_us) - these will be used while setting up replication. To collect table UUIDs, go to the `Tables` section in the Admin UI (127.0.0.1:7000).
+1. Collect partition table UUIDs from source universe (partition tables, here transactions_eu, transactions_in and transactions_us) - these will be used while setting up replication. To collect table UUIDs, go to the `Tables` section in the Admin UI (127.0.0.1:7000).
 
   ![xCluster_with_GP](/images/explore/yb_xcluster_table_uuids.png)
 
+  {{< note title="Note" >}}
+  xCluster doesn't support parent to parent table replication, You have to use only partition tables for replication setup.
```
hulien22, line 257: Perhaps something like this sounds better, since xcluster does support the replication, its just that the parent table doesn't have any actual data in it:

> Only the partition tables need to be specified for replication setup - it is not necessary to specify the parent table.

hulien22 (approved):

Issue #11158, opened 2022-01-20
[Docs] Document PostgreSQL logging options


sushantrmishra:
### Description

Postgres support https://www.postgresql.org/docs/13/runtime-config-logging.html#GUC-LOG-LINE-PREFIX 

We added more options for log_line_prefix 

%C = cloud name
%R = region / data center name
%Z = availability zone / rack name
%U = cluster UUID
%N = node and cluster name
%H = current hostname

We should add these options to our documentation.

sushantrmishra:
In https://github.com/yugabyte/yugabyte-db/issues/10331 , Implemented logic for suppressing non-Postgres logs from the Postgres file. This is done using an environment variable FLAGS_suppress_nonpg_logs which, when set to true, will trigger the suppression

./bin/yb-ctl start --tserver_flags="ysql_pg_conf_csv=\"suppress_nonpg_logs=true\""

polarweasel:
In sum, we have three logs we generate: master, t-server, and Postgres (which includes our client process output). Need to describe overall how YugabyteDB does its logging, the formats of our master and t-server logs, _and_ how our log-line prefix additions work.

m-iancu:
Just fyi, this is the list of tserver flags that we have documented https://docs.yugabyte.com/latest/reference/configuration/yb-tserver/

polarweasel:
Also, update the descriptions for [--enable_ysql](https://docs.yugabyte.com/latest/reference/configuration/yb-master/#enable-ysql) in master and tserver to say they should always match.

ddhodge:
#11869 for logging and logs

PR #11843, opened 2022-03-22
[docs] Postgresql.conf file settings

ddhodge:
Additional YB-specific log_line_prefix options, and suppress_nonpg_logs setting.
#11158

@netlify /latest/reference/configuration/yb-tserver/#using-postgresql-conf


docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -850,14 +862,41 @@ Default: `false`
...
+- %Z = availability zone / rack name
+- %U = cluster UUID
+- %N = node and cluster name
+- %H = current hostname
+
+For information on using log_line_prefix, refer to [log_line_prefix](https://www.postgresql.org/docs/11/runtime-config-logging.html#GUC-LOG-LINE-PREFIX) in the PostgreSQL documentation.
+
+### suppress_nonpg_logs (boolean)
+
+When set, suppresses logging of non-PostgreSQL output.
```
nathanhjli, line 896: I think we should mention that it suppress logging of non-PostgresSQL output to the Postgres log file. Current wording might make it sound like it prevents all non-PostgresSQL logs to all files.

lizayugabyte (commented):
Minor comments. Otherwise LGTM

docs/content/latest/reference/configuration/yb-tserver.md
```diff
@@ -850,14 +862,41 @@ Default: `false`
...
-When set to true, allows files to expire purely based on their value-level TTL expiration time (even if it is lower than the table TTL). This is useful for times where a file needs to expire earlier than its table-level TTL would allow. If no value-level TTL metadata is available, then table-level TTL will still be used. 
+When set to true, files expire based on their value-level TTL expiration time, even if it is lower than the table TTL. This is useful when a file needs to expire earlier than its table-level TTL would allow. If no value-level TTL metadata is available, then table-level TTL is still used.
 
 {{< warning title="Warning">}}
 Use of this flag can potentially result in expiration of live data - use at your discretion.
 {{< /warning >}}
 
 Default: `false`
 
+## Using postgresql.conf
```
xenown, line 873: I'm not too sure if the usage of this file is something we want to actively expose to customers. Afaik, users can apply this flag (https://docs.yugabyte.com/latest/reference/configuration/yb-tserver/#ysql-pg-conf) to set postgres settings. I think there is a gflag as well that customers can use. I think we should confirm with someone else whether it makes sense to have line 875. 
For line 873, maybe change it to "Configuring Postgres"
For line 877, maybe change it to "For information on available Postgres settings, refer to [Server Configuration](https://www.postgresql.org/docs/11/runtime-config.html) in the PostgreSQL documentation."

xenown (commented):
Added one comment under "Using postgresql.conf". log_line_prefix section lgtm.

docs/content/latest/reference/configuration/yb-tserver.md
ddhodge: cc @sushantrmishra

sushantrmishra: I would suggest to remove mentions postgres.conf since we use the tserver flags etc to modify the the config rather then directly modifying the content of it.

I would  suggest to put the headline as Postgres Logging options . 

Rest of the content looks good.

tanujnay112 (commented):
Left a few nits

ddhodge:
See #12206

PR #11844, opened 2022-03-23
[Docs] Convert 'Back up data' and 'Restore Data' pages to a single 'Export and Import Data' page

vkulichenko:
@netlify /preview/manage/backup-restore/


docs/content/preview/manage/backup-restore/export-import-data-ycql.md
```diff
@@ -0,0 +1,118 @@
...
+```sh
+ycqlsh -e "SOURCE '<file>'"
+```
+
+Where:
+- `<file>` – the path to the CQL script file to import the schema from
+
+### Import Data
+
+After the schema is imported, you can import the actual data by using the `COPY FROM` command that loads the data from one or more CSV files.
```
polarweasel, line 96: I wonder if we should just say "files" here instead of "CSV files", since comma-separated is only the default, not the only option?

PR #11917, opened 2022-03-29
[doc] PR for Changes to account setup for kubernetes config

lizayugabyte:
Addresses https://yugabyte.atlassian.net/browse/PLAT-3531


yogendra:
This will work.

But I think it requires additional proper example commands that can just work with a copy paste. For the variable part (<namespace>) sample values from top should be used.

yogendra (commented):
LGTM

PR #11929, opened 2022-03-30
[Docs] PR for YB Platform default AWS install - not clear if platform install happens after license upload

lizayugabyte:


sharkberto (commented):
Looks good!

PR #11943, opened 2022-03-31
[docs] Changed explanation of cpu cores needed for helm

sumitdatabase:
Changes made -->

1. Added explanation of why we need 8 core machines in k8s install on GKE using helm.
2. Edited the install of kubectl section - installing under gcloud puts it into a different directory that comes typically after the homebrew path. Thus installing under gcloud root directory may not do anything. Plus kubectl can be installed on Windows and Mac separately, separate from gcloud anyway.

### DIFF
--- a/docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
+++ b/docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
@@ -70,7 +70,7 @@ $ gcloud config set compute/zone us-west1-b
 
 - Install `kubectl`
 
-Refer to your specific [operating system](https://kubernetes.io/docs/tasks/tools/) for kubectl install.
+Refer to your specific [operating system](https://kubernetes.io/docs/tasks/tools/) for kubectl installation instructions.
 
 
 Note that GKE is usually 2 or 3 major releases behind the upstream/OSS Kubernetes release. This means you have to make sure that you have the latest kubectl version that is compatible across different Kubernetes distributions if that's what you intend to.
--- a/docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
+++ b/docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
@@ -45,9 +45,9 @@ You must have a GKE cluster that has Helm configured. If you have not installed 
 
 The YugabyteDB Helm Chart has been tested with the following software versions:
 
-- GKE running Kubernetes 1.18 (or later) with nodes such that a total of 12 CPU cores and 45 GB RAM can be allocated to YugabyteDB. This can be three nodes with 4 CPU core and 15 GB RAM allocated to YugabyteDB. `n1-standard-8` is the minimum instance type that meets these criteria.
+- GKE running Kubernetes 1.18 (or later). The helm chart you will use here to install yugabytedb requires 3 master servers and 3 t-servers each with 2 cpus - total 12 cpus. Thus you will need a kubernetes cluster with more than 12 cpus. With standard 3 nodes cluster, you will need each node with 8 cpus. 
+
 - Helm 3.4 or later
-- Docker image for YugabyteDB (`yugabytedb/yugabyte`) 2.1.0 or later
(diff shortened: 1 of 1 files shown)


docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
```diff
@@ -70,11 +70,8 @@ $ gcloud config set compute/zone us-west1-b
 
 - Install `kubectl`
 
-After installing Cloud SDK, install the `kubectl` command line tool by running the following command.
+Refer to your specific [operating system](https://kubernetes.io/docs/tasks/tools/) for kubectl install.
```
polarweasel, line 73: This doc is specific to GKE, though. Why remove this? I know you commented in the PR description that the path may vary, but it seems like that means we should just point out what the options are here, not punt to the k8s docs.

docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
```diff
@@ -45,9 +45,9 @@ You must have a GKE cluster that has Helm configured. If you have not installed
 
 The YugabyteDB Helm Chart has been tested with the following software versions:
 
-- GKE running Kubernetes 1.18 (or later) with nodes such that a total of 12 CPU cores and 45 GB RAM can be allocated to YugabyteDB. This can be three nodes with 4 CPU core and 15 GB RAM allocated to YugabyteDB. `n1-standard-8` is the minimum instance type that meets these criteria.
+- GKE running Kubernetes 1.18 (or later). The helm chart you will use here to install yugabytedb requires 3 master servers and 3 t-servers each with 2 cpus - total 12 cpus. Thus you will need a kubernetes cluster with more than 12 cpus. With standard 3 nodes cluster, you will need each node with 8 cpus.
```
polarweasel, line 48: Reworded a bit, but I'm still not clear about what the last sentence (about standard nodes) is trying to say.

Also pls verify that changing "requires" to "creates" is correct here!

```suggestion
- GKE running Kubernetes 1.18 (or later). The helm chart you use to install YugabyteDB creates 3 master servers and 3 tablet servers, each with 2 CPU cores, for a total of 12 CPU cores. This mean you need a Kubernetes cluster with more than 12 CPU cores. With a standard 3-node cluster, each node requires 8 CPU cores. 
```

polarweasel (changes requested):
Couple comments and changes...

docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
sumitdatabase: When someone installs kubectl on their machine, typically they would follow the instruction to install one kubectl cli, common for all k8s use cases across all clouds and on-prem. You will not have one kubectl for gcloud and another for aws and another for azure. 

So following the [doc](https://kubernetes.io/docs/tasks/tools/#kubectl) for example if using mac, you could use homebrew to install it. Or for windows, chocolatey ( similar to homebrew ) is another option. If  it is installed to a homebrew location then we do not need a kubectl installation under the gcloud install directory. 

For example on my mac,  I have kubectl in two places

- /opt/homebrew/bin/kubectl
- /opt/google-cloud-sdk/bin/kubectl

But since my $PATH variable lists /opt/homebrew/bin before /google-cloud-sdk/bin, when I type kubectl it runs the one under /opt/homebrew only. So the one one under gcloud diorectory is not used ( unless specify the full path )

Now the gcloud itself is installed now using homebrew, but using separate [install](https://cloud.google.com/sdk/docs/install) . Part of the install lets you specify which you want to gcloud to install.

We probably do not get too much into gcloud install ( :-) after all we are cloud native )..

polarweasel: Makes sense!

docs/content/latest/deploy/kubernetes/single-zone/gke/helm-chart.md
sumitdatabase: Hi @polarweasel  - I over-wrote on your suggestion - also changed from "This mean" to "This means". And changed the last sentence. Let me know if it looks good ? Thanks for your suggestions again!

PR #11954, opened 2022-04-01
Same change on helm-chart.md ( stable vs latest ) as #11943

sumitdatabase:
Making same change on helm-chart.md file around clarification on gke k8s cluster nodes - on stable branch as we did on latest branch last night


sumitdatabase:
Per conversation with Alex Ball, identical changes going in stable versus latest version of documentation.

PR #11971, opened 2022-04-04
[docs] minor change to tpcc download command

kerneljake:
removing strict requirement on $HOME makes copy button more useful


polarweasel (approved):
Thanks @kerneljake!

polarweasel:
@kerneljake good to go. But before you merge, could you make this same change in /stable as well? (No need for a re-review after)

PR #11988, opened 2022-04-05
[docs] rename Latest release series to Preview

polarweasel:
Lots of changes here:

* the /latest folder is now /preview
* all links to /latest are now /preview
* menus adjusted
* all community-slack signup links replaced with a shortcode so we don't have to keep changing it globally
* existing server-side redirects updated, and a catch-all added

I'm pretty sure we're going to see some external links break with this one, but I think we can adjust as we go.

Note: The PR preview link is deliberately pointing at a /latest URL with an anchor to test out the server-side redirect. Your browser should arrive at `/preview/quick-start/install/macos/#download-yugabytedb`.

@netlify /latest/quick-start/install/macos/#download-yugabytedb


ddhodge (approved):
LGTM

PR #12072, opened 2022-04-07
[Platform][Kubernetes client] Use pull secret from disk

bhavin192:
Fixes some issues from PLAT-2746.

The pullSecret is actually present on the disk, we were trying to load
it using getResourceAsStream, which was returning null. This resulted
in pull secret not getting created in the destination namespace. With
this change the pull secret is loaded from the disk and passed to the
Kubernetes client library.

It also modifies the logic to create Kubernetes client to account for
blank KUBECONFIG value. This variable is set to a blank string when
user doesn't provide a kubeconfig during provider creation. In such
cases we fall back to in-cluster credentials.

Scenarios tested:
- Created a couple of universes with yb.use_kubectl set to false. The
  pull secret was created successfully in the destination namespace.
- Tried to create universes using a provider without KUBECONFIG
  file. Worked as expected.


sb-yb (approved):
Thanks for catching this!

PR #12074, opened 2022-04-07
[docs] add -master_addresses to yb-admin commands

ankit-yb:
Added -master_addresses <master_addresses> in the admin commands.


polarweasel (commented):
LGTM, but please do this in /preview as well as /stable 👍

PR #12118, opened 2022-04-12
[doc] Doc changes needed for renaming platform based on the new nomenclature

lizayugabyte:
Part 1 of addressing https://yugabyte.atlassian.net/browse/PLAT-3424  (most files under yugabyte-platform folder and related images; top-level dir renamed to preview)


polarweasel:
I think you should merge this one it's good to go. We're close enough to the target date that we won't confuse too many people for too long. (And I assume part 2 is changing the product name _outside_ of the Yugabyte-platform directory?)

PR #12154, opened 2022-04-13
[docs] update some server-side redirects

polarweasel:
Adjusting some redirects...

The PR preview link should land you on the same page under /preview by way of two redirects. Verify with `curl -I` to see what the chain is...

@netlify /latest/yugabyte-cloud/cloud-secure-clusters/add-connections/


ddhodge (approved):
LGTM

PR #12208, opened 2022-04-20
[docs] release 2.12.4.0

polarweasel:
@netlify /preview/releases/release-notes/v2.12/#v2.12.4.0


polarweasel:
.4 was an internal release, so its release notes went into .5.

PR #12209, opened 2022-04-20
[docs] Add supported libraries as tile style in Drivers and ORMs docs landing page

samiahmedsiddiqui:
@netlify /preview/drivers-orms/

### DIFF
--- a/managed/ui/src/components/xcluster/ConfigureReplicationModal.tsx
+++ b/managed/ui/src/components/xcluster/ConfigureReplicationModal.tsx
@@ -99,7 +99,7 @@ export function ConfigureReplicationModal({ onHide, visible, currentUniverseUUID
   );
 
   const { data: tables, isLoading: isTablesLoading } = useQuery(
-    [currentUniverseUUID, 'tables'],
+    ['xcluster',currentUniverseUUID, 'tables'],
     () => fetchTablesInUniverse(currentUniverseUUID).then((res) => res.data)
   );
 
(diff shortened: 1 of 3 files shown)


polarweasel:
(Updated description. The Netlify preview entry point will change with the next commit to this branch)

docs/content/preview/drivers-orms/_index.md
```diff
@@ -29,20 +29,59 @@ Along with the smart drivers, YugabyteDB also supports upstream [PostgreSQL driv
...
+	<a href="csharp/" class="grid-child" title="C#">
+		<div>
+			<img src="/logos/csharp.svg" alt="C#" title="C#" />
+		</div>
+		C#
+	</a>
+	<a href="go/" class="grid-child" title="Go">
+		<div>
+			<img src="/logos/go.svg" alt="Go" title="Go" />
+		</div>
```
polarweasel, line 42: Will this work correctly here?
```suggestion
    <div class="icon">
      <i class="fa-brands fa-golang"></i>
    </div>
```

polarweasel (commented):
Looking good... If it turns out Fontawesome won't support full-colour icons, then the static images are OK. We just went through some trouble to make better use of it, though, and remove some outdated logo images, so I'd rather use a library for this if possible...

samiahmedsiddiqui:
@polarweasel Fontawesome doesn't support full-colour icons so that's why I have used static images. I know that it is a trouble to keep the icons up-to-date and Fontawesome also knows that, that's why they don't provide colour icons though.

polarweasel (approved):
I think this is a solid improvement over what's there now, so let's go with it.

PR #12277, opened 2022-04-26
WaitForServer.java error message

kerneljake:
#9133 changed the connectivity test from icmp to /dev/tcp, but the error message still refers to ping.  Removing the ping reference avoids the erroneous debugging step of enabling icmp when another protocol is actually required.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/subtasks/WaitForServer.java
+++ b/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/subtasks/WaitForServer.java
@@ -55,7 +55,7 @@ public class WaitForServer extends ServerSubTaskBase {
       closeClient(client);
     }
     if (!ret) {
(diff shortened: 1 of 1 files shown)


managed/src/main/java/com/yugabyte/yw/commissioner/tasks/subtasks/WaitForServer.java
```diff
@@ -55,7 +55,7 @@ public void run() {
       closeClient(client);
     }
     if (!ret) {
-      throw new RuntimeException(getName() + " did not respond to pings in the set time.");
+      throw new RuntimeException(getName() + " did not respond in the set time.");
```
iSignal, line 58: the ping here referred to a separate RPC called ping that exists on the db server. But agree that it can be confusing, accepted.

iSignal (approved):

PR #12530, opened 2022-05-17
[Docs] Update row-level geo-partitioning docs for transaction promotion

es1024:
With 5f339ee6 (2.13.3/2.14 onwards), users no longer run into the `Nonlocal tablet accessed in local transaction` error by default, and thus the `Running distributed transactions` section of the Row-Level Geo-Partitioning explore page is no longer relevant. This PR removes that section and adds documentation for gflags related to geo-partitioned transactions.


polarweasel (approved):
LGTM. Merge when you're ready, @es1024

PR #12535, opened 2022-05-18
[Docs] YSQL: Explicitly specify that the yb_enable_read_committed_iso…

pkj415:
…lation gflag is a tserver gflag


chrisckchang (commented):
these changes look good! reduces confusion for users as to whether this is a master or t-server flag.

PR #12739, opened 2022-06-01
[Docs] Update compression flags

tylarb:
- Move "compression_type" documentation to a location under
  "performance" because it's not networking related
- Small text / readability enhancements


ddhodge (approved):
Backported to stable and 2.8

Issue #12744, opened 2022-06-02
[CDCSDK] Enable CDC support for User Defined Types


vaibhav-yb:
This issue is created so as to track the progress of supporting the User Defined Types (UDT) with Change Data Capture (CDC).

vaibhav-yb:
Currently the tserver crashes with the following stacktrace if there's a column with any UDT:

```
*** Aborted at 1654317972 (unix time) try "date -d @1654317972" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x3e800004001) received by PID 16385 (TID 0x7fbc356e1700) from PID 16385; stack trace: ***
    @     0x7fbc4631cc20 (unknown)
    @     0x7fbc45f7c37f __GI_raise
    @     0x7fbc45f66db5 __GI_abort
    @     0x7fbc4d42a3bc  ExceptionalCondition (src/postgres/src/backend/utils/error/assert.c:54)
    @     0x7fbc4d4222da  SearchSysCache1 (src/postgres/src/backend/utils/cache/syscache.c:1603)
    @     0x7fbc4d2ba095  enum_out (src/postgres/src/backend/utils/adt/enum.c:158)
    @     0x7fbc4d363874  anyenum_out (src/postgres/src/backend/utils/adt/pseudotypes.c:163)
    @     0x7fbc4d43710e  FunctionCall1Coll (src/postgres/src/backend/utils/fmgr/fmgr.c:1135)
    @     0x7fbc4d438c9e  OutputFunctionCall (src/postgres/src/backend/utils/fmgr/fmgr.c:1767)
    @     0x7fbc4d4aa4a2  DecodeDatum (src/postgres/src/backend/ybgate/ybgate_api.c:788)
    @     0x7fbc4d954697  yb::docdb::set_decoded_string_value(unsigned long, char const*, yb::DatumMessagePB*, char const*) (src/yb/docdb/docdb_pgapi.cc:274)
    @     0x7fbc4d9556f8  yb::docdb::set_string_value(unsigned long, char const*, yb::DatumMessagePB*) (src/yb/docdb/docdb_pgapi.cc:568)
    @     0x7fbc4d950ac7  yb::docdb::SetValueFromQLBinaryHelper(yb::QLValuePB, int, yb::DatumMessagePB*) (src/yb/docdb/docdb_pgapi.cc:1245)
[...]

Issue #11830, opened 2022-03-22
[CDCSDK] Tserver crash while trying to stream data from colocated tables


vaibhav-yb:
Jira Link: [DB-3627](https://yugabyte.atlassian.net/browse/DB-3627)
**Steps to reproduce:**
1. Create database: `CREATE DATABASE testdb WITH colocated = true;`
2. Create tables: `CREATE TABLE t1 (id INT PRIMARY KEY, name TEXT); CREATE TABLE t2 (id INT PRIMARY KEY, name VARCHAR(255));`
3. Create DB stream ID.
4. Start Debezium connector on table `t1` only.
The connector was able to discover the tablet but it did not process anything further.
5. Insert a record in t1: `INSERT INTO t1 VALUES (1, 'Vaibhav');` This will not result in any record even though the streaming is being done for this record.
6. Insert a record in t2: `INSERT INTO t2 VALUES (2, 'vaibhav');` This immediately results in a tserver crash with the following exception coming up in `tserver.err`:

```
F20220322 12:49:18 ../../src/yb/docdb/primitive_value.h:185] Check failed: IsString()
    @        0x10d2bbef3  google::LogDestination::LogToSinks()
    @        0x10d2bb15c  google::LogMessage::SendToLog()
    @        0x10d2bba55  google::LogMessage::Flush()
    @        0x10d2c08af  google::LogMessageFatal::~LogMessageFatal()
    @        0x10d2bca69  google::LogMessageFatal::~LogMessageFatal()
    @        0x10ffeb6c8  yb::docdb::PrimitiveValue::GetString()
    @        0x1155a327f  yb::docdb::PrimitiveValue::ToQLValuePB()
    @        0x10dde5640  yb::cdc::AddColumnToMap()
    @        0x10dde5a8b  yb::cdc::AddPrimaryKey()
    @        0x10dde853a  yb::cdc::PopulateCDCSDKWriteRecord()
    @        0x10ddeee26  yb::cdc::GetChangesForCDCSDK()
    @        0x10dd49d1a  yb::cdc::CDCServiceImpl::GetChanges()
[...]

Issue #11829, opened 2022-03-22
[CDCSDK] CDCErrorException with CDC enabled on target table in xCluster replication


vaibhav-yb:
Jira Link: [DB-3011](https://yugabyte.atlassian.net/browse/DB-3011)
If we have xCLuster replication enabled on a table `test` from `source` universe to `target` universe and we insert an intent in the `source` universe, an exception is thrown by Debezium which is basically coming from yb-client itself:

```
org.yb.client.CDCErrorException: Server[ac622ee5d2ba4343895108ff4d1c73e9] CORRUPTION[code 2]: Error when decoding range components of a document key: Unexpected end of key when decoding document key
	at org.yb.client.TabletClient.dispatchCDCErrorOrReturnException(TabletClient.java:508)
	at org.yb.client.TabletClient.decode(TabletClient.java:439)
	at org.yb.client.TabletClient.decode(TabletClient.java:100)
	at org.jboss.netty.handler.codec.replay.ReplayingDecoder.callDecode(ReplayingDecoder.java:500)
	at org.jboss.netty.handler.codec.replay.ReplayingDecoder.messageReceived(ReplayingDecoder.java:435)
	at org.jboss.netty.channel.SimpleChannelUpstreamHandler.handleUpstream(SimpleChannelUpstreamHandler.java:70)
	at org.yb.client.TabletClient.handleUpstream(TabletClient.java:651)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
	at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendUpstream(DefaultChannelPipeline.java:791)
	at org.jboss.netty.handler.timeout.ReadTimeoutHandler.messageReceived(ReadTimeoutHandler.java:184)
	at org.jboss.netty.channel.SimpleChannelUpstreamHandler.handleUpstream(SimpleChannelUpstreamHandler.java:70)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:559)
	at org.yb.client.AsyncYBClient$TabletClientPipeline.sendUpstream(AsyncYBClient.java:2504)
	at org.jboss.netty.channel.Channels.fireMessageReceived(Channels.java:268)
[...]

Issue #11116, opened 2022-01-18
[CDCSDK] Status of changes related to enabling of CDC on Read Replicas


vaibhav-yb:
This issue will track the developments regarding enabling of Change Data Capture on Read Replica clusters.

**Current Status:**
Investigation under progress

mrajcevic01:
Have a customer that is looking to do CDC off of a read replica cluster in order to avoid putting any additional stress on their primary cluster. Have not run into this otherwise to date, but can see this becoming a common ask. We have seen this in the past with traditional monolithic architectures where standy-by clusters are used for backups, reporting, etc.

apatrida:
@mrajcevic01 in our case we would do both as we replicate data to in memory structures (REDIS) and big data stores in each data center (or differently per data center).  For example in primary / replica data centers that take traffic:

* PRIMARY AND/OR REPLICA DB -> stream processing -> memory structures
* PRIMARY AND/OR REPLICA DB -> search

And in addition in one data center that manages a data lake the additional

* REPLICA DB -> data lake

PR #12752, opened 2022-06-02
[docs] [CDCSDK] Updating CDC docs with missing limitations

vaibhav-yb:
This PR aims to add/improve the following sections in the docs about CDC:
* Limitations
  * No support for User Defined Types (UDT) - #12744 
  * Undefined behavior on downgrading version - #12800 
  * No support for CDC on target tables in xCluster replication - #11829
  * No support for Read Replicas - #11116
  * No support for Colocated tables - #11830
* GFlags
  * `enable_update_local_peer_min_index`
  * `cdc_intent_retention_ms`
* Section and notes about configuring Debezium connector when SSL is enabled on a YugabyteDB cluster.
* Description about `YBExtractNewRecordState`, the custom SMT we provide in order to extract the values from the JSON records.


docs/content/preview/reference/configuration/yb-tserver.md
```diff
@@ -880,6 +880,18 @@ Enable streaming of TRUNCATE record for a table on which CDC is active.
 
 Default: `false`
 
+##### --cdc_intent_retention_ms
+
+The time period after which the intents will be cleaned up if there is no client polling for the change records.
```
polarweasel, line 885:
```suggestion
The time period, in milliseconds, after which the intents will be cleaned up if there is no client polling for the change records.
```

docs/content/preview/explore/change-data-capture/_index.md
```diff
@@ -143,13 +138,27 @@ The snapshot feature uses the `cdc_snapshot_batch_size` GFlag. This flag's defau
...
 * DROP and TRUNCATE commands aren't supported. If a user tries to issue these commands on a table while a stream ID is there for the table, the server might crash, the behaviour is unstable. Issues for TRUNCATE [10010](https://github.com/yugabyte/yugabyte-db/issues/10010) and DROP [10069](https://github.com/yugabyte/yugabyte-db/issues/10069).
 * If a stream ID is created, and after that a new table is created, the existing stream ID is not able to stream data from the newly created table. The user needs to create a new stream ID. Issue [10921](https://github.com/yugabyte/yugabyte-db/issues/10921).
+* CDC is not supported on a target table for xCluster replication [11829](https://github.com/yugabyte/yugabyte-db/issues/11829).
+* The support for DDLs is not there yet.
 * A single stream can only be used to stream data from one namespace only.
 * Enabling CDC on tables created using previous versions of YugabyteDB is not supported, even after YugabyteDB is upgraded to version 2.13 or higher.
+* There should be a primary key on the table you want to stream the changes from.
```
polarweasel, line 156: Isn't this a prerequisite? You moved it from that section, and I assume things won't work right without it...

docs/content/preview/explore/change-data-capture/_index.md
```diff
@@ -143,13 +138,27 @@ The snapshot feature uses the `cdc_snapshot_batch_size` GFlag. This flag's defau
...
 * DROP and TRUNCATE commands aren't supported. If a user tries to issue these commands on a table while a stream ID is there for the table, the server might crash, the behaviour is unstable. Issues for TRUNCATE [10010](https://github.com/yugabyte/yugabyte-db/issues/10010) and DROP [10069](https://github.com/yugabyte/yugabyte-db/issues/10069).
 * If a stream ID is created, and after that a new table is created, the existing stream ID is not able to stream data from the newly created table. The user needs to create a new stream ID. Issue [10921](https://github.com/yugabyte/yugabyte-db/issues/10921).
+* CDC is not supported on a target table for xCluster replication [11829](https://github.com/yugabyte/yugabyte-db/issues/11829).
+* The support for DDLs is not there yet.
 * A single stream can only be used to stream data from one namespace only.
 * Enabling CDC on tables created using previous versions of YugabyteDB is not supported, even after YugabyteDB is upgraded to version 2.13 or higher.
```
polarweasel, line 155: Move this under the other bullet about version problems (line 149).

docs/content/preview/explore/change-data-capture/_index.md
```diff
@@ -143,13 +138,27 @@ The snapshot feature uses the `cdc_snapshot_batch_size` GFlag. This flag's defau
...
+
+In the current implementation, information related to the columns for the UDTs will not be there in the messages published on Kafka topic.
+
+{{< /note >}}
+
+* CDC behaviour is undefined on downgrading from a CDC supported version (2.13 and newer) to an unsupported version (2.12 and older) and upgrading it back. Issue [12800](https://github.com/yugabyte/yugabyte-db/issues/12800)
 * DROP and TRUNCATE commands aren't supported. If a user tries to issue these commands on a table while a stream ID is there for the table, the server might crash, the behaviour is unstable. Issues for TRUNCATE [10010](https://github.com/yugabyte/yugabyte-db/issues/10010) and DROP [10069](https://github.com/yugabyte/yugabyte-db/issues/10069).
 * If a stream ID is created, and after that a new table is created, the existing stream ID is not able to stream data from the newly created table. The user needs to create a new stream ID. Issue [10921](https://github.com/yugabyte/yugabyte-db/issues/10921).
+* CDC is not supported on a target table for xCluster replication [11829](https://github.com/yugabyte/yugabyte-db/issues/11829).
+* The support for DDLs is not there yet.
```
polarweasel, line 153: Or "very incomplete"?
```suggestion
* Support for DDL commands is incomplete.
```

docs/content/preview/integrations/cdc/debezium.md
```diff
@@ -123,6 +123,44 @@ curl -i -X POST -H "Accept:application/json" -H "Content-Type:application/json"
 }'
 ```
 
+{{< note title="Note" >}}
+
+In case you have a SSL enabled cluster, you need to provide the root certificate to the `database.sslrootcert` configuration property as well.
```
polarweasel, line 128:
```suggestion
If you have an SSL enabled cluster, you need to provide the path to the root certificate in the `database.sslrootcert` configuration property.
```

docs/content/preview/integrations/cdc/debezium.md
```diff
@@ -123,6 +123,44 @@ curl -i -X POST -H "Accept:application/json" -H "Content-Type:application/json"
...
+
+{{< /note >}}
+
+**In case of SSL enabled YugabyteDB cluster, follow these steps:**
+
+* Copy the certificate file to your docker container (assuming that the file exists on the root directory of your machine)
```
polarweasel, line 134: Docker is a product name, so we should capitalize it (unless they don't, but I think they do...)
```suggestion
* Copy the certificate file to your Docker container (assuming that the file exists on the root directory of your machine)
```

polarweasel (changes requested):
Very good start...

docs/content/preview/explore/change-data-capture/_index.md
vaibhav-yb: I was asked to move this out of prerequisite to limitations. Do you think I should also mention something similar under prerequisites as well?

docs/content/preview/integrations/cdc/debezium.md
vaibhav-yb: Yes, they do capitalize it. Thank you for pointing out!

docs/content/preview/explore/change-data-capture/_index.md
polarweasel: Yes, I would put it back in prereqs and reword to say something like "Currently, there should be a primary key... This limitation will be removed in a future version of YugabyteDB."

docs/content/preview/explore/change-data-capture/debezium-connector-yugabytedb.md
```diff
@@ -911,6 +911,26 @@ You can send this configuration with a `POST` command to a running Kafka Connect
...
+{{< note title="Note" >}}
+
+We use a custom record extractor in order to make the sinks understand the format we are sending the data in i.e. `YBExtractNewRecordState`. For example, if you are using a JDBC Sink Connector, you need to add two more properties to the sink configuration:
+
+| Property | Value |
+| :--- | :--- |
+| `transforms` | `unwrap` |
+| `transforms.unwrap.type` | `io.debezium.connector.yugabytedb.transforms.YBExtractNewRecordState` |
+
+More about `YBExtractNewRecordState` is mentioned in the next section.
```
polarweasel, line 923:
```suggestion
We use a custom record extractor (`YBExtractNewRecordState`) so that the sinks understand the format in which we are sending the data. For example, if you are using a JDBC sink connector, you need to add two more properties to the sink configuration:

| Property | Value |
| :------- | :---- |
| `transforms` | `unwrap` |
| `transforms.unwrap.type` | `io.debezium.connector.yugabytedb.transforms.YBExtractNewRecordState` |

See the following section for more details on `YBExtractNewRecordState`.
```

docs/content/preview/explore/change-data-capture/debezium-connector-yugabytedb.md
```diff
@@ -911,6 +911,26 @@ You can send this configuration with a `POST` command to a running Kafka Connect
...
+{{< /note >}}
+
+#### `YBExtractNewRecordState` SMT
+
+Unlike the Debezium Connector for PostgreSQL, we only send the `after` image of the "set of columns" that are modified. PostgreSQL sends the complete `after` image of the row which has changed. So by default if the column was not changed, it is not a part of the payload we send and the default value is set to `null`.
```
polarweasel, line 928:
```suggestion
The YugabyteDB Debezium connector sends an `after` image containing _only modified columns_. (The PostgreSQL Debezium connector's `after` image contains _all_ columns of a modified row). By default, if the column hasn't changed, it's not a part of the payload, and the default value is set to `null`.
```

docs/content/preview/explore/change-data-capture/debezium-connector-yugabytedb.md
```diff
@@ -911,6 +911,26 @@ You can send this configuration with a `POST` command to a running Kafka Connect
...
+
+Now, to differentiate between the case where a column was set `null` or was not modified, we change the value type to a struct. In this structure, an unchanged column will be `{'value': null}`, whereas a changed column that was set to `null` will be `{'value': null, 'set': true}`.
```
polarweasel, line 930:
```suggestion
To differentiate between the case where a column is set to `null` and the case in which it's not modified, we change the value type to a struct. In this structure, an unchanged column is `{'value': null}`, whereas the column changed to a null value is `{'value': null, 'set': true}`.
```

docs/content/preview/explore/change-data-capture/debezium-connector-yugabytedb.md
```diff
@@ -949,6 +969,16 @@ The APIs used to fetch the changes are set up to work with TLSv1.2 only. Make su
 
 {{< /note >}}
 
+{{< note title="Note" >}}
+
+If you have a YugabyteDB cluster with SSL enabled, all you need to do is obtain the root certificate and provide the path of the file to the `database.sslrootcert` configuration property. You can follow these links to get the certificates to your universe:
+
+* [Local deployments](../../secure/tls-encryption/_index)
+* [YB Anywhere](../../yugabyte-platform/security/enable-encryption-in-transit#connect-to-a-ysql-endpoint-with-tls)
+* [YB Managed](../../yugabyte-cloud/cloud-secure-clusters/cloud-authentication#download-your-cluster-certificate)
```
polarweasel, line 978: Links should always have a trailing slash, even if you're following the final path element with an anchor link. _index is the default file, so you just omit that. And here are the corrected relative paths :) 
```suggestion
* [Local deployments](../../../secure/tls-encryption/)
* [YB Anywhere](../../../yugabyte-platform/security/enable-encryption-in-transit/#connect-to-a-ysql-endpoint-with-tls)
* [YB Managed](../../../yugabyte-cloud/cloud-secure-clusters/cloud-authentication/#download-your-cluster-certificate)
```

docs/content/preview/explore/change-data-capture/debezium-connector-yugabytedb.md
```diff
@@ -949,6 +969,16 @@ The APIs used to fetch the changes are set up to work with TLSv1.2 only. Make su
 
 {{< /note >}}
 
+{{< note title="Note" >}}
+
+If you have a YugabyteDB cluster with SSL enabled, all you need to do is obtain the root certificate and provide the path of the file to the `database.sslrootcert` configuration property. You can follow these links to get the certificates to your universe:
```
polarweasel, line 974:
```suggestion
If you have a YugabyteDB cluster with SSL enabled, need to obtain the root certificate and provide the path of the file in the `database.sslrootcert` configuration property. You can follow these links to get the certificates for your universe:
```

docs/content/preview/integrations/cdc/debezium.md
```diff
@@ -123,6 +123,44 @@ curl -i -X POST -H "Accept:application/json" -H "Content-Type:application/json"
...
+    "database.password": "yugabyte",
+    "database.dbname" : "yugabyte",
+    "database.server.name": "dbserver1",
+    "table.include.list":"public.test",
+    "database.streamid":"d540f5e4890c4d3b812933cbfd703ed3",
+    "snapshot.mode":"never",
+    "database.sslrootcert":"/kafka/root.crt"
+  }
+}'
+```
```
polarweasel, line 162: Bulleted lists imply that ordering is not significant. This is a procedure, so we need to number the steps.
```suggestion
If you have an SSL enabled YugabyteDB cluster, do the following:

1. Copy the certificate file to your Docker container (assuming that the file exists on the root directory of your machine):

    ```sh
    docker cp ~/root.crt connect:/kafka/
    ```

1. Deploy the connector configuration:

    ```sh
    curl -i -X POST -H "Accept:application/json" -H "Content-Type:application/json" \
      localhost:8083/connectors/ \
      -d '{
      "name": "ybconnector",
      "config": {
        "connector.class": "io.debezium.connector.yugabytedb.YugabyteDBConnector",
        "database.hostname":"'$IP'",
        "database.port":"5433",
        "database.master.addresses": "'$IP':7100",
        "database.user": "yugabyte",
        "database.password": "yugabyte",
        "database.dbname" : "yugabyte",
        "database.server.name": "dbserver1",
        "table.include.list":"public.test",
        "database.streamid":"d540f5e4890c4d3b812933cbfd703ed3",
        "snapshot.mode":"never",
        "database.sslrootcert":"/kafka/root.crt"
      }
    }'
    ```
```

docs/content/preview/explore/change-data-capture/debezium-connector-yugabytedb.md
```diff
@@ -911,6 +911,26 @@ You can send this configuration with a `POST` command to a running Kafka Connect
...
+
+We have provided an SMT (Single Message Transformer for Kafka) `YBExtractNewRecordState` that interprets these values and sends the record in the format i.e removes the unmodified columns from the JSON message. In this way, the record which is transformed by 'YBExtractNewRecordState' is compatible with all the sink implementations. We took this approach so that the schema doesn't change with each new record and it can work with "Schema Registry". For example, if we keep changing the record to only include the value of columns then the schema of each record will be different (the total number unique schemas will be a result of making all possible combinations of columns) and thus we would require sending 'schema' with all the records. This will not work with "Schema Registry" where once a schema is registered, the record can only contain payload with schema 'version' and Kafka will make sure that it is interpreted correctly.
```
polarweasel, line 932: I'm not sure I completely follow the logic, so please make sure my rewording here makes sense!
```suggestion
A schema registry requires that, once a schema is registered, records must contain only payloads with that schema version. If you're using a schema registry, the YugabyteDB Debezium connector's approach can be problematic, as the schema may change with every message. For example, if we keep changing the record to only include the value of modified columns, the schema of each record will be different (the total number unique schemas will be a result of making all possible combinations of columns) and thus would require sending a schema with every record.

To avoid this problem when you're using a schema registry, use the `YBExtractNewRecordState` SMT (Single Message Transformer for Kafka), which interprets these values and sends the record in the correct format (by removing the unmodified columns from the JSON message). Records transformed by `YBExtractNewRecordState` are compatible with all sink implementations. This approach ensures that the schema doesn't change with each new record and it can work with a schema registry.
```

Issue #11968, opened 2022-04-04
[Docs] Update YSQL SELECT syntax to include explicit locking clauses


pkj415:
### Description

In https://www.postgresql.org/docs/current/sql-select.html, there is a clause at the end (copied below) which is used for explicit locking. YSQL supports this clause but doesn't have it in the docs.

`[ FOR { UPDATE | NO KEY UPDATE | SHARE | KEY SHARE } [ OF table_name [, ...] ] [ NOWAIT | SKIP LOCKED ] [...] ]`

The clause we need is `[ FOR { UPDATE | NO KEY UPDATE | SHARE | KEY SHARE } [ OF table_name [, ...] ] [ SKIP LOCKED ] [...] ]` since we don't have NOWAIT support yet.

We need this 2.13, 2.12, 2.8 and 2.6 branches.

PR #12805, opened 2022-06-07
[Docs] [#11968] Update YSQL SELECT syntax to include NOWAIT clause + …

pkj415:
…update stable release docs to match preview


polarweasel (approved):
LGTM. Thanks @pkj415 !

PR #12825, opened 2022-06-08
[Docs] remove reference to docker swarm which is practically not used anymore - from preview and stable

sumitdatabase:
Files changed both under stable and latest versions - just to remove the reference of docker swarm as an option to install .


polarweasel (commented):
Couple of things:

1. You need to actually delete the file, not just the links to it :)
1. Please search and make sure nothing links to that file.

Ping us in #docs channel if you need help with either!

PR #12830, opened 2022-06-09
[docs] fix broken shortcode end tags

polarweasel:
**Problem**: A recent Hugo update showed that we've been incorrectly closing a lot of shortcode tags. Tags should be of the form `{{% ... %}}`, but we had a bunch (all for includeMarkdown) with a slash in the closing, like this: `{{% ... /%}}`.

Older versions of Hugo just ignored the bad slash, but recent ones don't. This meant that over 1000 of our EBNF grammar and syntax diagrams displayed the markdown source, instead of syntax-highlighted EBNF or a rendered SVG diagram.

**Command** run to fix:

```bash
cd docs/content
find . -type f -exec perl -pi -e 's#(\{\{% .+ )(\/%\}\})#$1%\}\}#g' {} +
```

**Testing** is in two parts:

1. Verify that the PR preview correctly shows syntax-highlighted EBNF and syntax diagrams. Netlify is currently configured to use Hugo 0.96.0.
2. Update Hugo to 0.100.2, and re-verify.

@netlify /preview/api/ysql/the-sql-language/statements/ddl_create_database/#grammar

### DIFF
--- a/docs/content/preview/api/ysql/exprs/aggregate_functions/invocation-syntax-semantics.md
+++ b/docs/content/preview/api/ysql/exprs/aggregate_functions/invocation-syntax-semantics.md
@@ -35,10 +35,10 @@ The following six diagrams, [`select_start`](../../../syntax_resources/grammar_d
 
 <div class="tab-content">
   <div id="grammar" class="tab-pane fade show active" role="tabpanel" aria-labelledby="grammar-tab">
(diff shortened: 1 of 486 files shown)


polarweasel:
The right-nav problem seems to be due to indentation. Hugo has made a number of changes to shortcode indentation handling in the last few versions, so I guess that's not a huge surprise.

Seems to be working now. Please add eyeballs and confirm!

Issue #12861, opened 2022-06-13
[DocDB] Enable https web UI for master/tserver when node-to-node encryption in transit is enabled


iSignal:
Jira Link: [DB-2622](https://yugabyte.atlassian.net/browse/DB-2622)
### Description

When node-to-node encryption is enabled, a server cert is provided to yugabyte for RPC communication. We can also use this certificate to enable a https web UI for the master and tserver. Typically these interfaces do not show sensitive data but in some specific scenarios like LDAP configuration, the gflags can contain sensitive information. 

@bmatican @es1024

es1024:
There is the `webserver_certificate_file` gflag to enable https on the UI, but this must be set to point to the certificate manually.

iSignal:
Thanks @es1024 . Will file a Platform issue for orchestrating both together.

PR #12865, opened 2022-06-13
[Docs] Add docs on gflags for webserver SSL/authentication

es1024:
Added documentation for the `webserver_certificate_file`, `webserver_authentication_domain`, and `webserver_password_file` gflags that were present in all past releases but not documented.

Context: #12861

### DIFF
--- a/docs/content/preview/reference/configuration/yb-master.md
+++ b/docs/content/preview/reference/configuration/yb-master.md
@@ -160,22 +160,19 @@ Default: The `www` directory in the YugabyteDB home directory.
 
 ##### --webserver_certificate_file
 
-Location of the SSL certificate file (in .pem format) to use for the web server. If empty, SSL is
-not enabled for the web server.
+Location of the SSL certificate file (in .pem format) to use for the web server. If empty, SSL is not enabled for the web server.
 
 Default: `""`
 
 ##### --webserver_authentication_domain
 
-Domain used for .htpasswd authentication. This should be used in conjunction with
-[`--webserver_password_file`](#webserver-password-file).
+Domain used for .htpasswd authentication. This should be used in conjunction with [`--webserver_password_file`](#webserver-password-file).
 
 Default: `""`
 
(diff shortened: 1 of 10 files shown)


docs/content/preview/reference/configuration/yb-master.md
```diff
@@ -158,6 +158,27 @@ Monitoring web server home.
 
 Default: The `www` directory in the YugabyteDB home directory.
 
+##### --webserver_certificate_file
+
+Location of the SSL certificate file (in .pem format) to use for the web server. If empty, SSL is
+not enabled for the web server.
+
+Default: `""`
```
polarweasel, line 166: Probably clearer to say this, rather than some quotation marks. If you agree, do for all. :)
```suggestion
No default value.
```

docs/content/preview/reference/configuration/yb-master.md
```diff
@@ -158,6 +158,27 @@ Monitoring web server home.
 
 Default: The `www` directory in the YugabyteDB home directory.
 
+##### --webserver_certificate_file
+
+Location of the SSL certificate file (in .pem format) to use for the web server. If empty, SSL is
+not enabled for the web server.
```
polarweasel, line 164: Please don't hard-wrap lines. Either one sentence per line, or one paragraph.
```suggestion
Location of the SSL certificate file (in .pem format) to use for the web server. If empty or omitted, SSL is not enabled for the web server.
```

polarweasel (changes requested):
Left a couple of comments and suggestions that will apply to every change...

docs/content/preview/reference/configuration/yb-master.md
es1024: No default value sounds to me like it's unset (and therefore is required to be set), versus set to a placeholder value, which does not have to be overridden.

polarweasel (approved):

Issue #12866, opened 2022-06-13
[OSSexp] Add Yugabyted UI repo as part of the yugabyte-db code base


nchandrappa:
### Description

For development purposes, `Yugabyted UI` project was being maintained in a separate repo for the initial MVP. Now we are moving the repo as part of the core `yugabyte-db` as the Yugabyted UI code base is now in a stable state. Also, this is a required step for getting the yugabyte-ui binary as part of the release process.

PR #12868, opened 2022-06-13
[#12866] Adding yugabyted UI code base as part of the yugabyte-db repo

nchandrappa:
For development purposes, `Yugabyted UI` project was being maintained in a separate repo for the initial MVP. Now we are moving the repo as part of the core `yugabyte-db` as the Yugabyted UI code base is now in a stable state. Also, this is a required step for getting the yugabyte-ui binary as part of the release process.


svarnau:
This seems to be a generated file? ./apiserver/cmd/server/.docs/api/openapi.yaml
Is it meant to be checked in versus generated at build time?

PR #12907, opened 2022-06-15
[doc] PR for Remove a step that's no longer required

lizayugabyte:
Restart is no longer required, so this step is removed.


ddhodge (approved):
LGTM

PR #12927, opened 2022-06-15
[docs] add more google servers to CSP

polarweasel:
Make sure the only complaints are about netlify...

@netlify /preview/yugabyte-cloud/cloud-quickstart/qs-add/


polarweasel:
This still prevents some google stuff from happening, and I don't think we care.

PR #12944, opened 2022-06-17
Update .fossa-local.yml

sambit-yb:
Adding yugabyted-ui to fossa yaml file.


sambit-yb:
using the alternate one.

PR #12963, opened 2022-06-20
[docs][cloud] API Key management

ddhodge:
API Keys Admin page

@netlify /preview/yugabyte-cloud/cloud-admin/cloud-admin-apikeys/


jagdish-yb (commented):
1. User's privileges i.e. Role is applied to the key he generates. If after creating a key, user role downgraded to account-dev, then all his keys will also get downgraded.
2. A keys mapping to user role or deletion of keys on user deletion is applicable to current release only. It will be changed in upcoming release and keys will be independent of creator users. May be we can explicitly mention it.

jagdish-yb (approved):
LGTM

lizayugabyte (approved):
A couple minor comments.

PR #12981, opened 2022-06-22
[docs] adjust support shortcodes

polarweasel:
Change the support shortcodes so they render correctly when nested in other shortcodes (such as notes). This is, incidentally, already how the `slack-link` shortcode works.


ddhodge (approved):
LGTM

PR #13039, opened 2022-06-24
[Docs] Update instructions to build on AlmaLinux 8

def-:
for open source contributors


docs/netlify.toml
```diff
@@ -165,6 +165,13 @@
   to = "/stable/quick-start/explore/ysql/"
   force = true
 
+# Switch from CentOS to AlmaLinux, TODO: Update this in stable too once it hits stable
```
polarweasel, line 168: No need to adjust in stable, since the contribute section always points to preview. See line 125 :)
```suggestion
# Switch from CentOS to AlmaLinux
```

PR #13151, opened 2022-07-05
[Docs] Post deploy fixes

samiahmedsiddiqui:
Issues:

1. #13155 
2. #13157

### DIFF
--- a/docs/assets/scss/_admonition.scss
+++ b/docs/assets/scss/_admonition.scss
@@ -6,10 +6,6 @@
   }
 }
 
-.admonition a {
-  font-weight: 700;
-}
-
 main .td-content .admonition {
   margin: 20px 0 0 0;
   padding: 25px 32px 30px;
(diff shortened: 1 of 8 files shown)


bllewell:
I looked at the result [HERE](https://deploy-preview-13151--infallible-bardeen-164bc9.netlify.app/preview/api/ysql/syntax_resources/grammar_diagrams/). The issue seems to be fixed. It's an improvement to have the list of rules on the right hand nav panel rather than at the top of the single diagram page. The "copy" button by the side of the "grammar" form of each rule seems to be a solution to a problem that nobody has. I suppose that, very rarely, one might want to show a rule in an email, Slack message, forum posting, or the like. But in that case it's very easy to select with the cursor and copy in the ordinary way.

I don't s'pose that the "copy" button is harmful—but I'd still recommend removing it.

polarweasel:
> I don't s'pose that the "copy" button is harmful—but I'd still recommend removing it.

That's something we'll want to do in the RRDiagram generator, rather than here.

@bllewell if you're sure nobody's going to want to single-click copy the EBNF, I'll update the generator.

polarweasel (approved):
All looks good to me...

PR #13175, opened 2022-07-06
[Docs] PR for Add a note to make a bundle when there is an intermidiate cert for CA-signed certs

lizayugabyte:
Addresses https://yugabyte.atlassian.net/browse/PLAT-4480


ddhodge (approved):
LGTM

PR #13219, opened 2022-07-08
[Docs] Updated PITR docs for earlier versions

vkulichenko:
Most of the changes are a simple copy of the content from the current version all the way down to 2.6.

In addition:
- Limitations are updated for different versions to reflect the correct state of those versions.
- 2.6 page is updated to reflect the fact that it only supports YCQL, but not YSQL.

@netlify /preview/manage/backup-restore/point-in-time-recovery/


vkulichenko:
Most of the changes are a simple copy of the content from the current version all the way down to 2.6.

In addition:
- Limitations are updated for different versions to reflect the correct state of those versions.
- 2.6 page is updated to reflect the fact that it only supports YCQL, but not YSQL.

polarweasel (changes requested):
Basically: copy/paste from /preview so you have the current state of things.

polarweasel (approved):
LGTM. Thanks for hanging in there!

PR #13227, opened 2022-07-10
[docs][cloud] Troubleshooting connections

ddhodge:
Add note on Remaining connections slots are reserved error

@netlify /preview/yugabyte-cloud/cloud-troubleshoot/#remaining-connections-slots-are-reserved


ddorian (approved):
LGTM. But wait for a review from Liza too.

PR #13256, opened 2022-07-11
[docs] fix misbehaving syntax tabs

polarweasel:
@netlify /preview/api/ysql/exprs/window_functions/invocation-syntax-semantics/


bllewell:
With the caveat that you mentioned about per-syntax-inclusion independent toggling between "grannar" and "diagram", It works fine. Huge improvement!

PR #13313, opened 2022-07-14
[docs] read replicas GA as of 2.12

polarweasel:
Read replicas and follower reads are both GA from 2.12 onwards. Need to adjust in /v2.12, /stable, and /preview.


abhishekmishragithub:
LGTM 🚀

abhishekmishragithub (approved):

PR #13350, opened 2022-07-19
[docs] add quick-start back into stable

polarweasel:
Note that this does NOT change the left-nav link in stable or any other version: those will continue to point to the Managed quick-start in /preview.

Links to any version below stable should get redirected to the Managed quick-start in /preview. (Bit of testing required to ensure that links to /stable/quick-start/ don't get redirected.)

@netlify /stable/quick-start/


ddhodge (approved):
LGTM

PR #13442, opened 2022-07-26
[docs] [CoreDB] show max connections

### LINKED ISSUE #13420: [Docs] document max_connections
### Description

2.14.0.0 and 2.15.1.0 introduced a change in the `show max_connections` statement's output.

In the new approach, if you're running as a superuser, it reports the total number of connections (as defined by the `--ysql_max_connections` TServer flag, with a default of 300). But if you're a non-superuser, it reports the number of connections you can use — say, 300 minus the number of superuser-only reserved connections.

Right now we only really talk about this on the Tserver reference page and in Karthik's new `limiting-connections.md`. A parenthetical note on the latter page, and a sentence on the former, is probably enough.

ALSO... this paragraph in `cmd_show.md` needs a review. I wonder if it's still accurate?

> Although the values of a parameter can be set, displayed, and reset, the effect of these parameters are not yet supported in Yugabyte. The factory-settings or default behaviors will be used for the moment.

aishwarya24:
Resolves #13420 
Changes are in preview and stable.
@netlify /preview/develop/quality-of-service/limiting-connections/


docs/content/preview/api/ysql/the-sql-language/statements/cmd_show.md
```diff
@@ -44,7 +44,8 @@ Use the `SHOW` statement to display the value of a run-time parameter.
 
 ## Semantics
 
-- Although the values of a parameter can be set, displayed, and reset, the effect of these parameters are not yet supported in Yugabyte. The factory-settings or default behaviors will be used for the moment.
+The parameter values in YSQL can be set and typically take effect the same way as in PostgreSQL. However, since YugabyteDB uses a different storage engine ([DocDB](../../../../../architecture/layered-architecture/#docdb)), many configurations related to the storage layer will not have the same effect in YugabyteDB as in PostgreSQL. For example, configurations related to connection and authentication, query planning, error reporting and logging, run-time statistics, client connection defaults, and so on, should work as in PostgreSQL.
```
polarweasel, line 47:
```suggestion
The parameter values in YSQL can be set and typically take effect the same way as in PostgreSQL. However, because YugabyteDB uses a different storage engine ([DocDB](../../../../../architecture/layered-architecture/#docdb)), many configurations related to the storage layer will not have the same effect in YugabyteDB as in PostgreSQL. For example, configurations related to connection and authentication, query planning, error reporting and logging, run-time statistics, client connection defaults, and so on, should work as in PostgreSQL.

```

docs/content/preview/develop/quality-of-service/limiting-connections.md
```diff
@@ -18,61 +18,106 @@ Each connection to a YugabyteDB cluster uses CPU and memory, so it is important
...
-You can display the limits as shown below.
-```sql
-select datname, datconnlimit from pg_database where datname =’test_connection’ ;
-```
-```
-       datname     | datconnlimit
------------------+--------------
- test_connection |            1
-```
+- Set connection limit for database as follows:
```
polarweasel, line 87:
```suggestion
- Set the database connection limit:
```

docs/content/preview/reference/configuration/yb-tserver.md
```diff
@@ -533,7 +533,7 @@ Default: Uses the YSQL display format.
 
 Specifies the maximum number of concurrent YSQL connections.
 
-Default: `300`
+Default: `300` for superusers. Non-superuser roles see only the connections available for use, while superusers see all connections, including those reserved for non-superusers.
```
polarweasel, line 536: Also shouldn't this be "those reserved for superusers"?
```suggestion
Default: 300 for superusers. Non-superuser roles see only the connections available for use, while superusers see all connections, including those reserved for non-superusers.
```

polarweasel (approved):
Good to go once you make the little changes (and same changes in stable; I only marked in preview)

PR #13447, opened 2022-07-26
[Docs] Fixed Python command used in PITR examples

### LINKED ISSUE #13433: [Docs] PITR example flow doesn't work with Python 3
### Description

The [Explore PITR](https://docs.yugabyte.com/preview/explore/cluster-management/point-in-time-recovery-ysql/) page uses Python's `print` command to print out the current timestamp that is later used for restoration. In Python 3, this command was [converted to a function](https://stackoverflow.com/questions/826948/syntax-error-on-print-with-python-3), so the syntax provided in the documentation doesn't work.

Need to document both syntax flavors, or figure out a way that works in both Python 2 and 3.

vkulichenko:
Fixes #13433 

`print(...)` syntax (with parenthesis) works in both Python2 and Python 3 - updated the command with this format across all versions.

In addition, removed the YSQL page for 2.6, as only YCQL PITR is supported in that version.

@netlify /preview/explore/cluster-management/point-in-time-recovery-ysql/


aishwarya24 (approved):
LGTM!

PR #13481, opened 2022-07-29
[docs] remove very deprecated "start_pgsql_proxy"

polarweasel:
We deprecated the `--start_pgsql_proxy` flag in v2.2, but never replaced it everywhere in the docs with `--enable_ysql`. This fixes that in all YugabyteDB versions (well, all the versions on the main docs site: 2.4 and up).

@netlify /preview/reference/configuration/yb-tserver/#example


ionthegeek (approved):
Looks good! This covers all the instances of the flag I was able to turn up.

Issue #13288, opened 2022-07-13
[DocDB] Remove "using std::string" from .h files


hari90:
Jira Link: [DB-2925](https://yugabyte.atlassian.net/browse/DB-2925)
### Description

string -> std::string in header files (I think there might be a using std::string somewhere in a global header file if this compiles as is, which is wrong and should be fixed separately).

mbautin:
A script that helps with automatically fixing compilation errors: https://gist.githubusercontent.com/mbautin/15cf3d9817a9cbc65915bc1cbc0bbcb8/raw

PR #13491, opened 2022-07-30
Removing "using" from header files

mbautin:
The "using" directive should not appear in header files because it will affect all source files that include them. Removing the "using" directive from header files in our codebase involves adding a lot of `std::` and `using std::...` to many header and source files.


hari90:
Fixes #13288

hari90 (approved):

PR #13529, opened 2022-08-05
Update docker.md

jonrosner:
Why would we pull the preview-build but then run latest? This makes you download two containers.


ddorian:
Shouldn't you change it in the whole page? Example on line 109 on the same file

jonrosner:
True. I changed line 109 as well.

PR #13629, opened 2022-08-15
[Docs] Fix highlight in the right nav

samiahmedsiddiqui:


polarweasel (approved):
So much nicer now. Thank you!

PR #13666, opened 2022-08-17
[doc] PR for Document Advanced Master Metrics

lizayugabyte:
Addresses https://yugabyte.atlassian.net/browse/PLAT-4892 and https://yugabyte.atlassian.net/browse/PLAT-5109

Preview: https://deploy-preview-13666--infallible-bardeen-164bc9.netlify.app/preview/yugabyte-platform/troubleshoot/universe-issues/#master-server-advanced


lizayugabyte:
@anmalysh-yb I added new metrics to the doc. While doing that, I noticed that some of the existing metrics changed/were missing, so I added them based on what's in the UI on dev portal. 
When you see "Information pending" while reviewing the doc, could you, please, let me know what description we should provide for these metrics

anmalysh-yb:
It turns out that we can show latency in milliseconds or microseconds, based on the actual latency value.
So, in all the places, where we say that metric is in milliseconds - we should say that it's in milliseconds or microseconds.
Another option is to completely remove these mentions as we show actual unit on the Y axis.

Will try to add data, where information is pending, to the ticket itself.

anmalysh-yb (approved):
Added details for some more metrics in the ticket.

PR #13669, opened 2022-08-17
[docs] update CVE list

polarweasel:
Adding this CVE to the list.

Additionally, I've deleted the contents of this page in every version but preview. That plus the warning on older pages should be sufficient to get people to the right place.

@netlify /preview/secure/vulnerability-disclosure-policy/#security-tracker-cve-list


sambit-yb (approved):

sambit-yb (approved):
CVE ID updated .

PR #13716, opened 2022-08-22
[docs][cloud] Add note on staging clusters

ddhodge:
Add note on using staging clusters on Planning page

@netlify /preview/yugabyte-cloud/cloud-basics/create-clusters-overview/#yugabytedb-version

### DIFF
--- a/docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters-overview.md
+++ b/docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters-overview.md
@@ -19,9 +19,9 @@ The following best practices are recommended for production clusters.
 | Feature | Recommendation |
 | :--- | :--- |
 | [Provider and region](#provider-and-region) | Deploy your cluster in a virtual private cloud (VPC), with the same provider and in the same region as your application VPC. YugabyteDB Managed supports AWS and GCP.<br>Multi-region clusters must be deployed in VPCs. You need to create the VPCs before you deploy the cluster. Refer to [VPC network](../cloud-vpcs/). |
-| [Fault tolerance](#fault-tolerance) | Availability zone (AZ) level - minimum of three nodes across multiple AZs, with a replication factor of 3. |
+| [Fault tolerance](#fault-tolerance) | Region or Availability zone (AZ) level - minimum of three nodes across multiple regions or AZs, with a replication factor of 3. |
 | [Sizing](#sizing) | For most production applications, at least 3 nodes with 4 to 8 vCPUs per node.<br>Clusters support 10 simultaneous connections per vCPU. For example, a 3-node cluster with 4 vCPUs per node can support 10 x 3 x 4 = 120 connections.<br>When scaling your cluster, for best results increase node size up to 16 vCPUs before adding more nodes. For example, for a 3-node cluster with 4 vCPUs per node, scale up to 8 or 16 vCPUs before adding a fourth node. |
-| [YugabyteDB version](#yugabytedb-version) | Use the **Stable** release track. |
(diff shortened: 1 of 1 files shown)


docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters-overview.md
```diff
@@ -76,27 +76,39 @@ For a list of supported regions, refer to [Cloud provider regions](../../release
...
-- **Region level**. Yugabyte supports multi-region clusters with regional fault tolerance. Contact {{% support-cloud %}} for help configuring regional fault tolerance.
-
 Although you can't change the cluster fault tolerance after the cluster is created, you can scale horizontally as follows:
 
+- For Region level, you can add or remove nodes in increments of 1 per region; all regions have the same number of nodes.
 - For Availability zone level, you can add or remove nodes in increments of 3.
 - For Node level, you can add or remove nodes in increments of 1.
 
-For production clusters, Availability zone level is recommended.
+For production clusters, multi-region clusters providing region level fault tolerance  is recommended.
```
daniel-yb, line 91: Is this true? Aren't there latency considerations for customers depending on their requirements? I've seen that a lot of customers for YBA for example go with AZ level fault tolerance for this reason.

docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters-overview.md
```diff
@@ -113,6 +125,12 @@ If you need a feature from a preview release (that isn't yet available in a stab
 
 Yugabyte manages upgrades for you. After you choose a track, database upgrades continue to take releases from the track you chose. For multi-node clusters, Yugabyte performs a rolling upgrade without any downtime. You can manage when Yugabyte performs maintenance and upgrades by configuring the [maintenance window](../../cloud-clusters/cloud-maintenance/) for your cluster.
 
+#### Staging clusters
+
+Yugabyte tests every version in the stable branch for backwards compatibility. However, it's good practice to first test database updates against your pre-production environment (aka development, testing, staging, or canary environment) to ensure compatibility before upgrading your production clusters.
+
+Create a staging cluster (this can be smaller than your production cluster) and configure your pre-production environment to connect to it. Be sure to configure the [maintenance windows](../../cloud-clusters/cloud-maintenance/) for the staging and production cluster so that you can validate updates against your applications in your pre-production environment before updating your production cluster.
+
```
daniel-yb, line 133: I don't think this is really possible yet in a bullet-proof way. For example while a customer may set their maintenance window for staging cluster on Mondays and production cluster on Wednesdays, if we schedule the upgrade on the Tuesday, the prod cluster will still have maintenance scheduled first. We have a design internally to implement maintenance priority of clusters within the same account, but we probably won't get around to implementing it for a few more months.

docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters-overview.md
vitroth: Yeah I would replace this with something more like "Which fault tolerance level is appropriate depends heavily on your application architecture and design.  If you are unsure of the correct setting for your situation, please contact Yugabyte Support and we will work with you to understand your use case and provide guidance  on what settings to use."

docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters-overview.md
vitroth: Agreed, but we could mitigate that with an internal policy of scheduling maintenance windows based on a weekly cycle.

Issue #520, opened 2018-10-29
using jsonb in python


schlemihl:
i am starting to explore yugabyte as an alternative to cassandra.
i'd love to use the jsonb data type from python/cassandra_driver.
i can insert json data, but i can't query it yet...
`NotSupportedError: Unknown data type code 0x0080. Have to skip entire result set.`

certainly, standard cassandra doesn't know a jsonb data type, so does the python cassandra driver while parsing the result. i am looking into the python driver, and i wonder, if one could hack/misuse the user defined data type mechanism to add code to parse jsonb? i have not yet find it - i am so far no cassandra guy!

do you know how i could use/query jsonb from python?

kmuthukk:
hi @schlemihl 

The YB fork of the python driver for Cassandra already includes the JSONB support.

https://github.com/YugaByte/cassandra-python-driver

While this is not yet packages as a pip install, you can do a manual install via:

```
git clone https://github.com/YugaByte/cassandra-python-driver.git
cd cassandra-python-driver/
python setup.py install
```
and verify via:
```
python -c 'import cassandra; print cassandra.__version__'
```

Will try to send you a sample JSON example soon.

kmuthukk:
@schlemihl: Here's a simple example illustrating use of JSONB type in YCQL (YugaByte)
using a simple execute statement and also using a prepared statement.

```
#!/usr/local/bin/python

from cassandra.cluster import Cluster

def main():
  # Create the cluster connection.
  cluster = Cluster(['127.0.0.1'])
  session = cluster.connect()

  # Create the keyspace.
  session.execute('CREATE KEYSPACE IF NOT EXISTS ybdemo;')
  print "Created keyspace ybdemo"

  # Create the table.
  session.execute(
    """
     CREATE TABLE IF NOT EXISTS ybdemo.employee (id int PRIMARY KEY,
                                                 name varchar,
                                                 age int,
                                                 profile jsonb);
    """)
  print "Created table employee"

  # Insert a row.
  print "Simple insert..."
  session.execute(
    """
    INSERT INTO ybdemo.employee (id, name, age, profile)
    VALUES (100, 'John', 35, '{ "address": {"zip": 12345, "state": "CA"},
                                "hobbies": ["skating", "soccer"],
                                "language": "English"}');
    """)
  print "Inserted 1 row..";
  # Use a prepared stmt
  print "Use a prepared insert stmt to insert 5 more rows..."
  # Insert a row.
[...]

PR #13772, opened 2022-08-25
[docs] Add note about YCQL requiring Yugabyte-specific client drivers

def-:
Was recently noticed two times in QA, and there are existing bugs about this: https://github.com/yugabyte/yugabyte-db/issues/520
So there seems to be some confusion on this topic, best to clear that up in documentation.

### DIFF
--- a/docs/content/preview/reference/drivers/ycql-client-drivers.md
+++ b/docs/content/preview/reference/drivers/ycql-client-drivers.md
@@ -16,6 +16,8 @@ The following client drivers are supported for use with the [Yugabyte Cloud Quer
 
 For tutorials on building a sample application with the following client drivers, click the relevant link included below for each driver.
 
+Note that for YCQL you have to use the Yugabyte-specific client drivers, otherwise you can run into `NotSupportedError: Unknown data type code 0x0080. Have to skip entire result set.` errors.
+
 ## C/C++
 
 ### Yugabyte C/C++ Driver for YCQL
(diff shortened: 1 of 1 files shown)


docs/content/preview/reference/drivers/ycql-client-drivers.md
```diff
@@ -16,6 +16,8 @@ The following client drivers are supported for use with the [Yugabyte Cloud Quer
 
 For tutorials on building a sample application with the following client drivers, click the relevant link included below for each driver.
 
+Note that for YCQL you have to use the Yugabyte-specific client drivers, otherwise you can run into `NotSupportedError: Unknown data type code 0x0080. Have to skip entire result set.` errors.
```
Arjun-yb, line 19: this issue is not with most of the datatypes, I guess it is only with JSONB type, it is good to specify it clearly.

vijayjoshi16: Yes, for the JSONB datatype the original `cassandra` driver does not support select query. But it is supported in `yb-cassandra-driver`. @def- and I were discussing about it a few days ago and we had to remove the existing installation of the original `cassandra` driver and install the `yb-cassandra-driver` again to get rid of this issue.

def-: As discussed with Oleg, there are also other differences, can be slower or work incorrectly. Jsonb might just be the most obvious failure. So I think we should keep the recommendation general.

Arjun-yb (approved):

docs/content/preview/reference/drivers/ycql-client-drivers.md
polarweasel, line 19: Let's make this more generic and more prominent. How's this?
```suggestion
{{< note title="Use YugabyteDB client drivers" >}}
You should always use the YugabyteDB YCQL client drivers. Using generic Cassandra drivers can lead to errors and performance issues.
{{< /note >}}
```

polarweasel (approved):
Good as-is, better with my suggestion :)

Issue #13788, opened 2022-08-28
[Docs] leader lease duration flag isn't documented - it should be 


tylarb:
### Description

Under "Raft flags" we're missing the really important leader lease duration

https://docs.yugabyte.com/preview/reference/configuration/yb-master/#raft-flags

from `yb-master --help` 

```
    -leader_lease_duration_ms (Leader lease duration. A leader keeps
      establishing a new lease or extending the existing one with every
      UpdateConsensus. A new server is not allowed to serve as a leader (i.e.
      serve up-to-date read requests or acknowledge write requests) until a
      lease of this duration has definitely expired on the old leader's side.)
      type: int32 default: 2000
```

A note should be added to these docs, like: 

"""
leader lease duration should be longer than the heartbeat interval, and less than the multiple of `--leader_failure_max_missed_heartbeat_periods` multiplied by `--raft_heartbeat_interval_ms`
"""

(reference thread https://app.slack.com/client/T0YK2G6N9/C9T4FKYSV/thread/C9T4FKYSV-1614193263.043100 )

tylarb:
This flag should also be documented on the tserver page since it's needed there as well - https://docs.yugabyte.com/preview/reference/configuration/yb-tserver/#raft-heartbeat-interval-ms

PR #13800, opened 2022-08-29
[docs] Add documentation for leader lease configuration

ddhodge:
Add --leader_lease_duration_ms flag
See #13788

@netlify /preview/reference/configuration/yb-master/#leader-lease-duration-ms


tylarb:
@ddhodge   content looks :+1:  

Needs to additionally be in yb-tserver documentation though - 

We maybe should add a note to each of these leader election flags saying "the following flags need to be identical between master and tserver"

tylarb:
@ddhodge   can you sync with someone from DST  team on what the list of "following flags should be the same" should be? All of the follower settings related to this thread... but it's worth a review if there's a couple more

ddhodge:
Will be adding to TServer doc and backporting to previous versions once wording is finalized.

docs/content/preview/reference/configuration/yb-master.md
```diff
@@ -287,6 +287,14 @@ For read replica clusters, set the value to `10` in all `yb-tserver` and `yb-mas
 
 Default: `6`
 
+##### --leader_lease_duration_ms
+
+The leader lease duration, in milliseconds. A leader keeps establishing a new lease or extending the existing one with every consensus update. A new server is not allowed to serve as a leader (that is, serve up-to-date read requests or acknowledge write requests) until a lease of this duration has definitely expired on the old leader's side.
```
robertsami, line 292: > A new server is not allowed to serve as a leader (that is, serve up-to-date read requests or acknowledge write requests) until a lease of this duration has definitely expired on the old leader's side.

this isn't completely accurate, as we have a mechanism whereby the leader can relinquish it's lease. so i might write this:

> A new server is not allowed to serve as a leader (that is, serve up-to-date read requests or acknowledge write requests) until a lease of this duration has definitely expired on the old leader's side, or the old leader has explicitly acknowledged the new leader's lease.

though this also brushes over some details. so maybe a third option is:

> This lease allows the leader to safely serve reads for the duration of its lease, even during a network partition. See: https://docs.yugabyte.com/preview/architecture/transactions/single-row-transactions/#leader-leases-reading-the-latest-data-in-case-of-a-network-partition

robertsami (changes requested):
why are we adding this flag in the yb-master docs and not yb-tserver?

tylarb:
@robertsami  all of the other flags related here are actually in the master and tserver documentation - I believed that they should be the same in both places? Is that correct?

robertsami:
> @robertsami all of the other flags related here are actually in the master and tserver documentation - I believed that they should be the same in both places? Is that correct?

leader_lease_duration_ms should be in both places with the same description

robertsami:
> [](#commits-pushed-87af0f9)
> [ @ddhodge ](/ddhodge)
> ` [add to Tserver + stable 2.12 2.8](/yugabyte/yugabyte-db/pull/13800/commits/87af0f9c48468e13f883ad3190673b8adff5ba49) `
> ` [87af0f9](/yugabyte/yugabyte-db/pull/13800/commits/87af0f9c48468e13f883ad3190673b8adff5ba49) `
> [ @ddhodge ](/ddhodge)
> ` [Merge branch 'master' into docs_leader_lease_duration_flag](/yugabyte/yugabyte-db/pull/13800/commits/4818b602c540210694f3de516dbb559073cbbc10) `
> Verified
> ` [4818b60](/yugabyte/yugabyte-db/pull/13800/commits/4818b602c540210694f3de516dbb559073cbbc10) `
> [ @ddhodge ](/ddhodge)
> ` [format](/yugabyte/yugabyte-db/pull/13800/commits/7bb9c03a3d11d3866b93982bf80e7d7977f4ce9d) `
> ` [7bb9c03](/yugabyte/yugabyte-db/pull/13800/commits/7bb9c03a3d11d3866b93982bf80e7d7977f4ce9d) `
> [@robertsami](/robertsami)
> Member
> ### ** [robertsami](/robertsami)**

@tylarb are you asking if `leader_lease_duration_ms` must have the same value in both places? that i'm not sure of. i dont see why that would need to be the case, but @amitanandaiyer may be able to comment on that

amitanandaiyer:
I don't think the raft/leader_lease values need to be the same across tservers and masters. They can be independently set.

However, I don't see a use case where they need to be different either.

tylarb:
I did a quick review, and the changes LGTM at this point

PR #13838, opened 2022-09-01
[docs][cloud] Use include for quick start

ddhodge:
Use an include file for YB Managed quick start

@netlify /preview/yugabyte-cloud/cloud-quickstart/


polarweasel (approved):
LGTM! Let's get this merged, then we can improve the approach once Docsy 0.5.0 is released.

PR #13871, opened 2022-09-05
[Docs] Disable Homepage redirect

samiahmedsiddiqui:


polarweasel (changes requested):
To do:
1. Use more appropriate icons for the three Products boxes
2. YugabyteDB Core should be just YugabyteDB
3. Get started button top right should go to Managed signup page

polarweasel (approved):
Looks good to me...

PR #13960, opened 2022-09-09
Update tablet splitting limitations.

SrivastavaAnubhav:
`tablet splitting should be disabled during an index backfill` -> This is now fixed and in 2.14.0.
Restoring to arbitrary times for PITR is fixed in 2.14.1.


aishwarya24 (approved):
LGTM

PR #13995, opened 2022-09-13
[docs] [reference] removed Read Committed from v2.12

aishwarya24:
@netlify /preview/architecture/transactions/


polarweasel:
@pkj415 removing the architecture doc still leaves the release notes, and mentions in 15 other files. Is that OK?

pkj415:
The reason I was thinking of removing the 2.12 docs page is that -- read committed has a few things missing in 2.12 and the docs are slightly imprecise now as we have learnt more things over time (they even don't reflect 2.12 release's read committed exactly).

But your point made me rethink -- and the below screenshot on the current 2.12 page anyway has a warning with redirect to the 2.15 docs, so I think we can skip removing it.
Screen Shot 2022-09-19 at 11 09 35 AM

aishwarya24:
Closing the PR since as per comments.

PR #14021, opened 2022-09-15
[docs] Add class in table to remove alter colors

samiahmedsiddiqui:


polarweasel (approved):
This is better than it was. Let's go with it for now, and then figure out a better overall approach to presenting multi-node steps.

PR #14150, opened 2022-09-22
[Docs] PR for Minor edits YugabyteDB Upgrades doc

lizayugabyte:
Formatting and other minor edits


lizayugabyte:
@charleswang234 I added a section called "Perform rolling upgrades". However, most of the google doc which I used as a base seems to describe internal implementation of what's going on behind the scenes during a rolling upgrade. I'm not sure the customers are expected to do any of this, and it's not clear what they are supposed to do to perform rolling upgrades of YugabyteDB.

charleswang234:
All these steps are what happens when you select the rolling upgrades in the universe page on yb-anywhere/platform: 
image

This may not be the place to put all this information, since these are the implicit steps of what happens when we perform a rolling upgrade on yb-anywhere. It seems like the documentation here is about manually performing upgrades by sshing into nodes and manually installing the newer software, etc.

Let me know if you still have any questions

lizayugabyte:
@charleswang234 Yes, I realize that, and that's why I'm so confused about https://yugabyte.atlassian.net/browse/PLAT-5471 "Currently we don’t spell out the exact sequence in which rolling upgrades should be performed when using OSS... My proposal is to improve the OSS documentation". The only source document this ticket points to is the google doc YB Anywhere Rolling Software Upgrade Steps which describes the inner workings of YBA and nothing at all about what the YBDB customer needs to do to perform rolling upgrades. Do you have any information on that? If no, do you know whom I should contact?

ddhodge (approved):
LGTM

PR #14154, opened 2022-09-23
[Docs][CDCSDK] Fixing a typo error in a configuration property

vaibhav-yb:
The PR contains the diff to fix an error in a configuration property for the Debezium Connector for YugabyteDB where the property was mentioned as `column.truncate.to_length_.chars` which should be `column.truncate.to._length_.chars` - missing period after the word "to".


ddhodge (approved):
LGTM

PR #14223, opened 2022-09-28
Prevent scroll top in programming-language tab

samiahmedsiddiqui:


polarweasel (commented):
Works correctly now! Thank you!

PR #14271, opened 2022-09-30
[docs] typo fix for tablespaces

### LINKED ISSUE #14033: [Docs] Tablespaces
### Description

In Section Leader preference on page: https://docs.yugabyte.com/preview/explore/ysql-language-features/going-beyond-sql/tablespaces/,
statement
```
If all the nodes in us-east-1a are unavailable, they we will fallback to the next preferred zone us-east2 which only has a 12ms latency.
```
should be changed to 
```
If all the nodes in us-east-1b are unavailable, they we will fallback to the next preferred zone us-east-2a which only has a 12ms latency.
```

aishwarya24:
resolves #14033


ddhodge (approved):
LGTM

PR #14425, opened 2022-10-12
[docs] use docsy's new readfile shortcode

polarweasel:
This is a temporary fix until Docsy 0.5.0 lands.

@netlify /preview/tabpane-include/


polarweasel:
Hey @ddhodge I've added Docsy 0.5.0's `readfile` shortcode here. Want to play around and see if it's fixed the inclusion problem with tabpanes?

polarweasel:
As of fc65836, `tabpane` inclusions render now, but the behaviour of the tabs and their content is weird. One pane's tabs will switch, but the other pane's tabs won't. I think it's the first one you click. But then clicking the second pane's tabs will switch the first pane's _content_, but not its _tab headers_.

polarweasel:
Fixed in Docsy 0.5.1, which we're now using on the live site. No need for this PR.

Issue #14484, opened 2022-10-15
[DocDB] yb_build.sh fails on Ubuntu 20.04


jteplitz:
Jira Link: [DB-3878](https://yugabyte.atlassian.net/browse/DB-3878)
### Description

**Overview**

When following the steps outlined [here](https://docs.yugabyte.com/preview/contribute/core-database/build-from-src-ubuntu/) to build from source on Ubuntu, the build script downloads the dependencies for almalinux and then fails.

**Steps to reproduce**

These steps are taken directly from the docs:
1. Install prerequisites
```
sudo apt-get update
sudo apt-get install uuid-dev libbz2-dev libreadline-dev maven ninja-build \
                     cmake curl rsync python3-pip python3-venv zip autoconf libtool \
                     pkg-config libssl1.0-dev libicu-dev bison flex \
                     libncurses5-dev
```

Clone the repo and run `./yb_build.sh release`

**Results**

Initially the build scripts failed to find my C compiler:
```-- Check for working C compiler: /opt/yugabyte-db/build-support/compiler-wrappers/cc -- broken
CMake Error at /usr/share/cmake-3.16/Modules/CMakeTestCCompiler.cmake:60 (message):
  The C compiler

    "/opt/yugabyte-db/build-support/compiler-wrappers/cc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /opt/yugabyte-db/build/release-clang12-linuxbrew-dynamic-ninja/CMakeFiles/CMakeTmp

    Run Build Command(s):/usr/bin/ninja cmTC_1cab4 && [1/2] Building C object CMakeFiles/cmTC_1cab4.dir/testCCompiler.c.o
    FAILED: CMakeFiles/cmTC_1cab4.dir/testCCompiler.c.o
    /opt/yugabyte-db/build-support/compiler-wrappers/cc    -o CMakeFiles/cmTC_1cab4.dir/testCCompiler.c.o   -c testCCompiler.c
    /opt/yugabyte-db/build-support/compiler-wrappers/../common-build-env.sh: line 1102: cc_executable: unbound variable
    ninja: build stopped: subcommand failed.
```
Despite clang and gcc being in `/usr/bin`
```
root@208d3277a61c:/opt/yugabyte-db# which clang
/usr/bin/clang
root@208d3277a61c:/opt/yugabyte-db# which gcc
/usr/bin/gcc
```
After explicitly pointing yb_build at clang12, I get this error with boost:
```
[...]

PR #14602, opened 2022-10-21
[docs] [#14484] Fix Ubuntu build instructions

mbautin:
Remove some unneeded packages and do not mention an outdated version of OpenSSL. Add --no-linuxbrew until we make sure we automatically disable the use of Linuxbrew on Ubuntu.


svarnau (approved):
LGTM

PR #14616, opened 2022-10-24
[docs] correct numbers in table fragment section

aishwarya24:


docs/content/preview/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -198,7 +198,7 @@ SELECT COUNT(*) FROM sample_table WHERE yb_hash_code(x,y) >= 4600 and yb_hash_co
 (1 row)
 ```
 
-Because we use what can be assumed to be a uniformly distributed hash function to partition our rows, we can assume that this is a count of approximately 1/128 of all the rows. Therefore, multiplying this count, 78 by 128 gives us a good estimate of the total number of rows (9984 in this case) in the table without querying and iterating over all tablets.
+Because we use what can be assumed to be a uniformly distributed hash function to partition our rows, we can assume that this is a count of approximately 1/128 of all the rows. Therefore, multiplying this count, 74 by 128 gives us a good estimate of the total number of rows (9472 in this case) in the table without querying and iterating over all tablets.
```
tanujnay112, line 201: Nit: It should be 
`count, 74, by 128 gives us...`
I forgot that extra comma there originally

docs/content/v2.8/api/ysql/exprs/func_yb_hash_code.md
```diff
@@ -192,7 +192,7 @@ SELECT COUNT(*) FROM sample_table WHERE yb_hash_code(x,y) >= 4600 and yb_hash_co
 (1 row)
 ```
 
-Since, we use what can be assumed to be a uniformly distributed hash function to partition our rows, we can assume that this is a count of approximately 1/128 of all the rows. Therefore, multiplying this count, 78 by 128 gives us a good estimate of the total number of rows (9984 in this case) in the table without querying and iterating over all tablets.
+Because we use what can be assumed to be a uniformly distributed hash function to partition our rows, we can assume that this is a count of approximately 1/128 of all the rows. Therefore, multiplying this count, 74 by 128 gives us a good estimate of the total number of rows (9472 in this case) in the table without querying and iterating over all tablets.
```
tanujnay112, line 195: Ditto

tanujnay112 (changes requested):

PR #14650, opened 2022-10-25
Replace header guards with #pragma once

mbautin:


hari90 (approved):
First set of files LGTM. This is so big that the entire page does not even load for me.

hari90:
Can you share the script used to do this?

mbautin:
Script: https://gist.githubusercontent.com/mbautin/dd5e27d8a49c36f0ef85147e9e6a3325/raw

(had to apply changes to gutil one by one).

PR #14697, opened 2022-10-27
[docs] [migrate] Add a troubleshoot page 

### LINKED ISSUE #439: Expose raw file sizes in UI.
Today, we expose on the http://<tablet-server>:9000/metrics the rocksdb_total_sst_file_size metric for each tablet. This is the compressed size of the on-disk files.

This is then rolled up for each TS and also exposed via yb-master's http://yb-master:7000/tablet-servers under the "Total SST File Sizes" column.

It would be very useful to show the raw (uncompressed) size of the data as an additional column as this will be good indicator of the benefits of prefix compression + on-disk compression that the system does under the covers.

aishwarya24:
Build preview : https://deploy-preview-14697--infallible-bardeen-164bc9.netlify.app/preview/migrate/troubleshoot/
- resolves https://github.com/yugabyte/yb-voyager/issues/439
This page includes the known issues and possible solutions. Also, unsupported features is moved here.


aishwarya24:
Closing this one . We have a troubleshoot page with a different format.

PR #14944, opened 2022-11-10
 [ONLY 2.17.0] Fix flags lint rule name

hari90:
Lint rule gflag_register_flag_validator has been renamed to flags


bmatican (approved):
Please prefix title with [ONLY 2.X] so it's somewhat consistent in git log

Issue #11475, opened 2022-02-14
[docs] Docs enhancement proposal - Command blocks with editable variable names. 


yogendra:
https://docs.yugabyte.com/latest/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes/

## Issue description
Command blocks with editable variable names. Most of the time we have to copy the commands, paste in a text file and change it. In some cases, a value/text (Example: `<namespace>`) has to be typed more than once in a command, and also in multiple commands for same activity (Example: create sa and role bindings). Some of these commands do not show any errors right away, but cause a later step to fail. This can be eliminated with UX. 

## Suggested improvement

1. Please look at https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-zonal-cluster#gcloud for a very good example.

2. You could also consider using variables instead of placeholders.

streddy-yb:
cc @baba230896 @bhavin192

PR #14951, opened 2022-11-10
[docs][PLAT-5834] Use env vars in K8s service account creation steps

bhavin192:
Summary:
The <namespace> placeholder and the <SA_NAMESPACE> from the RBAC YAML were creating confusion on which one to replace while running the commands.

With this change we set an environment variable and use it across all the commands, so that users can just set one variable and easily execute all the commands.

Fixes https://github.com/yugabyte/yugabyte-db/issues/11475

Test Plan:
- Tried all the commands from this page with the variable value set and without it.

### DIFF
--- a/docs/content/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes.md
+++ b/docs/content/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes.md
@@ -79,15 +79,19 @@ Before you install YugabyteDB on a Kubernetes cluster, perform the following:
 
 The secret of a service account can be used to generate a `kubeconfig` file. This account should not be deleted once it is in use by YugabyteDB Anywhere. 
 
-Set the `YBA_NAMESPACE` environment variable to the namespace where you have installed YugabyteDB Anywhere. This variable will used by the rest of the commands from this document.
+Set the `YBA_NAMESPACE` environment variable to the namespace where your YugabyteDB Anywhere is installed, as follows:
 
 ```sh
 export YBA_NAMESPACE="yb-platform"
 ```
 
+Note that the `YBA_NAMESPACE` variable is used in the commands throughout this document.
+
 Run the following `kubectl` command to apply the YAML file:
 
 ```sh
+export YBA_NAMESPACE="yb-platform"
+
 kubectl apply -f https://raw.githubusercontent.com/yugabyte/charts/master/rbac/yugabyte-platform-universe-management-sa.yaml -n ${YBA_NAMESPACE}
 ```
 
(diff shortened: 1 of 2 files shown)


docs/content/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes.md
```diff
@@ -77,12 +77,18 @@ Before you install YugabyteDB on a Kubernetes cluster, perform the following:
 
 ### Service account
 
-The secret of a service account can be used to generate a `kubeconfig` file. This account should not be deleted once it is in use by YugabyteDB Anywhere. *namespace* in the service account creation command can be replaced with the desired namespace in which to install YugabyteDB.
+The secret of a service account can be used to generate a `kubeconfig` file. This account should not be deleted once it is in use by YugabyteDB Anywhere. 
+
+Set the `YBA_NAMESPACE` environment variable to the namespace where you have installed YugabyteDB Anywhere. This variable will used by the rest of the commands from this document.
```
lizayugabyte, line 82: Suggestion:

Set the `YBA_NAMESPACE` environment variable to the namespace where your YugabyteDB Anywhere is installed, as follows:

```sh
export YBA_NAMESPACE="yb-platform"
```

Note that the `YBA_NAMESPACE` variable is used in the commands thoughout this document.

lizayugabyte (approved):
Looks good. One comment.

docs/content/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes.md
```diff
@@ -77,12 +77,18 @@ Before you install YugabyteDB on a Kubernetes cluster, perform the following:
...
+
+```sh
+export YBA_NAMESPACE="yb-platform"
```
iSignal, line 85: though repetitive, it would be better to include this export in each shell command so that folks can copy-paste directly.

iSignal (approved):

docs/content/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes.md
bhavin192: Yep, but the only problem with that is the users have to modify the environment variable everytime for each command if they are using a different value than what we have recommended in this document.

bhavin192: So, I checked what happens when we don't set the value and run the commands. None of them succeeds but give weird errors, which will make it hard to debug. 
I'm adding default value to all the variable expansions. 
```
${YBA_NAMESPACE:-yb-platform}
```
That way if they provide anything, it will default to yb-platform. 
What do you think about this idea?

docs/content/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes.md
bhavin192: Thanks! This is much better.

docs/content/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes.md
iSignal: Makes sense.

PR #14960, opened 2022-11-11
[Docs] clean up list elements and other formatting

samiahmedsiddiqui:
- List elements added incorrectly using HTML tags and due to that it is not showing as per new theme.
- Add rightNav on _indexpage and removed from content

### DIFF
--- a/docs/content/preview/deploy/kubernetes/single-zone/oss/helm-chart.md
+++ b/docs/content/preview/deploy/kubernetes/single-zone/oss/helm-chart.md
@@ -259,7 +259,7 @@ kubectl get services --namespace yb-demo
 Expect output similar to the following:
 
 ```output
-NAME                 TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                                        AGE
+NAME                 TYPE           CLUSTER-IP      EXTERNAL-IP        PORT(S)                                        AGE
 ...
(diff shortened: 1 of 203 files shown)


docs/content/preview/deploy/kubernetes/single-zone/oss/helm-chart.md
```diff
@@ -259,7 +259,7 @@ kubectl get services --namespace yb-demo
 Expect output similar to the following:
 
 ```output
-NAME                 TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                                        AGE
+NAME                 TYPE           CLUSTER-IP      EXTERNAL-IP        PORT(S)                                        AGE
```
polarweasel, line 262: Where did this one come from? And does it need to be replicated in all the other helm-chart files in this and the other versions?

polarweasel (approved):
LGTM! One question outstanding, then we should be good to merge.

docs/content/preview/deploy/kubernetes/single-zone/oss/helm-chart.md
samiahmedsiddiqui: @polarweasel On couple of instances, it is already aligned. Let me check all the helm-chart files.

samiahmedsiddiqui: @polarweasel Applied this change in couple of other instances (especially in helm-chart files) in all versions.

polarweasel:
Thanks for the excellent cleanup with this PR @samiahmedsiddiqui!

PR #15030, opened 2022-11-16
[docs] [reference ] Removed an incorrect yb master evict statement

### LINKED ISSUE #15003: [Docs] yb-master documentation incorrectly states that masters evict failed followers
### Description

In the yb-msater documentation, it describes the --follower_unavailable_consdered_failed_sec flag:
https://docs.yugabyte.com/preview/reference/configuration/yb-master/#follower-unavailable-considered-failed-sec

In this section, it mentions `The follower is then evicted from the configuration and the data is re-replicated elsewhere.` but this is not true for masters.

1. Master followers will not be evicted from the RAFT group because `--evict_failed_followers` is set to `false` by default for masters.
2. It is not possible to automate master recovery without some outside watcher (e.g. YBA) because individual masters can't know whether the problem is with them or with other master(s) (e.g. if there is a network partition). The masters themselves will never attempt to re-replicate their data elsewhere.

We should probably just remove `The follower is then evicted from the configuration and the data is re-replicated elsewhere.` from this section of the docs.

aishwarya24:
resolves #15003 
@netlify /preview/reference/configuration/yb-master/#raft-flags


aishwarya24:
@ionthegeek : I have removed the incorrect statement. I wonder if we need to document the flag you mentioned in the issue : `--evict_failed_followers`

ionthegeek (approved):
Changes look good.

It would be good to have the --evict_failed_followers flag documented but we probably don't want customers changing it. It is false by default for masters because we can't auto-recover a master but it is true by default for tservers because the failed follower will be evicted and the data will be re-replicated.

aishwarya24:
Will create another PR for the --evict_failed_followers flag.

PR #15089, opened 2022-11-21
[docs] Quick Start - replace build app example

ddhodge:
Replace Build an Application with simple apps examples to support multiple languages in Quick Start

@netlify /preview/quick-start/


aishwarya24 (approved):
LGTM

PR #15116, opened 2022-11-22
[docs] change version and add limitation

### LINKED ISSUE #14799: [Docs] Connect an application
### Description

There are a couple of issues with this doc.
1. postgres needs to be above version 11, so 12 or higher.

2.You can not install the yugabyte driver on top of the current psycopg2 driver. You need to uninstall the pyscopg2 driver. Remove the cache then install the driver(this **didnt** work for me but was recommended by a software engineer)

3. The instructions this doc forwards you to says you should install the binary package, this will cause errors when trying to install the yugabyte version

4.I installed successfully on Ubuntu 20.04 with  following commands

sudo apt-get update
sudo apt-get upgrade
sudo apt-get install python3-dev
sudo apt-get install libpq-dev
sudo apt-get update
sudo apt-get upgrade
wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
sudo apt-get install postgresql-12
export PATH=/usr/lib/postgresql/12/bin/:$PATH
sudo apt-get install python-is-python3
sudo apt install python3-pip
pip3 install psycopg2-yugabytedb

5. I am currently working with 2 customers who are experiencing issues with this driver documentation

6. Maybe we can advise in the doc to not install the binary driver or to remove the psycopg2 driver if already installed for customers?

7.Maybe we can have a short script available like above for different OS?

aishwarya24:
resolves #14799

### DIFF
--- a/docs/content/preview/drivers-orms/python/postgres-psycopg2.md
+++ b/docs/content/preview/drivers-orms/python/postgres-psycopg2.md
@@ -202,7 +202,7 @@ Query returned: John, 35, Python
 
 If there is no output or you get an error, verify the parameters included in the connection string.
 
-### Limitation
+### Limitations
 
 Currently, [PostgreSQL psycopg2 driver](https://github.com/psycopg/psycopg2) and [Yugabyte Psycopg2 smart driver](https://github.com/yugabyte/psycopg2) _cannot_ be used in the same environment.
 
--- a/docs/content/preview/drivers-orms/python/postgres-psycopg2.md
+++ b/docs/content/preview/drivers-orms/python/postgres-psycopg2.md
@@ -202,6 +202,10 @@ Query returned: John, 35, Python
 
 If there is no output or you get an error, verify the parameters included in the connection string.
 
+### Limitation
+
+Currently, [PostgreSQL psycopg2 driver](https://github.com/psycopg/psycopg2) and [Yugabyte Psycopg2 smart driver](https://github.com/yugabyte/psycopg2) _cannot_ be used in the same environment.
+
 ## Learn more
 
 - [PostgreSQL Psycopg2 driver reference](../../../reference/drivers/python/postgres-psycopg2-reference/)
--- a/docs/content/preview/drivers-orms/python/yugabyte-psycopg2.md
+++ b/docs/content/preview/drivers-orms/python/yugabyte-psycopg2.md
@@ -51,7 +51,7 @@ To start building your application, make sure you have met the [prerequisites](.
 
 Building Psycopg2 requires a few prerequisites (a C compiler and some development packages). Check the [installation instructions](https://www.psycopg.org/docs/install.html#build-prerequisites) and [the FAQ](https://www.psycopg.org/docs/faq.html#faq-compile) for details.
 
-The YugabyteDB Psycopg2 requires PostgreSQL version 12 or above (preferably 14).
+The YugabyteDB Psycopg2 requires PostgreSQL version 12 or later (preferably 14).
 
 After you've installed the prerequisites, install psycopg2-yugabytedb like any other Python package, using pip to download it from [PyPI](https://pypi.org/project/psycopg2-yugabytedb/):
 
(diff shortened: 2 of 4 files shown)


docs/content/preview/drivers-orms/python/postgres-psycopg2.md
```diff
@@ -202,6 +202,10 @@ Query returned: John, 35, Python
 
 If there is no output or you get an error, verify the parameters included in the connection string.
 
+### Limitation
```
lizayugabyte, line 205: I think we usually call these sections "Limitations", even if there's only one listed

docs/content/preview/drivers-orms/python/yugabyte-psycopg2.md
```diff
@@ -51,7 +51,7 @@ To start building your application, make sure you have met the [prerequisites](.
 
 Building Psycopg2 requires a few prerequisites (a C compiler and some development packages). Check the [installation instructions](https://www.psycopg.org/docs/install.html#build-prerequisites) and [the FAQ](https://www.psycopg.org/docs/faq.html#faq-compile) for details.
 
-The YugabyteDB Psycopg2 requires PostgreSQL version 11 or above (preferably 14).
+The YugabyteDB Psycopg2 requires PostgreSQL version 12 or above (preferably 14).
```
lizayugabyte, line 54: version 12 or later

docs/content/preview/drivers-orms/python/yugabyte-psycopg2.md
```diff
@@ -195,6 +195,10 @@ Query returned: John, 35, Python
 
 If there is no output or you get an error, verify the parameters included in the connection string.
 
+### Limitation
```
lizayugabyte, line 198: Suggestion: 
Limitations

docs/content/preview/reference/drivers/python/postgres-psycopg2-reference.md
```diff
@@ -199,3 +199,7 @@ conn = psycopg2.connect("host=<hostname> port=5433 dbname=yugabyte user=<usernam
 ```
 
 The difference between `verify-ca` and `verify-full` depends on the policy of the root CA. If you're using a public CA, verify-ca allows connections to a server that somebody else may have registered with the CA. Because of this behavior, you should always use verify-full with a public CA. If you're using a local CA, or even a self-signed certificate, using verify-ca may provide enough protection, but the best security practice is to always use verify-full.
+
+## Limitation
```
lizayugabyte, line 203: Limitations

docs/content/preview/reference/drivers/python/yugabyte-psycopg2-reference.md
```diff
@@ -55,7 +55,7 @@ Learn how to perform common tasks required for Python application development us
 
 Building Psycopg2 requires a few prerequisites (a C compiler and some development packages). Check the [installation instructions](https://www.psycopg.org/docs/install.html#build-prerequisites) and [the FAQ](https://www.psycopg.org/docs/faq.html#faq-compile) for details.
 
-The YugabyteDB Psycopg2 requires PostgreSQL version 11 or above (preferably 14).
+The YugabyteDB Psycopg2 requires PostgreSQL version 12 or above (preferably 14).
```
lizayugabyte, line 58: later instead of above

docs/content/preview/reference/drivers/python/yugabyte-psycopg2-reference.md
```diff
@@ -189,6 +189,10 @@ When you're done experimenting, run the following command to destroy the local c
 ./bin/yb-ctl destroy
 ```
 
+### Limitation
```
lizayugabyte, line 192: Limitations

docs/content/preview/drivers-orms/python/postgres-psycopg2.md
aishwarya24: Thanks thats good to know.

PR #15163, opened 2022-11-28
[doc] PR for Document prerequisites for k8s YBA to understand where core dumps go to

lizayugabyte:
Addresses https://yugabyte.atlassian.net/browse/PLAT-6041

### DIFF
--- a/docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
+++ b/docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
@@ -50,7 +50,9 @@ You prepare the host as follows:
 
   Replicated installs a compatible Docker version if it is not pre-installed on the host. The currently supported Docker version is 20.10.n.
 
-- For a Kubernetes-based installation, you need to ensure that the host can pull container images from the [Quay.io](https://quay.io/) container registry. For details, see [Pull and push YugabyteDB Docker images to private container registry](#pull-and-push-yugabytedb-docker-images-to-private-container-registry).
+- For a Kubernetes-based installation, you need to ensure that the host can pull container images from the [Quay.io](https://quay.io/) container registry. For details, see [Pull and push YugabyteDB Docker images to private container registry](#pull-and-push-yugabytedb-docker-images-to-private-container-registry). 
+
+  In addition, you need to ensure that core dumps are enabled and configured on the underlying Kubernetes node. For details, see [Specify ulimit and remember the location of core dumps](#specify-ulimit-and-remember-the-location-of-core-dumps). 
 
 ### Airgapped hosts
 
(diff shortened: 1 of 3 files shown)


lizayugabyte:
@iSignal Could you review this doc PR, please?

docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
```diff
@@ -50,7 +50,9 @@ You prepare the host as follows:
 
   Replicated installs a compatible Docker version if it is not pre-installed on the host. The currently supported Docker version is 20.10.n.
 
-- For a Kubernetes-based installation, you need to ensure that the host can pull container images from the [Quay.io](https://quay.io/) container registry. For details, see [Pull and push YugabyteDB Docker images to private container registry](#pull-and-push-yugabytedb-docker-images-to-private-container-registry).
+- For a Kubernetes-based installation, you need to ensure that the host can pull container images from the [Quay.io](https://quay.io/) container registry. For details, see [Pull and push YugabyteDB Docker images to private container registry](#pull-and-push-yugabytedb-docker-images-to-private-container-registry).
```
iSignal, line 53: @lizayugabyte : do you think we can create tabs on this page for Replicated vs Kubernetes, like we have on other pages like https://docs.yugabyte.com/preview/yugabyte-platform/install-yugabyte-platform/install-software/kubernetes/

That will allow us to specify separate pre-reqs for kubernetes clearly.

docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
```diff
@@ -69,7 +71,31 @@ Installing YugabyteDB Anywhere on Airgapped hosts, without access to any Interne
...
+
+```sh
+cat /proc/sys/kernel/core_pattern
+```
+
+If the value of `core_pattern` contains a `|` pipe symbol (for example, `|/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E`), the core dump is being redirected to a specific collector on the underlying Kubernetes node, with the location depending on the exact collector. To be able to retrieve core dump files in case of a crash within the Kubernetes pod, it is important that you understand where these files are written.
+
+If the value of `core_pattern` is a literal path of the form `/var/tmp/core.%p`, no action is required on your part, as core dumps will be copied by the YugabyteDB node to the persistent volume directory `/mnt/disk0/cores` for future analysis. 
+
+Note that both ulimits and sysctl are inherited from Kubernetes nodes and cannot be changed for an individual pod.
```
amannijhawan, line 96: Can we also add a blurb about systemd-coredump, some new k8s nodes might be using this to manage coredumps on the node
https://www.freedesktop.org/software/systemd/man/systemd-coredump.html

lizayugabyte: added

docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
```diff
@@ -69,7 +71,31 @@ Installing YugabyteDB Anywhere on Airgapped hosts, without access to any Interne
...
+
+A Kubernetes-based installation of YugabyteDB Anywhere requires you to address concerns related to security and core dump collection.
+
+#### Specify ulimit and remember the location of core dumps
+
+The core dump collection in Kubernetes requires special care due to the fact that `core_pattern` is not isolated in cgroup drivers.
+
+You need to ensure that core dumps are enabled on the underlying Kubernetes node. Running the `ulimit -c` command within a Kubernetes pod or node must produce a large non-zero value or the `unlimited` value as an output. For more information, see [How to enable core dumps](https://www.ibm.com/support/pages/how-do-i-enable-core-dumps). 
+
+To be able to locate your core dumps, you should be aware of the fact that the location to which core dumps are written depends on the sysctl `kernel.core_pattern` setting. For more information, see [Linux manual: core(5)](https://man7.org/linux/man-pages/man5/core.5.html#:~:text=Naming of core dump files).
```
bhavin192, line 84: The Linux manual link is broken in the preview.

bhavin192 (approved):
The changes look good to me.

docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
lizayugabyte: @iSignal @bhavin192 This would require creating separate .md files for different installation types. Links to this section would be broken (and there are quite a few of them, including from external docs such as blogs). 
I restructured the "Prepare the host" section in such a way that installation types appear in the right-side menu. Please take a look and let me know if that's acceptable for now.

docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
lizayugabyte: Yes, I saw it and fixed in the updated files

docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites.md
lizayugabyte: @iSignal Yes, I agree, restructuring of prerequisites/prepare content has been on my "future improvements" list. I suggest merging the current PR to make the core dumps information public, and then I'll create a new doc ticket to address restructuring (I'll copy your input to this new ticket). Would that work?

iSignal: Yes, that makes sense, let's get this checked in for now.

PR #15195, opened 2022-12-01
Fix shellcheck warnings and lower the required CMake version

mbautin:
- Fix shellcheck warnings in run-test.sh and yb_build.sh.
- Change required CMake version to 3.17.3.


svarnau:
second commit seems to have un-did the first commit?

PR #15198, opened 2022-12-02
[docs][explore] Update Fault tolerance with new app

ddhodge:
Update the Fault Tolerance topic to use the updated workload generator application

@netlify /preview/explore/fault-tolerance/macos/


docs/content/preview/explore/fault-tolerance/macos.md
```diff
@@ -17,74 +17,62 @@ menu:
...
+YugabyteDB can automatically handle failures and therefore provides [high availability](../../../architecture/core-functions/high-availability/). The recovery point objective (RPO) for the tablets in a YugabyteDB cluster is 0, meaning no data is lost in the failover to another zone. The recovery time objective (RTO) is 3 seconds, which is the time window for completing the failover and becoming operational out of the remaining zones.
+
+This tutorial demonstrates how YugabyteDB can continue to do reads and writes even in case of node failures. You create YSQL tables with a replication factor (RF) of 3, which allows a [fault tolerance](../../../architecture/docdb-replication/replication/) of 1. This means the cluster remains available for both reads and writes even if a node or an entire zone or region fails. However, if another node were to fail (bringing the number of failures to two), writes become unavailable on the cluster to preserve data consistency.
```
aishwarya24, line 22:
```suggestion
This tutorial demonstrates how YugabyteDB can continue to do reads and writes even in case of node failures. You create YSQL tables with a replication factor (RF) of 3, which allows a [fault tolerance](../../../architecture/docdb-replication/replication/#fault-tolerance/) of 1. This means the cluster remains available for both reads and writes even if a node or an entire zone or region fails. However, if another node were to fail (bringing the number of failures to two), writes become unavailable on the cluster to preserve data consistency.
```

aishwarya24: You are linking to the page at the end in Learn more anyway. So you might as well get more specific with the link here.

aishwarya24 (approved):

PR #15228, opened 2022-12-06
[docs] Remove formatting from table

ddhodge:
Remove some custom formatting from table. Minor edits.

@netlify /preview/explore/transactions/distributed-transactions-ysql/

### DIFF
--- a/docs/content/preview/explore/transactions/distributed-transactions-ysql.md
+++ b/docs/content/preview/explore/transactions/distributed-transactions-ysql.md
@@ -3,7 +3,7 @@ title: Distributed transactions
 headerTitle: Distributed transactions
 linkTitle: Distributed transactions
 description: Distributed transactions in YugabyteDB.
-headcontent: Explore distributed transactions in YugabyteDB.
+headcontent: Learn how a distributed transaction works in YugabyteDB
 menu:
   preview:
     name: Distributed transactions
@@ -48,7 +48,7 @@ CREATE TABLE accounts (
 );
 ```
 
-Insert some sample data into the table.
+Insert some sample data into the table:
 
 ```sql
 INSERT INTO accounts VALUES ('John', 'savings', 1000);
@@ -57,7 +57,7 @@ INSERT INTO accounts VALUES ('Smith', 'savings', 2000);
 INSERT INTO accounts VALUES ('Smith', 'checking', 50);
 ```
 
-The table should look as follows:
+Display the contents of the table:
 
 ```sql
 yugabyte=# SELECT * FROM accounts;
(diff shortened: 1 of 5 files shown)


docs/content/preview/explore/transactions/distributed-transactions-ysql.md
```diff
@@ -90,54 +90,62 @@ The following table explains what happens at each step.
 
 <table>
```
lizayugabyte, line 91: While you are in this file, could you fix it up a little? For example, in "Create a table":

_Insert some sample data into the table._ --> Insert some sample data into the table:

_The table should look as follows:_      and then there's a SELECT statement

docs/content/preview/explore/transactions/distributed-transactions-ysql.md
```diff
@@ -90,54 +90,62 @@ The following table explains what happens at each step.
...
   <tr>
```
lizayugabyte, line 92: The command+description info doesn't look good in a table. Commands need the full width of the page. This table should be converted to a list.

docs/content/preview/explore/transactions/distributed-transactions-ysql.md
```diff
@@ -201,7 +209,7 @@ where transaction_mode is one of:
   [ NOT ] DEFERRABLE
 ```
```
lizayugabyte, line 210: In the "Run a transaction" section:

_This is shown diagrammatically in the following illustration._ --> The following diagram shows the sequence of events that occur when a transaction is running:
(or something like that)

docs/content/preview/explore/transactions/distributed-transactions-ysql.md
```diff
@@ -201,7 +209,7 @@ where transaction_mode is one of:
...
```
lizayugabyte, line 211: In the "Run a transaction" section:

_After the above transaction succeeds, the table should look as follows:_   and then there's a SELECT statement. We also need to get rid of "above"

docs/content/preview/explore/transactions/distributed-transactions-ysql.md
```diff
@@ -201,7 +209,7 @@ where transaction_mode is one of:
...
 
-### `transaction_mode`
```
lizayugabyte, line 204: In the "Concurrency control" section, we should unwrap the Note element

docs/content/preview/explore/transactions/distributed-transactions-ysql.md
```diff
@@ -201,7 +209,7 @@ where transaction_mode is one of:
...
+### transaction_mode
 
 The `transaction_mode` can be set to one of the following options:
```
lizayugabyte, line 215: _Users can perform read or write operations._ and _Users can perform read operations only._   Who are these users? Shouldn't it be "you"?

_As an example, trying to do a write operation such as creating a table or inserting a row in a READ ONLY transaction would result in an error as shown below._ -->  As an example, trying to do a write operation such as creating a table or inserting a row in a READ ONLY transaction would result in an error:

docs/content/preview/explore/transactions/distributed-transactions-ysql.md
```diff
@@ -226,7 +234,7 @@ yugabyte=# CREATE TABLE example(k INT PRIMARY KEY);
 ERROR: cannot execute CREATE TABLE in a read-only transaction
 ```
 
-### `DEFERRABLE` transactions
+### DEFERRABLE transactions
 
 The `DEFERRABLE` transaction property in YSQL is similar to PostgreSQL in that has no effect unless the transaction is also `SERIALIZABLE` and `READ ONLY`.
```
lizayugabyte, line 239: _The DEFERRABLE transaction property in YSQL is similar to PostgreSQL in that has no effect unless the transaction is also SERIALIZABLE and READ ONLY._    -- ... in that has no effect unless ...  -- doesn't sound right

lizayugabyte (commented):
Some comments

PR #15247, opened 2022-12-07
[docs] 404 and sitemap/RSS fixes

polarweasel:
Several things going on here:

- a number of 404 fixes
- adding a redirect or two
- adding a missing link
- adjusting the format of the sitemap template, to make it output valid sitemaps and to omit pages marked as `private` in their front matter
- adjusting the format of the RSS template, to trim down the entries and make them have correct dates


polarweasel:
At this point the link checker is showing no errors, so I'm going to call things good for now.

PR #15271, opened 2022-12-09
[docs] Add an example to audit logging docs for concurrent transactions

dataindataout:
Based on a demo from my recent YFTT, I'm adding an example of audit logging of multiple concurrent transactions.

Also added a few notes based on findings in YFTT preparation and audience questions.

@netlify /preview/quick-start/


lizayugabyte (approved):

polarweasel (approved):
Thanks @dataindataout! Please add these changes into `docs/content/preview/secure/audit-logging/audit-logging-ysql.md` as well, and we should be good to go.

PR #15497, opened 2023-01-03
[docs] Add voyager redirects to preview

aishwarya24:
Make stable(2.16) and 2.14 of migrate section point to preview.


ddhodge (approved):
LGTM

PR #15518, opened 2023-01-05
[docs] Change earlier version logic

samiahmedsiddiqui:


docs/data/currentVersions.json
```diff
@@ -32,6 +33,7 @@
       "appVersion": "2.14.5.0-b18",
       "isStable": true,
       "isLTS": true,
+      "isSTS": true,
```
polarweasel, line 36: A stable version can be either STS or LTS, but not both.
```suggestion
      "isSTS": false,
```

docs/layouts/partials/earlier-version-warning.html
```diff
@@ -8,15 +9,17 @@
...
-    <p class="admonition-title">Attention</p>
-    <p>This page documents an earlier version. <a href="{{ $previewUrl }}">Go to the preview {{ $previewVersion }} version.</a></p>
-  </div>
-  {{ end }}
-{{ end }}
\ No newline at end of file
+  {{ range .Site.Data.currentVersions.dbVersions }}
+    {{- if or (eq $docVersion .series) (eq $docVersion .alias)  -}}
+      <div class="admonition warning">
+        {{- if and (eq .isStable true) (eq .isSTS true) -}}
```
polarweasel, line 15: This logic should be "if isStable is true AND (if isSTS is true OR isLTS is true)"

polarweasel: And perhaps we can simplify to just "if isSTS is true OR isLTS is true".

polarweasel (approved):
Works just right. Thanks @samiahmedsiddiqui!

PR #15522, opened 2023-01-05
[doc] PR for Add resource group role necessary for Azure provider

lizayugabyte:
Addresses https://yugabyte.atlassian.net/browse/PLAT-5861

Based on the comments in the Jira ticket, only the wording around the disk size needed to be changed.


chidmuthu (changes requested):
We should specify that the disk we want to expand is the OS disk.

chidmuthu:
https://learn.microsoft.com/en-us/azure/virtual-machines/linux/expand-disks?tabs=ubuntu#increase-the-size-of-the-os-disk

chidmuthu (approved):

PR #15604, opened 2023-01-09
[docs] use current version as title in version selector

samiahmedsiddiqui:


docs/layouts/partials/navbar-version-selector.html
```diff
@@ -1,5 +1,19 @@
...
+  {{ range .Site.Params.versions }}
+    {{ if hasPrefix $pagePermalink .url }}
+      {{ $versionName = .version }}
+      {{- if in .version "(Preview)" -}}
+        {{ $versionName = replace .version "(Preview)" "<span class='metadata-tag-gray'>Preview</span>" }}
+      {{- else if in .version "(STS)" -}}
+        {{ $versionName = replace .version "(STS)" "<span class='metadata-tag-green'>STS</span>" }}
+      {{- else if in .version "(LTS)" -}}
+        {{ $versionName = replace .version "(LTS)" "<span class='metadata-tag-green'>LTS</span>" }}
+      {{ end }}
```
polarweasel, line 13:
```suggestion
      {{- else if in .version "Unsupported" -}}
        {{ $path = "/" }}
      {{ end }}
```

polarweasel (approved):
LGTM. There's an issue on the site root where the URLs in the version get a double slash (such as `/preview//`), but it seems to be a self-correcting problem as the browser just drops the second slash.

PR #15623, opened 2023-01-10
Prometheus config based on kubernetes_sd_configs

hau21um:
When using original configuration, if cluster architecture change (adding/removing masters/tservers), it is also needed to change prometheus config. With proposed changes, prometheus will get the count of masters/tservers automatically. Should be working with current Grafana dashboard as well.

### DIFF
--- a/docs/content/preview/explore/observability/prometheus-integration/macos.md
+++ b/docs/content/preview/explore/observability/prometheus-integration/macos.md
@@ -55,7 +55,45 @@ global:
 scrape_configs:
   - job_name: "yugabytedb"
     metrics_path: /prometheus-metrics
+    kubernetes_sd_configs:
+      - role: pod    
     relabel_configs:
+      # only scrape the pods where "component" label is "yugabytedb"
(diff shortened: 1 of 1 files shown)


bhavin192:
Hello @hau21um, thank you for the PR. 
This Prometheus configuration is for a local test setup as described at the top of the page. This setup uses the tool `yugabyted` to spin-up a local YugabyteDB cluster using 127.0.0.1/2/3, hence the Prometheus configuration has those IPs. Since this is localhost and not running on Kubernetes, your change will not be relevant to this particular documentation page. 

For Kubernetes we have Helm chart with [support for Prometheus Operator's ServiceMonitor](https://github.com/yugabyte/charts/blob/2596d9f04a4340cf02b420d5f5df3d8abb8750cd/stable/yugabyte/values.yaml#L198), or you can also [pass on extra annotations/labels](https://github.com/yugabyte/charts/blob/2596d9f04a4340cf02b420d5f5df3d8abb8750cd/stable/yugabyte/values.yaml#L313-L317) for your specific setup, so that Prometheus will scrape the YugabyteDB pods.

PR #15713, opened 2023-01-16
[Docs] PR for Edit the Architecture section files for clarity+formatting

lizayugabyte:
Notes to reviewers:

- DocDB vs. RocksDB vs. YugabyteDB --- these seem to be used interchangeably, which would confuse the reader.
- time, hybrid time --- these imply value, but almost never accompanied by the word value. "lower" and "higher" was used in the docs, which I changed to "earlier" and "later" because these can be applied to time.
- end user, user --- these mean "you" in most cases and should be corrected, but I didn't do it because I thought I'd do another pass, then kinda ran out of time. Feel free to make suggestions on how rephrase when you see them.

Note to Alex: there's already a bunch of conflicts. Do you mind resolving them on the command line during review?


ddhodge (approved):
misc tidyups

polarweasel (commented):
I've gotten partway through and made a bunch of suggested changes and asked some questions. Sorry I haven't been able to get any further through, but happy to take another pass on the next round.

PR #15765, opened 2023-01-20
[docs][cloud] PR for We should call out leader and follower placement for multi-region clusters

ddhodge:
Add caveats related to data in tablespaces and replication when a preferred region is specified.

@netlify /preview/yugabyte-cloud/cloud-basics/create-clusters/create-clusters-multisync/

### DIFF
--- a/docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters/create-clusters-geopartition.md
+++ b/docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters/create-clusters-geopartition.md
@@ -28,6 +28,8 @@ To view your cluster tablespaces, you can enter the following command:
 SELECT * FROM pg_tablespace;
 ```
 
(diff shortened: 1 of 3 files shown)


docs/content/preview/yugabyte-cloud/cloud-basics/create-clusters/create-clusters-geopartition.md
```diff
@@ -28,6 +28,8 @@ To view your cluster tablespaces, you can enter the following command:
 SELECT * FROM pg_tablespace;
 ```
 
+Note that data placed in regions in tablespaces is not replicated to other regions, and remains subject to the fault tolerance of the cluster (Node- or Availability Zone-level).
```
mrajcevic01, line 31: Wondering if we can rephrase "Note that data placed in regions in tablespaces" because it could be misleading. You can set up tablespaces to be multi-region if you wanted. Maybe something along the lines of "Note that data pinned to a single region via tablespaces"

lizayugabyte (approved):

PR #15906, opened 2023-01-31
[docs] Smart driver topology fallback and server refresh

ddhodge:
Update smart driver page and driver pages with:
- how to specify fallback locations for topology keys
- how to change the server refresh interval

@netlify /preview/drivers-orms/smart-drivers/


docs/content/preview/drivers-orms/go/yb-pgx.md
```diff
@@ -100,21 +100,22 @@ The following table describes the connection parameters required to connect, inc
 | password | User password | yugabyte
 | dbname | Database name | yugabyte
 | `load_balance` | [Uniform load balancing](../../smart-drivers/#cluster-aware-connection-load-balancing) | Defaults to upstream driver behavior unless set to 'true'
-| `yb_servers_refresh_interval` | Interval in seconds to refresh the node list | 300
+| `yb_servers_refresh_interval` | Interval in seconds to refresh the servers list | 300
```
ashetkar, line 103: We can add that this is applicable only when load-balance is set to true.

ashetkar (approved):

docs/content/preview/drivers-orms/go/yb-pgx.md
```diff
@@ -49,6 +49,8 @@ type: docs
 
 The [YugabyteDB PGX smart driver](https://pkg.go.dev/github.com/yugabyte/pgx) is a distributed Go driver for [YSQL](/preview/api/ysql/) based on [jackc/pgx](https://github.com/jackc/pgx/), with a additional [connection load balancing](../../smart-drivers/) features.
```
ashetkar, line 50: Not sure we use word `distributed` for a driver.

docs/content/preview/drivers-orms/go/yb-pgx.md
```diff
@@ -100,27 +101,29 @@ The following table describes the connection parameters required to connect, inc
...
-url := fmt.Sprintf("%s?load_balance=true", baseUrl)
+url := fmt.Sprintf("%s?load_balance=true&yb_servers_refresh_interval=240", baseUrl)
 conn, err := pgx.Connect(context.Background(), url)
 ```
 
 The following is an example connection string for connecting to YugabyteDB with topology-aware load balancing:
 
 ```sh
-postgres://username:password@localhost:5433/database_name?load_balance=true&
+postgres://username:password@localhost:5433/database_name?load_balance=true& \
```
ashetkar, line 126: If this is expected to be copied as-is, then `\` won't be useful in a Go app. Instead you can either put it in a single line or use `+`
```
url := "postgres://username:password@localhost:5433/database_name?load_balance=true&" +
    "topology_keys=cloud1.region1.zone1,cloud1.region1.zone2"
```

docs/content/preview/drivers-orms/go/yb-pgx.md
```diff
@@ -143,6 +146,8 @@ You can specify multiple hosts in the connection string to provide alternative o
 postgres://username:password@host1:5433,host2:5433,host3:5433/database_name?load_balance=true
 ```
 
+To query the cluster for a list of available hosts, use the `yb_servers()` function.
```
ashetkar, line 149: We don't expect users to query the yb_servers() function. It is the driver who does it. This statement may confuse readers. Same holds true on other language pages.

docs/content/preview/drivers-orms/java/yugabyte-jdbc.md
```diff
@@ -57,6 +57,8 @@ type: docs
 
 For Java applications, the JDBC driver provides database connectivity through the standard JDBC application program interface (APIs) available on the Java platform.
 
+The driver package includes a `YBClusterAwareDataSource` class that uses one initial contact point for the YugabyteDB cluster as a means of discovering all the nodes and, if required, refreshing the list of live endpoints with every new connection attempt. The refresh is triggered if stale information (by default, older than 5 minutes) is discovered.
```
ashetkar, line 60: I see that in the code example below, we really don't use this `YBClusterAwareDataSource` class or the Hikari APIs. We can remove them from this page.

The JDBC application can be written either with the `DriverManager.getConnection()` API or the `DataSource` API. On this page, we are using the former way.

docs/content/preview/drivers-orms/smart-drivers.md
```diff
@@ -28,6 +25,18 @@ Yugabyte has developed the following smart drivers, available as open source sof
...
+## Key features
+
+YugabyteDB smart drivers have the following key features.
+
+| Feature | Notes |
+| :--- | :--- |
+| Multiple hosts | As with the upstream driver (with the exception of node.js), you can specify multiple hosts for the initial connection, thereby avoiding dropped connections in the case where the primary host is unavailable. |
+| [Cluster aware](#cluster-aware-connection-load-balancing) | Smart drivers perform automatic uniform connection load balancing<br/>After the driver establishes an initial connection, it fetches the list of available servers from the cluster and distributes connections evenly across them. |
+| [Servers refresh interval](#servers-refresh-interval) | By default, the driver refreshes the list of available servers every five minutes. The interval is configurable (with the exception of Python). |
+| [Topology aware](#topology-aware-connection-load-balancing) | In cases where you want to restrict connections to specific geographies for lower latency, you can target specific regions and zones, along with fallback zones, across which to balance connections. |
```
ashetkar, line 37: I would move 'Topology aware' one place above, right after 'Cluster aware'.
'Servers refresh interval' is not really a feature but a way to configure a property used by 'Cluster aware' and 'Topology aware' features.

docs/content/preview/drivers-orms/smart-drivers.md
```diff
@@ -72,6 +81,14 @@ With cluster-aware (also referred to as uniform) connection load balancing, conn
 
 For example, if a client application creates 100 connections to a YugabyteDB cluster consisting of 10 nodes, then the driver creates 10 connections to each node. If the number of connections is not exactly divisible by the number of servers, then a few may have 1 less or 1 more connection than the others. This is the client view of the load, so the servers may not be well balanced if other client applications are not using a smart driver.
 
+A connection works as follows:
+
+- The driver makes an initial connection to the host specified in the URL/connection string. You can specify multiple hosts to act as backups if the connection to the primary host fails.
+- The driver fetches information about the cluster nodes using the `yb_servers()` function. By default, this list is refreshed every 5 minutes, or when a new connection request is received.
```
ashetkar, line 87: nit (to be precise):
```
By default, this list is refreshed every 5 minutes, and this time is checked when a new connection request is received.
```

docs/content/preview/drivers-orms/smart-drivers.md
```diff
@@ -80,28 +97,37 @@ For example, using the Go smart driver, you would turn on load balancing as foll
...
+
+For example, using the Go smart driver, you can change the interval to four minutes (specified in seconds) as follows:
 
-The application must use the same connection URL to create every connection it needs, so that the distribution happens equally.
+```go
+"postgres://username:password@host:5433/database_name?load_balance=true&yb_servers_refresh_interval=240"
+```
 
-Note that the nodes in the cluster must be accessible. If, for example, the cluster has multiple regions deployed in separate VPCs, your application would need access to all the regions, typically via peering.
+Note this feature is not available in the YugabyteDB Python Smart Driver.
```
ashetkar, line 118: I would add word 'currently' here. We'll add this property soon.

docs/content/preview/drivers-orms/smart-drivers.md
```diff
@@ -110,11 +136,27 @@ For example, using the Go driver, you would set the parameters as follows:
...
+
+```go
+"postgres://username:password@localhost:5433/database_name?load_balance=true& \
+    topology_keys=cloud1.region1.*"
+```
 
-You still need to specify load balance as true to enable the topology-aware connection load balancing.
+#### Fallback topology keys
+
+To specify fallback locations in cases where a location is unavailable, add `:n` to the topology key, where n is an integer indicating priority. The following example sets `zone1` as the topology key, and zones 2 and 3 as fallbacks (in that order) if `zone1` can't be reached:
```
ashetkar, line 148: IMO, we can introduce the fallback use case before the syntax. Something like:
So users can specify fallback placements when they specify topology-keys. So when no nodes in the primary placements are available, the driver can attempt connections to nodes in the fallback placements.
Also, if topology-keys without preference-value/priority is provided (i.e. no explicit fallback placement is specified), the driver falls back to nodes in the entire cluster in case the nodes in those specified placements are not available.

docs/content/preview/drivers-orms/java/yugabyte-jdbc.md
ddhodge: Right, it is the Reference page example that uses YBClusterAwareDataSource; I will move it back there. Generally though, this information isn't provided in a single location which could be inconvenient.

ashetkar (approved):

docs/content/preview/drivers-orms/go/yb-pgx.md
```diff
@@ -47,7 +47,9 @@ type: docs
 
 </ul>
 
-The [YugabyteDB PGX smart driver](https://pkg.go.dev/github.com/yugabyte/pgx) is a distributed Go driver for [YSQL](/preview/api/ysql/) based on [jackc/pgx](https://github.com/jackc/pgx/), with a additional [connection load balancing](../../smart-drivers/) features.
+The [YugabyteDB PGX smart driver](https://pkg.go.dev/github.com/yugabyte/pgx) is a Go driver for [YSQL](/preview/api/ysql/) based on [jackc/pgx](https://github.com/jackc/pgx/), with a additional [connection load balancing](../../smart-drivers/) features.
```
polarweasel, line 50:
```suggestion
The [YugabyteDB PGX smart driver](https://pkg.go.dev/github.com/yugabyte/pgx) is a Go driver for [YSQL](/preview/api/ysql/) based on [jackc/pgx](https://github.com/jackc/pgx/), with additional [connection load balancing](../../smart-drivers/) features.
```

docs/content/preview/drivers-orms/go/yb-pgx.md
```diff
@@ -47,7 +47,9 @@ type: docs
...
+
+The driver makes an initial connection to the first contact point provided by the application to discover all the nodes in the cluster and, if required, refreshes the list of live endpoints with every new connection attempt. The refresh is triggered if stale information (by default, older than 5 minutes) is discovered.
```
polarweasel, line 52:
```suggestion
The driver makes an initial connection to the first contact point provided by the application to discover all the nodes in the cluster. If the driver discovers stale information (by default, older than 5 minutes), it refreshes the list of live endpoints with every new connection attempt.
```

lizayugabyte (changes requested):

PR #15991, opened 2023-02-07
Adding python symlink troubleshooting step

angoothachap:


docs/content/preview/contribute/docs/docs-build.md
```diff
@@ -115,6 +115,8 @@ When the build is done, the `yugabyte-db/docs/public` folder contains a full HTM
 
 ## Troubleshooting
 
+* If either `npn start` or `npm run build` exit with the error `npm ERR! env: python: No such file or directory`, for a homebrew python install, create a symlink by executing `sudo ln -s /opt/homebrew/bin/python3 /opt/homebrew/bin/python`
```
ddorian, line 118:
```suggestion
* If either `npm start` or `npm run build` exit with the error `npm ERR! env: python: No such file or directory`, for a homebrew python install, create a symlink by executing `sudo ln -s /opt/homebrew/bin/python3 /opt/homebrew/bin/python` 
```

polarweasel:
I've taken a different approach, without needing sudo, in [another PR](https://github.com/yugabyte/yugabyte-db/pull/16026). I suggest we take that one rather than this, as it should avoid ever needing the troubleshooting step. :)

PR #16042, opened 2023-02-08
Docs: Fix some grammatical errors in the Read Committed isolation doc…

pkj415:
…s page


docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
+  Note that the clock skew between all nodes in the cluster is always in a [max_clock_skew_usec](/preview/reference/configuration/yb-tserver/#max-clock-skew-usec) bound due to clock synchronization algorithms.
+* For writes with a write timestamp higher than `T1` + `max_clock_skew`, the database can be sure that they were done after the read timestamp was chosen. But for writes at a time between `T1` and `T1` + `max_clock_skew`, node `N2` can find itself in an ambiguous situation such as the following:
 
-  * It should still return the data if the client issued the read after the data was committed, because it could be the same client connecting to YSQL from a different node and the following guarantee needs to be maintained; the database always returns data that was committed in the past.
+  * It should return the data even if the client issued the read from a different node after writing the data since the following guarantee needs to be maintained: the database always returns data that was committed in the past (here past means the user-perceived past not based machine clocks).
```
aishwarya24, line 39:
```suggestion
  * It should return the data even if the client issued the read from a different node after writing the data, because the following guarantee needs to be maintained: the database always returns data that was committed in the past (past refers to the user-perceived past, and not based on machine clocks).
```

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
+* Assuming that node `N1` reads from node `N2`, it could be the case that there exists some data written on node `N2` with a write timestamp `T2` (> `T1`) but it had been written before the read was issued. This can happen because of clock skew where the physical clock on node `N2` might be running ahead of node `M`, and hence the write which was actually done in the past, still has a write timestamp higher than `T1`.
 
-  Note that the clock skew between all nodes in the cluster is always in a [`max_clock_skew`](/preview/reference/configuration/yb-tserver/#max-clock-skew-usec) bound due to clock synchronization algorithms.
-* For writes at some time later than `T1` + `max_clock_skew`, the database can be sure that they were done after the read timestamp was chosen on any node. But for writes at a time between `T1` and `T1` + `max_clock_skew`, node `N2` can find itself in an ambiguous situation such as the following:
+  Note that the clock skew between all nodes in the cluster is always in a [max_clock_skew_usec](/preview/reference/configuration/yb-tserver/#max-clock-skew-usec) bound due to clock synchronization algorithms.
```
aishwarya24, line 36:
```suggestion
  Note that the clock skew between all nodes in the cluster is always in a [max_clock_skew_usec](../../../reference/configuration/yb-tserver/#max-clock-skew-usec) bound due to clock synchronization algorithms.
```

aishwarya24 (approved):
LGTM . Left minor suggestions.

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
 * The node `N1` might collect data from many shards on different physical YB-TServers. In this pursuit, it will issue requests to many other nodes to read data.
-* Assuming that node `N1` reads from node `N2`, it could be the case that there exists some data written on node `N2` at time `T2` (> `T1`) but had been written before the read was issued. This can happen because of clock skew where the physical clock on node `N2` might be running slightly ahead of node `M`, and hence the write which was actually done in the past, still has a timestamp higher than `T1`.
+* Assuming that node `N1` reads from node `N2`, it could be the case that there exists some data written on node `N2` with a write timestamp `T2` (> `T1`) but it had been written before the read was issued. This can happen because of clock skew where the physical clock on node `N2` might be running ahead of node `M`, and hence the write which was actually done in the past, still has a write timestamp higher than `T1`.
```
lizayugabyte, line 34: Suggestion:

* Assuming that node `N1` reads from node `N2`, it is possible that data had already been written on node `N2` with a write timestamp `T2` (> `T1`) before the read was issued. This could be caused by clock skew if the physical clock on node `N2` ran ahead of node `M`, resulting in the write done in the past still having a write timestamp later than `T1`.

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
 
-  Note that the clock skew between all nodes in the cluster is always in a [`max_clock_skew`](/preview/reference/configuration/yb-tserver/#max-clock-skew-usec) bound due to clock synchronization algorithms.
-* For writes at some time later than `T1` + `max_clock_skew`, the database can be sure that they were done after the read timestamp was chosen on any node. But for writes at a time between `T1` and `T1` + `max_clock_skew`, node `N2` can find itself in an ambiguous situation such as the following:
+  Note that the clock skew between all nodes in the cluster is always in a [max_clock_skew_usec](/preview/reference/configuration/yb-tserver/#max-clock-skew-usec) bound due to clock synchronization algorithms.
+* For writes with a write timestamp higher than `T1` + `max_clock_skew`, the database can be sure that they were done after the read timestamp was chosen. But for writes at a time between `T1` and `T1` + `max_clock_skew`, node `N2` can find itself in an ambiguous situation such as the following:
```
lizayugabyte, line 37: Suggestion:

* For writes with a write timestamp later than `T1` + `max_clock_skew`, the database can be sure that these writes were done after the read timestamp had been chosen. But for writes at a time between `T1` and `T1` + `max_clock_skew`, node `N2` can find itself in an ambiguous situation, such as the following:

docs/content/preview/architecture/transactions/read-committed.md
lizayugabyte, line 39: Suggestion:

 * It should return the data even if the client issued the read from a different node after having written the data, because the following guarantee needs to be maintained: the database always returns data that was committed in the user-perceived past (as opposed to past in machine clocks).

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
 
-  * It should still return the data if the client issued the read after the data was committed, because it could be the same client connecting to YSQL from a different node and the following guarantee needs to be maintained; the database always returns data that was committed in the past.
+  * It should return the data even if the client issued the read from a different node after writing the data since the following guarantee needs to be maintained: the database always returns data that was committed in the past (here past means the user-perceived past not based machine clocks).
 
   * It should not return the data if the write was actually performed in the future, that is, after the read point (also known as snapshot) was chosen.
```
lizayugabyte, line 41: Suggestion:

  * It should not return the data if the write was performed in the future (that is, after the read point had been chosen).

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
 
-* If node `N2` finds writes in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier, node `N2` avoids giving incorrect results and raises a `Read restart` error.
+* Hence, if node `N2` finds writes in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier, and to also avoid reading data that was actually written after the read was issued, node `N2` raises a `Read restart` error.
```
lizayugabyte, line 43: Suggestion:

* If node `N2` finds writes in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier and to avoid reading data that was written after the read had been issued, node `N2` raises a read restart error.

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
 
 Some distributed databases handle this uncertainty due to clock skew by using algorithms to maintain a tight bound on the clock skew, and then taking the conservative approach of waiting out the clock skew before acknowledging the commit request from the client for each transaction that writes data. YugabyteDB instead uses various mechanisms internally to reduce the scope of this ambiguity, and if there is still ambiguity in rare scenarios, the error is surfaced to the client.
```
lizayugabyte, line 45: Suggestion:

Some distributed databases handle uncertainty due to clock skew by using algorithms to maintain a tight bound on the clock skew, and then taking the conservative approach of waiting out the clock skew before acknowledging the commit request from the client for each transaction that writes data. YugabyteDB, however, uses various internal mechanisms to reduce the scope of this ambiguity, and if there is still ambiguity in rare cases, the error is surfaced to the client.

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
 
-However, for Read Committed isolation, YugabyteDB has stronger mechanisms in place to ensure that the ambiguity is always resolved internally, and `Read restart` errors are not surfaced to the client. In Read Committed transactions, clients do not have to add any retry logic for `Read restart` errors (similarly to serialization errors).
+However, for Read Committed isolation, YugabyteDB has stronger mechanisms in place to ensure that the ambiguity is always resolved internally, and `Read restart` errors are not surfaced to the client. In Read Committed transactions, clients do not have to add any retry logic for `Read restart` errors (similar to serialization errors).
```
lizayugabyte, line 47: For Read Committed isolation, YugabyteDB has stronger protection to ensure that the ambiguity is resolved internally and read restart errors are not surfaced to the client. In Read Committed transactions, clients do not have to add any retry logic for read restart errors (similarly to serialization errors).

docs/content/preview/architecture/transactions/read-committed.md
```diff
@@ -27,24 +27,24 @@ The following two key semantics set apart Read Committed isolation from Repeatab
...
 
-* A client starts a distributed transaction by connecting to YSQL on a node `N1` in the YugabyteDB cluster and issuing a statement which reads data from multiple shards on different physical YB-TServers in the cluster. For this issued statement, the read point which defines the snapshot of the database at which the data will be read, is picked on a YB-TServer node `M` based on the current time of that YB-TServer. Depending on the scenario, node `M` could be the same as `N1` or not, but that is not relevant to this discussion. Consider `T1` to be the chosen read time.
+* A client starts a distributed transaction by connecting to YSQL on a node `N1` in the YugabyteDB cluster and issues a statement which reads data from multiple shards on different physical YB-TServers in the cluster. For this issued statement, the read point which defines the snapshot of the database at which the data will be read, is picked on a YB-TServer node `M` based on the current time of that YB-TServer. Depending on the scenario, node `M` could be the same as `N1` or not, but that is not relevant to this discussion. Consider `T1` to be the chosen read time.
```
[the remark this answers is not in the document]
pkj415: incorporated the suggestion except the change from `cluster` to `universe`. Cluster seems better here since this is in the architecture section and readers might be able to related more with `cluster`.

About `.... the read point that .... of that YB-TServer` -- I couldn't find a good way to reword/ break.... Is it fine as is?

docs/content/preview/architecture/transactions/read-committed.md
pkj415: incorporated the suggestion with this minor deviation: I used `.... for writes with a write timestamp between .... ` instead of `... for writes at a time between ... ` since the former is more accurate. Because we are specifically talking about a `write timestamp` here

docs/content/preview/architecture/transactions/read-committed.md
pkj415: incorporated the suggestion. Plus I added one more line to it since there was one critical piece missing in this explanation.

docs/content/preview/architecture/transactions/read-committed.md
pkj415: incorporated, but also added a phrase in there to make it accurate, earlier it was missing

docs/content/preview/architecture/transactions/read-committed.md
pkj415: changed `this certainty` to `this specific uncertainty` instead of removing this because that is important. There could be many other uncertainties but we want to refer this one. Let me know what you think

docs/content/preview/architecture/transactions/read-committed.md
pkj415: I think the existing one is better...? -- because we want to convey that there are multiple mechanisms and the word `always` is redundant but kept to emphasize to the reader that the errors never occur

docs/content/preview/architecture/transactions/read-committed.md
pkj415: nice suggestion! made it much better

docs/content/preview/architecture/transactions/read-committed.md
pkj415: done

docs/content/preview/architecture/transactions/read-committed.md
pkj415: done

PR #16093, opened 2023-02-13
[docs] adjust URL for YB blogs

polarweasel:
Adjust all blog links to point to www.yugabyte.com/blog/


polarweasel:
Checked some random pages, all the blog links work

PR #16264, opened 2023-02-28
[docs][explore] Destroy local cluster instructions

ddhodge:
Add setup instructions on destroying local clusters for cleanup

@netlify /preview/explore/


aishwarya24 (approved):
This looks great ! Definitely clearer than having it on the explore index page. LGTM

PR #16381, opened 2023-03-10
Architecting for HA of transactions

premkumr:
- Adding a page for HA of Transctions
- Helps users to handle Error Codes correctly


premkumr:
Abandoning this PR because of too many merge conflicts with master

PR #16619, opened 2023-03-30
[PLAT-7792] Disable play enhancement for subsequent Play 2.8 + Play Ebean 2.x migration

anmalysh-yb:
Summary:
Latest play 2.8 and play-ebean versions does not work with play enhancement. Also, latest sbt does not work with latest play-enhancement version. Basically, project reached EOL 3-4 years ago.
We have to get rid of it to migrate to up-to-date play stack.

One issue is that play-ebean expects getters/setters in models - it further enhances them with it's logic. So, we have to encapsulate all ebean model fields. This is done carefully as sometimes there are conflicts with existing get/set methods in models, which are not used as getter/setters now. Another issue is that some task params use inheritance and re-define some fields in child classes. It works in case getters/setters are used. But does not work with public fields - as there are 2 separate fields under the hood - and you may set one but get the other one, which is unset. Encapsulated these as well.

Test Plan:
Unit tests.
Universe create/delete.
Backup create/restore.
itests.

Reviewers: sb-yb, vbansal, svarshney

Subscribers: yugaware

Differential Revision: https://phabricator.dev.yugabyte.com/D23706


Vars-07 (commented):
lgtm.

Vars-07 (approved):

anmalysh-yb:
Landed changes via Phabricator

PR #16620, opened 2023-03-30
[PLAT-7793] Up-version to Play 2.8, play-ebean 6.2.0, flyway-play 7.37.0 and sbt 1.7.2

anmalysh-yb:
Summary:
Migrating to latest framework versions to simplify security patches, etc.
Major part of the diff is changing the way we access Request object in controller method - now it has to be passed explicitly.
Also some changes here and there to be compatible with the current framework APIs.
For flyway - all migrations were moved to _default folders + standard Flyway initializer is used.
The only custom part left is the ability to fulle disable running migrations for perf-advisor DB in YBM environment.

Test Plan:
Create universe.
Create backup/restore backup.
Regular and SSO login.
itests run

Reviewers: sb-yb, svarshney, vbansal

Subscribers: sanketh, yugaware

Differential Revision: https://phabricator.dev.yugabyte.com/D24013


managed/src/main/java/com/yugabyte/yw/models/Universe.java
```diff
@@ -434,9 +434,6 @@ public static Universe saveDetails(
   public static void delete(UUID universeUUID) {
     // First get the universe.
     Universe universe = Universe.getOrBadRequest(universeUUID);
-    // Make sure this universe has been locked.
-    // TODO: fixme. Useless check. java asserts are turned off by default in production code!!!
-    assert !universe.universeDetails.updateInProgress;
```
sb-yb, line 439: lol

managed/src/main/resources/swagger.json
```diff
@@ -16,17 +16,14 @@
           "type" : "string"
         },
         "hostVpcId" : {
-          "readOnly" : true,
```
sb-yb, line 19: Why all these changes to readOnly property?

managed/src/main/resources/logback-sbt.xml
```diff
@@ -0,0 +1,150 @@
+<!-- https://www.playframework.com/documentation/latest/SettingsLogger -->
```
sb-yb, line 1: How is this different from logback.xml? 
I checked the link but there was only mention of logback.xml nothing about logback-sbt.xml.
Maybe add some comments here explaining the purpose.

sb-yb (commented):
swagger spec changes for readonly are a concern. 
Did these happen because now there is an implicit setter?

Otherwise looks good.

sb-yb (approved):
Approving to unblock you. But please check/fix the `readOnly` changes in swagger spec b4 landing.

managed/src/main/resources/swagger.json
anmalysh-yb: Seems like new swagger plugin does not respect accessMode = READ_ONLY for some reason.
Will dig into that.

managed/src/main/resources/logback-sbt.xml
anmalysh-yb: So, the context here is:
In new sbt version it has special classloader for build processing, which has very limited set of classes in it.
And play-ebean plugin is initialised during the build, which causes logging config to be parsed and processed.
As our default logback.xml file contains this:

`        <filter class="ch.qos.logback.core.filter.EvaluatorFilter">
          <evaluator>
            <expression>
              (((String) mdc.get("logType")) == "audit") &amp;&amp;
              (${auditLogOutputToFile:-true} == false)
            </expression>
          </evaluator>
          <OnMatch>DENY</OnMatch>
        </filter>
  `
during logging config processing this expression is transformed to java code, which returns ch.qos.logback.classic.Level and is then compiled by https://janino-compiler.github.io/janino/.

But, the issue is that Janino(or Logback itself, don't rememder exactly) gets classloader for compilation process using SomeClass.class.getClassLoader() - which returns SBT internal classloader with only JVM and internal SBT classes in it - and hence an Exception is thrown during compilation that ch.qos.logback.classic.Level class not found.
  
Spent a day to figure this out and find a fix for that. :)

So, as a fix - I'm just creating special logback config for all the SBT operations and set it via .sbtopts. It's a copy of logback.xml, but without the above expression. This also affects dev mode - so decided to keep it as close to original logging config as possible to have similar logging in dev vs prod.

sb-yb: Maybe say that: 
This config is used during compilation configured via .sbtOpts

managed/src/main/resources/swagger.json
anmalysh-yb: Huh, found the root cause finally and fixed.
Described the fix in SwaggerGenTest class comment - really tricky issue.

managed/src/main/resources/logback-sbt.xml
anmalysh-yb: Added comment.

managed/src/main/resources/swagger.json
anmalysh-yb: Adding issue and fix description here as well for reference:

`
    // So, this one is really tricky. The only play2.8 compatible swagger library I've found,
    // with the same functionality as our previous library, is
    // https://github.com/dwickern/swagger-play. It works fine, except that readOnly fields are
    // not serialized in the resulting swagger.json.
    // The reason is in this "workaround" for the issue with readOnly field they faced:
    // https://github.com/dwickern/swagger-play/commit/b04e4a2e145a277592d29c09d482ccb92138fa14#diff-a74181fef5ee291b8ecaf4773e3e4df1e335e0f0f3640b51f60298e00a143ba2
    // Basically, in marks AbstractProperty.readOnly() method with @JsonIgnore annotation.
    // At the same time - jackson-module-scala adds ScalaAnnotationIntrospector
    // to Swagger ObjectMapper, which makes it treat readOnly() method as getter for readOnly field.
    // As a result - jackson finds all getters/setters for readOnly field across the whole type
    // hierarchy, and treats this field as @JsonIgnore as a result of that.
    //
    // Tried to undo the above "workaround" - but this really causes InvalidDefinitionException
    // during serialization. Decided to not dig into it further to find a better fix for this
    // serialization cycle issue and instead replaced Swagger ObjectMapper with our default mapper
    // with ReferenceSerializationConfigurer.serializeAsComputedRef, which swagger itself adds by
    // default.
`

anmalysh-yb:
Already landed

Issue #16659, opened 2023-04-03
[Docs] yb-admin - add_transaction_tablet options are wrong


srinivasa-vasu:
### Description

`yb-admin -master_addresses <master-addresses> add_transaction_tablet <keyspace> <table_name>` - The arguments to the `add_transaction_tablet` operation is not correct. It requires `table_id` instead of `keyspace` and `table_name`.

```
yb-admin -master_addresses 127.0.0.2:7100 add_transaction_tablet
Error running add_transaction_tablet: Invalid argument (yb/tools/yb-admin_cli.cc:78): Invalid arguments for operation
Usage: yb-admin add_transaction_tablet  <table_id>
```

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

PR #16660, opened 2023-04-03
[docs] update yb-admin add_transaction_tablet ops argument

srinivasa-vasu:
`yb-admin -master_addresses <master-addresses> add_transaction_tablet <keyspace> <table_name>` - The arguments to the `add_transaction_tablet` operation is not correct. It requires `table_id` instead of `keyspace` and `table_name`.

Updated the doc to reflect `table_id`

PR for the issue id #16659


aishwarya24:
This has been fixed here: https://github.com/yugabyte/yugabyte-db/pull/17097

PR #16764, opened 2023-04-07
Clarify Peer Cert management docs

sb-yb:
We need to clarify that both active and standby root CA certs have to be added on the active instance.

### DIFF
--- a/docs/content/preview/yugabyte-platform/administer-yugabyte-platform/high-availability.md
+++ b/docs/content/preview/yugabyte-platform/administer-yugabyte-platform/high-availability.md
@@ -79,7 +79,7 @@ After the active instance has been configured, you can configure one or more sta
 
 1. Click **Create**.
 
-1. Switch to the active instance, and then switch to **Instance Configuration**. Click **Add Instance**, enter the new standby instance's IP address or hostname, including the HTTP or HTTPS protocol prefix and port if you are not using the default of 80 or 443. If the HTTPS protocol is enabled, you can use the **Manage Peer Certificates** dialog to set up a root certificate of other instances.
+1. Switch to the active instance, and then switch to **Instance Configuration**. Click **Add Instance**, enter the new standby instance's IP address or hostname, including the HTTP or HTTPS protocol prefix and port if you are not using the default of 80 or 443. If the HTTPS protocol is enabled, you can use the **Manage Peer Certificates** dialog to set up a root CA certificate of all the instances. Make sure that on the active instance we have added root CA certs for the **active as well as all the standby**. 
 
 1. Click **Continue** on the **Add Standby Instance** dialog.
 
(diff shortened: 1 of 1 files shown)


docs/content/preview/yugabyte-platform/administer-yugabyte-platform/high-availability.md
```diff
@@ -79,7 +79,7 @@ After the active instance has been configured, you can configure one or more sta
 
 1. Click **Create**.
 
-1. Switch to the active instance, and then switch to **Instance Configuration**. Click **Add Instance**, enter the new standby instance's IP address or hostname, including the HTTP or HTTPS protocol prefix and port if you are not using the default of 80 or 443. If the HTTPS protocol is enabled, you can use the **Manage Peer Certificates** dialog to set up a root certificate of other instances.
+1. Switch to the active instance, and then switch to **Instance Configuration**. Click **Add Instance**, enter the new standby instance's IP address or hostname, including the HTTP or HTTPS protocol prefix and port if you are not using the default of 80 or 443. If the HTTPS protocol is enabled, you can use the **Manage Peer Certificates** dialog to set up a root CA certificate of all the instances. Make sure that on the active instance we have added root CA certs for the **active as well as all the standby**.
```
ddhodge, line 82: do these caveats also not apply to the equivalent step under Configure the active instance?

also, the wording is unclear - do you add one certificate for all instances ("set up a root CA certificate for all the instances"), or one per instance ("added root CA certs for the **active as well as all the standby**")?

ddhodge, line 82:
```suggestion
1. Switch to the active instance, and then switch to **Instance Configuration**. Click **Add Instance**, enter the new standby instance's IP address or hostname, including the HTTP or HTTPS protocol prefix and port if you are not using the default of 80 or 443.

1. If the HTTPS protocol is enabled, you can use the **Manage Peer Certificates** dialog to set up a root CA certificate for all the instances. Make sure that on the active instance you added root CA certificates for the active as well as all the standbys. 
```

chidmuthu:
I think really you just need to add the CA cert of the original active to the standby after it has been promoted. I think only the first peer certificate is used because I found that order of certs mattered.

docs/content/preview/yugabyte-platform/administer-yugabyte-platform/high-availability.md
sb-yb: If the instances are using self signed certificate then it will be separate certificate.  But many of our customers use enterprise CA that is common across their org. So all the instances will have same root CA certificate. You can still add it for each instance.
But describing all of this is getting into a tutorial of what is a root CA. So I'm not sure we want to give that much of detail.

sb-yb:
> I think really you just need to add the CA cert of the original active to the standby after it has been promoted. I think only the first peer certificate is used because I found that order of certs mattered.

That works but that means after first promote system goes into error state until you fix it by adding old primary's rootCA cert to the list. Hence less than desirable workflow.

sb-yb:
@ddhodge @chidmuthu sent phabricator diff instead:
https://phabricator.dev.yugabyte.com/D24339

Also improved the doc.

Issue #16883, opened 2023-04-17
[YCQL] BaseCQLTest.java setUpCqlClient method bug?


WangPingGang:
Jira Link: [DB-6222](https://yugabyte.atlassian.net/browse/DB-6222)
### Description

attemptsMade is always 0 and never change it value?
```
    int attemptsMade = 0;
    boolean waitSuccessful = false;

    while (attemptsMade < 30) {
      int numPeers = execute("SELECT peer FROM system.peers").all().size();
      if (numPeers >= expectedNumPeers) {
        waitSuccessful = true;
        break;
      }
      LOG.info("system.peers still contains only " + numPeers + " entries, waiting");
      Thread.sleep(1000);
    }
```

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-6222]: https://yugabyte.atlassian.net/browse/DB-6222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dr0pdb:
@WangPingGang Thanks for reporting this!

Yes, it looks like a bug. If you're interested, please open a PR which fixes this by incrementing the value inside the for loop.

PR #16886, opened 2023-04-17
[#16883][YCQL]Fix java unit test setUpCqlClient maybe endless loop

WangPingGang:
attemptsMade is not change in loop.
So add attemptsMade++;
```
    while (attemptsMade < 30) {
      int numPeers = execute("SELECT peer FROM system.peers").all().size();
      if (numPeers >= expectedNumPeers) {
        waitSuccessful = true;
        break;
      }
      LOG.info("system.peers still contains only " + numPeers + " entries, waiting");
      Thread.sleep(1000);
    }
```


dr0pdb (approved):
Thank you!

PR #16896, opened 2023-04-18
use `on conflict` in the examples

premkumr:
- In the On Conflict page, we missed using the clause in the examples.


ddhodge (approved):
LGTM

PR #16951, opened 2023-04-20
[docs] [yba] added aws graviton instance type

aishwarya24:
docs for https://yugabyte.atlassian.net/browse/PLAT-2651
@netlify /preview/yugabyte-platform/configure-yugabyte-platform/supported-os-and-arch/


chidmuthu:
@iSignal isn't that only enabled with the new UI? When will that become default?

aishwarya24:
@iSignal @chidmuthu - I think this needs to be covered as part of the configure provider PR : https://github.com/yugabyte/yugabyte-db/pull/17036/files#r1178173100.

ddhodge:
@iSignal @aishwarya24 The page currently says 
"YugabyteDB Anywhere supports both x86 and ARM (aarch64) CPU architectures. [...] If you plan to deploy YugabyteDB on AWS Graviton-based EC2 instances, use a custom AMI certified for 64-bit ARM (arm64) architecture."
What would need to change here?

iSignal:
> @iSignal @aishwarya24 The page currently says "YugabyteDB Anywhere supports both x86 and ARM (aarch64) CPU architectures. [...] If you plan to deploy YugabyteDB on AWS Graviton-based EC2 instances, use a custom AMI certified for 64-bit ARM (arm64) architecture." What would need to change here?

As you can see from the screenshot above, the user can now just check "Default Aarch64" instead of requiring a custom AMI for aarch64 architecture. We should update the docs to suggest both these options.

As @chidmuthu pointed out, this UI will go live in 2.18 and in the next 2.19 release, so we probably want to match the timing to the release.

ddhodge (approved):

PR #17193, opened 2023-05-05
Relax tablet-splitting restrictions for XCluster

lingamsandeep:
Relax the tablet-splitting restriction with XCluster as it is supported for all releases 2.14 and above.


aishwarya24 (approved):
@lingamsandeep LGTM . I have also applied these changes till 2.14 in docs.

PR #17224, opened 2023-05-08
[docs] updated version number for workload generator

aishwarya24:


aishwarya24:
This is being fixed in #17528

PR #17392, opened 2023-05-17
[docs][CDCSDK] Moving GFlag under correct yb-master reference

vaibhav-yb:
A GFlag `enable_delete_truncate_cdcsdk_table` was incorrectly listed under `yb-tserver` while it is a `yb-master` flag. This PR fixes the listing of the flag's information.


ddhodge (approved):
LGTM

PR #17320, opened 2023-05-12
[docs] update MCS note and EA tags

aishwarya24:
Summary
- Added Early access tags for YBA support for [MCS](https://deploy-preview-17320--infallible-bardeen-164bc9.netlify.app/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/kubernetes/#configure-kubernetes-multi-cluster-environment) and [CDC](https://deploy-preview-17320--infallible-bardeen-164bc9.netlify.app/preview/explore/change-data-capture/). 
- addressed https://yugabyte.atlassian.net/browse/IDEA-46?atlOrigin=eyJpIjoiNjM4NTY0NjE5ZWQ2NDFhZmI5Mjg4MTcxMmNkMGQ5ZGYiLCJwIjoiamlyYS1zbGFjay1pbnQifQ
- Note to self - Todo- update EA link for Performance advisor along with CD and MCS once Dwight's release page is merged.

PR #17569, opened 2023-05-26
[docs] Mark CDC Early Access

ddhodge:
Add Early Access tag to CDC pages

@netlify /preview/explore/change-data-capture/


ddhodge:
duplicate of #17320

PR #17613, opened 2023-05-31
[yugabyted][UI] Read replica integration

Spikatrix:
- Nodes page shows read replica nodes with a tag alongside primary nodes
 - The settings page shows primary and read replica region/zone info
 - The overview and performance page now has region, node and cluster type (primary/readreplica) selectors to filter data


nchandrappa (approved):
LGTM.

PR #17666, opened 2023-06-02
[Doc][DocDB] improve architecture isolation levels page

mdbridge:
Changes:
* internal locking in DocDB section:
  * removed references to nonexistent pure-write transactions
  * summarized the conflict table in English
  * explained why write locks don't block other writers and what that means for concurrent blind writes
  * explained that snapshot lock is really a combination of the other two locks

* added section explicitly explaining locking granularities, including which levels we can take lock set

* added an explanation of what strong and weak locks are accomplishing and how they work
* English explanation of the giant conflict table

* removed text talking about how snapshot isolation transactions modifying different columns in the same row can run concurrently
  * Piyush says we are removing this functionality

Same changes to both preview and stable versions:
```
diff preview/architecture/transactions/isolation-levels.md stable/architecture/transactions/isolation-levels.md
7c7
<   preview:
---
>   stable:
```

### DIFF
--- a/docs/content/preview/architecture/transactions/isolation-levels.md
+++ b/docs/content/preview/architecture/transactions/isolation-levels.md
@@ -24,17 +24,19 @@ Transaction isolation level support differs between the YSQL and YCQL APIs:
 - [YSQL](../../../api/ysql/) supports Serializable, Snapshot, and Read Committed<sup>$</sup> isolation levels.
 - [YCQL](../../../api/ycql/dml_transaction/) supports only Snapshot isolation using the `BEGIN TRANSACTION` syntax.
 
-<sup>$</sup> Read Committed support is currently in [Beta](/preview/faq/general/#what-is-the-definition-of-the-beta-feature-tag). This level is supported only if the YB-TServer flag `yb_enable_read_committed_isolation` is set to `true`. By default, this flag is `false`, in which case the Read Committed isolation level of YugabyteDB's transactional layer falls back to the stricter Snapshot isolation (together with YSQL Read Committed and Read Uncommitted also in turn using the Snapshot isolation). The default isolation level for the YSQL API is essentially Snapshot because Read Committed, which is the YSQL API and PostgreSQL syntactic default, maps to Snapshot isolation.
+Similarly to PostgreSQL, you can specify Read Uncommitted for YSQL, but it behaves the same as Read Committed.
+
+<sup>$</sup> Read Committed support is currently in [Beta](/preview/faq/general/#what-is-the-definition-of-the-beta-feature-tag). This level is supported only if the YB-TServer flag `yb_enable_read_committed_isolation` is set to `true`. By default, this flag is `false`, in which case the Read Committed isolation level of YugabyteDB's transactional layer falls back to the stricter Snapshot isolation. The default isolation level for the YSQL API is essentially Snapshot because Read Committed, which is the YSQL API and PostgreSQL syntactic default, maps to Snapshot isolation.
 
 ## Internal locking in DocDB
 
 In order to support the three isolation levels, the lock manager internally supports the following three types of locks:
 
-- Snapshot isolation write lock is taken by a snapshot isolation (and also read committed) transaction on values that it modifies.
-
 - Serializable read lock is taken by serializable transactions on values that they read in order to guarantee they are not modified until the transaction commits.
 
 - Serializable write lock is taken by serializable transactions on values they write.
+
+- Snapshot isolation write lock is taken by a snapshot isolation (and also read committed) transaction on values that it modifies.
 
 The following matrix shows conflicts between these types of locks at a high level:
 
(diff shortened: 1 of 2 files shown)


docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+```
+
+then we could lock at any of the following:
+- the entire tablet
+- all rows having h1=2, h2=3
+- all rows having h1=2, h2=3, r1=4
+- the row having h1=2, h2=3, r1=4, r2=5
+- column v of the row having h1=2, h2=3, r1=4, r2=5
+
+With YCQL, granularities exist below the column level; for example only one key of a column map can be locked.
```
ddhodge, line 92:
```suggestion
With YCQL, granularities exist below the column level; for example, only one key of a column map can be locked.
```

docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+
+## Efficiently detecting conflicts between locks of different granularities
+
+The straightforward way to handle locks of different granularities would be to have a map from location to lock types.  However, this is too inefficient for detecting conflicts: attempting, for example, to add a lock at the tablet level would require checking for locks at every row and column in that tablet.
+
+To make conflict detection efficient, YugabyteDB stores extra information at each location about the locks at lower granularities below it.  In particular, it takes a normal lock at the original granularity and weaker versions of that lock at all the granularities that enclose the original granularity.  We call the normal locks _strong_ locks and the weaker variants _weak_ locks.
```
ddhodge, line 98:
```suggestion
To make conflict detection efficient, YugabyteDB stores extra information at each location about the locks at granularities below it.  In particular, it takes a normal lock at the original granularity and weaker versions of that lock at all the granularities that enclose the original granularity. The normal locks are called _strong_ locks and the weaker variants _weak_ locks.
```

docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+
+As an example, pretend we have only tablet- and row-level granularities. To take a serializable write lock at the row level, we would take a strong write lock at the row level and a weak write lock at the tablet level.  To take a serializable read lock at the tablet level, we would just take a strong read lock at the tablet level.
```
ddhodge, line 100:
```suggestion
As an example, suppose you have only tablet- and row-level granularities. To take a serializable write lock at the row level, you would take a strong write lock at the row level and a weak write lock at the tablet level.  To take a serializable read lock at the tablet level, you would just take a strong read lock at the tablet level.
```

docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+
+By using the following conflict rules, we can decide if two original locks would conflict based only on whether or not their strong/weak locks at any location would conflict:
```
ddhodge, line 102:
```suggestion
Using the following conflict rules, you can decide if two original locks would conflict based only on whether or not their strong/weak locks at any location would conflict:
```

docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+
+- two strong locks conflict iff they would've conflicted as normal locks
```
ddhodge, line 104:
```suggestion
- two strong locks conflict if they would've conflicted as normal locks
```

docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+- two weak locks never conflict
+- a strong lock conflicts with a weak lock iff they would've conflicted as normal locks
```
ddhodge, line 106:
```suggestion
- a strong lock conflicts with a weak lock if they would've conflicted as normal locks
```

docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+
+That is, for each location that would have two locks, would they conflict under the above rules?  There is no need to enumerate the sublocations of any location.
+
+In our example, we would detect a conflict at the tablet level between the strong read lock and the weak write lock.  If we had instead tried to take the serializable read lock at the row level then there would've been no conflict at the tablet level (both locks there are weak) but there might have been at the row level if the two locks were taken on the same row.
```
ddhodge, line 110:
```suggestion
In the example, you would detect a conflict at the tablet level between the strong read lock and the weak write lock.  If you had instead tried to take the serializable read lock at the row level, then there would've been no conflict at the tablet level (both locks there are weak) but there might have been at the row level if the two locks were taken on the same row.
```

docs/content/preview/architecture/transactions/isolation-levels.md
```diff
@@ -67,11 +67,49 @@ The following matrix shows conflicts between locks of different types at a high
...
+
+When we include the strong/weak distinction, the full conflict matrix becomes:
```
ddhodge, line 112:
```suggestion
Including the strong/weak distinction, the full conflict matrix becomes:
```

ddhodge (approved):

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: :+1:

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: style guide says to prefer active voice

mdbridge: good: removing "lower", however

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: The current version reads better to me because the connotations are that  we are temporarily changing the system.  Somehow I don't feel "you" has permission to do that.

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: dumping the "by",  but I think we is better here because it includes the system unlike "you".

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: that was not a typo,  but I will spell it out to avoid confusion (if and only if)

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: ditto

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: made changes except left it as we because of the including the system

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: :+1:

docs/content/preview/architecture/transactions/isolation-levels.md
ddhodge: Style guide not mention we don't use we, I forget. But we don't.

docs/content/preview/architecture/transactions/isolation-levels.md
ddhodge: What's really doing the work here - the database? As noted, 'we' is typically avoided, but I'm not that fussed about it.

docs/content/preview/architecture/transactions/isolation-levels.md
ddhodge: another one for the style guide - spell out acronyms I've never heard of heh

docs/content/preview/architecture/transactions/isolation-levels.md
ddhodge: ditto

pkj415:
Listing my comments here, I find it easier this way than to use github's reviewing process:

```
1. remove "(together with YSQL Read Committed and Read Uncommitted also in turn using the Snapshot isolation)"
2. Add a note saying this - "Similar to PostgreSQL, Read Uncommitted on YSQL behaves the same as Read Committed"
3a. Can we reverse the ordering of description for the lock types as: serializable reads, serializable writes and snapshot writes?
4. "of a column map" -> "of a column of map data type"?
5. "granularities below it" -> "granularities coarser than it"
6. "To make conflict detection efficient, YugabyteDB stores extra information at each location about the locks at granularities below it" -> the word "location" seems a bit weird here. How about this -

   i) In the section "Locking granularities", we slip in the words "lockable object" (maybe by saying "... then we could lock at any of the following objects" or ".... these are the possible lockable objects").
   ii) "a map from location to lock types." -> "a map from locked object to lock type"
   iii) "To make conflict detection efficient, YugabyteDB stores extra information at each location about the locks at granularities below it." -> "To make conflict detection efficient, YugabyteDB takes extra locks on all coarser granularities of a locked object."
   iv) "Using the following conflict rules, we can decide if two original locks would conflict based only on whether or not their strong/weak locks at any location would conflict:" -> "Using the following conflict rules, we can detect for conflicts by just checking lock types on the same object instead of checking all sub-objects too:"
   v) "two strong locks conflict if and only if they would've conflicted as normal locks": both normal and strong mean the same as mentioned 2 paragraphs earlier, so this statement seems to give no info, maybe we say that they conflict as per the matrix in "Internal locking in DocDB"?
   vi) "If we had instead tried to take the serializable read lock at the row level, then there would've been no conflict at the tablet level (both locks there are weak)" --> also because both would be read locks right? So not really useful?
   vii) "but there might have been at the row level if the two locks were taken on the same row." --> you mean there might have been a conflict if a read lock was taken on one row and write on another?

7. Maybe now, or in another PR, we can add a section with a table that maps INSERT, UPDATE, DELETE and the 4 row-level lock statements to the finer lock types. It will also need explaining that for to implement the 4 row-level lock statements, we allow directly taking a weak mode on an "object" and take "weak" modes on all coarser granularities of that object.
```

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: :+1:

docs/content/preview/architecture/transactions/isolation-levels.md
mdbridge: technically an abbreviation and an extremely common one in math at that but who knows what our readers are familiar with

pkj415 (approved):
LGTM!

docs/content/preview/architecture/transactions/isolation-levels.md
ddhodge: is no abbr. not in the style guide? no etc, eg, ie, etc. spell it all out thx!

mdbridge:
talked with @pkj415  and plan to address his issue 7 in a successor pull request

PR #17670, opened 2023-06-02
[Doc][DocDB] Fix unavailability window in documentation

mdbridge:
Documentation currently says two seconds, which is wrong.  It should say three seconds.

Fixing both preview and stable.


Huqicheng (approved):
lgtm

ddhodge (approved):

PR #17928, opened 2023-06-26
[WIP] Set PGGate AUH events in PGPROC 

vrajat:
Major work to be done:
* Add metadata for events
* Add support ofr YB PGGate events in pgstat.c so that it can print the right event names.


hbhanawat (approved):
ybcstat_get_wait_event_component is another function that we might need. We can take it in another PR.

PR #17952, opened 2023-06-27
[yugabyted][UI] Alerts and activities integration

Spikatrix:


nchandrappa:
Changes required to the Activities page,

- In the in-progress activities section, we need to add a progress bar for displaying the percentage completion based on the total tuples and completed tuples metrics.
-  For the Activities page UX, we need to add a white box similar to the setting pages to align with the overall design.
- On the alerts page, we need to add an info icon explaining the alert details for each configurable alert.

nchandrappa:
It has been tested and works according to design.

PR #17978, opened 2023-06-28
Fetch TServer UUID and send AUH Metadata along with the Perform protobuf

abhinab-yb:
This diff fetches TServer UUID and stores it in the pg_session object so that it can be used to fill up the outgoing rpcs and also be accessible from PG.

Also, another field is added to the PgPerformOptionsPB called AUHMetadataPB, which is required for the necessary data to be sent from PG to TServer.

### DIFF
--- a/src/postgres/src/backend/tcop/postgres.c
+++ b/src/postgres/src/backend/tcop/postgres.c
@@ -44,6 +44,7 @@
 #include "commands/async.h"
 #include "commands/portalcmds.h"
 #include "commands/prepare.h"
+#include "common/ip.h"
 #include "executor/spi.h"
 #include "jit/jit.h"
 #include "libpq/libpq.h"
@@ -73,6 +74,7 @@
 #include "tcop/pquery.h"
 #include "tcop/tcopprot.h"
 #include "tcop/utility.h"
+#include "utils/builtins.h"
 #include "utils/inval.h"
 #include "utils/relcache.h"
 #include "utils/catcache.h"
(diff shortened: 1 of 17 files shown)


src/yb/client/client.cc
```diff
@@ -1065,6 +1065,13 @@ Status YBClient::GetYsqlCatalogMasterVersion(uint64_t *ysql_catalog_version) {
   return Status::OK();
 }
 
+Status YBClient::GetTServerUUID(std::string &node_uuid) {
```
amitanandaiyer, line 1068: Generally, out params should be passed as a ptr, instead of a reference.

better yet, just return the uuid.

abhinab-yb: Updated

src/postgres/src/backend/tcop/postgres.c
```diff
@@ -5129,6 +5139,73 @@ PostgresMain(int argc, char *argv[],
...
+			}
+			else if (beentry->st_clientaddr.addr.ss_family == AF_UNIX)
+			{
+					/*
+						* Unix sockets always reports NULL for host and -1 for
+						* port, so it's possible to tell the difference to
+						* connections we have no permissions to view, or with
+						* errors.
+						*/
+					ereport(LOG, (errmsg("PG client ip and port not found")));
```
vrajat, line 5194: Different error messages will help to figure out which if condition or fn failed. Can you log different strings in the ereport mentioning what went wrong?

src/postgres/src/backend/tcop/postgres.c
```diff
@@ -5129,6 +5139,73 @@ PostgresMain(int argc, char *argv[],
...
+			{
+				/* Unknown address type, should never happen */
+				ereport(LOG, (errmsg("PG client ip and port not found")));
+			}
+		}
+
+		if (IsYugaByteEnabled())
+			HandleYBStatus(YBCPgSetAUHMetadata(remote_host, atoi(remote_port)));
+		else
+		 	ereport(LOG, (errmsg("yugabyte not enabled yet")));
```
vrajat, line 5206: Is this required? I think there are multiple places where the code moves on silently if not run as YB

vrajat (approved):

PR #18027, opened 2023-06-30
[WIP] APIs to fetch TServer AUH Data

abhinab-yb:


vrajat (approved):

src/yb/client/client-internal.cc
```diff
@@ -2621,6 +2622,37 @@ Result<bool> YBClient::Data::CheckIfPitrActive(CoarseTimePoint deadline) {
   return resp.is_pitr_active();
 }
 
+std::vector<std::string> YBClient::Data::ActiveUniverseHistory() {
```
amitanandaiyer, line 2625: What is the use-case for this function?

src/yb/tserver/pg_client.proto
```diff
@@ -673,3 +675,48 @@ message PgCancelTransactionRequestPB {
...
+  // Not used for "connection setup" calls.
+  RemoteMethodPB remote_method = 2;
+
+  // Propagate the timeout as specified by the user. Note that, since there is some
+  // transit time between the client and server, if you wait exactly this amount of
+  // time and then respond, you are likely to cause a timeout on the client.
+  uint32 timeout_millis = 3;
+}
+
+message RpcCallInProgressPB {
```
amitanandaiyer, line 710: Where is this used?

Is this cut-pasted from somewhere? can't we just include and use those PBs?

src/yb/client/tablet_server.h
```diff
@@ -83,5 +83,16 @@ struct YBTabletServerPlacementInfo {
   }
 };
 
+struct YBActiveUniverseHistoryInfo {
+  std::string wait_state;
+
+  template <class PB>
+  static YBActiveUniverseHistoryInfo FromPB(const PB& pb) {
+    return YBActiveUniverseHistoryInfo {
+      .wait_state = pb,
```
amitanandaiyer, line 92: Generally, FromPB/ToPB is used for searializing/desearilzing from PB to a class and vice-versa.

wait_state is a string, how does it get assigned to a PB?

src/yb/rocksdb/util/thread_posix.cc
```diff
@@ -238,4 +238,15 @@ int ThreadPool::UnSchedule(void* arg) {
   return count;
 }
 
+std::vector<std::string> ThreadPool::GetThreadpoolWaitStates() {
+  std::vector<std::string> res;
+  PthreadCall("lock", pthread_mutex_lock(&mu_));
```
amitanandaiyer, line 243: What is this lock/unlock protecting?

src/yb/client/tablet_server.h
abhinab-yb: Right now, the wait_state in RpcCallInProgressPB is just a string, so the PB (in FromPB) is just a std::string. But I think this will change in the future and we will have something like

`WaitStatePB wait_state = 8`

In that case, this FromPB can be modified to deserialize WaitStatePB in the future.

src/yb/client/client-internal.cc
abhinab-yb: I followed the other examples and they seemed to call into client-internal, so I followed the same pattern. But you are right, this can be done from client.cc itself.

src/yb/rocksdb/util/thread_posix.cc
abhinab-yb: This lock/unlock should protect `bg_wait_states_` similar to how they protect `bgthreads_`. I added these when the updates were also being protected by the same lock, but now the updates are not protected (in this PR), I think it's safe to be removed.

abhinab-yb: This lock also protects bg_wait_states_ being inserted into or removed from.

src/yb/client/client-internal.cc
```diff
@@ -81,6 +81,7 @@
 #include "yb/rpc/proxy.h"
 #include "yb/rpc/rpc.h"
 #include "yb/rpc/rpc_controller.h"
+#include "yb/rpc/rpc_introspection.pb.h"
```
hbhanawat, line 84: Is this needed?

src/yb/client/client-internal.h
```diff
@@ -49,6 +49,7 @@
 
 #include "yb/rpc/rpc_fwd.h"
 #include "yb/rpc/rpc.h"
+#include "yb/rpc/rpc_introspection.pb.h"
```
hbhanawat, line 52: is this needed?

amitanandaiyer (approved):

src/yb/client/client-internal.cc
abhinab-yb: No, removing

src/yb/client/client-internal.h
abhinab-yb: No, removing

PR #18046, opened 2023-07-02
RocksDB Background Flush and Compaction AUH Instrumentation

abhinab-yb:


src/yb/rocksdb/db/db_impl.cc
```diff
@@ -3285,19 +3285,23 @@ void DBImpl::SchedulePendingCompaction(ColumnFamilyData* cfd) {
   TEST_SYNC_POINT("DBImpl::SchedulePendingCompaction:Done");
 }
 
-void DBImpl::BGWorkFlush(void* db) {
+void DBImpl::BGWorkFlush(void* db, yb::util::WaitStateInfoPtr wait_state) {
```
amitanandaiyer, line 3288: Where are we actually using this call, with 2 params? I can't find any caller for this.

Also wondering why we need to pass in the parameter and set the status here -- wouldn't the same be achieved by setting Wait_state at the start/end of BackgroundCallFlush method?

Similarly for BackgroundCallCompaction.

abhinab-yb: The caller is in thread_posix.cc line 136.

Since there are multiple threads which can do flush and compaction, a single wait_state object won't be enough. We need a wait_state object per thread, and a way to access the wait states of all the threads. I found the best place for that to be in the class ThreadPool (in thread_posix.h).

There are 2 ways to have this wait_state object in the thread pool, one is to go from BGWorkFlush to the thread pool to create the wait state object, this won't change the function signature, but will require locks when we are creating and destroying this wait state object. And we have to create and destroy wait state object every time flush and compaction runs.

The other way is to create one wait state object per thread, that way we don't have to acquire any extra locks, but doing this requires us to somehow pass this wait state object from the thread pool to BGWorkFlush, that's why the function signature changes. After completion of the job, we just have to update the status to unused.

amitanandaiyer: We'll probably need some kind of locking ... to ensure that DumpRpcs get the latest state. But, keeping that aside, even with having the vector, this seems like something that can be avoided.

Before calling
```
133: (*function)(arg);
```
we can call the appropriate `ADOPT_WAIT_STATE(...);`

and the functions in Compact/Flush can update the "ThreadLocal" wait state using SET_WAIT_STATE in the compaction/flush function calls.

src/yb/rocksdb/db/db_impl.cc
```diff
@@ -3431,6 +3435,9 @@ void DBImpl::BackgroundCallFlush(ColumnFamilyData* cfd) {
   bool made_progress = false;
   JobContext job_context(next_job_id_.fetch_add(1), true);
 
+  SET_WAIT_STATE_METADATA(yb::util::AUHMetadata{.request_id = std::to_string(next_job_id_.load())});
```
amitanandaiyer, line 3438: Instead of replacing the whole AUHMetadata, we could just  update the request_id accordingly:

`yb::util::WaitStateInfo::CurrentWaitState().metadata().request_id = ...`

amitanandaiyer (approved):
accepted with comments.

PR #18072, opened 2023-07-04
Circular buffer size flag: yb_auh.circular_buf_size_kb 

hbhanawat:
- Changed the flag to be circular buffer size in kb rather than the count of items in circular buffer
- Added remaining parameters to the circular buffer and to the auh view.


hbhanawat:
merging it after discussing it with @vrajat and @amitanandaiyer

PR #18109, opened 2023-07-06
[Docs] Highlight output code blocks

samiahmedsiddiqui:


premkumr (approved):
LGTM.

We should add an example for this in the contribute section..

samiahmedsiddiqui:
@ddhodge Can you please add an example for this.

PR #18113, opened 2023-07-06
Fix query_id from PG to TServer and add flush and compaction wait event class

abhinab-yb:
pg_stat_statements doesn't store the query id in the `Query` object in case of utility statements. This PR fixes that and also setting the query_id correctly in AUH metadata.

A new wait event class is also introduced for the background rocksdb flush and compaction workers.

### DIFF
--- a/src/postgres/src/backend/parser/analyze.c
+++ b/src/postgres/src/backend/parser/analyze.c
@@ -130,7 +130,10 @@ parse_analyze(RawStmt *parseTree, const char *sourceText,
 		(*post_parse_analyze_hook) (pstate, query);
 
 	if (IsYugaByteEnabled() && query->queryId)
+	{
 		YBCSetQueryId(query->queryId);
+		MyProc->queryid = query->queryId;
+	}
 
 	free_parsestate(pstate);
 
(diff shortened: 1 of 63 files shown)


src/postgres/src/backend/parser/analyze.c
```diff
@@ -156,12 +159,6 @@ parse_analyze_varparams(RawStmt *parseTree, const char *sourceText,
 
 	query = transformTopLevelStmt(pstate, parseTree);
 
-	if (IsYugaByteEnabled())
```
vrajat, line 159: Why was this code deleted? If you have moved it somewhere else, can you also set queryid in `MyProc` ?

src/yb/yql/pggate/pg_session.cc
```diff
@@ -668,7 +668,7 @@ Result<PerformFuture> PgSession::Perform(BufferableOperations&& ops, PerformOpti
       caching_info.mutable_lifetime_threshold_ms()->set_value(*cache_options.lifetime_threshold_ms);
     }
   }
-
+
```
vrajat, line 671: nit: Whitespace change

vrajat (changes requested):
Can you add back `MyProc->queryid = ...` and also add it to the new places where queryid is set.

amitanandaiyer (approved):
lgtm

PR #18135, opened 2023-07-07
Update AUH data format

abhinab-yb:


src/yb/client/tablet_server.h
```diff
@@ -84,12 +85,18 @@ struct YBTabletServerPlacementInfo {
 };
 
 struct YBActiveUniverseHistoryInfo {
```
amitanandaiyer, line 87: Looks very Similar to wait_state.h/WaitStateInfo.

Why do we need this separately here?

src/yb/rpc/yb_rpc.cc
```diff
@@ -335,11 +335,7 @@ bool YBInboundCall::DumpPB(const DumpRunningRpcsRequestPB& req,
     auto wait_state = this->wait_state();
     if (wait_state) {
       // TBD: Add WaitStateInfoPB to the response instead of the string.
```
amitanandaiyer, line 337: can get rid of the TBD

src/yb/client/client.cc
```diff
@@ -2211,7 +2211,7 @@ Result<bool> YBClient::CheckIfPitrActive() {
   return data_->CheckIfPitrActive(deadline);
 }
 
-std::vector<std::string> YBClient::ActiveUniverseHistory() {
+std::vector<WaitStateInfoPB> YBClient::ActiveUniverseHistory() {
```
amitanandaiyer, line 2214: Don't have to fix this here.

But, it may be better to pass in the response/pb and have the function  "add" WaitStateInfoPBs to the passed in PB, than construct a vector of results, then the caller adds it to the pb.

Less memory allocations and copying of the auh-data.

amitanandaiyer (approved):
looks good on the docdb side. 
May want to get an accept from hemanth/rajat for the pgstatfunc changes.

PR #18077, opened 2023-07-04
[docs][ybm] Update AWS PSE CLI instructions

ddhodge:
Update AWS CLI instructions for creating an interface endpoint on AWS

@netlify /preview/yugabyte-cloud/cloud-basics/cloud-vpcs/managed-endpoint-aws/

PR #18208, opened 2023-07-13
Add information on AZoneID

shinji62:
Added information on Aws AZoneID, to avoid issues when choosing the AZ in the customer account.


shinji62:
Merge into #18077

PR #18232, opened 2023-07-13
Updated secondary index section in Explore

KavyaShivashankar:
1. Separated Secondary index docs for YSQL and YCQL into their respective tabs
2. Added links for YCQL docs in feature list for secondary index and secondary index with JSONB


aishwarya24:
LGTM. Thanks @KavyaShivashankar

PR #18444, opened 2023-07-27
Add method name in wait_event_aux for TServer RPCs

abhinab-yb:
There are some RPCs in the TServer which are not from Perform() like CreateTablet, FinishTransaction, Heartbeat, and even ActiveUniverseHistory. Right now, these don't have metadata like top_level_request_id, top_level_node_id, query_id.

This PR adds the method name in the aux_info so we can identify these.

### DIFF
--- a/src/postgres/contrib/yb_auh/yb_auh.c
+++ b/src/postgres/contrib/yb_auh/yb_auh.c
@@ -205,7 +205,7 @@ static void tserver_collect_samples(TimestampTz auh_sample_time)
   for (int i = 0; i < numrpcs; i++) {
     auh_entry_store(auh_sample_time, rpcs[i].metadata.top_level_request_id,
                     rpcs[i].metadata.current_request_id, rpcs[i].wait_status_code,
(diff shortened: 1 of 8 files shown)


src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -205,7 +205,7 @@ static void tserver_collect_samples(TimestampTz auh_sample_time)
   for (int i = 0; i < numrpcs; i++) {
     auh_entry_store(auh_sample_time, rpcs[i].metadata.top_level_request_id,
                     rpcs[i].metadata.current_request_id, rpcs[i].wait_status_code,
-                    rpcs[i].aux_info.tablet_id, rpcs[i].metadata.top_level_node_id,
+                    rpcs[i].aux_info.method, rpcs[i].metadata.top_level_node_id,
```
amitanandaiyer, line 208: I think it may be more useful to have tablet-id for slicing/dicing purposes than the method name, while exporting to the circular buffer/archival storage.

But feel free to use the method name for now, while we are debugging/understanding auh

amitanandaiyer (approved):
lgtm

PR #18453, opened 2023-07-27
Increase documented disk space requirement to match product checks

### LINKED ISSUE #18158: [Docs] Prerequisites - YBA Installer
### Description

In the hardware prerequisites for YBA Installer, it lists the minimum disk size as 200GB.
https://docs.yugabyte.com/stable/yugabyte-platform/install-yugabyte-platform/prerequisites/installer/

The preflight check fails if there are less than 215GB available:

```
INFO[2023-07-10T18:31:37Z] Running preflight check 'install does not exist'
...
INFO[2023-07-10T18:31:37Z] Running preflight check 'disk-availability'
INFO[2023-07-10T18:31:37Z] preflight disk-availability failed: Availabile disk space on volume 203 GB is less than minimum required 215 GB
...
#  Check name             Status   Error
1  disk-availability      Critical Availabile disk space on volume 203 GB is less than minimum required 215 GB
...
FATAL[2023-07-10T18:31:37Z] preflight failed
```

The docs should be updated with the correct minimum disk size. If installing on the system disk, approximately 5GB are consumed by the OS.

ionthegeek:
Fixes #18158

### DIFF
--- a/docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites/installer.md
+++ b/docs/content/preview/yugabyte-platform/install-yugabyte-platform/prerequisites/installer.md
@@ -68,7 +68,7 @@ A node running YugabyteDB Anywhere is expected to meet the following requirement
 
 - 4 cores
 - 8 GB memory
-- 200 GB disk space
+- 215 GB disk space
 
 ## Other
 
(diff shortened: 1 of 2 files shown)


ddhodge:
@ionthegeek No margin for error? Should be backported to all stable releases (2.16 etc), or is this a recent requirement?

ionthegeek:
It might be good to have some margin of error but I'm not sure what an appropriate margin of error would be, so I just made the docs match the check in the software.

YBA Installer is new in 2.18 so this does not apply to older releases.

PR #18487, opened 2023-08-01
Fetch AUH data from directly from the tablet server instead of YBClient

abhinab-yb:


src/yb/tserver/tablet_server.cc
```diff
@@ -1193,5 +1193,36 @@ std::vector<yb::util::WaitStateInfoPtr> TabletServer::GetThreadpoolWaitStates()
   return tablet_manager_->GetThreadpoolWaitStates();
 }
 
+std::vector<WaitStateInfoPB> TabletServer::ActiveUniverseHistory() const {
```
amitanandaiyer, line 1196: This could even go to the base class, so that master inherits the same.

but, i think it is ok for now -- as we aren't expecting calls to the master.

amitanandaiyer (approved):
accepted with comments.

src/yb/tserver/tablet_server.cc
abhinab-yb: It's defined in the base class too

PR #18545, opened 2023-08-03
Add AUH Metadata to select and insert CQL queries

abhinab-yb:
This PR adds AUH Metadata to select and insert queries in CQL.

The query id we have allocated is 8 bytes, while the CQL query id is 32 bytes, this PR takes the first 8 bytes out of this 32 bytes.


src/yb/util/wait_state.cc
```diff
@@ -22,6 +22,27 @@ namespace yb::util {
 // __thread WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 thread_local WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 
+void AUHMetadata::QueryIdHexToUInt(std::string queryId) {
```
amitanandaiyer, line 25: Is this supposed to return the UInt query_id?

amitanandaiyer: sounds like Update/Set QueryIdFromHex may be a better name.

also use query_id_str instead of CamelCase

abhinab-yb: Yeah, this will update the uint query_id in AUHMetadata. Updated the PR with your comments

src/yb/util/wait_state.cc
```diff
@@ -22,6 +22,27 @@ namespace yb::util {
 // __thread WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 thread_local WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 
+void AUHMetadata::SetQueryIdFromHex(std::string query_id_str) {
+  int size = std::min(16, (int)query_id_str.length());
```
amitanandaiyer, line 26: Do we need to re-invent this?

Why not just use
stoi/stoul/boost::lexical_cast()

https://www.geeksforgeeks.org/different-ways-to-convert-hex-string-to-integer-in-cpp-stl/

src/yb/util/wait_state.h
```diff
@@ -112,6 +112,8 @@ struct AUHMetadata {
   int64_t current_request_id = 0;
   std::string client_node_ip;
 
+  void SetQueryIdFromHex(std::string queryId);
```
amitanandaiyer, line 115: can become a normal SetQueryId(uint )
which can be called as 
SetQueryId(stoi("foo", 0, 16);

amitanandaiyer (approved):
one comment inline.

abhinab-yb:
Closing this and opening [#18678](https://github.com/yugabyte/yugabyte-db/pull/18678).

PR #18574, opened 2023-08-07
Metadata for internal flush/compactions

amitanandaiyer:
[tag internal rpc/compactions/flush as internal](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/a25e7152fd63d54d899ee5f8c9061571e2403e9e) 
[a25e715](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/a25e7152fd63d54d899ee5f8c9061571e2403e9e)
@amitanandaiyer
[more wait-classes](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/d6a8a7232e47ed83429b3c0dc4235daca6549333) 
[d6a8a72](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/d6a8a7232e47ed83429b3c0dc4235daca6549333)
@amitanandaiyer
[Set tablet aux in ConsensusServiceImpl](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/96e84899c152cc5f3408c63b0428971dd716f2d4)
[96e8489](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/96e84899c152cc5f3408c63b0428971dd716f2d4)
@amitanandaiyer
[Fix Created should actually be Handling.](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/41ee21e578dd22db5ecb0c371b27505ee96520de)

### DIFF
--- a/src/yb/consensus/raft_consensus.cc
+++ b/src/yb/consensus/raft_consensus.cc
@@ -1506,6 +1506,14 @@ Status RaftConsensus::Update(
     return STATUS(IllegalState, "Rejected: --TEST_follower_reject_update_consensus_requests "
                                 "is set to true.");
   }
+  auto wait_state = util::WaitStateInfo::CurrentWaitState();
+  if (wait_state) {
+    wait_state->UpdateMetadata(
+      yb::util::AUHMetadata{
+          .top_level_node_id = "consensus",
+          .top_level_request_id = {0, 1},
+          .query_id = -1});
+  }
   SET_WAIT_STATUS(util::WaitStateCode::Updating);
 
   TEST_PAUSE_IF_FLAG(TEST_follower_pause_update_consensus_requests);
--- a/src/yb/rocksdb/util/thread_posix.cc
+++ b/src/yb/rocksdb/util/thread_posix.cc
@@ -176,7 +176,11 @@ void ThreadPool::StartBGThreads() {
     CHECK_OK(yb::Thread::Create(
         category_name, std::move(thread_name),
         [this, tid]() { this->BGThread(tid); }, &thread));
-    bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>());
+    bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>(
+                                yb::util::AUHMetadata{
+                                    .top_level_node_id = "rocksdb",
+                                    .top_level_request_id = {0, 2},
+                                    .query_id = -2}));
     bgthreads_.push_back(thread);
   }
 }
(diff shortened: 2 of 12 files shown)


src/yb/consensus/raft_consensus.cc
```diff
@@ -1506,6 +1506,14 @@ Status RaftConsensus::Update(
...
                                 "is set to true.");
   }
+  auto wait_state = util::WaitStateInfo::CurrentWaitState();
+  if (wait_state) {
+    wait_state->UpdateMetadata(
+      yb::util::AUHMetadata{
+          .top_level_node_id = "consensus",
+          .top_level_request_id = {0, 1},
+          .query_id = -1});
+  }
```
abhinab-yb, line 1516: IMO top level node id should be the actual node id, and we can reserve some top level request ids for special kind of requests like these, e.g. `{0, 0}` will be for flush and compaction, `{0, 1}` for consensus.

Also, where are the client node ip and request id being set?

src/yb/rocksdb/util/thread_posix.cc
```diff
@@ -176,7 +176,11 @@ void ThreadPool::StartBGThreads() {
     CHECK_OK(yb::Thread::Create(
         category_name, std::move(thread_name),
         [this, tid]() { this->BGThread(tid); }, &thread));
-    bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>());
+    bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>(
+                                yb::util::AUHMetadata{
+                                    .top_level_node_id = "rocksdb",
+                                    .top_level_request_id = {0, 2},
+                                    .query_id = -2}));
```
abhinab-yb, line 183: same comment as above

abhinab-yb: Also since these metadata fields (along with client node ip) will be same for all the threads, can we just set them in the response PB instead of here?

src/yb/consensus/raft_consensus.cc
amitanandaiyer: Consensus ops are not specifically tied to any specific query. The consensus rpc's get merged for different queries, besides we don't actually persist the AUH/metadata into the WAL logs, so the context is lost when entering the consensus layer.

So I don't think top level node id per se makes a lot of sense -- the way we interpret it for other "queries".

Just trying to populate it with something, so that it stands out, as something non-empty.

src/yb/rocksdb/util/thread_posix.cc
amitanandaiyer: We are already maintaining WaitStateInfo per thread as bg_wait_states_ . Why leave it empty and handle this specially elsewhere? Better to update it to reflect the metadata for the items handled in these threads.

amitanandaiyer:
https://github.com/yugabyte/yugabyte-db/pull/18597 is the new one prepared against the merged yb_auh

PR #18575, opened 2023-08-07
Add support to fetch CQL Server AUH data

abhinab-yb:


src/yb/master/master_tserver.h
```diff
@@ -80,7 +80,9 @@ class MasterTabletServer : public tserver::TabletServerIf,
 
   std::vector<yb::util::WaitStateInfoPtr> GetThreadpoolWaitStates() const override;
 
-  std::vector<WaitStateInfoPB> ActiveUniverseHistory() const override;
+  void ActiveUniverseHistory(tserver::PgActiveUniverseHistoryResponsePB* resp) const override {}
+
+  void GetCQLServerMessenger(tserver::CQLServerMessenger messenger) override {};
```
hbhanawat, line 85: Do we need these changes?

src/yb/yql/pggate/util/ybc_stat.cpp
```diff
@@ -88,6 +88,9 @@ ybcstat_get_wait_event_type(uint32_t wait_event_info) {
         case YB_FLUSH_AND_COMPACTION:
             event_type = "Flush and Compaction";
             break;
+        case YB_CQL_WAIT_STATE:
```
hbhanawat, line 91: Should we have a separate wait event component for YCQL just like PG?
We can then have the following values
wait event component: YCQL
wait event Type: Query processing 
wait event:   Parse/Analyze/Execute

src/yb/yql/pggate/util/ybc_stat.cpp
```diff
@@ -88,6 +88,9 @@ ybcstat_get_wait_event_type(uint32_t wait_event_info) {
...
+            event_type = "CQL Server";
+            break;
```
hbhanawat, line 93: I don't see wait events for DMLRead and DMLWrite i.e. when CQLServer has sent an RPC to tserver.

src/yb/yql/cql/ql/ql_processor.cc
```diff
@@ -428,6 +431,7 @@ void QLProcessor::ExecuteAsync(const ParseTree& parse_tree, const StatementParam
       return;
     }
   }
+  SET_WAIT_STATUS(util::WaitStateCode::Execute);
```
hbhanawat, line 434: When is this wait state cleared?

src/yb/master/master_tserver.h
abhinab-yb: CQL server is not accessible from TServer, so the second function installs a callback to get the CQL Server callback. 

The first function, I think that makes it a little more efficient

src/yb/yql/pggate/util/ybc_stat.cpp
abhinab-yb: Makes sense, I'll update it

src/yb/yql/pggate/util/ybc_stat.cpp
abhinab-yb: Adding these

src/yb/master/master_tserver.h
hbhanawat: Didn't get it. These calls are in master. Is it because of inheritance that you need them here?

abhinab-yb: Yeah, these are also included in master because of inheritance

hbhanawat: The base class should have this function rather than all the inheriting classes.

abhinab-yb: The base class has only virtual functions, so there needs to be a no-op definition here. The other examples are similar too

PR #18582, opened 2023-08-07
Fetch CQL AUH data

abhinab-yb:


src/yb/master/master_tserver.h
```diff
@@ -80,7 +80,9 @@ class MasterTabletServer : public tserver::TabletServerIf,
 
   std::vector<yb::util::WaitStateInfoPtr> GetThreadpoolWaitStates() const override;
 
-  std::vector<WaitStateInfoPB> ActiveUniverseHistory() const override;
+  void ActiveUniverseHistory(tserver::PgActiveUniverseHistoryResponsePB* resp) const override {}
+
+  void GetCQLServerMessenger(tserver::CQLServerMessenger messenger) override {};
```
hbhanawat, line 85: This function should somehow be removed from here. It isn't a correct place. @amitanandaiyer can you give some ideas?

src/yb/rpc/messenger.h
```diff
@@ -50,6 +50,8 @@
 #include "yb/rpc/proxy_context.h"
 #include "yb/rpc/scheduler.h"
 
+#include "yb/tserver/pg_client.pb.h"
```
amitanandaiyer, line 53: not sure if we have dependencies from /rpc to /tserver.

we may create loops if we are not careful.

src/yb/yql/cql/ql/ql_processor.cc
```diff
@@ -428,7 +431,9 @@ void QLProcessor::ExecuteAsync(const ParseTree& parse_tree, const StatementParam
       return;
     }
   }
+  SET_WAIT_STATUS(util::WaitStateCode::Execute);
   executor_.ExecuteAsync(parse_tree, params, std::move(cb));
+  SET_WAIT_STATUS(util::WaitStateCode::ExecuteFinished);
```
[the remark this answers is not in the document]
amitanandaiyer: You could probably use SCOPED_ macros to set/reset to the old state.

We don't want to ever have no-state.  (so probably you meant reset to handling -- when you say clear).

Think of the states like a state machine. you'd move from one state to another.

Some of these are "CPU" states. Not necessarily "wait" states -- in terms of waiting.

src/yb/master/master_tserver.h
```diff
@@ -80,7 +80,9 @@ class MasterTabletServer : public tserver::TabletServerIf,
 
   std::vector<yb::util::WaitStateInfoPtr> GetThreadpoolWaitStates() const override;
 
-  std::vector<WaitStateInfoPB> ActiveUniverseHistory() const override;
+  void ActiveUniverseHistory(tserver::PgActiveUniverseHistoryResponsePB* resp) const override {}
```
amitanandaiyer, line 83: I think we should just add a very simple method to TabletServerIf like cql_server_messenger(); or cql_server() which returns null by default. In TServer, we'll have the CQL server register/update this to return it.

TServers/messenger shouldn't depend/know about what PB/request are created. Can have a method in /util which takes a messenger and populates the appropriate PB/response

src/yb/yql/cql/ql/exec/executor.cc
```diff
@@ -862,6 +863,7 @@ Status Executor::GetOffsetOrLimit(
 //   calls within the same process. These two different cases can be cleaned up later to avoid
 //   confusion.
 Status Executor::ExecPTNode(const PTSelectStmt *tnode, TnodeContext* tnode_context) {
+  SET_WAIT_STATUS(util::WaitStateCode::CQLRead);
```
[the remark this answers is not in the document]
abhinab-yb: The state will be updated to `ExecuteWaitingForCB`

src/yb/tserver/pg_client_service.cc
```diff
@@ -724,14 +725,28 @@ class PgClientServiceImpl::Impl {
...
+
   Status ActiveUniverseHistory(
       const PgActiveUniverseHistoryRequestPB& req, PgActiveUniverseHistoryResponsePB* resp,
       rpc::RpcContext* context) {
-    auto tserver_wait_states = tablet_server_.ActiveUniverseHistory();
-
-    for (auto wait_state : tserver_wait_states) {
-      resp->add_wait_states()->CopyFrom(wait_state);
-    }
+    GetRPCsWaitStates(resp, tablet_server_.GetMessenger(util::MessengerType::kTserver));
```
amitanandaiyer, line 748: Should we store the TServer/CQL wait states separately in the ResponsePB?

Seems like we are just merging them into one list.

src/yb/tserver/tablet_server.cc
```diff
@@ -1235,35 +1235,19 @@ std::vector<yb::util::WaitStateInfoPtr> TabletServer::GetThreadpoolWaitStates()
...
-
-  std::vector<WaitStateInfoPB> res;
-  
-  for (auto conns : dump_resp.inbound_connections()) {
-    for (auto call : conns.calls_in_flight()) {
-      if (call.has_header() && call.header().remote_method().method_name() != "ActiveUniverseHistory") {
-        res.push_back(call.wait_state());
-      }
-    }
+void TabletServer::GetCQLServerMessenger(CQLServerMessenger messenger) {
```
amitanandaiyer, line 1238: Should this be called Set instead of Get?

Also why is Messenger being sent by value? We should be storing a ptr, or a reference.

src/yb/tserver/pg_client_service.cc
abhinab-yb: We can separate them as tserver or cql by their wait states, so should we make still send them separately?

src/yb/tserver/tablet_server.cc
abhinab-yb: I followed the example of `RegisterCertificateReloader` which seemed to do it this way.

Updating the Get to Set

src/yb/tserver/pg_client_service.cc
amitanandaiyer: We can do this as a follow-up. But, say we get 1000 events from Server, but 2000 events from CQL.

If we decide to keep only 500 events per service/server. They'd need to be sampled accordingly. So better to keep them separate.

src/yb/tserver/tablet_server_interface.h
```diff
@@ -36,6 +36,7 @@ namespace tserver {
 
 using CertificateReloader = std::function<Status(void)>;
 using PgConfigReloader = std::function<Status(void)>;
+using CQLServerMessenger = std::function<rpc::Messenger*(void)>;
```
amitanandaiyer, line 39: This seems a convoluted way to get to messenger.

We could just hold/save the cql's messenger, instead of a ptr/callback.

amitanandaiyer (approved):

PR #18597, opened 2023-08-08
Metadata for internal flush/compactions (new)

amitanandaiyer:
[tag internal rpc/compactions/flush as internal](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/a25e7152fd63d54d899ee5f8c9061571e2403e9e)
[a25e715](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/a25e7152fd63d54d899ee5f8c9061571e2403e9e)
@amitanandaiyer
[more wait-classes](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/d6a8a7232e47ed83429b3c0dc4235daca6549333)
[d6a8a72](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/d6a8a7232e47ed83429b3c0dc4235daca6549333)
@amitanandaiyer
[Set tablet aux in ConsensusServiceImpl](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/96e84899c152cc5f3408c63b0428971dd716f2d4)
[96e8489](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/96e84899c152cc5f3408c63b0428971dd716f2d4)
@amitanandaiyer
[Fix Created should actually be Handling.](https://github.com/yugabyte/yugabyte-db/pull/18574/commits/41ee21e578dd22db5ecb0c371b27505ee96520de)

### DIFF
--- a/src/yb/rocksdb/util/thread_posix.cc
+++ b/src/yb/rocksdb/util/thread_posix.cc
@@ -178,7 +178,7 @@ void ThreadPool::StartBGThreads() {
         [this, tid]() { this->BGThread(tid); }, &thread));
     bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>(
                                 yb::util::AUHMetadata{
-                                    .top_level_node_id = "rocksdb",
+                                    .top_level_node_id = "localhost",
                                     .top_level_request_id = {0, 2},
                                     .query_id = -2}));
     bgthreads_.push_back(thread);
--- a/src/yb/rocksdb/util/thread_posix.cc
+++ b/src/yb/rocksdb/util/thread_posix.cc
@@ -176,7 +176,11 @@ void ThreadPool::StartBGThreads() {
     CHECK_OK(yb::Thread::Create(
         category_name, std::move(thread_name),
         [this, tid]() { this->BGThread(tid); }, &thread));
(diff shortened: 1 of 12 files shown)


src/yb/rocksdb/util/thread_posix.cc
```diff
@@ -176,7 +176,11 @@ void ThreadPool::StartBGThreads() {
     CHECK_OK(yb::Thread::Create(
         category_name, std::move(thread_name),
         [this, tid]() { this->BGThread(tid); }, &thread));
-    bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>());
+    bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>(
+                                yb::util::AUHMetadata{
+                                    .top_level_node_id = "localhost",
+                                    .top_level_request_id = {0, 2},
+                                    .query_id = -2}));
```
abhinab-yb, line 183: Re: discussion from previous PR

My initial thought was that these metadata fields are going to be the same for each thread, so instead of copying it for each thread, we can just populate it when we are sending this info to PG. But this will only make sense for std::string because we are not allocating memory. So now I think it's fine to update the metadata here.

abhinab-yb (approved):

PR #18667, opened 2023-08-11
[docs] [build] added  a devDependency "eslint-webpack-plugin"

aishwarya24:
With latest node updates to master and also hugo version update, after pulling the latest master changes , and running `npm start` locally, I got the following error: 
`[webpack-cli] Error: Cannot find module 'eslint-webpack-plugin'`

So, I installed `eslint-webpack-plugin` using the command `npm install --save-dev eslint-webpack-plugin`. 
The command installed and added the dependency entry in `package-lock.json` and `package.json`.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -22,7 +22,6 @@
         "eslint-config-xo-space": "0.34.0",
         "eslint-plugin-import": "2.28.0",
         "eslint-plugin-jsx-a11y": "6.7.1",
-        "eslint-webpack-plugin": "4.0.1",
         "hugo-algolia": "1.2.14",
         "node-sass": "9.0.0",
         "npm-run-all": "4.1.5",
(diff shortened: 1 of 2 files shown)


premkumr:
@aishwarya24 - I did not see this error when testing. are you merging this?

samiahmedsiddiqui:
@aishwarya24 `eslint-webpack-plugin` is already defined in the [package.json](https://github.com/yugabyte/yugabyte-db/blob/master/docs/package.json#L41). In actual, you need to run `npm install` command or `npm install eslint-webpack-plugin` command. If you merge this PR then it might break the build as for production site Netlify only install `dependencies` and your command has moved this package into `devDependencies` which works for the preview build but not for production build.

aishwarya24:
> @aishwarya24 `eslint-webpack-plugin` is already defined in the [package.json](https://github.com/yugabyte/yugabyte-db/blob/master/docs/package.json#L41). In actual, you need to run `npm install` command or `npm install eslint-webpack-plugin` command. If you merge this PR then it might break the build as for production site Netlify only install `dependencies` and your command has moved this package into `devDependencies` which works for the preview build but not for production build.

Thanks for this detailed explanation @samiahmedsiddiqui. Closing the PR.

PR #18671, opened 2023-08-11
[#19018] yb-admin get_xcluster_info should return valid JSON

kerneljake:
Currently `yb-admin get_xcluster_info` returns invalid JSON which makes parsing long output with tools like `jq` difficult.

Steps to reproduce

Before:
```
$ yb-admin get_xcluster_info | jq
parse error: Invalid numeric literal at line 1, column 63
```

After:
```
$ yb-admin get_xcluster_info | jq
{
  "version": 0,
  "xcluster_producer_registry": {},
  "consumer_registry": {}
}
```

### DIFF
--- a/src/yb/tools/yb-admin_client.cc
+++ b/src/yb/tools/yb-admin_client.cc
@@ -2105,8 +2105,8 @@ Status ClusterAdminClient::GetXClusterConfig() {
   MessageToJsonString(
       cluster_config.cluster_config().consumer_registry(), &consumer_registry_output);
   cout << Format(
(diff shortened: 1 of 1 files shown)


hari90 (changes requested):
Thank you for fixing this.
Can you please file a GH item for it as well?
And change the title to this format:

`[#<GH issue number>] xCluster: yb-admin get_xcluster_info should return valid JSON`

hari90 (approved):
:+1:

premkumr:
LGTM. Good fix. Thanks @kerneljake !

premkumr (approved):

PR #18690, opened 2023-08-15
[docs] Edit global apps and port to stable

ddhodge:
Edit the Global applications section in Develop and port the section to stable

@netlify /stable/develop/build-global-apps/


aishwarya24 (approved):
LGTM to suggestions. Apply the same to stable as well.

PR #18701, opened 2023-08-16
Fix compilation and assert failure

amitanandaiyer:


abhinab-yb (approved):

src/yb/util/wait_state.h
```diff
@@ -12,6 +12,7 @@
 //
 #pragma once
 
+#include <array>
```
hbhanawat, line 15: @amitanandaiyer is this change needed?

amitanandaiyer: Yes. Abhinab has used an array in this file.

The compiler on my laptop complains/requires this. Not sure why others don't complain.

PR #18704, opened 2023-08-16
Fix client node ip and top level node ip size

abhinab-yb:
This PR converts ipv4 addresses to 32 bit unsigned integers and top level node id to two 64 bit unsigned integer.


src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -436,9 +430,11 @@ pg_active_universe_history_internal(FunctionCallInfo fcinfo)
     else
       nulls[j++] = true;
 
-    // Top level node id
+    char top_level_node_id[33];
+    uint128_to_char(top_level_node_id, AUHEntryArray[i].top_level_node_id);
```
hbhanawat, line 434: Does this convert it to the tserver node id? At some point, I would like to add the capability to join this table with a tserver list.

abhinab-yb: Yes, the tserver node id is a 16 bytes hex, so it's broken into two 8 byte integers and propagated everywhere, this just converts these two 8 byte integers back to a single 16 byte hex

src/postgres/src/backend/utils/misc/pg_yb_utils.c
```diff
@@ -3575,30 +3575,98 @@ void YbSetIsBatchedExecution(bool value)
...
+	int digits = 0;
+	while (digit > 0)
+	{
+		digits++;
+		digit /= 10;
+	}
+	return digits;
+}
+
+void client_node_ip_to_string(char *client_node_ip,
```
amitanandaiyer, line 3627: Generally as per our coding guidelines, we'd want to have the input parameters first, then followed by output parameters.

So unless postgres requires a different convention, the output `char *client_node_ip` should become the 3rd parameter.

src/postgres/src/include/pg_yb_utils.h
```diff
@@ -906,11 +906,13 @@ OptSplit *YbGetSplitOptions(Relation rel);
 	} while (0)
 #endif
 
-void ProcSetNodeUUID(const char *);
+void ProcSetTopLevelNodeId(const uint64_t *);
 void ProcSetTopRequestId(const uint64_t *);
 
-void
-top_level_request_id_uint_to_char(char *top_level_request_id, const uint64_t top_level_request_id_uint[2]);
+void uint128_to_char(char *char_id, const uint64_t uint_id[2]);
```
amitanandaiyer, line 912: Fix order of input/output params. 

Input params should go before the output params in the order.

src/yb/consensus/raft_consensus.cc
```diff
@@ -1520,7 +1520,7 @@ Status RaftConsensus::Update(
   if (wait_state) {
     wait_state->UpdateMetadata(
       yb::util::AUHMetadata{
-          .top_level_node_id = "localhost",
+          .top_level_node_id = {0, 1},
```
amitanandaiyer, line 1523: does `{0, 1}` correspond to 127.0.0.1 ?

src/yb/tserver/pg_client.proto
```diff
@@ -275,7 +275,7 @@ message PgGetTServerUUIDRequestPB {
 
 message PgGetTServerUUIDResponsePB {
   AppStatusPB status = 1;
-  string node_uuid = 2;
+  repeated uint64 top_level_node_id = 2;
```
amitanandaiyer, line 278: How many times can this be repeated?

Consider using something like
`repeated bar arr    = 1 [length=10];`

src/yb/util/wait_state.cc
```diff
@@ -22,6 +22,32 @@ namespace yb::util {
 // __thread WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 thread_local WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 
+void AUHMetadata::set_client_node_ip(std::string &&endpoint) {
```
amitanandaiyer, line 25: Why do you need a RValue here?

Seems like a const & should suffice.

src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -47,8 +47,8 @@ typedef struct ybauhEntry {
   long request_id;
   uint32 wait_event;
   char wait_event_aux[16];
-  char top_level_node_id[16];
-  char client_node_ip[16];
+  uint64_t top_level_node_id[2];
```
amitanandaiyer, line 50: not a big fan of having uint64_t arrays.

conceptually find it simpler to use char/uint8 [] arrays of size 8/16.

can always get uint64_t foo = *(static_cast<uint64_t*>&bar[0]) to read the array as a unit64 number.

src/postgres/src/include/pg_yb_utils.h
amitanandaiyer: Even simpler would be to just return the output instead of passing in a pointer.

amitanandaiyer (commented):
few comments inline

src/postgres/src/include/pg_yb_utils.h
abhinab-yb: Returning the output would require to allocate memory using malloc/palloc and then deallocating later to avoid memory leak. Passing the pointer seems a simpler approach.

src/yb/consensus/raft_consensus.cc
abhinab-yb: Do you mean the client node ip ?

src/yb/tserver/pg_client.proto
abhinab-yb: There doesn't seem to be any such way to set the length of a repeated field

src/postgres/contrib/yb_auh/yb_auh.c
abhinab-yb: using uint8 arrays will make the code longer in places like this. Is this okay?

```
    bg_wait_states_.push_back(std::make_shared<yb::util::WaitStateInfo>(
                                yb::util::AUHMetadata{
                                    .top_level_node_id = {0, 1},
                                    .top_level_request_id = {0, 2},
                                    .query_id = -2}));
```

amitanandaiyer: it's ok for now.But I think it is cleaner to just track as bytes/chars.

We could make these constants in some file. But let's get to it later.

src/postgres/src/backend/utils/misc/pg_yb_utils.c
```diff
@@ -3575,30 +3575,98 @@ void YbSetIsBatchedExecution(bool value)
...
+		digits++;
+		digit /= 10;
+	}
+	return digits;
+}
+
+void client_node_ip_to_string(uint32 client_node_host,
+							  uint16 client_node_port,
+							  char *client_node_ip)
+{
```
amitanandaiyer, line 3630: Consider
` sprintf(client_node_ip, "%d.%d.%d.%d:%d", 
client_node_host >> 24,  255 & client_node_host >> 16, 255 & client_node_host >> 8, 255 & client_node_host, client_node_port);`

src/yb/client/client.cc
```diff
@@ -1081,11 +1081,20 @@ Status YBClient::GetYsqlCatalogMasterVersion(uint64_t *ysql_catalog_version) {
...
-    return data_->meta_cache_->local_tserver()->permanent_uuid();
+Result<std::vector<uint64_t>> YBClient::GetTServerUUID() {
+  vector<uint64_t> permanent_uuid(2);
 
-  return "";
+  if (data_->meta_cache_->local_tserver()) {
+    std::string tserver_uuid = data_->meta_cache_->local_tserver()->permanent_uuid();
+    if (!tserver_uuid.empty()) {
+      permanent_uuid[0] = std::stoull(tserver_uuid.substr(0, std::min(16, (int)tserver_uuid.size())), 0, 16);
+    }
```
amitanandaiyer, line 1091: else set to 0?

src/yb/client/client.cc
```diff
@@ -1081,11 +1081,20 @@ Status YBClient::GetYsqlCatalogMasterVersion(uint64_t *ysql_catalog_version) {
...
+    if (tserver_uuid.size() > 16) {
+      permanent_uuid[1] = std::stoull(tserver_uuid.substr(16, std::min(16, (int)tserver_uuid.size() - 16)), 0, 16);
+    }
+  }
```
amitanandaiyer, line 1095: same here?

src/yb/util/wait_state.cc
```diff
@@ -22,6 +22,32 @@ namespace yb::util {
 // __thread WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 thread_local WaitStateInfoPtr WaitStateInfo::threadlocal_wait_state_;
 
+void AUHMetadata::set_client_node_ip(const std::string &endpoint) {
+  client_node_host = 0;
+  client_node_port = 0;
+  int power = 1;
+  int index = (int)endpoint.size() - 1;
+  while (index >= 0 && endpoint[index] != ':') {
```
amitanandaiyer, line 30: prefer std existing functions to parse int for easier readability and conciseness.

amitanandaiyer (approved):

src/yb/client/client.cc
abhinab-yb: vector is initialized to 0 by default. I'll make it explicit

PR #18712, opened 2023-08-16
Set query-id for Catalog calls

amitanandaiyer:


hbhanawat (approved):
These changes look good. One question though - how do we differentiate between a catalog call vs dmlread/write? Do we have a separate wait event for them?

src/postgres/src/backend/access/index/genam.c
```diff
@@ -439,7 +439,16 @@ systable_getnext(SysScanDesc sysscan)
...
+		const int64 old_query_id = MyProc->queryid;
+		if (old_query_id == 0) {
+			const int64 kCatalogQueryId = -3;
+			YBCSetQueryId(kCatalogQueryId);
+		}
+		HeapTuple ret = ybc_systable_getnext(sysscan);
+		if (old_query_id == 0) {
+			YBCSetQueryId(0);
+		}
+		return ret;
```
abhinab-yb, line 451: There can be catalog requests after the query id is set. See this trace: http://10.150.1.75:16686/trace/09697fa78beea461028531060aaa06c1?uiFind=dd35d7791d4900f8

Should they be marked as -3 as well?

Also, query_id can be incorrectly set to 0 after the query_id has already been  calculated.

src/postgres/src/backend/utils/cache/relcache.c
```diff
@@ -5825,6 +5831,10 @@ RelationCacheInitializePhase3(void)
 		write_relcache_init_file(true);
 		write_relcache_init_file(false);
 	}
+	if (IsYugaByteEnabled() && old_query_id == 0)
+	{
+		YBCSetQueryId(0);
+	}
```
abhinab-yb, line 5837: Same as above comment, this can incorrectly set the query id to 0.

abhinab-yb (changes requested):

src/postgres/src/backend/utils/cache/relcache.c
amitanandaiyer: Can you expand on how? 

This code is run only if the existing query-id when entering this function is 0.

So in essence, we are only setting it to 0, if it were 0 when the control entered this function.

src/postgres/src/backend/access/index/genam.c
amitanandaiyer: `Also, query_id can be incorrectly set to 0 after the query_id has already been calculated.`
Don't think so. Unless the argument is that query_id is being set inside this function.

We are essentially resetting the QueryId to whatever it was earlier at line 442.

abhinab-yb: There are catalog requests which are part of a statement but for which the query_id has not been calculated yet. Similarly there are catalog requests which are made after query_id is calculated. Isn't it better for the query_id to be consistent across them?

src/postgres/src/backend/utils/cache/relcache.c
abhinab-yb: My bad, I missed line 5834.

src/postgres/src/backend/access/index/genam.c
amitanandaiyer: As discussed, I  think it is better to use the query-id for catalog requests -- when that is available, than use a random/fixed "-3". 

For the ones where we are making catalog-requests from places where we have no query-id, it is good to tag them with a different query-id so we can document it as a "known" query/rpc. Will help us catch any new/other catalog requests that may be made -- which we  might not have caught.

abhinab-yb: Agreed

abhinab-yb (approved):

amitanandaiyer:
> These changes look good. One question though - how do we differentiate between a catalog call vs dmlread/write? Do we have a separate wait event for them?

The query-id should be different. For catalog calls we'll have the kCatalogQueryId = -3

PR #18713, opened 2023-08-16
Simple test for iterating/debugging

amitanandaiyer:


src/yb/yql/pgwrapper/pg_mini-test.cc
```diff
@@ -450,6 +450,28 @@ TEST_F(PgMiniTest, Simple) {
   ASSERT_EQ(value, "hello");
 }
 
+TEST_F(PgMiniTest, YB_DISABLE_TEST_IN_TSAN(AUH)) {
+  auto conn = ASSERT_RESULT(Connect());
+
```
hbhanawat, line 455: How it is to be used @amitanandaiyer ?

PR #18714, opened 2023-08-17
[docs][ybm][DOC-88] Increase max connections from 10 to 15

ddhodge:
Increase max connections from 10 to 15

@netlify /preview/yugabyte-cloud/cloud-basics/create-clusters-overview/


aishwarya24 (approved):
LGTM

PR #18720, opened 2023-08-17
Add default CPU wait state in PG

abhinab-yb:


src/postgres/src/backend/postmaster/pgstat.c
```diff
@@ -3624,9 +3625,12 @@ pgstat_get_wait_event_type(uint32 wait_event_info)
 		case PG_WAIT_IO:
 			event_type = "IO";
 			break;
+		case PG_WAIT_CPU:
```
hbhanawat, line 3628: Let's call it PG_CPU. It is not a wait event.

src/postgres/src/backend/postmaster/pgstat.c
```diff
@@ -3705,9 +3709,16 @@ pgstat_get_wait_event(uint32 wait_event_info)
 				event_name = pgstat_get_wait_io(w);
 				break;
 			}
+		case PG_WAIT_CPU:
+			{
+				WaitEventCPU w = (WaitEventCPU) wait_event_info;
+
+				event_name = pgstat_get_wait_cpu(w);
```
hbhanawat, line 3716: pgstat_get_cpu_event

hbhanawat (approved):

PR #18727, opened 2023-08-17
[Docs] Node version update from 16 to 20

samiahmedsiddiqui:


aishwarya24 (approved):
Built the server locally from this branch and it works fine even though my local node version is 16. LGTM

PR #18742, opened 2023-08-17
Start Hugo without rendering older versions

premkumr:
- Decreases start time from `30s` to `15s`
- start hugo as `npm run fast`
- Will build preview and stable only


ddhodge (approved):
LGTM

PR #18750, opened 2023-08-18
add sample rate 

nandhiniS1212:


src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -468,17 +481,16 @@ pg_active_universe_history_internal(FunctionCallInfo fcinfo)
...
     else
       nulls[j++] = true;
-
+      
     tuplestore_putvalues(tupstore, tupdesc, values, nulls);
   }
   /* clean up and return the tuplestore */
   tuplestore_donestoring(tupstore);
   LWLockRelease(auh_entry_array_lock);
-
```
abhinab-yb, line 481: nit: removed line

src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -468,17 +481,16 @@ pg_active_universe_history_internal(FunctionCallInfo fcinfo)
...
     // TODO: sample rate is throwing an error in certain mac environments
     // Disabling it for now.
-    if (false && AUHEntryArray[i].sample_rate)
-      values[j++] = Int16GetDatum(AUHEntryArray[i].sample_rate);
+    if (AUHEntryArray[i].sample_rate)
+      values[j++] = Float8GetDatum(AUHEntryArray[i].sample_rate);
     else
       nulls[j++] = true;
-
+
```
abhinab-yb, line 488: nit: extra white space

hbhanawat (changes requested):
1. You are not sampling the events based on the sample size? If sample_size_procs is 5, only 5 events should be inserted in the circular buffer. I don't see that happening. 
2. Secondly, I don't think we need two variables sample_size_procs and sample_size_rpcs. For making it configurable you can follow the same approach as that of yb_auh.sampling_interval

abhinab-yb (approved):

PR #18763, opened 2023-08-18
[docs] Break up yb-admin reference page

ddhodge:
Break yb-admin command reference into 3 sections:
- Everything else
- Backup and snapshot
- CDC and xCluster

@netlify /preview/admin/yb-admin/


ddhodge:
Not proceeding with this for now

PR #18791, opened 2023-08-22
Support conditional setting of wait-state

amitanandaiyer:
Add a log line in set_state_if

 reduce spew. Use TRACE instead of VLOG

### DIFF
--- a/src/yb/util/wait_state.cc
+++ b/src/yb/util/wait_state.cc
@@ -71,7 +71,7 @@ void WaitStateInfo::set_state_if(WaitStateCode prev, WaitStateCode c) {
   if (!ret) {
     return;
   }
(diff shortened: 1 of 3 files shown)


src/yb/util/wait_state.cc
```diff
@@ -66,6 +66,32 @@ WaitStateCode WaitStateInfo::get_frozen_state() const {
   return code_;
 }
 
+void WaitStateInfo::set_state_if(WaitStateCode prev, WaitStateCode c) {
```
abhinab-yb, line 69: How does this work? I don't see the wait state code getting updated, or is it not supposed to do that at all?

abhinab-yb (approved):

PR #18792, opened 2023-08-22
Fix CQL Wait states so that the distribution of wait states across cql and tserver is in the same range

abhinab-yb:


src/yb/util/wait_state.h
```diff
@@ -138,7 +138,7 @@ YB_DEFINE_ENUM_TYPE(
 
     // CQL Wait Events
     ((Parse, YB_CQL_WAIT_STATE))(Analyze)(Execute)(ExecuteWaitingForCB)
-    (CQLRead)(CQLWrite)
+    (CQLRead)(CQLWrite)(CQLHandling)(CQLHandlingDone)(FlushAsync)(FlushAsyncDone)
```
amitanandaiyer, line 141: We have FlushAsync in YBClient as well, which gets called from other places.

Can you prepend these to disambiguate: "ExecutorFlushAsync..."

amitanandaiyer (approved):
one minor comment inline.

PR #18801, opened 2023-08-22
[DOC-91] Add YCQL timing feature

aishwarya24:
@netlify /preview/admin/ycqlsh/#timing

### DIFF
--- a/docs/content/preview/admin/ycqlsh.md
+++ b/docs/content/preview/admin/ycqlsh.md
@@ -12,6 +12,8 @@ menu:
     identifier: ycqlsh
     parent: admin
     weight: 20
+rightNav:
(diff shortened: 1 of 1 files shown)


ddhodge (approved):
LGTM
Is there a reason to not sort the commands alphabetically?
ycqlsh doesn't need to be inside ` ` when it isn't part of a command of some sort.

HarshDaryani896 (approved):
LGTM. thanks @aishwarya24

aishwarya24:
> LGTM Is there a reason to not sort the commands alphabetically? ycqlsh doesn't need to be inside ` ` when it isn't part of a command of some sort.

sorted them and removed the quotes.

PR #18804, opened 2023-08-22
Add tablet id for flush and compaction wait states

abhinab-yb:


src/yb/rocksdb/util/thread_posix.cc
```diff
@@ -243,7 +248,12 @@ int ThreadPool::UnSchedule(void* arg) {
 
 std::vector<yb::util::WaitStateInfoPtr> ThreadPool::GetThreadpoolWaitStates() {
   PthreadCall("lock", pthread_mutex_lock(&mu_));
-  std::vector<yb::util::WaitStateInfoPtr> res = bg_wait_states_;
+  std::vector<yb::util::WaitStateInfoPtr> res;
+  for (auto wait_state : bg_wait_states_) {
```
amitanandaiyer, line 252: Do we want to filter out the un-used ones, or keep them around?

src/yb/rocksdb/util/thread_posix.h
```diff
@@ -102,6 +102,7 @@ class ThreadPool {
     void (*function)(void*);
     void* tag;
     void (*unschedFunction)(void*);
+    yb::util::AUHAuxInfo aux_info;
```
amitanandaiyer, line 105: Why do we need this?

amitanandaiyer:
changes to tablet.cc makes sense. Why do you need any of the other changes?

When we enter the Flush/Compact states, we can track the tablet-id along with those states. Why bother with the changes to the remaining files?

src/yb/rocksdb/util/thread_posix.cc
abhinab-yb: I think it's better to filter them out, otherwise we keep seeing some wait states with -2 as the query id and everything else is null

src/yb/rocksdb/util/thread_posix.h
abhinab-yb: When we are scheduling bg events we also need to store the aux info which is in stored in the current threadlocal, otherwise we are going to lose it as this bg event is going to be picked up by some other thread.

abhinab-yb:
> changes to tablet.cc makes sense. Why do you need any of the other changes?
> 
> When we enter the Flush/Compact states, we can track the tablet-id along with those states. Why bother with the changes to the remaining files?

The thread which schedules the background task is different from the thread which executes the background task. Is there a better way to pass the aux info from one threadlocal to the other?

amitanandaiyer:
The thread which schedules the background task is different from the thread which executes the background task. Is there a better way to pass the aux info from one threadlocal to the other?

Why do we need to pass that information ? The compact or flush task which gets executed will be executing code/functions that already have the tablet-id information. We can just update/set the wait-state/aux information there.

amitanandaiyer:
you could consider using `db_options_.log_prefix ` 

```

  // A prefix for log messages, usually containing the tablet id.
  std::string log_prefix;
  ```

amitanandaiyer (approved):
lgtm! thanks

PR #18852, opened 2023-08-24
[docs] Put older releases under End of life

ddhodge:
Move older releases into subdirectory in left nav

@netlify /preview/releases/


aishwarya24 (commented):
some comments

PR #18878, opened 2023-08-28
Add example scenario for Read Replica

premkumr:
- Adding an example topology for Read Replicas


aishwarya24 (approved):
pushed some changes. LGTM

PR #18926, opened 2023-08-31
Fix negtive request id in TServer

abhinab-yb:


src/postgres/src/backend/utils/misc/pg_yb_utils.c
```diff
@@ -3590,7 +3590,8 @@ void ProcSetTopLevelRequestId(const uint64_t *top_level_request_id)
 
 void uint128_to_char(const uint64_t uint_id[2], char *char_id)
 {
-	sprintf(char_id, "%lx%lx", uint_id[0], uint_id[1]);
+	// lx breaks compile in mac
```
amitanandaiyer, line 3593: use 
`#if defined(__APPLE__)` ?

src/yb/yql/pggate/pg_client.cc
```diff
@@ -440,7 +440,8 @@ class PgClient::Impl {
     *req.mutable_options() = std::move(*options);
     PrepareOperations(&req, operations);
     // Amit: This is probably unnecessary. header_ already contains a call-id.
-    req.mutable_options()->mutable_auh_metadata()->set_current_request_id(reinterpret_cast<uint64_t>(&req));
+    // Abhinab: Using the header call id as the request id.
+    // req.mutable_options()->mutable_auh_metadata()->set_current_request_id(reinterpret_cast<uint64_t>(&req));
```
amitanandaiyer, line 444: Why is this commented?

abhinab-yb: There seems to be only 2 addresses which get reused everytime, so that means the request id also stays the same even with different requests. Also, when the call reaches tserver, we are using header.call_id anyway.

amitanandaiyer: Probably remove lines 442-444?

amitanandaiyer (approved):

PR #18952, opened 2023-09-01
Simplify Waitstates

amitanandaiyer:


src/yb/util/wait_state.h
```diff
@@ -85,76 +96,66 @@ YB_DEFINE_ENUM_TYPE(
...
        (BlockCacheReadFromDisk)
 
     // Perform Wait Events
-    ((DmlRead, YB_PG_WAIT_PERFORM)) (DmlWrite) (DmlReadWrite)
+    ((DmlRead, YB_PG_WAIT_PERFORM))
+    (DmlWrite)
+    (DmlReadWrite)
+    (PgPerformHandling) 
+    (PGWaitingOnDocdb)
+    (PGActiveOnCPU)
```
abhinab-yb, line 152: These events will come as part of PGGate and not TServer, is that fine?

abhinab-yb (approved):
approved with an inline comment

src/yb/util/wait_state.h
amitanandaiyer: Yes. That was intentional.

PR #18956, opened 2023-09-01
[docs] Update ldap-authentication-platform.md

kerneljake:
This is a bit pedantic, but the terminology gets confusing when talking about options, flags, and values when things are included inside other things.  The quoted text is an example instead of a literal value to be pasted.

### DIFF
--- a/docs/content/preview/yugabyte-platform/security/ldap-authentication-platform.md
+++ b/docs/content/preview/yugabyte-platform/security/ldap-authentication-platform.md
@@ -33,7 +33,7 @@ For information on using LDAP for authentication with YugabyteDB Anywhere, refer
 
 ## Bind to the LDAP server using TLS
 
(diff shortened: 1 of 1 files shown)


docs/content/preview/yugabyte-platform/security/ldap-authentication-platform.md
```diff
@@ -33,7 +33,7 @@ For information on using LDAP for authentication with YugabyteDB Anywhere, refer
 
 ## Bind to the LDAP server using TLS
 
-To bind to the LDAP server using TLS, you set the `ldaptls=1` option and the `ysql_hba_conf_csv` flag to the following value:
+To bind to the LDAP server using TLS, you include the `ldaptls=1` option in the `ysql_hba_conf_csv` flag as per the following example:
```
ddhodge, line 36:
```suggestion
To bind to the LDAP server using TLS, include the `ldaptls=1` option in the `ysql_hba_conf_csv` flag as per the following example:
```

ddhodge (approved):

PR #18984, opened 2023-09-05
Table id as the aux info in CQL reads and writes

abhinab-yb:


src/yb/yql/pggate/util/ybc_stat.cpp
```diff
@@ -107,7 +107,7 @@ ybcstat_get_wait_event_type(uint32_t wait_event_info) {
             event_type = "Flush and Compaction";
             break;
         case YB_CQL_WAIT_STATE:
-            event_type = "Query Processing";
+            event_type = "Query";
```
amitanandaiyer, line 110: Query seems too generic. Could as well be confused to a pg-query? 
Probably have CQL somewhere in the name?

src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -212,7 +212,8 @@ static void tserver_collect_samples(TimestampTz auh_sample_time, uint16 num_rpcs
     if(random() <= sample_rate * MAX_RANDOM_VALUE){
       auh_entry_store(auh_sample_time, rpcs[i].metadata.top_level_request_id,
                     rpcs[i].metadata.current_request_id, rpcs[i].wait_status_code,
-                    rpcs[i].aux_info.tablet_id, rpcs[i].metadata.top_level_node_id,
+                    (rpcs[i].aux_info.tablet_id[0] == '\0' ? rpcs[i].aux_info.table_id : rpcs[i].aux_info.tablet_id),
```
amitanandaiyer, line 215: So we seem to be populating the Aux with the table id, if the tablet-id is null.

Sounds like this should be the case -- tablet id being full while in a CQL state ... but I don't think this should be something we rely on. 

What if a thread starts processing a cql-request -- enters the local-tserver mode, and then gets back to the cql mode. Can it have both tablet-id and table-id in the aux_info?

A cleaner solution will be to have a few boolean functions

ShouldExportTabletId(..) /ShouldExportTableId(WaitEvent) ... type function to export one or the other depending on what wait-state we are in.

amitanandaiyer (approved):
approving to unblock. but I don't think we should rely on tablet-id being empty to export table-id. The decision should be based on what the wait-state is.

PR #19088, opened 2023-09-12
Use AUH hooks to set and unset some AUH metadata

abhinab-yb:


amitanandaiyer (approved):
looks good to me. but I'm not super familiar with postgres.

PR #19093, opened 2023-09-12
Updated Vitess comparison

rbranson:
More correctly representative of current-day Vitess capabilities. The changes made were to be as least invasive as possible while updating the wording to be factually correct.

More background: as of the last few years, Vitess presents a single logical view of the database to clients and has a SQL query planner which distributes queries according to where data is placed across shards. While Vitess does not currently support distributed transactions, it will route reads and writes properly to their shards without client involvement. It supports JOINs, aggregates, etc even in sharded mode.

While these queries will be executed properly, if developers are not aware of how queries will be routed according to data placement in the cluster, they may be performed as a "scatter" and thus will not scale as well as queries which align with data placement. However, this is a fundamental behavior of a shared-nothing distributed database. Vitess provides "vtexplain" functionality to help developers understand when this is occurring.

### DIFF
--- a/docs/content/preview/faq/comparisons/vitess.md
+++ b/docs/content/preview/faq/comparisons/vitess.md
@@ -15,9 +15,9 @@ type: docs
 
 Vitess is an automated sharding solution for MySQL. Each MySQL instance acts as a shard of the overall database. It uses etcd, a separate strongly consistent key-value store, to store shard location metadata such as which shard is located in which instance. Vitess uses a pool of stateless servers to route queries to the appropriate shard based on the mapping stored in etcd. Each instance uses standard MySQL source-replica replication to account for high availability.
 
-## No single logical SQL database
+## Single logical SQL database with caveats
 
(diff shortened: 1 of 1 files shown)


docs/content/preview/faq/comparisons/vitess.md
```diff
@@ -15,9 +15,9 @@ type: docs
...
 
-## No single logical SQL database
+## Single logical SQL database with caveats
 
-SQL features that access multiple rows of data spread across multiple shards are not allowed. Examples are distributed ACID transactions and JOINs. This means a Vitess cluster loses the notion of a single logical SQL database. Application developers have to be acutely aware of their sharding mechanism and account for those while designing their schema as well as while executing queries.
+While Vitess presents a single logical SQL database to clients, it does not support distributed ACID transactions across shards. Application developers have to be aware of their sharding mechanism and account for those while designing their schema as well as while executing queries as scalability may be impacted by queries which scatter across shards.
```
ddhodge, line 20:
```suggestion
While Vitess presents a single logical SQL database to clients, it does not support distributed ACID transactions across shards. Application developers have to be aware of their sharding mechanism and account for it while designing their schema, as well as while executing queries as scalability may be impacted by queries which scatter across shards.
```

ddhodge (approved):
LGTM thanks for the submission

PR #19217, opened 2023-09-20
[docs][ybm] Add Grafana support for Metrics export

ddhodge:
Add Grafana support for Metrics export

@netlify /preview/yugabyte-cloud/cloud-monitor/metrics-export/


TaranSinghania:
Can you also mention that list of current supported zones are as mentioned here? https://grafana.com/docs/grafana-cloud/monitor-infrastructure/otlp/send-data-otlp/

PR #19495, opened 2023-10-12
Add info on user_enforced consistency and index ttl

### LINKED ISSUE #1696: Document details of user enforced consistency
Used for higher performance of writes in inserts where consistency semantics can be relaxed. Typically used in streaming apps where high ingest is required.

## Details

**Main caveat: If an insert/update or batch of such operations fails, it is the app's responsibility to retry the operation so that the index is consistent.** Much like in a 2-table case, it would have been the apps responsibility to retry (in case of a failure between the update to the two tables) to make sure both tables are in sync again.

Main advantage of this approach rather than you having to use a second table is for UPDATE cases where the old index record has to be deleted from the table, and so would be an extra roundtrip to the database.

## Example

Notice that here the table is not marked with transactions enabled. And the index has the 'consistency_level' : 'user_enforced'  clause. # Example

```
create table example (
  ...,
  PRIMARY KEY (...)
) WITH CLUSTERING ORDER BY (...)
   AND transactions = { 'enabled' : false };

CREATE INDEX example_index ON example (...) INCLUDE (cluster_size)
    WITH CLUSTERING ORDER BY(...)
    AND transactions = { 'enabled' : false, 'consistency_level' : 'user_enforced' };
```

premkumr:
fixes #1696

### DIFF
--- a/docs/content/preview/api/ycql/ddl_create_index.md
+++ b/docs/content/preview/api/ycql/ddl_create_index.md
@@ -91,7 +91,7 @@ Where
 
 ## Semantics
 
--  An error is raised if transactions have not be enabled using the `WITH transactions = { 'enabled' : true }` clause on the table to be indexed. This is because secondary indexes internally use distributed transactions to ensure ACID guarantees in the updates to the secondary index and the associated primary key. More details [here](https://www.yugabyte.com/blog/yugabyte-db-1-1-new-feature-speeding-up-queries-with-secondary-indexes/).
+- An error is raised if transactions have not be enabled using the `WITH transactions = { 'enabled' : true }` clause on the table to be indexed. This is because secondary indexes internally use distributed transactions to ensure ACID guarantees in the updates to the secondary index and the associated primary key. More details [here](https://www.yugabyte.com/blog/yugabyte-db-1-1-new-feature-speeding-up-queries-with-secondary-indexes/).
 - An error is raised if `index_name` already exists in the associated keyspace unless the `IF NOT EXISTS` option is used.
 
 {{< note title="Note" >}}
@@ -111,7 +111,7 @@ CREATE INDEX ON orders (warehouse)
 ```
 
 {{< warning >}}
-When using an index without transactions enabled, it is the responsibility of the application to retry any insert/update failures to make sure that the table and index are in sync.
+When using an index without transactions enabled, it is the responsibility of the application to retry any insert/update/delete failures to make sure that the table and index are in sync.
 {{< /warning >}}
 
 ### PARTITION KEY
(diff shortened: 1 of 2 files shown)


docs/content/preview/api/ycql/ddl_create_index.md
```diff
@@ -91,16 +91,29 @@ Where
 
 ## Semantics
 
-- An error is raised if transactions have not be enabled using the `WITH transactions = { 'enabled' : true }` clause on the table to be indexed. This is because secondary indexes internally use distributed transactions to ensure ACID guarantees in the updates to the secondary index and the associated primary key. More details [here](https://www.yugabyte.com/blog/yugabyte-db-1-1-new-feature-speeding-up-queries-with-secondary-indexes/).
+-  An error is raised if transactions have not be enabled using the `WITH transactions = { 'enabled' : true }` clause on the table to be indexed. This is because secondary indexes internally use distributed transactions to ensure ACID guarantees in the updates to the secondary index and the associated primary key. More details [here](https://www.yugabyte.com/blog/yugabyte-db-1-1-new-feature-speeding-up-queries-with-secondary-indexes/).
```
ddorian, line 94:
```suggestion
- An error is raised if transactions have not be enabled using the `WITH transactions = { 'enabled' : true }` clause on the table to be indexed. This is because secondary indexes internally use distributed transactions to ensure ACID guarantees in the updates to the secondary index and the associated primary key. More details [here](https://www.yugabyte.com/blog/yugabyte-db-1-1-new-feature-speeding-up-queries-with-secondary-indexes/).
```

docs/content/preview/api/ycql/ddl_create_index.md
```diff
@@ -91,16 +91,29 @@ Where
...
+Indexes require transactions to have been enabled on the table. For cases where the table was created without enabling transactions, `consistency_level` has to be set to `user_enforced` like,
+
+```sql
+CREATE TABLE orders (id int PRIMARY KEY, warehouse int);
+CREATE INDEX ON orders (warehouse)
+      WITH transactions = { 'enabled' : false, 'consistency_level' : 'user_enforced' };
+```
+
+{{< warning >}}
+When using an index without transactions enabled, it is the responsibility of the application to retry any insert/update failures to make sure that the table and index are in sync.
```
ddorian, line 114:
```suggestion
When using an index without transactions enabled, it is the responsibility of the application to retry any insert/update/delete failures to make sure that the table and index are in sync.
```

docs/content/preview/api/ycql/ddl_create_index.md
```diff
@@ -116,6 +129,7 @@ When an index is created on an existing table, YugabyteDB will automatically bac
 - The `CLUSTERING ORDER BY` property can be used to set the ordering for each clustering column individually (default is `ASC`).
 - The `TABLETS = <num>` property specifies the number of tablets to be used for the specified YCQL index. Setting this property overrides the value from the [`--yb_num_shards_per_tserver`](../../../reference/configuration/yb-tserver/#yb-num-shards-per-tserver) option. For an example, see [Create an index specifying the number of tablets](#create-an-index-specifying-the-number-of-tablets).
 - Use the `AND` operator to use multiple index properties.
+- When setting a ttl on the index using `default_time_to_live`, please ensure that the ttl value is the same as that of the table's ttl. If they are different, it would lead to the index and the table being out of sync and would lead to unexpected behavior.
```
ddorian, line 132:
```suggestion
- When setting a TTL on the index using `default_time_to_live`, please ensure that the TTL value is the same as that of the table's TTL. If they are different, it would lead to the index and the table being out of sync and would lead to unexpected behavior.
```

We should also note that the TTL will not be transactional. Meaning, suppose you are reading 10K rows using an index, if the expiration is now, you may read ALIVE rows from the index, but when going to the table the rows would be expired.

ddorian (changes requested):
LGTM. Request a review from dwight/Aishwarya before commit.

dataindataout:
This doesn't cover the case of adding row-level TTL to a table having secondary indexes.

dataindataout:
Thank you for adding the information about row-level TTL and indexes. Looks good.

premkumr:
@ddorian, @dataindataout  could you approve the diff ? (if it looks good)

ddorian (approved):

PR #19636, opened 2023-10-24
[docs] Update create-universe-multi-zone.md

kerneljake:
This gives semantic parity to the first sentence of each of the two tabs: generic and kubernetes.  The existing generic phrasing seems to imply that you cannot create a kubernetes universe in one geographic region across multiple AZs.  But you can; all it takes is using the kubernetes provider -- which is explained on the other tab.


ddhodge (approved):
LGTM

PR #19702, opened 2023-10-27
Added table ID metadata collection

nandhiniS1212:


src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -384,6 +423,29 @@ static void auh_entry_store(TimestampTz auh_time,
...
+                       const char* name,
+                       uint32_t table_type,
+                       uint32_t relation_type,
+                       const char* namespace_id,
+                       const char* namespace_name,
+                       uint32_t database_type,
+                       const char* pgschema_name,
+                       bool colocated,
+                       const char* parent_table_id) {
+    ybtableInfo tableInfo;
```
acube123, line 436: What is this function doing? 

It seems to return nothing.

nandhiniS1212: It was created so that based on what the table ID information is to be collected for, I can populate the global fields later (Similar to the circular buffer), but I can remove it for now.

src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -255,6 +278,7 @@ yb_auh_main(Datum main_arg) {
 
     pg_collect_samples(auh_sample_time, auh_sample_size);
     tserver_collect_samples(auh_sample_time, auh_sample_size);
+    collect_table_info();
```
abhinab-yb, line 281: This shouldn't be called in the background worker, but rather when we query the view for table id metadata. yb_servers is a good example.

src/yb/master/catalog_manager.h
```diff
@@ -463,6 +463,9 @@ class CatalogManager : public tserver::TabletPeerLookupIf,
   Status GetColocatedTabletSchema(const GetColocatedTabletSchemaRequestPB* req,
                                   GetColocatedTabletSchemaResponsePB* resp);
 
+  Status ListTableInfo(const ListTablesRequestPB* req,
```
amitanandaiyer, line 466: This has the same signature as the one below? why are two separate ones needed.

Please consider merging them.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -745,6 +745,14 @@ class PgClient::Impl {
     return resp;
   }
 
+  Result<tserver::PgTableIDMetadataResponsePB>TableIDMetadata() {
```
amitanandaiyer, line 748: nit: space after return type.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -745,6 +745,14 @@ class PgClient::Impl {
...
+    tserver::PgTableIDMetadataRequestPB req; 
+    tserver::PgTableIDMetadataResponsePB resp;
+    RETURN_NOT_OK(proxy_->TableIDMetadata(req, &resp, PrepareController()));
+    //std::vector<master::ListTablesResponsePB::TableInfo> result(resp.tables().begin(), resp.tables().end());
```
amitanandaiyer, line 752: remove

src/yb/client/client.cc
```diff
@@ -2295,6 +2295,13 @@ Result<bool> YBClient::CheckIfPitrActive() {
   return data_->CheckIfPitrActive(deadline);
 }
 
+Result<ListTablesResponsePB> YBClient::ListTableInfo() {
```
amitanandaiyer, line 2298: Probably ok for the feature branch. But, you are duplicating the logic between this function and the one below.

You could merge the commonality between the two to reduce future maintenance. Add a todo if you aren't fixing it now, but will get to it when merging to the master branch.

src/postgres/contrib/yb_auh/yb_auh.c
```diff
@@ -483,6 +497,145 @@ ybauh_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
+  rsinfo->setResult = tupstore;
+  rsinfo->setDesc = tupdesc;
+
+  MemoryContextSwitchTo(oldcontext);
+
+  YBCTableIDMetadataInfo *infolist = NULL;
+	size_t size = 0;
+  
+  HandleYBStatus(YBCTableIDMetadata(&infolist, &size));
+  ybtableInfo* tableInfo = (ybtableInfo*)palloc(size * sizeof(ybtableInfo));
```
abhinab-yb, line 536: You can use `infolist` to print the view instead of allocating memory again.

abhinab-yb (approved):

PR #19723, opened 2023-10-30
auh global view

hbhanawat:
This script establishes connection with all the servers in the YB cluster and creates a global view for pg_stat_statements and pg_active_universe_history. Since password is needed, we need to pass it as custom.password parameter as show below           
Usage:                                                                                            
$ PGOPTIONS="-c custom.password=${yb_passwd}" bin/ysqlsh -U yugabyte -f auh_global_views.sql


amitanandaiyer (approved):
lgtm.

PR #19724, opened 2023-10-30
[Docs] Show the page title rather than first heading

samiahmedsiddiqui:
On search result, it should show the page title rather than first heading.


ddhodge (approved):
LGTM

PR #19739, opened 2023-10-30
[docs] YSQL max connections clarification

ddhodge:
Clarify YSQL Max connections

@netlify /preview/reference/configuration/yb-tserver/#ysql-max-connections


tylarb (approved):
LGTM

PR #19851, opened 2023-11-06
[docs] Fix multi-column index example

aishwarya24:
Slack thread: https://yugabyte.slack.com/archives/CDQB8MEAU/p1699266093788359

### DIFF
--- a/docs/content/preview/explore/indexes-constraints/secondary-indexes-ysql.md
+++ b/docs/content/preview/explore/indexes-constraints/secondary-indexes-ysql.md
@@ -157,7 +157,7 @@ To add a multi-column index to an existing table, you can use the following synt
 CREATE INDEX index_name ON table_name ((col2,col3,col4));
 ```
 
-If you include nested parenthesis between the columns as per the preceding syntax, you will have to perform search using all the columns col2,col3, and col4.
+If you enclose the columns in nested parentheses as per the preceding syntax, you will have to perform search using all the columns (col2, col3, and col4).
 
 However, if you add the multi-column index to an existing table using the following syntax:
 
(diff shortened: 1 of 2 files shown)


tylarb:
I feel we should expand on what the nested parens means in this example.

Both the original and the proposed are correct, from a syntax perspective, but the outcome is different in each case. 

If we adjust to to the proposed ```((col2, col3, col4))``` then the sentence:

"""
So, for the preceding syntaxes, you can perform search using the following column combinations:


```sql
(col2)
(col2,col3)
(col2,col3,col4)
```
"""


Is actually no longer true... you'd have to use all of col 2,3, and 4. 

The difference is that, for ```CREATE INDEX index_name ON table_name(col2,col3,col4);```

the hash partition is just col2, while ```CREATE INDEX index_name ON table_name((col2,col3,col4)); ``` has a hash partition on the unique col2, col3 and col4...

aishwarya24:
@tylarb - Thanks for explaining the different ways we can add the index. I have modified based on what I understood. Can you PTAL and let me know. Thanks!

docs/content/preview/explore/indexes-constraints/secondary-indexes-ysql.md
```diff
@@ -154,17 +154,26 @@ CREATE TABLE table_name (
 To add a multi-column index to an existing table, you can use the following syntax:
 
 ```sql
-CREATE INDEX index_name ON table_name(col2,col3,col4);
+CREATE INDEX index_name ON table_name ((col2,col3,col4));
 ```
 
-The column order is very important when you create a multi-column index in YSQL because of the structure in which the index is stored. As such, these indexes have a hierarchical order from left to right. So, for the preceding syntaxes, you can perform search using the following column combinations:
+If you include nested parenthesis between the columns as per the preceding syntax, you will have to perform search using all the columns col2,col3, and col4.
```
ddhodge, line 160:
```suggestion
If you enclose the columns in nested parentheses as per the preceding syntax, you will have to perform search using all the columns (col2, col3, and col4).
```

aishwarya24:
Closing it as its being addressed in another PR.

Issue #19920, opened 2023-11-09
[DocDB] style: rename AsyncClientInitialiser to AsyncClientInitializer


jasonyb:
Jira Link: [DB-8863](https://yugabyte.atlassian.net/browse/DB-8863)
### Description

There's already

- async_initializer.cc
- MakeClientInitializer

so "AsyncClientInitialiser" using an 's' is off.  Change it to 'z'.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-8863]: https://yugabyte.atlassian.net/browse/DB-8863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

maximilian0101:
Hello! I would like to work on this issue. Where in the docs can I find the mistake?

davixiao:
Apologies, didn't notice Max's comment. Feel free to disregard my PR if Max ends up making his own!

jasonyb:
I'm not going to throw away already-done work, so I will review @davixiao's PR.  Thank you both.

PR #19928, opened 2023-11-11
[#19920] style: rename AsyncClientInitialiser to AsyncClientInitializer

davixiao:
issue: https://github.com/yugabyte/yugabyte-db/issues/19920


jasonyb (approved):
Both empty:

```
git grep -i initialiser src/yb
find -iname '*initialiser*'
```

Thank you for your contribution!

PR #19931, opened 2023-11-13
[Docs] Search Enhancement: Display subhead beneath the title

samiahmedsiddiqui:
If there is a subhead, display it beneath the title like this (but without the highlight)
image


aishwarya24 (approved):
LGTM

PR #19936, opened 2023-11-14
[docs] Remove wrong flags from 2.14

ddhodge:
Remove some flags that are not available in 2.14

@netlify /v2.14/reference/configuration/yugabyted/


aishwarya24 (approved):
LGTM

Issue #17495, opened 2023-05-22
[YSQL] pg_stat_user_indexes.idx_scan incremented too early


jasonyb:
Jira Link: [DB-6653](https://yugabyte.atlassian.net/browse/DB-6653)
### Description

Commit cf99487cdffb6e08a25e340197d7edf7a29c0e58 supports pg_stat_user_indexes idx_scan column for lsm index, but it deviates from the status quo: rather than incrementing on amgettuple or amgetbitmap, it increments on ambeginscan, which is earlier.  In case an error is hit between ambeginscan and amgettuple/amgetbitmap, it would increment the stat when the expectation is to not.  An example is the "cannot use more than %d predicates in a table or index scan" error message called by amrescan, which is in between ambeginscan and amgettuple/amgetbitmap.  I believe the stat bump for lsm should be moved to `ybcingettuple` using `!ybScan->is_exec_done` as a marker for whether it is the first time we ran `ybcingettuple`, similar to how `btgettuple` uses `!BTScanPosIsValid(so->currPos)`.

Unpolished repro that hits that error:

```sh
bin/ysqlsh -c "create table many ($(for i in {1..99}; do echo -n "c$i int,"; done) k int primary key)"
bin/ysqlsh -c "create index on many ($(for i in {1..30}; do echo -n "c$i,"; done) c1 asc)"
bin/ysqlsh -c "select * from many where $(for i in {1..30}; do echo -n "c$i = 1 and c$i < 2 and c$i > 0 and "; done) true"
```

```
ERROR:  cannot use more than 64 predicates in a table or index scan
```

Each time the erroring `select` is run, observe the `idx_scan` count go up in `pg_stat_user_indexes`: `select * from pg_stat_user_indexes;`.

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-6653]: https://yugabyte.atlassian.net/browse/DB-6653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

akash-ind:
Hi Team, 
Can anyone assign me this issue?

jasonyb:
> Hi Team, Can anyone assign me this issue?

@akash-ind Feel free to submit a PR.  It's open for anyone, so no assignment.

[...]

PR #19963, opened 2023-11-16
[#17495] pg_stat_user_indexes.idx_scan incremented too early

davixiao:


src/yb/yql/pgwrapper/pg_libpq-test.cc
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
...
+  ASSERT_OK(conn.Execute(createTableStmt));
+  ASSERT_OK(conn.Execute(createIndexStmt));
+
+  ASSERT_NOK(conn.Fetch(predicateErrorQuery));
+  const auto noIncrementRows = ASSERT_RESULT((conn.FetchRows<int64_t>("select idx_scan from pg_stat_user_indexes")));
+  ASSERT_EQ(noIncrementRows, (decltype(noIncrementRows){0, 0}));
+
+  ASSERT_TRUE(ASSERT_RESULT(conn.HasIndexScan("select * from many where c1 = 1")));
+  const auto incrementRows = ASSERT_RESULT((conn.FetchRows<int64_t>("select idx_scan from pg_stat_user_indexes")));
+  ASSERT_EQ(incrementRows, (decltype(incrementRows){0, 0})); // should expect 0, 1
```
davixiao, line 220: @jasonyb any ideas why the count isn't incremented here when an index scan happens successfully? It gives
```
                          indexrelname                           | idx_scan
-----------------------------------------------------------------+----------
 many_pkey                                                       |        0
 many_c1_c2_c3_c4_c5_c6_c7_c8_c9_c10_c11_c12_c13_c14_c15_c16_idx |        1
 ```
 on my local, but test results seem to give 0, 0.

File: src/yb/yql/pgwrapper/pg_libpq-test.cc [hunk 2]
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
...
+
+  std::string predicateErrorQuery = "select * from many where ";
+  for (int i = 1; i <= 30; ++i)
+    predicateErrorQuery += Format("c$0 = 1 and c$0 < 2 and c$0 > 0 and ", i);
+  predicateErrorQuery += " true";
+
+  ASSERT_OK(conn.Execute(createTableStmt));
+  ASSERT_OK(conn.Execute(createIndexStmt));
+
+  ASSERT_NOK(conn.Fetch(predicateErrorQuery));
```
jasonyb, line 214:
```suggestion
  auto status = ResultToStatus(conn.Fetch(predicateErrorQuery));
  ASSERT_NOK(status);
  ASSERT_STR_CONTAINS(status.message().ToBuffer(), "cannot use more than <todo fill this>");
```

src/yb/yql/pgwrapper/pg_libpq-test.cc
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
   ASSERT_EQ(row, (decltype(row){1, "hello"}));
 }
 
+TEST_F(PgLibPqTest, PgStatIdxScanNoIncrementOnErrorTest) {
+  auto conn = ASSERT_RESULT(Connect());
+
+  std::string createTableStmt = "create table many (";
```
jasonyb, line 196: use snake case for non-constexpr vars as the whole file does that

jasonyb, line 196:
```suggestion
  std::string createTableStmt = "CREATE TABLE many (";
```

use proper capitalization in your SQL

src/yb/yql/pgwrapper/pg_libpq-test.cc
jasonyb: `conn.HasIndexScan` doesn't run the query

src/yb/yql/pgwrapper/pg_libpq-test.cc
jasonyb, line 196: should use stringstream: see `src/yb/yql/pgwrapper/pg_catalog_version-test.cc` for example

src/yb/yql/pgwrapper/pg_libpq-test.cc
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
...
+  predicateErrorQuery += " true";
+
+  ASSERT_OK(conn.Execute(createTableStmt));
+  ASSERT_OK(conn.Execute(createIndexStmt));
+
+  ASSERT_NOK(conn.Fetch(predicateErrorQuery));
+  const auto noIncrementRows = ASSERT_RESULT((conn.FetchRows<int64_t>("select idx_scan from pg_stat_user_indexes")));
+  ASSERT_EQ(noIncrementRows, (decltype(noIncrementRows){0, 0}));
+
+  ASSERT_TRUE(ASSERT_RESULT(conn.HasIndexScan("select * from many where c1 = 1")));
```
jasonyb, line 218: should demonstrate a large number of predicates + no error.  Get it as many predicates as you can before error.  Should be [the number of predicates in error case] - 1.  could share the same query (sub)string between both this non-error case and error case since they can be mostly the same.

src/yb/yql/pgwrapper/pg_libpq-test.cc
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
...
+
+  std::string createIndexStmt = "create index on many (";
+  for (int i = 1; i <= 30; ++i)
+    createIndexStmt += Format("c$0,", i);
+  createIndexStmt += "c1 asc)";
+
+  std::string predicateErrorQuery = "select * from many where ";
+  for (int i = 1; i <= 30; ++i)
+    predicateErrorQuery += Format("c$0 = 1 and c$0 < 2 and c$0 > 0 and ", i);
+  predicateErrorQuery += " true";
```
jasonyb, line 209: should try to find the lowest number of predicates that still hits error instead of this current 90.  I think it is 65.  Try storing `constexpr kMaxPredicates = 64` and using `kMaxPredicates + 1`.

src/yb/yql/pgwrapper/pg_libpq-test.cc
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
...
+
+  std::string createTableStmt = "create table many (";
+  for (int i = 1; i <= 30; ++i)
+    createTableStmt += Format("c$0 int,", i);
+  createTableStmt += "k int primary key)";
+
+  std::string createIndexStmt = "create index on many (";
+  for (int i = 1; i <= 30; ++i)
+    createIndexStmt += Format("c$0,", i);
+  createIndexStmt += "c1 asc)";
```
jasonyb, line 204: c1 shows up twice.  Logic can avoid that from happening, for example tracking whether it's the first iteration or not to determine whether comma should be added.  The first column in the index should have explicit "asc" in the syntax; the rest are implicit asc so that is not necessary, but you can still do that if lazy.

src/yb/yql/pgwrapper/pg_libpq-test.cc
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
   ASSERT_EQ(row, (decltype(row){1, "hello"}));
 }
 
+TEST_F(PgLibPqTest, PgStatIdxScanNoIncrementOnErrorTest) {
```
jasonyb, line 193: code comment what the test does

src/yb/yql/pgwrapper/pg_libpq-test.cc
```diff
@@ -190,6 +190,36 @@ TEST_F(PgLibPqTest, Simple) {
...
+  auto conn = ASSERT_RESULT(Connect());
+
+  std::string createTableStmt = "create table many (";
+  for (int i = 1; i <= 30; ++i)
```
jasonyb, line 197: 30 can be stored in `constexpr kNumColumns = 30;`

jasonyb (changes requested):
Thanks for the quick turnaround.

src/yb/yql/pgwrapper/pg_libpq-test.cc
davixiao: I've also tried the following, but these attempts also don't increase the count
```
ASSERT_OK(conn.Fetch("select * from many where c1 = 1"));
 ```
 and
 ```
 auto values = ASSERT_RESULT(conn.FetchRows<int32_t>("select c1 from many where c1 = 1"));
 ASSERT_EQ(values.size(), 0);
 ```

jasonyb: This was nontrivial information: stats are updated only periodically.  I have documented/updated the test myself.

jasonyb (approved):

jasonyb:
Sorry the test may have been more difficult than anticipated.  Got the tests in good shape.  Just waiting for internal test suites to run and then will check it in.  Thanks for contributing!

PR #19971, opened 2023-11-16
[docs] Backport delete table in Cassandra feature support 

aishwarya24:
@netlify /v2.18/explore/ycql-language/cassandra-feature-support/


ddhodge (approved):
I feel like seeing H3s in the rightnav would be nice for these pages, but otherwise LGTM

aishwarya24:
> I feel like seeing H3s in the rightnav would be nice for these pages, but otherwise LGTM

I agree , when I was editing it , I missed seeing them in the right nav actually. Will remove it.

PR #19995, opened 2023-11-21
[DOC-164] Changed path for test certs

aishwarya24:


ddhodge (approved):
LGTM but does it apply to earlier versions?

PR #20066, opened 2023-11-29
Remove yb client wait states except CQL Query

abhinab-yb:


src/yb/client/batcher.cc
```diff
@@ -312,7 +312,7 @@ void Batcher::CombineError(const InFlightOp& in_flight_op) {
 }
 
 void Batcher::LookupTabletFor(InFlightOp* op) {
-  PUSH_WAIT_STATUS(util::WaitStateCode::LookingUpTablet);
+  // PUSH_WAIT_STATUS(util::WaitStateCode::LookingUpTablet);
```
amitanandaiyer, line 315: I would also comment out the YBClient wait codes in wait_state.h to ensure that we aren't mistakenly using it elsewhere.

Issue #14322, opened 2022-10-04
[Docs]  On-Prem Unresponsive Node directions need clarification 


CharlotteRose:
### Description

Currently our documentation touches on removing an unresponsive node from a Universe. 

For On-Prem customers the instructions are lacking the added clarifying point that nodes will not be cleaned by the YBA/Platform. Based on recent feedback we are requesting a link to the 

[ Current Information ]

https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#delete-node

When this happens, new Master leaders are elected for the underlying data shards, but since the universe enters a partially under-replicated state, it would not be able to tolerate additional failures. To remedy the situation, you can eliminate the unreachable node by taking actions in the following sequence:

    Step 1: [Remove node](https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#remove-node)
    Step 2: [Release instance](https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#release-instance)
    Step 3: [Delete node](https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#delete-node)

[ Current Information ]

https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#delete-node

When this happens, new Master leaders are elected for the underlying data shards, but since the universe enters a partially under-replicated state, it would not be able to tolerate additional failures. To remedy the situation, you can eliminate the unreachable node by taking actions in the following sequence:

    Step 1: [Remove node](https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#remove-node)
    Step 2: [Release instance](https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#release-instance)
    Step 3: [Delete node](https://docs.yugabyte.com/preview/yugabyte-platform/manage-deployments/remove-nodes/#delete-node)

For On-Prem customers, if you desire to reuse the node, manual cleaning of the node will be required. Please see our section on [Delete database server nodes ](https://docs.yugabyte.com/preview/yugabyte-platform/configure-yugabyte-platform/set-up-cloud-provider/on-premises/#delete-database-server-nodes) for remaining steps to delete On-Prem Nodes.

PR #20076, opened 2023-11-29
[docs][yba] Clarify removal of nodes for on-prem

ddhodge:
Removing unresponsive on-prem nodes requires additional step
#14322

@netlify /preview/yugabyte-platform/manage-deployments/remove-nodes/


aishwarya24 (approved):
LGTM

Issue #19516, opened 2023-10-13
[DocDB] Fix retryable requests logs to follow standard format


rthallamko3:
Jira Link: [DB-8309](https://yugabyte.atlassian.net/browse/DB-8309)
### Description

Some retryable requests logs should have a tablet / peer prefix like all the raft logs

The following logs need to be fixed:

`I1013 00:08:27.656975  2504 retryable_requests.cc:258] Initialized RetryableRequestsManager, found a file ? no, wal dir=/mnt/d0/yb-data/tserver/wals/table-878e704957a046dc99f6e5de29ef2fdd/tablet-d4302530b520446ca7b8b923e3b6c4a9
`

It should follow the convention 
`I1013 00:13:16.665539  2509 tablet_bootstrap.cc:1527] T 8c0741edf81c4e6c941c8ba7e9fc315c P e1b27e86c7354dcd9a17d926ff14d325: Dumping replay state to log at the end of PlaySegments
`

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-8309]: https://yugabyte.atlassian.net/browse/DB-8309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ajd12342:
Hi @rthallamko3 !
I am Anuj Diwan, a Computer Science PhD student at UT Austin. I am part of a team along with @arjunrs1 (Arjun Somayazulu) and we're taking a graduate [Distributed Systems course](https://github.com/vijay03/cs380d-f23). For our course project, we are interested in contributing to Yugabyte. Could we work on this issue? Any pointers for us to get started would be appreciated as well. 

Thanks and regards,
Anuj.

rthallamko3:
@ajd12342 , Refer to Yugabyte contribution page for instructions - https://docs.yugabyte.com/preview/contribute/core-database/checklist/ Feel free to take up this issue. cc @Huqicheng in case you need references to this particular issue. If you get past the initial aspects, we can identify a couple more items in this area.

ajd12342:
Hi @Huqicheng @rthallamko3 , thanks, we will take this one up. A couple of questions:
[...]

PR #20083, opened 2023-11-29
Fix retryable requests logs to follow standard format

ajd12342:
Adds tablet ID and FS Manager UUID to logging code in retryable_requests.cc, fixing #19516 .

### DIFF
--- a/src/yb/consensus/retryable_requests.cc
+++ b/src/yb/consensus/retryable_requests.cc
@@ -40,7 +40,10 @@
 #include "yb/util/rw_mutex.h"
 #include "yb/util/status_format.h"
 
+#include "yb/gutil/strings/substitute.h"
+
 using namespace std::literals;
+using strings::Substitute;
 
 // We use this limit to prevent request range from infinite grow, because it will block log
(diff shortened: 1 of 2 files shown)


src/yb/consensus/retryable_requests.cc
```diff
@@ -244,18 +244,37 @@ std::ostream& operator<<(std::ostream& out, const ReplicateData& data) {
...
+
+RetryableRequestsManager::RetryableRequestsManager(
+    const TabletId& tablet_id, FsManager* const fs_manager, const std::string& wal_dir,
+    const MemTrackerPtr& mem_tracker, const std::string log_prefix)
+    : tablet_id_(tablet_id),
+      fs_manager_(fs_manager),
+      dir_(wal_dir),
+      retryable_requests_(std::make_unique<RetryableRequests>(mem_tracker, std::move(log_prefix))),
+      log_prefix_(std::move(log_prefix)) {
+    // Several tests (e.g. tablet_peer-test) erroneously set the log_prefix argument to ""
```
Huqicheng, line 259: I think we can explicitly set the log_prefix from upper layer, otherwise, leave it as empty.

Issue #16812, opened 2023-04-11
[DocDB] Add observability in raft replication


robertsami:
Jira Link: [DB-6159](https://yugabyte.atlassian.net/browse/DB-6159)
### Description

We have `performing_update_mutex_` and `peer_lock_` in `consensus_peers.cc`. If these get held for too long raft heartbeats will go unserved and cause leadership loss and re-elections. We should explicitly log any time these locks are held too long to make it more clear that this is happening in cases where we see rapid/intermittent leadership loss

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-6159]: https://yugabyte.atlassian.net/browse/DB-6159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ajd12342:
Hi @robertsami !
I am Anuj Diwan, a Computer Science PhD student at UT Austin. I am part of a team along with @arjunrs1 (Arjun Somayazulu) and we're taking a graduate [Distributed Systems course](https://github.com/vijay03/cs380d-f23). For our course project, we are interested in contributing to Yugabyte. This issue is related to our course material. Could we work on this issue? Any pointers for us to get started would be appreciated as well. 

Thanks and regards,
Anuj.

rthallamko3:
@ajd12342 , Refer to Yugabyte contribution page for instructions - https://docs.yugabyte.com/preview/contribute/core-database/checklist/ Feel free to take up this issue. cc @Huqicheng in case you need references to this particular issue. If you get past the initial aspects, we can identify a couple more items in this area.

ajd12342:
@Huqicheng Thanks, we will take this issue up. Our initial idea is to add a field to the lock that logs the timestamp at which it was locked (which naturally gets reset after an unlock and lock). Then, whenever someone queries the state of the lock or tries to get the lock, we can compare the current timestamp to this timestamp in order to decide whether to log that it has been taking too long. Is this reasonable?

ajd12342:
Hi @rthallamko3 could you assign this one to us if all looks good?

Huqicheng:
@ajd12342 Sounds good. Is it possible to implement a wrapper of the lock instead of adding the field directly to the current lock?

PR #20110, opened 2023-11-30
Add observability in raft replication

ajd12342:
Adds 2 new wrapper classes, AtomicTryMutexWithTimestamp (around AtomicTryMutex) and simple_spinlock_with_timestamp (around simple_spinlock) that implement the desired functionality in #16812 . Note that no additional mutex is needed to ensure thread safety of the lock_timestamp_ variable because the mutex itself guarantees thread safety (this is why we perform an unlock _after_ the timestamp check).

### DIFF
--- a/src/yb/util/locks.h
+++ b/src/yb/util/locks.h
@@ -84,7 +84,7 @@ class CAPABILITY("mutex") simple_spinlock {
   DISALLOW_COPY_AND_ASSIGN(simple_spinlock);
 };
 
-// Saves timestamp of locking in order to log when it is held for too long
+// Saves timestamp of locking in order to log when it is held for too long.
 class CAPABILITY("mutex") simple_spinlock_with_timestamp {
  public:
   simple_spinlock_with_timestamp() {}


src/yb/util/atomic.h
```diff
@@ -444,6 +445,43 @@ class AtomicTryMutex {
...
+
+  // Set the threshold duration for warnings
+  void set_warning_threshold(long long threshold_ms) {
+    warning_threshold_ms_ = threshold_ms;
+  }
+
+ private:
+  AtomicTryMutex mutex_;
+  std::chrono::steady_clock::time_point lock_timestamp_;
+  long long warning_threshold_ms_ = 1000;  // Default threshold
```
Huqicheng, line 482: Please define the threashold as a GFlag,
long long warning_threshold_ms_ = 1000;
see
DEFINE_RUNTIME_int64(cdc_intent_retention_ms, 4 * 3600 * 1000,
    "Interval up to which CDC consumer's checkpoint is considered for retaining intents."
    "If we haven't received an updated checkpoint from CDC consumer within the interval "
    "specified by cdc_checkpoint_opid_interval, then CDC does not consider that "
    "consumer while determining which op IDs to delete from the intent.");

src/yb/util/locks.h
```diff
@@ -84,6 +84,51 @@ class CAPABILITY("mutex") simple_spinlock {
   DISALLOW_COPY_AND_ASSIGN(simple_spinlock);
 };
 
+// Saves timestamp of locking in order to log when it is held for too long
```
Huqicheng, line 87: nit: Comment should end with `.`

src/yb/util/atomic.h
```diff
@@ -444,6 +445,43 @@ class AtomicTryMutex {
   std::atomic<bool> locked_{false};
 };
 
+class AtomicTryMutexWithTimestamp {
```
Huqicheng, line 448: I was thinking something like

template <class LockType>
class LockWithTimestamp {
  LockType lock_;
  ...
}

When using it, LockWithTimestamp<simple_spinlock> or LockWithTimestamp<AtomicTryMutex> could be defined.

src/yb/util/atomic.h
ajd12342: For this, should we add the GFlag to atomic.h itself or move both the GFlag and the class definition to atomic.cc ?

Huqicheng: @ajd12342 Oh, the flag should be moved to atomic.cc

src/yb/util/atomic.h
Huqicheng: I will recommend declare and define this class under src/yb/util/locks.h and src/yb/util/locks.cc

Issue #17280, opened 2023-05-11
[Docs] "Create server certificates" page has wrong instructions on subjectAltName


WhiteWind:
### Description

With the provided configs subjectAltName does not appear in resulting node certificates.

Fixes: 
Add `copy_extensions = copy` under [my_ca] in ca.conf
Add `req_extensions = req_ext` under [req] in node.conf
Change case: `SubjectAltName` to `subjectAltName` in node.conf

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

PR #20132, opened 2023-12-03
Update server-certificates.md

jrnker:
Fixed case issue in subjectAltName
Added necessary options for openssl ca to generate SAN certificates


ddhodge:
@Vars-07 what about the additional changes in #17280
- Add copy_extensions = copy under [my_ca] in ca.conf
- Add req_extensions = req_ext under [req] in node.conf

Vars-07:
@ddhodge first requirement we don't need as that is for openssl ca, which we don't mention in our doc, & the second requirement will be taken care off by
`-batch \
           -extfile <node-ip-address>/node.conf \
           -extensions req_ext`

Issue #13702, opened 2022-08-21
[DocDB] GetConsensusState could also return current index.


amitanandaiyer:
### Description

GetConsensusState is used by scripts such as yugatool to generate tablet-reports.

However the information printed wrt term/index are inaccurate:

The index printed which comes from RaftConfigPB corresponds to the index at which the last config-change took place.

It would be more relevant to have the current-index information be added to the proto.
Also, for the leader, it would be useful to add information about whether or not the initial no-op has been replicated.

cc : @tvesely @eugeneckim @tylarb

mohatagarvit:
Hi,
As per the conversation in slack, @Huqicheng @rthallamko3, [I](https://github.com/mohatagarvit) and [Kulin Shah](https://github.com/kulinshah98) found #13700 and this issue related and would want to start work on this. Can you share any details or pointers to get started?

rthallamko3:
@amitanandaiyer , Can you add some code pointers for this task? cc @Huqicheng

Huqicheng:
GetConsensusState request is handled by
```cpp
void ConsensusServiceImpl::GetConsensusState(const consensus::GetConsensusStateRequestPB *req,
                                             consensus::GetConsensusStateResponsePB *resp,
                                             rpc::RpcContext context) {
  DVLOG(3) << "Received GetConsensusState RPC: " << req->DebugString();

  RpcScope scope(tablet_manager_, "GetConsensusState", req, resp, &context);
  if (!scope) {
    return;
  }
  ConsensusConfigType type = req->type();
  if (PREDICT_FALSE(type != CONSENSUS_CONFIG_ACTIVE && type != CONSENSUS_CONFIG_COMMITTED)) {
    HandleErrorResponse(resp, &context,
        STATUS(InvalidArgument, Substitute("Unsupported ConsensusConfigType $0 ($1)",
                                           ConsensusConfigType_Name(type), type)));
    return;
  }
  LeaderLeaseStatus leader_lease_status;
  *resp->mutable_cstate() = scope->ConsensusState(req->type(), &leader_lease_status);
  resp->set_leader_lease_status(leader_lease_status);
}
```
And it's returning RaftConfigPB as part of the response
```
// A set of peers, serving a single tablet.
message RaftConfigPB {
  // The index of the operation which serialized this RaftConfigPB through
  // consensus. It is set when the operation is consensus-committed (replicated
  // to a majority of voters) and before the consensus metadata is updated.
[...]

PR #20133, opened 2023-12-03
Extend GetConsensusState to return current index.

kulinshah98:
Extend GetConsensusState to return current index. Solves [Issue #13702](https://github.com/yugabyte/yugabyte-db/issues/13702)

### DIFF
--- a/src/yb/consensus/consensus.h
+++ b/src/yb/consensus/consensus.h
@@ -280,13 +280,15 @@ class Consensus {
   // leader lease status captured under the same lock.
   virtual ConsensusStatePB ConsensusState(
       ConsensusConfigType type,
-      LeaderLeaseStatus* leader_lease_status = nullptr) const = 0;
+      LeaderLeaseStatus* leader_lease_status = nullptr, 
+      ListAllOpIdPB* opid_list = nullptr) const = 0;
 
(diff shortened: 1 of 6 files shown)


src/yb/consensus/consensus.h
```diff
@@ -280,13 +280,15 @@ class Consensus {
   // leader lease status captured under the same lock.
   virtual ConsensusStatePB ConsensusState(
       ConsensusConfigType type,
-      LeaderLeaseStatus* leader_lease_status = nullptr) const = 0;
+      LeaderLeaseStatus* leader_lease_status = nullptr, 
+      ListAllOpIdPB* opid_list = nullptr) const = 0;
```
Huqicheng, line 284: Would ConsensusWatermarksPB be more clear?

src/yb/consensus/metadata.proto
```diff
@@ -139,6 +139,20 @@ message ConsensusStatePB {
   required RaftConfigPB config = 3;
 }
 
+message ListAllOpIdPB {
+  // Operation ID for the latest request received.
+  optional int64 last_received_opid_index = 1;
```
Huqicheng, line 144: Instead of op index (int64), OpIdPB (term, index) looks more clear. As working on this item, let's make the debug info as detailed as possible.

PR #20172, opened 2023-12-05
[DOC-173] Removed config backup target under YBA Configure section

aishwarya24:


ddhodge:
Remove tile from config landing page

PR #20257, opened 2023-12-10
[yugabyte#10948] Sys Catalog Compaction and Flush should work across Peers.

### LINKED ISSUE #10948: [docdb] Sys catalog compaction and flush should work across peers
In https://github.com/yugabyte/yugabyte-db/issues/8730, we added commands `flush_sys_catalog` and `compact_sys_catalog`. However, these commands only work at the leader sys catalog tablet. We should enhance these commands to work across peers, not just at the leader

albertchokor:
- Fixes #10948 to extend sys catalog compaction and flush to work across peers
- Updated existing `AdminCliTest` unit tests `FlushSysCatalog` and `CompactSysCatalog` to test all masters in cluster

Please check MasterAdminProxy instantiation, could not find similar cases where MasterAdminProxy RPC is invoked on all peers as  well. Follows other similar code for getting all masters and instantiating proxies to invoke RPCs.

### DIFF
--- a/src/yb/tools/yb-admin-test.cc
+++ b/src/yb/tools/yb-admin-test.cc
@@ -947,8 +947,7 @@ TEST_F(AdminCliTest, FlushSysCatalog) {
     master_addr_strs.push_back(ToString(addr));
   }
 
-  auto client = ASSERT_RESULT(
-      YBClientBuilder().master_server_addrs(master_addr_strs).Build());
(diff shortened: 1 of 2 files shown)


robertsami:
@albertchokor thanks for the PR! can you run the tests with the following command (change filter for the other test) and ensure it passes?

```
./yb_build.sh --cxx-test tools_yb-admin-test --gtest_filter="AdminCliTest.FlushSysCatalog"
```

PR #20430, opened 2024-01-03
[doc][yba] Add YBA installer to Upgrade

ddhodge:
Add YBA Installer to Upgrade section.
DOC-186

@netlify /preview/yugabyte-platform/upgrade/


docs/content/preview/yugabyte-platform/upgrade/upgrade-yp-installer.md
```diff
@@ -0,0 +1,55 @@
...
+  </li>
+
+  <li>
+    <a href="../upgrade-yp-kubernetes/" class="nav-link">
+      <i class="fa-regular fa-dharmachakra" aria-hidden="true"></i>Kubernetes</a>
+  </li>
+
+</ul>
+
+To upgrade using YBA Installer, first download the version of YBA Installer corresponding to the version of YBA you want to upgrade to.
```
chidmuthu, line 33: Maybe call out you can only upgrade to a higher version and downgrading is NOT supported, so this is a one way action.

docs/content/preview/yugabyte-platform/upgrade/upgrade-yp-installer.md
```diff
@@ -0,0 +1,55 @@
...
+$ cd yba_installer_full-{{<yb-version version="preview" format="build">}}/
+```
+
+When ready to upgrade, run the `upgrade` command from the untarred directory of the target version of the YBA upgrade:
+
+```sh
+$ sudo ./yba-ctl upgrade
+```
+
+The upgrade takes a few minutes to complete. When finished, use the status command to verify that YBA has been upgraded to the target version:
```
chidmuthu, line 49: In case of errors, we leave around the previous version software to make it easy to switch back. If there are any issues with upgrade they should immediately contact support as we don't have any rollback mechanism built into the product.

ddhodge: So it is there, but they shouldn't do anything with the old version, just contact support if they want to roll back?
Should we mention how to get rid of it after they are sure about success of upgrade, or anything like that.

docs/content/preview/yugabyte-platform/upgrade/upgrade-yp-replicated.md
```diff
@@ -52,3 +73,31 @@ If you are performing an upgrade to YugabyteDB Anywhere version 2.14 or later, t
...
+
+3. Log in to the Replicated Admin Console at <https://:8800/> and navigate to **Settings** to load the new license file, as per the following illustration:
+
+   ![Airgap Settings](/images/yp/airgap-settings.png)
+
+   Change the two directories to match the ones you used. For example, enter `/opt/yugabyte/releases/yugaware-{{<yb-version version="stable" format="build">}}/` in the **Update Path** field and `/home/{user}/` in the **License File** field.
+
+   Replicated detects updates based on the updated path information and applies them in the same way it does for connected YugabyteDB Anywhere installations.
+
+4. Proceed with the YugabyteDB Anywhere upgrade process by following instructions provided in [Upgrade using Replicated](#upgrade-using-replicated).
```
aishwarya24, line 101:
```suggestion
1. Proceed with the YugabyteDB Anywhere upgrade process by following instructions provided in [Upgrade using Replicated](#upgrade-using-replicated).
```

docs/content/preview/yugabyte-platform/upgrade/upgrade-yp-replicated.md
```diff
@@ -52,3 +73,31 @@ If you are performing an upgrade to YugabyteDB Anywhere version 2.14 or later, t
...
+
+5. Upgrade your YugabyteDB universe by following instructions provided in [Upgrade the YugabyteDB software](../../manage-deployments/upgrade-software/).
```
aishwarya24, line 103:
```suggestion
1. Upgrade your YugabyteDB universe by following instructions provided in [Upgrade the YugabyteDB software](../../manage-deployments/upgrade-software/).
```

aishwarya24 (approved):
minor comments.

docs/content/preview/yugabyte-platform/upgrade/upgrade-yp-installer.md
chidmuthu: Yeah, we maybe don't even want to tell them it's there. We will GC anything older than the most recent version so at one time the most a customer could see is two versions in the /opt/yugabyte/software directory. But, they are not advised to switch to the old software dir without supports help. Technically if the upgrade doesn't complete we shouldn't swap over to the new software so when they navigate to the running YBA after a failed upgrade they should still see the old version, but in case there are unforeseen issues it would be a support call if upgrade fails and they appear to be upgraded. If upgrade fails and they are still the same version then it's probably they need to fix some issue on the node and rerun upgrade.

ddhodge: OK I added a note to contact support if there are problems and left it at that

docs/content/preview/yugabyte-platform/upgrade/upgrade-yp-replicated.md
```diff
@@ -40,13 +40,15 @@ To perform the migration, you must first upgrade your installation to v2.20.1 or
...
 ## Upgrade using Replicated
 
 To start the upgrade, log in to the Replicated Admin Console via <https://:8800> and then perform the following:
 
 1. Navigate to **Dashboard** and click **View release history** to open **Release History**, as shown in the following illustration:
 
-  ![image](/images/yb-platform/upgrade-replicated1.png)
+    ![Release History](/images/yb-platform/upgrade-replicated1.png)
 
 1. Find the required release version and click the corresponding **Install**.
```
aishwarya24, line 53: Indent the text after this point. "If the required version is not in the list, cli....."

aishwarya24 (commented):
minor comment

PR #20733, opened 2024-01-22
[Docs] LangChain / OpenAI Tutorial

BrettHoyer:


dmagda:
@ddhodge @BrettHoyer 

Could you please do a hot fix?

1. I would put the "Start the Flask server" section before the "Review the application". Justification: show how the app works first so that the readers can experience the app. And only then explain internals in the "Review Application ").
2. Also, I would rename: Start the Flask Server => Start the App

PR #21050, opened 2024-02-13
Add a link to the xcluster limitations in the "explore" section of xC…

lingamsandeep:
…luster replication.

Add a link to the limitations in the "explore" section of xCluster replication.


ddhodge (approved):
LGTM

PR #21067, opened 2024-02-14
[docs] added note for tableplus integration 

aishwarya24:
Slack request - https://yugabyte.slack.com/archives/CDHEH3N64/p1707913262385649?thread_ts=1707781745.782159&cid=CDHEH3N64

PR #21108, opened 2024-02-20
Add JSONB exception to Tableplus usage

premkumr:
- JSONB is not supported via the default Cassandra driver


premkumr:
Duplicate PR. #21067

PR #21179, opened 2024-02-23
Update persistence.md

rthallamko3:
Make it explicit that the Packed Row flag for YSQL is on by default for new installs starting 2.20.0


ddhodge (approved):
LGTM

Issue #19511, opened 2023-10-12
[DocDB] Track memory for traces when enable_tracing is turned on


rthallamko3:
### Description

When tracing is enabled by turning on the gflag - enable_tracing, significant amount of memory is used up by traces (Proportional to the number of active connections/RPCs). This memory is untracked and can appear as a leak. Its better to track the memory.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

PR #21185, opened 2024-02-24
[#19511] Track Trace class memory  when enable_tracing is turned on 

ztalarick:
issue [#19511](https://github.com/yugabyte/yugabyte-db/issues/19511)

### DIFF
--- a/src/yb/util/memory/arena.cc
+++ b/src/yb/util/memory/arena.cc
@@ -221,7 +221,9 @@ void ArenaBase<Traits>::AddComponentUnlocked(Buffer buffer, Component* component
   if (!second_ && component->next()) {
     second_ = component;
   }
+  size_t bytes_before = arena_footprint_;
   arena_footprint_ += component->full_size();
+  TrackMemoryDifference(bytes_before, arena_footprint_);
   if (PREDICT_FALSE(arena_footprint_ > FLAGS_arena_warn_threshold_bytes) && !warned_) {
     LOG(WARNING) << "Arena " << reinterpret_cast<const void *>(this)
                  << " footprint (" << arena_footprint_ << " bytes) exceeded warning threshold ("
@@ -253,13 +255,17 @@ void ArenaBase<Traits>::Reset(ResetMode mode) {
     // detection of memory-related bugs (invalid shallow copies, etc.).
     size_t last_size = current->full_size();
     current->Destroy(buffer_allocator_);
+    UntrackMemory(arena_footprint_);
     arena_footprint_ = 0;
     ReleaseStoreCurrent(nullptr);
     AddComponentUnlocked(NewBuffer(last_size, 0));
 #else
+    size_t bytes_before = arena_footprint_;
     arena_footprint_ = current->full_size();
+    TrackMemoryDifference(bytes_before, arena_footprint_);
 #endif
   } else {
+    UntrackMemory(arena_footprint_);
     arena_footprint_ = 0;
   }
 }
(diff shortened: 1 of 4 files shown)


src/yb/util/memory/arena.h
```diff
@@ -284,6 +287,11 @@ class ArenaBase {
   // allocated and added to the arena's list. Also covers any other
   // mutation of the component data structure (eg Reset).
   mutable mutex_type component_lock_;
+
+  MemTrackerPtr mem_tracker_ = nullptr;
+  void TrackMemory(size_t bytes);
+  void UntrackMemory(size_t bytes);
+  void TrackMemoryDifference(size_t bytes_before, size_t bytes_after);
```
yusong-yan, line 294: I just realized that `MemTracker::Consume` can also take a negative integer as input, so in this case, we don't need this member function anymore.

src/yb/util/memory/arena.h
```diff
@@ -284,6 +287,11 @@ class ArenaBase {
   // allocated and added to the arena's list. Also covers any other
   // mutation of the component data structure (eg Reset).
   mutable mutex_type component_lock_;
+
+  MemTrackerPtr mem_tracker_ = nullptr;
+  void TrackMemory(size_t bytes);
+  void UntrackMemory(size_t bytes);
```
yusong-yan, line 293: Can we rename these two functions to `ConsumeMemory` and `ReleaseMemory()`.

src/yb/util/memory/arena.cc
```diff
@@ -253,13 +255,17 @@ void ArenaBase<Traits>::Reset(ResetMode mode) {
...
     size_t last_size = current->full_size();
     current->Destroy(buffer_allocator_);
+    UntrackMemory(arena_footprint_);
     arena_footprint_ = 0;
     ReleaseStoreCurrent(nullptr);
     AddComponentUnlocked(NewBuffer(last_size, 0));
 #else
+    size_t bytes_before = arena_footprint_;
     arena_footprint_ = current->full_size();
+    TrackMemoryDifference(bytes_before, arena_footprint_);
```
yusong-yan, line 266: Replace by `TrackMemory`. Also remember to use `static_cast<int64_t>` when substracting the before and after value as the result might be negative.

src/yb/util/memory/arena.h
yusong-yan, line 294: In C++, the common convention is to define member functions before member variables in a class definition.

src/yb/util/trace.h
```diff
@@ -283,6 +284,12 @@ class Trace : public RefCountedThreadSafe<Trace> {
   std::vector<scoped_refptr<Trace> > child_traces_;
 
   DISALLOW_COPY_AND_ASSIGN(Trace);
+
+  MemTrackerPtr mem_tracker_ = nullptr;
+  void init_mem_tracker();
+  void set_arena_memtracker();
+  void track_tracer_memory(size_t bytes);
+  void untrack_tracer_memory(size_t bytes);
```
yusong-yan, line 292: ConsumeMemory and ReleaseMemory

src/yb/util/memory/arena.cc
```diff
@@ -253,13 +255,17 @@ void ArenaBase<Traits>::Reset(ResetMode mode) {
...
 #endif
   } else {
+    UntrackMemory(arena_footprint_);
```
yusong-yan, line 269: Btw, we should also release `arena_footprint_` memory at destructor `ArenaBase<Traits>::~ArenaBase()`.

src/yb/util/memory/arena.cc
```diff
@@ -221,7 +221,9 @@ void ArenaBase<Traits>::AddComponentUnlocked(Buffer buffer, Component* component
   if (!second_ && component->next()) {
     second_ = component;
   }
+  size_t bytes_before = arena_footprint_;
   arena_footprint_ += component->full_size();
+  TrackMemoryDifference(bytes_before, arena_footprint_);
```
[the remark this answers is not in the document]
ztalarick: is component->full_size(); negative when the arena gets smaller?

src/yb/util/memory/arena.cc
ztalarick: I am a little confused here. Is that what is being done?

src/yb/util/memory/arena.cc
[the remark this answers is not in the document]
yusong-yan: component->full_size() cannot be negative be cause its return type is size_t, which is unsigned integer.

More high level, you can think `current_` in ArenaBase as a linklist head. And this function basically adds a new node to the linklist, thus we need to increase the memory consumption.

src/yb/util/memory/arena.cc
yusong-yan: In `ArenaBase<Traits>::~ArenaBase()` destructor, `AcquireLoadCurrent()->Destroy(buffer_allocator_);` deallocates all components(inside of this function, it usea while loop to delete all nodes from the linkedlist), so we should call `UntrackMemory(arena_footprint_);` in the destructor.

PR #21242, opened 2024-02-29
[Docs] Change underscore with hyphen in Algolia search query

samiahmedsiddiqui:
Whenever we have an underscore in the search term, convert it to hypen and send to alogolia. So searching for `gflags` will result only in the pages that they are in.

This change is toggleable by changing `algolia_sequence_expressions` parameter from the configuration.
https://github.com/yugabyte/yugabyte-db/pull/21242/files#diff-b2bf863a6d69d3893c459b482a68fa340b0d93542358c0993d9ac79cb7caad05R6


premkumr (approved):
LGTM

ddhodge (approved):

PR #21483, opened 2024-03-14
Added ‘Migrate from PostgreSQL’ in Quickstart with summary and linking to Voyager

ymahajan:
Added ‘Migrate from PostgreSQL’ in Quickstart with summary and linking to Voyager


aishwarya24 (approved):
LGTM

PR #21671, opened 2024-03-25
[Docs] Repeated tabs with the same look

samiahmedsiddiqui:
Allow multiple tabs to show as same size, not the different size as shown here:
https://docs.yugabyte.com/preview/explore/fault-tolerance/handling-region-failures/#third-region-failure

Add `repeatedTabs="true"` parameter on the `nav/tabs` shortcode where you like to have the tabs to be same size.

Current shortcode example:
```
{{<nav/tabs list="local,anywhere" active="local"/>}}
```

Shortcode with repeated tabs example:

```
{{<nav/tabs list="local,anywhere" active="local" repeatedTabs="true" />}}
```

### DIFF
--- a/docs/assets/scss/_yb_tabs.scss
+++ b/docs/assets/scss/_yb_tabs.scss
@@ -185,7 +185,7 @@ ul.nav.nav-tabs-yb:not(.yb-pills) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


premkumr (approved):
LGTM

PR #21885, opened 2024-04-08
Updated the interpretation of log_stop_retaining_min_disk_mb

### LINKED ISSUE #21849: [DocDB] Flag log_stop_retaining_min_disk_mb is internally used as kb
### Description

This gflag has unit mb, but it's internally interpreted as kb. Thus, the mismatch can lead to unexpected behavior when users try to modify it value.

### Issue Type

kind/bug

AyoubKaz07:
### Description
The gflag log_stop_retaining_min_disk_mb was internally interpreted as kilobytes (kb) instead of megabytes (mb), leading to a unit mismatch. This inconsistency could result in unexpected behavior when users attempted to modify the value of this flag.

### Changes
Updated the interpretation of log_stop_retaining_min_disk_mb to use megabytes internally to align with its specified unit.

Fixes: #21849, #19391

### DIFF
--- a/docs/content/preview/reference/configuration/yb-tserver.md
+++ b/docs/content/preview/reference/configuration/yb-tserver.md
@@ -1167,7 +1167,7 @@ Default: `86400`
 
 Stop retaining logs if the space available for the logs falls below this limit, specified in megabytes. As with `log_max_seconds_to_retain`, this flag is ignored if a log segment contains unflushed entries.
 
-Default: `102400`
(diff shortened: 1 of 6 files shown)


hari90 (changes requested):
We cannot land this. Customers use have overridden this flag with the value in kb. They will be broken due to this change.

Best we can do is update the flag description, and deprecate this and add a new flag with the proper name.

PR #21912, opened 2024-04-10
Architecture revamp - Phase 2

premkumr:
- Removing core functions
- Better Hybrid Clocks
- More Concepts
- Raft
- CDC


mdbridge:
its Raft not RAFT

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Log replication
+
+The leader is responsible for replicating log entries to the followers. When a client sends a request to the leader, the leader appends the request to its local log and then sends the log entry to the followers. The followers then append the log entry to their local logs and send an acknowledgment back to the leader. Once the leader receives a majority of acknowledgments, it considers the log entry to be committed and can apply it to its state machine.
+
+## Replication of the write operation
+
+All write operations are handled by the tablet leader. These are the squence of operations that happen during the replication of a write.
+
+1. The leader first takes a lock on the key using a local in-memory lock manager and reads data if necessary (for read-modify-write or conditional update operations). Note that this locking mechanism does not exist on followers.
```
mdbridge, line 48: actually many locks have to be taken as we have to lock the prefixes usually
for snapshot isolation there's also some validation required to make sure nothing has already been written between the read time and the current time (simplifying)

 this is also assuming single-shard transaction's so no intents have to be written as part of the locking
 we do however, have to check for intents from other distributed transactions regardless

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+1. The batch of changes to be written to [DocDB](../../docdb/) is prepared. This write batch is very close to the final set of RocksDB key-value pairs to be written, only lacking the final [hybrid timestamp](../../transactions/transactions-overview#hybrid-logical-clocks) at the end of each key.
+1. The leader appends the batch to its Raft log and chooses a [hybrid timestamp](../../transactions/transactions-overview#hybrid-logical-clocks) for the write operation.
+1. The data is replicated via Raft to its peers.
+1. After receiving a `replication successful` callback from the Raft subsystem, the leader applies the write batch to its local DocDB.
```
mdbridge, line 52: I think you want to say something about using the hybrid timestamp of the log entry as the timestamps for the batch entries

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+## Log replication
+
+The leader is responsible for replicating log entries to the followers. When a client sends a request to the leader, the leader appends the request to its local log and then sends the log entry to the followers. The followers then append the log entry to their local logs and send an acknowledgment back to the leader. Once the leader receives a majority of acknowledgments, it considers the log entry to be committed and can apply it to its state machine.
+
+## Replication of the write operation
+
+All write operations are handled by the tablet leader. These are the squence of operations that happen during the replication of a write.
+
+1. The leader first takes a lock on the key using a local in-memory lock manager and reads data if necessary (for read-modify-write or conditional update operations). Note that this locking mechanism does not exist on followers.
+1. The batch of changes to be written to [DocDB](../../docdb/) is prepared. This write batch is very close to the final set of RocksDB key-value pairs to be written, only lacking the final [hybrid timestamp](../../transactions/transactions-overview#hybrid-logical-clocks) at the end of each key.
```
mdbridge, line 49: you probably want to simplify this but one of the batches for the log can include writes from many concurrent write requests

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+1. The leader appends the batch to its Raft log and chooses a [hybrid timestamp](../../transactions/transactions-overview#hybrid-logical-clocks) for the write operation.
+1. The data is replicated via Raft to its peers.
+1. After receiving a `replication successful` callback from the Raft subsystem, the leader applies the write batch to its local DocDB.
+1. Responds with success to the user.
```
mdbridge, line 53: I'm not actually sure if we respond with success before or after writing to local DB
the data is durable as soon as the log entry is replicated on a majority of participants

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+1. The next update from the leader notifies followers that the entry has been committed, and the followers apply the write batch to their RocksDB instances.
+
+## Leader election
+
+When the system starts up, each node is assigned an initial election term, which is typically a non-negative integer. When a  follower does not receive a heartbeat within the election timeout period it becomes a candidate to start a new election term and attempt to become the new Leader. Each node votes for a candidate in a particular election term. The candidate that receives votes from a majority of the nodes is elected as the new leader. Whenever a new leader is elected, the election term is incremented. This ensures that each term has a unique identifier, and it helps the nodes in the system differentiate between the various terms.
```
mdbridge, line 58: sometimes there isn't a majority winner so there isn't a winner in that term.  This will result in a new election being called for the next term

mdbridge, line 58: it's more like when the tablet (read Raft group here) is created, it starts with a term of (I don't know which) 0 or 1.    The last term number is persistent and survives restarts.

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+
+If a node receives a request from a leader with a stale term (i.e., a term that is lower than the node's current term), the node will reject the request and inform the leader of the newer term. This helps prevent outdated leaders from making changes to the replicated log. RAFT ensures that the replicated log remains consistent across the nodes by requiring that all log entries be associated with a specific term. This means that a node will only accept log entries that are from the current term or a previous term
```
mdbridge, line 60: ensures is too strong here as a lot of other things are required as well.
the last sentence is just wrong.  A node that has been down for a while when coming up will get caught up with log records for everything it missed, which can include log records from quite a few different terms.

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+
+The Leader continues to send heartbeat messages to maintain its leadership status. If a Follower does not receive a heartbeat within the election timeout period, a new election is triggered.
+
+## Consistency guarantees
+
+RAFT ensures that the replicated log remains consistent across all the nodes in the cluster. It achieves this by enforcing a rule that the leader can only append new entries to the log, and it must ensure that those entries are consistent with the log of the majority of the nodes. If a follower's log is not consistent with the leader's log, the leader will override the follower's log with its own.
```
mdbridge, line 66: cluster->universe.  The read replicas are also part of the Raft groups
only append->append only
the leader can remove entries at the end of its log that are not majority replicated
new entries don't have consistency issues with the nodes because they won't be shared by those nodes


actually, I'm not sure what you mean by "append new entries to the log".  You mean to the local log of the leader or to the virtual log defined by the majority replicated entries in all the nodes?

this paragraph needs work

docs/content/preview/architecture/docdb-replication/raft.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Fault tolerance
+
+RAFT is designed to be fault-tolerant, meaning that it can continue to operate even if some of the nodes in the cluster fail. As long as a majority of the nodes in the cluster are alive, the system can continue to make progress and maintain the consistency of the replicated log. If the leader fails, a new leader will be elected, and the system will continue to operate without interruption.
```
mdbridge, line 70: please don't use the word cluster here.  We cannot tolerate the loss of a minority of the nodes in a cluster, only the loss of a minority of nodes in every Raft group.
availability here requires the majority of nodes are able to talk to each other
interruption is a tricky word; there is a several second hiccup when leaders failed in an unplanned way

mdbridge: that is, we are okay as long as there is no raft group for which a majority of nodes are lost

docs/content/preview/architecture/docdb-replication/raft.md
mdbridge: actually this should be Raft group instead of cluster as well
I tend to use the word participant for the members of a raft group rather than nodes

aishwarya24 (approved):
LGTM. Pushed a bunch of edits.

PR #21929, opened 2024-04-11
Enhancing copy throughput by retrieving datums with checkpointing available in TupleTableSlot

naanagon:
In the sample table we have almost all char columns and few varchar columns for both of these data types attlen in pg_attribute table is -1.  Since we have varlen attributes in nocachegetattr function it is going to this part of the code [(nocachecode,](https://github.com/naanagon/yugabyte-db/blob/master/src/postgres/src/backend/access/common/heaptuple.c#L607) [ybcModifyTable](https://github.com/yugabyte/yugabyte-db/blob/master/src/postgres/src/backend/executor/ybcModifyTable.c#L326)) . In this way the time complexity is O(n*n) where n is number of attributes.

there is an another for loop in ([nocachecode](https://github.com/naanagon/yugabyte-db/blob/master/src/postgres/src/backend/access/common/heaptuple.c#L607)) as it is doing the same work every time, whey can't we build a prefix sum array once for the row and just use that for all the attributes.

### DIFF
--- a/src/postgres/src/backend/bootstrap/bootstrap.c
+++ b/src/postgres/src/backend/bootstrap/bootstrap.c
@@ -823,9 +823,13 @@ InsertOneTuple(Oid objectid)
 	tuple = heap_form_tuple(tupDesc, values, Nulls);
 	if (objectid != (Oid) 0)
 		HeapTupleSetOid(tuple, objectid);
-
 	if (IsYugaByteEnabled())
-		YBCExecuteInsert(boot_reldesc, tupDesc, tuple, ONCONFLICT_NONE);
+	{
+		TupleTableSlot *slot = MakeSingleTupleTableSlot(tupDesc);
+		ExecStoreHeapTuple(tuple, slot, false);
+		YBCExecuteInsert(boot_reldesc, slot, ONCONFLICT_NONE);
+		ExecDropSingleTupleTableSlot(slot);
+	}
 	else
 		simple_heap_insert(boot_reldesc, tuple);
 
--- a/src/postgres/src/backend/commands/createas.c
+++ b/src/postgres/src/backend/commands/createas.c
@@ -627,8 +627,7 @@ intorel_receive(TupleTableSlot *slot, DestReceiver *self)
 	if (IsYBRelation(myState->rel))
 	{
 		YBCExecuteInsert(myState->rel,
-						 RelationGetDescr(myState->rel),
-						 tuple,
+						 slot,
 						 ONCONFLICT_NONE);
 	}
 	else
--- a/src/postgres/src/backend/commands/matview.c
+++ b/src/postgres/src/backend/commands/matview.c
@@ -486,8 +486,7 @@ transientrel_receive(TupleTableSlot *slot, DestReceiver *self)
 	if (IsYBRelation(myState->transientrel))
 	{
 		YBCExecuteInsert(myState->transientrel,
-						 RelationGetDescr(myState->transientrel),
-						 tuple,
+						 slot,
 						 ONCONFLICT_NONE);
 	}
 	else
--- a/src/postgres/src/backend/commands/tablecmds.c
+++ b/src/postgres/src/backend/commands/tablecmds.c
@@ -5419,8 +5419,7 @@ ATRewriteTable(AlteredTableInfo *tab, Oid OIDNewHeap, LOCKMODE lockmode)
 			{
 				if (IsYBRelation(newrel))
 					YBCExecuteInsert(newrel,
-									 RelationGetDescr(newrel),
-									 tuple,
+					                 newslot,
 									 ONCONFLICT_NONE);
 				else
 					heap_insert(newrel, tuple, mycid, hi_options, bistate);
@@ -18211,8 +18210,7 @@ YbATCopyTableRowsUnchecked(Relation old_rel, Relation new_rel,
 		ExecStoreHeapTuple(tuple, newslot, false);
 
 		/* Write the tuple out to the new relation */
-		YBCExecuteInsert(new_rel, newslot->tts_tupleDescriptor, tuple,
-						 ONCONFLICT_NONE);
+		YBCExecuteInsert(new_rel, newslot, ONCONFLICT_NONE);
 
 		MemoryContextReset(econtext->ecxt_per_tuple_memory);
 
--- a/src/postgres/src/backend/executor/nodeModifyTable.c
+++ b/src/postgres/src/backend/executor/nodeModifyTable.c
@@ -560,7 +560,7 @@ ExecInsert(ModifyTableState *mtstate,
 				 * locked and released in this call.
 				 * TODO(Mikhail) Verify the YugaByte transaction support works properly for on-conflict.
 				 */
-				newId = YBCHeapInsert(slot, tuple, blockInsertStmt, estate);
+				newId = YBCHeapInsert(slot, blockInsertStmt, estate);
 
 				/* insert index entries for tuple */
 				recheckIndexes = ExecInsertIndexTuples(slot, tuple,
@@ -629,7 +629,7 @@ ExecInsert(ModifyTableState *mtstate,
 			if (IsYBRelation(resultRelationDesc))
 			{
 				MemoryContext oldContext = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
-				newId = YBCHeapInsert(slot, tuple, blockInsertStmt, estate);
+				newId = YBCHeapInsert(slot, blockInsertStmt, estate);
 
 				/* insert index entries for tuple */
 				if (YBCRelInfoHasSecondaryIndices(resultRelInfo))
@@ -1573,7 +1573,8 @@ ExecUpdate(ModifyTableState *mtstate,
 		ModifyTable *plan = (ModifyTable *) mtstate->ps.plan;
 		if (is_pk_updated)
 		{
-			YBCExecuteUpdateReplace(resultRelationDesc, planSlot, tuple, estate);
+			slot->tts_tuple->t_ybctid = YBCGetYBTupleIdFromSlot(planSlot);
+			YBCExecuteUpdateReplace(resultRelationDesc, slot, estate);
 			row_found = true;
 		}
 		else
--- a/src/postgres/src/backend/executor/ybcModifyTable.c
+++ b/src/postgres/src/backend/executor/ybcModifyTable.c
@@ -265,13 +265,15 @@ static void YBCExecWriteStmt(YBCPgStatement ybc_stmt,
 }
 
 static Oid YBCApplyInsertRow(
-    YBCPgStatement insert_stmt, Relation rel, TupleDesc tupleDesc,
-    HeapTuple tuple, OnConflictAction onConflictAction, Datum *ybctid,
+    YBCPgStatement insert_stmt, Relation rel, TupleTableSlot *slot, 
+    OnConflictAction onConflictAction, Datum *ybctid,
     YBCPgTransactionSetting transaction_setting) {
 	Oid            relfileNodeId    = YbGetRelfileNodeId(rel);
 	AttrNumber     minattr          = YBGetFirstLowInvalidAttributeNumber(rel);
 	int            natts            = RelationGetNumberOfAttributes(rel);
 	Bitmapset      *pkey            = YBGetTablePrimaryKeyBms(rel);
+	TupleDesc	   tupleDesc        = RelationGetDescr(rel);
+	HeapTuple 	   tuple            = slot->tts_tuple;
 
 	/* Generate a new oid for this row if needed */
 	if (rel->rd_rel->relhasoids)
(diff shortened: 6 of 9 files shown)


src/postgres/src/backend/access/common/heaptuple.c
```diff
@@ -657,6 +657,217 @@ nocachegetattr(HeapTuple tuple,
...
+		 * attribute.  If we have a cached offset, we can use it.
+		 */
+		att = TupleDescAttr(tupleDesc, attnum);
+		if (att->attcacheoff >= 0)
+			return fetchatt(att, tp + att->attcacheoff);
+
+		/*
+		 * Otherwise, check for non-fixed-length attrs up to and including
+		 * target.  If there aren't any, it's safe to cheaply initialize the
+		 * cached offsets for these attrs.
```
jasonyb, line 734: It appears that, ever since the beginning as far as Git history shows, caching attr offsets was disabled for variable-length attributes.  I don't know why exactly that is the case, but it calls into question whether your caching is safe.

src/postgres/src/backend/access/common/heaptuple.c
```diff
@@ -657,6 +657,217 @@ nocachegetattr(HeapTuple tuple,
...
+		 */
+		if (HeapTupleHasVarWidth(tuple))
+		{
+			int			j;
+
+			for (j = 0; j <= attnum; j++)
+			{
+				if (TupleDescAttr(tupleDesc, j)->attlen <= 0)
```
jasonyb, line 742: Supposing it is safe to cache varlen offsets for the case you are targetting, I think a cleaner solution is to reuse the existing attcacheoff rather than duplicate code.  You could add a YB exception here (perhaps by a boolean argument to the function) so that slow is not set to true, and the caller can reset these attcacheoff back to -1 when done?  Note I haven't given the idea much thought and there could be holes.

src/postgres/src/backend/access/common/heaptuple.c
```diff
@@ -657,6 +657,217 @@ nocachegetattr(HeapTuple tuple,
...
+		int			j = 1;
+
+		/*
+		 * If we get here, we have a tuple with no nulls or var-widths up to
+		 * and including the target attribute, so we can use the cached offset
+		 * ... only we don't have it yet, or we'd not have got here.  Since
+		 * it's cheap to compute offsets for fixed-width columns, we take the
+		 * opportunity to initialize the cached offsets for *all* the leading
+		 * fixed-width columns, in hope of avoiding future visits to this
+		 * routine.
```
jasonyb, line 763: It seems this is the O(n^2) behavior in case all your columns are not fixed-width.

jasonyb (commented):
It seems the core issue is with upstream PG code, unless this access pattern by YB is discouraged by upstream PG:

```
for (AttrNumber attnum = minattr; attnum <= natts; attnum++)
	column->datum = heap_getattr(tuple, attnum, tupleDesc, &column->is_null);
```

src/postgres/src/backend/access/common/heaptuple.c
naanagon: Since they are varlen attributes the offsets of attributes will changes from one tuple and another tuple and i think that is reason offsets aren't cached.

src/postgres/src/backend/access/common/heaptuple.c
naanagon: slow becomes true as we have varlen attributes. This is the loop that is causing O(n^2) - [link](https://github.com/yugabyte/yugabyte-db/pull/21929/files/59af390717316ddf7655db232c94a54d3f35ef6d#diff-7c28cbfd32f6fe08301e4f1386e4437ad7c84f538434b9676e668564d3861be0R607)

src/postgres/src/backend/access/common/heaptuple.c
naanagon: It is safe to cacheoffsets for each row. Even i was thinking in the same line but i was stuck on how to reset to -1  only for this varlen case at ybcModifyTable level.  But if we throw exception then that requires changes at many places which is hard for now to review.. Let me also rethink on this.

naanagon (commented):
comments

andrei-mart:
It is correctly pointed out that repeatable calls to heap_getattr are generally inefficient.
However, in cases like this, when all or most of values are fetched, you can simply use heap_deform_tuple.
Same or even better efficiency as heap_getattr with cache, no need to add anything to the library.

naanagon:
> It is correctly pointed out that repeatable calls to heap_getattr are generally inefficient. However, in cases like this, when all or most of values are fetched, you can simply use heap_deform_tuple. Same or even better efficiency as heap_getattr with cache, no need to add anything to the library.

Thanks andrei for taking a quick look. Based on our suggestion given in slack, using TupleTableSlot is the best option. I have updated the pr accordingly. Didn't combine tupDesc and slot because in classes like matview.c the tupleDesc is retrieved from diff struct like DR_transientrel, which i'm thinking may not be updated in slot.

Please lmk if i'm missing anything.

src/postgres/src/backend/bootstrap/bootstrap.c
```diff
@@ -824,8 +824,12 @@ InsertOneTuple(Oid objectid)
 	if (objectid != (Oid) 0)
 		HeapTupleSetOid(tuple, objectid);
 
-	if (IsYugaByteEnabled())
-		YBCExecuteInsert(boot_reldesc, tupDesc, tuple, ONCONFLICT_NONE);
+	if (IsYugaByteEnabled()){
```
andrei-mart, line 827: New line before the opening brace

src/postgres/src/backend/commands/createas.c
```diff
@@ -626,8 +626,9 @@ intorel_receive(TupleTableSlot *slot, DestReceiver *self)
 
 	if (IsYBRelation(myState->rel))
 	{
+		slot->tts_tupleDescriptor=RelationGetDescr(myState->rel);
```
andrei-mart, line 629: ExecSetSlotDescriptor should be used.
However, it is a bad idea to modify passed in  slot. 
Why do we think that slot does not have a descriptor already?

src/postgres/src/backend/commands/matview.c
```diff
@@ -485,8 +485,9 @@ transientrel_receive(TupleTableSlot *slot, DestReceiver *self)
 	tuple = ExecMaterializeSlot(slot);
 	if (IsYBRelation(myState->transientrel))
 	{
+		slot->tts_tupleDescriptor=RelationGetDescr(myState->transientrel);
```
andrei-mart, line 488: Same as above

src/postgres/src/backend/commands/tablecmds.c
```diff
@@ -5417,11 +5417,13 @@ ATRewriteTable(AlteredTableInfo *tab, Oid OIDNewHeap, LOCKMODE lockmode)
 			/* Write the tuple out to the new relation */
 			if (newrel)
 			{
-				if (IsYBRelation(newrel))
+				if (IsYBRelation(newrel)){
```
andrei-mart, line 5420: New line before brace

src/postgres/src/backend/commands/tablecmds.c
```diff
@@ -5417,11 +5417,13 @@ ATRewriteTable(AlteredTableInfo *tab, Oid OIDNewHeap, LOCKMODE lockmode)
...
+					newslot->tts_tupleDescriptor=RelationGetDescr(newrel);
```
andrei-mart, line 5421: Unsafe tuple descriptor assignment again

src/postgres/src/backend/executor/nodeModifyTable.c
```diff
@@ -1573,6 +1573,7 @@ ExecUpdate(ModifyTableState *mtstate,
 		ModifyTable *plan = (ModifyTable *) mtstate->ps.plan;
 		if (is_pk_updated)
 		{
+			planSlot->tts_tupleDescriptor=RelationGetDescr(resultRelationDesc);
```
andrei-mart, line 1576: and again

src/postgres/src/backend/executor/ybcModifyTable.c
```diff
@@ -265,14 +265,14 @@ static void YBCExecWriteStmt(YBCPgStatement ybc_stmt,
...
-    YBCPgStatement insert_stmt, Relation rel, TupleDesc tupleDesc,
+    YBCPgStatement insert_stmt, Relation rel, TupleTableSlot *slot, 
     HeapTuple tuple, OnConflictAction onConflictAction, Datum *ybctid,
     YBCPgTransactionSetting transaction_setting) {
 	Oid            relfileNodeId    = YbGetRelfileNodeId(rel);
 	AttrNumber     minattr          = YBGetFirstLowInvalidAttributeNumber(rel);
 	int            natts            = RelationGetNumberOfAttributes(rel);
 	Bitmapset      *pkey            = YBGetTablePrimaryKeyBms(rel);
-
+	TupleDesc		tupleDesc       = slot->tts_tupleDescriptor;
```
andrei-mart, line 275: if slot->tts_tupleDescriptor is NULL you probably can use RelationGetDescr(rel)

andrei-mart (changes requested):

src/postgres/src/backend/commands/createas.c
naanagon: I haven't used ExecSetSlotDescriptor as this clears the tuple from the slot. 
slot should have descriptor otherwise it won't work in slot_getattr as that retrieves tupleDesc from slot->tts_tupleDescriptor. But, I have done a check

RelationGetDescr(rel)!=slot->tts_tupleDescriptor -> this is true as it has different tdtypeid and tdrefcount. So, I tried to update it.

Based on your comments i have done a recheck and I hope we aren't using tdtypeid and tdrefcount anywhere.

andrei-mart:
```
yugabyte=# CREATE TABLE test (r1 int, r2 int, i int, PRIMARY KEY (r1 ASC, r2 DESC));
CREATE TABLE
yugabyte=# INSERT INTO test VALUES (1, 1, 1), (2, 2, 2), (3, 3, 3);
INSERT 0 3
yugabyte=# INSERT INTO test  VALUES (3, 3, 31) ON CONFLICT (r1, r2) DO UPDATE SET r1 = EXCLUDED.i, r2 = EXCLUDED.i;
INSERT 0 1
yugabyte=# select * from test;
 r1 | r2 | i  
----+----+----
  1 |  1 |  1
  2 |  2 |  2
 31 | 31 | 31
(3 rows)
```
Expected result is 
```
yugabyte=# select * from test;
 r1 | r2 | i  
----+----+---
  1 |  1 | 1
  2 |  2 | 2
 31 | 31 | 3
(3 rows)
```
There are more failures of `./yb_build.sh --java-test 'org.yb.pgsql.TestPgUpdatePrimaryKey'`

src/postgres/src/backend/executor/nodeModifyTable.c
```diff
@@ -1573,7 +1573,8 @@ ExecUpdate(ModifyTableState *mtstate,
 		ModifyTable *plan = (ModifyTable *) mtstate->ps.plan;
 		if (is_pk_updated)
 		{
-			YBCExecuteUpdateReplace(resultRelationDesc, planSlot, tuple, estate);
+			slot->tts_tuple->t_ybctid = YBCGetYBTupleIdFromSlot(planSlot);
+			YBCExecuteUpdateReplace(resultRelationDesc, slot, tuple, estate);
```
naanagon, line 1577: There is a TODO already here to pass a transformed slot. I guess slot is the right option instead of planSlot

src/postgres/src/backend/executor/ybcModifyTable.c
```diff
@@ -265,14 +265,20 @@ static void YBCExecWriteStmt(YBCPgStatement ybc_stmt,
...
-    HeapTuple tuple, OnConflictAction onConflictAction, Datum *ybctid,
+    YBCPgStatement insert_stmt, Relation rel, TupleTableSlot *slot, 
+    OnConflictAction onConflictAction, Datum *ybctid,
     YBCPgTransactionSetting transaction_setting) {
 	Oid            relfileNodeId    = YbGetRelfileNodeId(rel);
 	AttrNumber     minattr          = YBGetFirstLowInvalidAttributeNumber(rel);
 	int            natts            = RelationGetNumberOfAttributes(rel);
 	Bitmapset      *pkey            = YBGetTablePrimaryKeyBms(rel);
+	TupleDesc		tupleDesc       = slot->tts_tupleDescriptor;
+	HeapTuple 	    tuple           = slot->tts_tuple;
```
naanagon, line 276: Based on the definition of ExecMaterializeSlot, I haven't used it. So, ybctid and oid are updated in tuple itself.

src/postgres/src/backend/commands/tablecmds.c
```diff
@@ -18211,7 +18210,7 @@ YbATCopyTableRowsUnchecked(Relation old_rel, Relation new_rel,
 		ExecStoreHeapTuple(tuple, newslot, false);
 
 		/* Write the tuple out to the new relation */
-		YBCExecuteInsert(new_rel, newslot->tts_tupleDescriptor, tuple,
+		YBCExecuteInsert(new_rel, newslot,
```
andrei-mart, line 18213: Nit: the statement is shorter now, can fit into single line.

src/postgres/src/backend/executor/ybcModifyTable.c
```diff
@@ -342,7 +347,7 @@ static Oid YBCApplyInsertRow(
 	if (IsCatalogRelation(rel))
 	{
 		MarkCurrentCommandUsed();
-		CacheInvalidateHeapTuple(rel, tuple, NULL);
+		CacheInvalidateHeapTuple(rel, slot->tts_tuple, NULL);
```
andrei-mart, line 350: If I'm not missing something, local variable tuple is defined here and equal to slot->tts_tuple

andrei-mart (approved):

PR #21949, opened 2024-04-12
[docs] Move Triggers, follower reads and tablespaces in explore section

aishwarya24:
- Moved Triggers to Advanced features
- Moved Follower reads and Tablespaces to multi-region deployments


aishwarya24:
Will create another PR for this.

PR #22077, opened 2024-04-19
[docs] Fix coredb contributor instructions for alma, centos

jaki:
Update CoreDB contributor instructions, which went out of date some time
ago.

- Add patchelf package to installation instructions.  It was always
  required, but what likely happened is that it used to be implicitly
  installed as part of a dependency of a different required package, and
  that is no longer the case today.
- Note that there may be compilation issues with centos + gcc.


ddhodge (approved):
LGTM

PR #22114, opened 2024-04-23
Making product text non-breaking

premkumr:
- Previously "YB Anywhere" could be broken into 2 lines when displayed. 
- This will ensure they come together in one line
- Can be seen in the TA page


ddhodge:
This is fine - but we don't typically use 'YB Anywhere' - it's either YugabyteDB Anywhere or YBA

PR #22124, opened 2024-04-23
Change the basedir to ${HOME}/var/node from /tmp/ybd

premkumr:
- update setup/local shortcode
- added cluster/cmd shortcode for stop & destroy
```
# to stop 3 nodes
{{%cluster/cmd op="stop"%}}

# to stop node 2
{{%cluster/cmd op="stop" nodes="2"%}}

# to destroy node 3
{{%cluster/cmd op="destroy" nodes="3"%}}
```
- updated multiple files to use these shortcodes


aishwarya24 (approved):
LGTM

PR #22130, opened 2024-04-24
[Docs] Update website responsive behavior

samiahmedsiddiqui:
When resizing the window with the left nav collapsed, the sizing code narrows the text column while there is still a lot of whitespace. The text column should never exceed its max width for readability, but the resizing could be more optimal.
docs-resize


ddhodge (approved):
LGTM

PR #22220, opened 2024-05-01
Move work_mem away from pg_hints

premkumr:
- From @FranckPachot : `work_mem` is an execution time parameter but hints apply only on planning phase


aishwarya24 (approved):
LGTM

FranckPachot (commented):
This is not correct. This doc page is about hints. What we need is show an example of using the 'set' hint with a parameter that impacts the query planner. `work_mem` has nothing to do in this page because it is an execution-time parameter.

To summarize, the issue
- is not about setting `work_mem` in one way or another
- it is about setting a query planner parameter

PR #22313, opened 2024-05-08
[docs] PackedRow gflag documentation updates

rthallamko3:
PackedRow gflag documentation updates - Update the gflags page to indicate that packed row is GA. It was updated to GA in 2.20 timeframe, We missed updating the gflags page.


aishwarya24 (approved):
LGTM

PR #22385, opened 2024-05-14
[Docs] Remove extra space when H3/H4 is hidden on rightnav

samiahmedsiddiqui:
There is an extra space when H3/H4 is hidden on the rightnav.
Screenshot 2024-05-10 at 1 34 12 PM


premkumr (approved):
LGTM

Issue #17751, opened 2023-06-13
[CDCSDK] Turn flag: enable_update_local_peer_min_index to true by default


adithya-yb:
Jira Link: [DB-6847](https://yugabyte.atlassian.net/browse/DB-6847)
The flag: "enable_update_local_peer_min_index" was turned to false by default since CDC was facing an exception . But the original issue has been fixed, now we need to turn this flag on by default, since this enables an optimisation which reduces the amount of RPCs , and hence network load on clusters using xCluster / CDC. 

[DB-6847]: https://yugabyte.atlassian.net/browse/DB-6847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #22397, opened 2024-05-15
[Docs] Remove enable_update_local_peer_min_index from docs

hari90:
enable_update_local_peer_min_index has been enabled by default since 2.18 (#17751). The docs pages have an outdated default value. Further this flag is no longer useful to be tweaked by the customer, and in recent versions setting it to false may have unintended consequences. Hence removing references to the flag.


dr0pdb (approved):

ddhodge (approved):
LGTM

PR #22424, opened 2024-05-16
[DOC-319] [2024.1] Docs for tablet limits

aishwarya24:
Deploy preview: https://deploy-preview-22424--infallible-bardeen-164bc9.netlify.app/preview/architecture/docdb-sharding/tablet-splitting/#tablet-limits


docs/content/preview/reference/configuration/yb-master.md
```diff
@@ -355,6 +355,29 @@ Each tablet replica generally requires 700 MiB of this memory.
...
+
+The number of tablet replicas that each core on a YB-TServer can support. This is used by both YB-Masters and YB-TServers.
+
+Default: `0` for no limit.
+
+##### tablet_replicas_per_gib_limit
+
+The number of tablet replicas that each GiB reserved by YB-TServer for tablet overheads can support. This is used by both YB-Masters and YB-TServers.
+
+Default: 1024 * (7/10) (corresponding to an overhead of roughly 700 KiB per tablet)
```
druzac, line 380: I'm not sure these belong here. The text 

> Enables/disables the feature. No limits at all will be enforced if this is false.

doesn't make sense in this context - what feature? I'd perhaps move these to the `Sharding flags` section. Or possibly the "tablet splitting" section so the flag layout is parallel to the feature description layout, but tablet splitting doesn't quite feel right.

Also the text for the `enforce_tablet_replica_limits` flag should be something like - 
`Enables/disables blocking of requests which would bring the total number of tablets in the system over a limit. See <link> for more information.`

docs/content/preview/reference/configuration/yb-master.md
```diff
@@ -355,6 +355,29 @@ Each tablet replica generally requires 700 MiB of this memory.
...
+
+##### tablet_replicas_per_core_limit
+
+The number of tablet replicas that each core on a YB-TServer can support. This is used by both YB-Masters and YB-TServers.
+
+Default: `0` for no limit.
+
+##### tablet_replicas_per_gib_limit
+
+The number of tablet replicas that each GiB reserved by YB-TServer for tablet overheads can support. This is used by both YB-Masters and YB-TServers.
```
druzac, line 378: nit:
YB-TServer -> YB-TServers

docs/content/preview/reference/configuration/yb-tserver.md
```diff
@@ -314,6 +314,17 @@ Each tablet replica generally requires 700 MiB of this memory.
 
 Default: `0` unless [`--use_memory_defaults_optimized_for_ysql`](#use-memory-defaults-optimized-for-ysql) is true.
 
+##### tablet_replicas_per_core_limit
+
+The number of tablet replicas that each core on a YB-TServer can support. This is used by both YB-Masters and YB-TServers.
```
druzac, line 319: I hadn't realized we have a page for yb-master flags and a page for yb-tserver flags. In that case, we should probably remove the last sentence:

> This is used by both YB-Masters and YB-TServers.

Here and elsewhere.

druzac (changes requested):

docs/content/preview/reference/configuration/yb-master.md
aishwarya24: Moving it to sharding flags.

docs/content/preview/architecture/docdb-sharding/tablet-splitting.md
```diff
@@ -313,6 +313,67 @@ In the following example, a three-node cluster is created and uses a YCSB worklo
...
+create table foo (i int primary key, j int) split into 10 tablets;
+```
+
+```output
+ERROR:  Invalid table definition: Error creating table yugabyte.foo on the master: The requested number of tablet replicas (30) would cause the total running tablet replica count (102) to exceed the safe system maximum (93)
+```
+
+As a best practice, YugabyteDB has pre-computed sensible defaults for the memory limiting flags. By setting the [use_memory_defaults_optimized_for_ysql](../../../reference/configuration/yb-tserver/#use-memory-defaults-optimized-for-ysql) flag to true, the [tablet_overhead_size_percentage](../../../reference/configuration/yb-tserver/#tablet-overhead-size-percentage) is set with a sensible default as well. So, to use the memory limits, the only flag that needs to be set is [enforce_tablet_replica_limits](../../../reference/configuration/yb-master/#enforce-tablet-replica-limits).
+
+It is recommended to use the pre-computed memory defaults. Setting `use_memory_defaults_optimized_for_ysql` reserves memory for PostgreSQL, which is wasteful if not using YSQL. In this case, set the `tablet_overhead_size_percentage` flag to 10 on all YB-TServers.
```
druzac, line 375: This is in the `Example` section currently. In my template it's in the `Best Practices` section. Is this change deliberate?

aishwarya24: I just added it as a conclusion for the topic.. but I see that you are viewing it as part of the example section, so ill update it accordingly.

aishwarya24:
Created PR - https://github.com/yugabyte/yugabyte-db/pull/22569

PR #22428, opened 2024-05-16
Update point-in-time-recovery limitations

druzac:
Add a blurb to the PITR documentation discouraging users from issuing DDLs while a PITR is in flight.


premkumr (approved):
LGTM

ddhodge:
@druzac should this not go in preview also? What about earlier versions?

PR #22530, opened 2024-05-24
DOC-65-Transaction-prepare-bind-syntax-errors

jasonriddell:
I corrected these typos.

> 1. the example won't compile, it is missing a ';'..' - I believe that to be to complete the String.format - requiring: tablename, key2, value2); <- closing ')'
> 
> 2.And a comment about 'k2' not previously being used in the BoundStatement - requiring the earlier String.format second " INSERT INTO.." to be " INSERT INTO %s (k, v) VALUES (:k2, :v2);" +. <- using the 'k2' value used here.


ddhodge:
Replaced by #22592

Issue #18360, opened 2023-07-21
[YSQL] Wrong results for IndexScan yb_hash_code(i) > -1


jasonyb:
Jira Link: [DB-7345](https://yugabyte.atlassian.net/browse/DB-7345)
### Description

Yet another yb_hash_code bug (see issue #18347).

Using same schema as before,

```sql
CREATE TABLE tt (i int, j int);
CREATE INDEX ON tt (i, j);
INSERT INTO tt VALUES (1, 2);
```

Index scan selects of the form `yb_hash_code(i) > <negative_number>` fail.

```sql
EXPLAIN /*+IndexScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) > -1;
```

```
                                QUERY PLAN
--------------------------------------------------------------------------
 Index Scan using tt_i_j_idx on tt  (cost=0.00..129.00 rows=1000 width=8)
   Index Cond: (yb_hash_code(i) > '-1'::integer)
(2 rows)
```

```sql
/*+IndexScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) > -1;
```

```
 i | j
---+---
(0 rows)
```

IndexOnlyScan fails, too.  Other negative numbers also fail.  Zero and positive numbers work fine.  `>= 0` works; `>= -1` doesn't.  `yb_hash_code` should always return a non-negative number since it returns 16-bit uint values.

Sequential scan is fine:

```sql
/*+SeqScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) > -1;
```

```
 i | j
---+---
 1 | 2
(1 row)
```

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-7345]: https://yugabyte.atlassian.net/browse/DB-7345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

parkhi:
Looking into this issue.

parkhi:
The issue seems to be happening here at this line https://github.com/yugabyte/yugabyte-db/blob/0ee77be23078517871825b6e7928ae688fbddccb/src/postgres/src/backend/access/yb_access/yb_scan.c#L2278

[...]

PR #22551, opened 2024-05-27
[#18360] YSQL: Fix wrong results for IndexScan yb_hash_code(i) > -1

parkhi:
**Issue #18360** : [YSQL] Wrong results for IndexScan yb_hash_code(i) > -1

**RootCause** : Hash code keys are stored in datatype ListCell which allows negative intergers as data while normal keys are stored in datatype ScanKey which accepts data in the format of uint64_t. When a negative number like -1 is passed in the sql query , it gets converted into uint64_t type as 18446744073709551615. The query internally becomes : 

/*+IndexScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) > 18446744073709551615 and code tries to filter out records accordingly which fails.

### DIFF
--- a/src/postgres/src/test/regress/expected/yb_hash_code.out
+++ b/src/postgres/src/test/regress/expected/yb_hash_code.out
@@ -975,4 +975,16 @@ EXPLAIN (COSTS OFF, TIMING OFF, SUMMARY OFF, ANALYZE)/*+IndexScan(tt)*/ SELECT *
  1 | 2
 (1 row)
 
+/*+IndexScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) >  2^31 - 1;
+ i | j 
+---+---
+(0 rows)
+
+/*+IndexScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) >  2^31;
+ i | j 
+---+---
+(0 rows)
+
+
+
 DROP TABLE tt;


src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -2316,7 +2316,7 @@ YbBindHashKeys(YbScanDesc ybScan)
 		Assert(YbIsHashCodeSearch(key));
 		YbBound bound = {
 			.type = YB_YQL_BOUND_VALID,
-			.value = key->sk_argument
+			.value = (lc->data.int_value<0)?0:key->sk_argument
```
tanujnay112, line 2319: Please leave a space around all the operators. Please look around the codebase for examples of how `ListCell` instances are accessed.

Can you achieve the same thing by only checking something on `key->sk_argument`?

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -2316,7 +2316,7 @@ YbBindHashKeys(YbScanDesc ybScan)
 		Assert(YbIsHashCodeSearch(key));
 		YbBound bound = {
 			.type = YB_YQL_BOUND_VALID,
-			.value = key->sk_argument
+			.value = (key->sk_argument > INT_MAX) ? 0 : key->sk_argument
```
tanujnay112, line 2319: What about the case where it's below `INT_MIN`?
Why is this `>` and not `>=`?
Is there a test case against `key->sk_argument` being equal to `INT_MAX`?

hari90:
The title should be in this format:

> [#18360] YSQL: Fix wrong results for IndexScan yb_hash_code(i) > -1

src/postgres/src/backend/access/yb_access/yb_scan.c
tanujnay112: >As of now, I have only added the test cases present on the bug as you can see in file [yb_hash_code.sql] , Should I add the case where the user is passing INT_MAX as input instead of a negative integer?

Yes, please add more test cases as requested.

parkhi: @tanujnay112 INT_MAX is not accepted by yugabyte sql command line and since this value can be different depending on whether machine is 32 bit or 64 , it can not be hard coded in test cases. Can you suggest some way to integrate this test in here?
Also I tested locally with yb_hash_code(i) > 2^31 - 1 (INT_MAX for my machine) and yb_hash_code(i) >  2^31 (INT_MAX+1) and they both resulted in 0 rows while yb_hash_code(i) > -1 resulted in 2 rows. Since lc->data.int_value is of type signed int , when value bigger than INT_MAX is passed , it is not read as it is but only 32 bits are read.

parkhi:
> The title should be in this format:
> 
> > [#18360] YSQL: Fix wrong results for IndexScan yb_hash_code(i) > -1

Done

src/postgres/src/backend/access/yb_access/yb_scan.c
tanujnay112: Yes please add those as test cases.

parkhi: @tanujnay112 Added the tests as suggested.

andrei-mart: Hi @parkhi, I've been looking into other yb_hash_code-related issue, basically, `yb_hash_code(hkey) < 65536` was reported not working properly.
I triaged the report and went to GitHub to check if it was reported before and found this work in progress.
Your fix is in right place, but you are missing a few points. 
First of all, the yb_hash_code returns a value from the range [0..65535]. If compared to an out of range constant, it is either "all" or "nothing", depending on the operation and whether the constant is above or below the range. 
It seems convenient to have out of bound condition checked when bound is applied to the total range in the YbApplyStartBound and YbApplyEndBound functions. The "Start" and "End" there defines the side, and out of bound value won't change the range, the function would just return true or false if the condition is respectively "all" or "nothing".

Second point (minor) is that uint64_t is not the best data type for the YbBound.value. It basically takes the value from the query, which is signed. The "long" value would allow typecast without data loss.

src/postgres/src/test/regress/sql/yb_hash_code.sql
```diff
@@ -208,3 +208,14 @@ CREATE INDEX t_x_hash_y_asc_idx ON t (x HASH, y ASC);
...
 SELECT yb_hash_code(x), y FROM t WHERE yb_hash_code(x) = 2675 AND y IN (5, 6);
 DROP TABLE t;
+
+-- Issue #18360
+CREATE TABLE tt (i int, j int);
+CREATE INDEX ON tt (i, j);
+INSERT INTO tt VALUES (1, 2);
+EXPLAIN (COSTS OFF, TIMING OFF, SUMMARY OFF, ANALYZE)/*+IndexScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) > -1;
+/*+IndexScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) > -1;
+/*+SeqScan(tt)*/ SELECT * FROM tt WHERE yb_hash_code(i) > -1;
```
andrei-mart, line 218: Suggested test cases:
yb_hash_code(i) compared to various constant out of the [0..65535] range: negative values as well as positive values around max(uint16), max(int32), max(uint32), etc.
Use various strict and non-strict inequality operators.

PR #22644, opened 2024-06-03
[Docs] Send the click events to Algolia from search

samiahmedsiddiqui:
Send the click events to Algolia so that the CTR (Click-through Rate) is calculated right.


docs/src/algolia-search.js
```diff
@@ -2,9 +2,38 @@ import algoliasearch from 'algoliasearch';
...
   const ignoreClickOnMeElement = document.querySelector('body:not(.td-searchpage) .search-area');
   const searchInput = document.getElementById('search-query');
 
+  /**
+   * Decode the cookie and return the appropriate cookie value if found
+   * Otherwise empty string is returned.
+   *
+   * return string
+   */
+  function getCookie(cname) {
```
premkumr, line 17: We have cookie processing on 5 different files(index.js, search-banner.js, track-utm, sidebar-tree and this one). Can we have single library to do this?

premkumr (approved):

docs/src/algolia-search.js
samiahmedsiddiqui: @premkumr Sure, let's do that but it is something not particularly related to the Algolia click events so let's merge this PR and work on cookie processing in a separate PR.

PR #22719, opened 2024-06-05
Update data-modeling-ysql.md

utkarsh-agrawal-jm:


premkumr (approved):
Thanks @utkarsh-agrawal-jm !

PR #22766, opened 2024-06-06
[docs] Update consumer safe time metric unit from microseconds to milliseconds

Jethro-M:
These consumer safe time metrics are reported from yugabyte db in milliseconds and not microseconds: https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/master/xcluster/xcluster_consumer_metrics.cc#L40

Have also verified by stopping the source universe in a DR setup and observing that the reported number is indeed in milliseconds.

@netlify /preview/launch-and-manage/monitor-and-alert/metrics/replication


aishwarya24 (approved):
LGTM

PR #22836, opened 2024-06-12
[Docs] Sort feature to tables (Where fulfills the criteria)

samiahmedsiddiqui:
**Criteria for Sortable Tables:**
- **Table Class:** Table must possess a `sortable` class.
- **Table Header:** Table must possess a header row containing column labels. Empty headers will disable sorting for that specific column. For example: 
https://deploy-preview-22836--infallible-bardeen-164bc9.netlify.app/preview/explore/linear-scalability/horizontal-vs-vertical-scaling/#differences
- **Number of rows:** Table must contain at least two data rows (excluding the header row).
- **Non-Empty Columns:** Each column participating in the sorting process must contain data. Empty columns (all blank cells) will not be included in the sorting functionality. For example: 
https://deploy-preview-22836--infallible-bardeen-164bc9.netlify.app/preview/explore/observability/active-session-history/#ysql

### DIFF
--- a/docs/assets/scss/_tables.scss
+++ b/docs/assets/scss/_tables.scss
@@ -116,7 +116,6 @@ main {
           font-size: 12px;
           line-height: 24px;
           padding: 12.75px 15px;
-          min-width: 140px;
 
           &.empty-th .sort-btn {
             display: none;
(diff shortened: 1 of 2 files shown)


aishwarya24 (approved):
Verified it locally. LGTM

PR #23133, opened 2024-07-05
[doc] join flag syntax

ddhodge:
yugabyted --join flag enforces dns syntax
typo fix in 2.14

DOC-413
DOC-204

stable/reference/configuration/yugabyted


aishwarya24 (approved):
LGTM

Issue #23120, opened 2024-07-03
[YSQL] Read Only Deferrable in Serializable isolation is waiting for concurrent write transactions to commit


pkj415:
Jira Link: [DB-12053](https://yugabyte.atlassian.net/browse/DB-12053)
### Description

Consider the following steps in order:

```
Session 1:

yugabyte=# BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE READ ONLY DEFERRABLE;
BEGIN

Session 2:
yugabyte=# BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
BEGIN
yugabyte=# UPDATE test SET v=v+1 WHERE k=1;
UPDATE 1

Session 1:
yugabyte=# SELECT * FROM test;
<waits for the concurrent transaction in Session 2 to commit>
```

The SELECT in Session 1 shouldn't really wait. SERIALIZABLE READ ONLY internally uses snapshot isolation as an optimization, so no read locks are taken by the SELECT in this case (unlike any regular read in serializable isolation level).

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-12053]: https://yugabyte.atlassian.net/browse/DB-12053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pao214:
Landed on master, 2024.1, and 2.20.

PR #23228, opened 2024-07-16
[docs] Add basic troubleshooting steps for read restart errors.

pao214:
DOC-408

@netlify /preview/architecture/transactions/read-restart-error/

### DIFF
--- a/docs/content/preview/architecture/transactions/read-restart-error.md
+++ b/docs/content/preview/architecture/transactions/read-restart-error.md
@@ -32,6 +32,35 @@ A detailed scenario that explains how clock skew can result in the above mention
 
   * It should return the data even if the client issued the read from a different node after writing the data, because the following guarantee needs to be maintained: the database should always return data that was committed in the past (past refers to the user-perceived past, and not based on machine clocks).
 
-  * It should not return the data if the write is performed in the future (that is, after the read point had been chosen) **and** had a write timestamp later than the read point. Because if that is allowed, everything written in future would be trivially visible to the read. Note that the latter condition is important because it is okay to return data that was written after the read point was picked, if the write timestamp was earlier than the read point (this doesn't break and consistency or isolation guarantees).
+  * It should not return the data if the write is performed in the future (that is, after the read point had been chosen) **and** had a write timestamp later than the read point. Because if that is allowed, everything written in future would be trivially visible to the read. Note that the latter condition is important because it is okay to return data that was written after the read point was picked, if the write timestamp was earlier than the read point (this doesn't break any consistency or isolation guarantees).
 
 * If node `N2` finds writes in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier, and to avoid reading data with a later write timestamp which was also actually written after the read had been issued, node `N2` raises a `Read restart` error.
+
+## Troubleshooting
+
+You can handle and mitigate read restart errors using the following techniques:
+
+- Implement retry logic in the application. Application retries can help mitigate read restart errors. Moreover, a statement or a transaction may fail in other ways such as transaction conflicts or infrastructure failures. Therefore, a retry mechanism is strongly recommended for a cloud-native, distributed database such as YugabyteDB.
+- Use SERIALIZABLE READ ONLY DEFERRABLE mode whenever possible. Read restart errors usually occur when the query is a SELECT statement with a large output footprint and there are concurrent writes that satisfy the SELECT statement.
+
+  Using DEFERRABLE will avoid a read restart error altogether. However, the tradeoff is that the statement waits out the maximum permissible clock skew before reading the data (which is max_clock_skew_usec that has a default of 500ms). This is not an issue for large SELECT statements running in the background because latency is not a priority.
+
+  Examples:
+
+  Set transaction properties at the session level.
+  ```sql
+  SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL SERIALIZABLE READ ONLY DEFERRABLE;
+  SELECT * FROM large_table;
+  ```
+
+  Enclose the offending query within a transaction block.
+  ```sql
+  BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE READ ONLY DEFERRABLE;
+  SELECT * FROM large_table;
+  COMMIT;
+  ```
+- Using read only, deferrable transactions is not always feasible, either because the query is not read only, or the query is part of a read-write transaction, or because an additional 500ms of latency is not acceptable. In these cases, try increasing the value of `ysql_output_buffer_size`.
+
+  This will enable YugabyteDB to retry the query internally on behalf of the user. As long as the output of a statement hasn't crossed ysql_output_buffer_size to result in flushing partial data to the external client, the YSQL query layer retries read restart errors for all statements in a Read Committed transaction block, for the first statement in a Repeatable Read transaction block, and for any standalone statement outside a transaction block. As a tradeoff, increasing the buffer size also increases the memory consumed by the YSQL backend processes, resulting in a higher risk of out-of-memory errors.
+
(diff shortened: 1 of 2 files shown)


docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -32,6 +32,12 @@ A detailed scenario that explains how clock skew can result in the above mention
...
+  * It should not return the data if the write is performed in the future (that is, after the read point had been chosen) **and** had a write timestamp later than the read point. Because if that is allowed, everything written in future would be trivially visible to the read. Note that the latter condition is important because it is okay to return data that was written after the read point was picked, if the write timestamp was earlier than the read point (this doesn't break any consistency or isolation guarantees).
 
 * If node `N2` finds writes in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier, and to avoid reading data with a later write timestamp which was also actually written after the read had been issued, node `N2` raises a `Read restart` error.
+
+## Troubleshooting
+
```
ddhodge, line 40:
```suggestion

You can handle and mitigate read restart errors using the following techniques:

```

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -32,6 +32,12 @@ A detailed scenario that explains how clock skew can result in the above mention
...
+1. Implement retry logic in the application. Application retries can help with mitigating read restart errors. Moreover, there are other ways in which a statement or a transaction may fail such as transaction conflicts or infrastructure failures. Therefore, a retry mechanism is strongly recommended for a cloud-native, distributed database such as YugabyteDB.
```
ddhodge, line 41:
```suggestion
- Implement retry logic in the application. Application retries can help mitigate read restart errors. Moreover, a statement or a transaction may fail in other ways, such as transaction conflicts or infrastructure failures. Therefore, a retry mechanism is strongly recommended for a cloud-native, distributed database such as YugabyteDB.
```

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -32,6 +32,12 @@ A detailed scenario that explains how clock skew can result in the above mention
...
+2. Use SERIALIZABLE, READ ONLY, DEFERRABLE mode whenever possible. Read restart errors usually occur when the query is a SELECT statement with a large output footprint. These are better run with the above-mentioned transaction characteristics anyway. Using DEFERRABLE will avoid a read restart error altogether. However, the tradeoff is that the statement waits out the clock skew before reading the data. This is not an issue for large SELECT statements. Tip: use `SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL SERIALIZABLE READ ONLY DEFERRABLE;` to change the default transaction properties.
```
ddhodge, line 42:
```suggestion
- Use SERIALIZABLE, READ ONLY, or DEFERRABLE mode whenever possible. Read restart errors usually occur when the query is a SELECT statement with a large output footprint. These are better run with the above-mentioned transaction characteristics anyway.

  Using DEFERRABLE will avoid a read restart error altogether. However, the tradeoff is that the statement waits out the clock skew before reading the data. This is not an issue for large SELECT statements.
  
  Tip: use `SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL SERIALIZABLE READ ONLY DEFERRABLE;` to change the default transaction properties.
```

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -32,6 +32,12 @@ A detailed scenario that explains how clock skew can result in the above mention
...
+3. Using read only, deferrable transactions is not always feasible, either because the query is not read only, or because it is not acceptable to wait out the clock skew. In these cases, try increasing the value of `ysql_output_buffer_size`. This will enable YugabyteDB to retry the query internally on behalf of the user. As a tradeoff, more memory is used by the YSQL backend processes, increasing the risk of out-of-memory errors. Be aware that this is not a silver bullet. For example, the COPY command can still raise a read restart error even though the command has a one line output. Increasing `ysql_output_buffer_size` is not useful in this scenario. The application must retry the COPY command instead. Another example is DMLs such as INSERT/UPDATE/DELETE. These do not have enough output to overflow the buffer size. However, when these statements are executed as part of an explicit transaction block (e.g. BEGIN; ... INSERT ... COMMIT;), a read restart error cannot be retried internally by YugabyteDB. The onus is on the application to ROLLBACK and retry the transaction.
```
ddhodge, line 43:
```suggestion
- Using read only, deferrable transactions is not always feasible, either because the query is not read only, or because it is not acceptable to wait out the clock skew. In these cases, try increasing the value of the `ysql_output_buffer_size` flag. This enables YugabyteDB to retry the query internally on behalf of the user. As a tradeoff, the YSQL backend processes use more memory, increasing the risk of out-of-memory errors.

  Be aware that this is not a silver bullet. For example, the COPY command can still raise a read restart error even though the command has a one line output. Increasing `ysql_output_buffer_size` is not useful in this scenario. The application must retry the COPY command instead. Another example is DMLs such as INSERT/UPDATE/DELETE. These do not have enough output to overflow the buffer size. However, when these statements are executed as part of an explicit transaction block (for example, BEGIN; ... INSERT ... COMMIT;), a read restart error cannot be retried internally by YugabyteDB. The onus is on the application to ROLLBACK and retry the transaction.
```

ddhodge: maybe the flag should also be added to the yb-tserver reference, doesn't appear to be documented

docs/content/preview/architecture/transactions/read-restart-error.md
pkj415: .... Read restart errors usually occur when the query is a SELECT statement with a large output footprint and there are concurrent writes that satisfy the SELECT statement.


"These are better run with the above-mentioned transaction characteristics anyway." -- I am not sure we should say this because if customers start using this, they won't go back once we make big improvements to reducing read restart errors? Also, there is https://github.com/yugabyte/yugabyte-db/issues/23120. On a side note, I realize that we need to prioritize this issue too.

However, the tradeoff is that the statement waits out the maximum permissible clock skew before reading the data (which is ysql_max_clock_skew_usecs that has a default of 500ms?)

.... This is not an issue for large SELECT statements because of the large latency of the read?

docs/content/preview/architecture/transactions/read-restart-error.md
pkj415: Using read only, deferrable transactions is not always feasible, either because the query is not read only, or the query is part of a read-write transaction or because ...

This will enable YugabyteDB to retry the query internally on behalf of the user. As long as the output of a statement hasn't crossed `ysql_output_buffer_size ` to result in flushing partial data to the external client, the YSQL query layer retries read restart errors for all statements in a Read Committed transaction block, for the first statement in a Repeatable Read transaction block and for any standalone statement outside a transaction block.

"However, when these statements are executed as part of an explicit transaction block (e.g. BEGIN; ... INSERT ... COMMIT;), a read restart error cannot be retried internally by YugabyteDB" -- this statement isn't right, we do retry these statements in RC txns and if they are the first statement in a RR txn.

docs/content/preview/architecture/transactions/read-restart-error.md
pao214: Not including "or" since this is PostgreSQL specific syntax and it does not have that.

docs/content/preview/architecture/transactions/read-restart-error.md
pao214: Filed https://github.com/yugabyte/yugabyte-db/issues/23232 for documentation of ysql_output_buffer_size flag.

ddhodge: What about the comment by Piyush "- this statement isn't right, we do retry these statements in RC txns and if they are the first statement in a RR txn"

pao214: @ddhodge  I've changed the wording to say REPEATABLE READ transaction and that the statement is in the middle of the transaction.

ddhodge (approved):
stable? other versions

pao214:
> stable? other versions?

Sushant has some additional suggestions. I'll sync with him offline.

pao214:
Made the change to stable as well (only other version).

PR #23256, opened 2024-07-22
[docs] Explain an example read ambiguity scenario with a diagram.

pao214:
@netlify /preview/architecture/transactions/read-restart-error/


docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,26 +15,68 @@ rightNav:
...
+  ![Read Restart Error](/images/architecture/txn/read_restart_error.svg)
+
+  In the above setup, there are three tablet servers, namely, `TSERVER 1`, `TSERVER 2`, and `TSERVER 3`. The data for the tokens table is hosted on servers TSERVER 2, and TSERVER 3. The query layer on TSERVER 1 is serving the SQL requests.
+
+  Moreover, TSERVER 3's clock is running 5 units of time ahead of TSERVER 2's clock because of clock skew.
+
+1. First, an INSERT of token 29 is issued to YSQL on TSERVER 1.
+2. This INSERT is routed to the tablet hosted on TSERVER 3. The operation picks a commit time of `T2=103` even though the "true time" is `98` (step 2).
+3. TSERVER 1 acknowledges the INSERT.
+4. Now, that the INSERT command is complete, the SELECT command is issued.
```
ddhodge, line 54:
```suggestion
4. Now that the INSERT command is complete, the SELECT command is issued.
```

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,26 +15,68 @@ rightNav:
...
+
+Follow the diagram below for a deeper discussion on the order of operations leading to this anomaly.
+
+  ![Read Restart Error](/images/architecture/txn/read_restart_error.svg)
+
+  In the above setup, there are three tablet servers, namely, `TSERVER 1`, `TSERVER 2`, and `TSERVER 3`. The data for the tokens table is hosted on servers TSERVER 2, and TSERVER 3. The query layer on TSERVER 1 is serving the SQL requests.
+
+  Moreover, TSERVER 3's clock is running 5 units of time ahead of TSERVER 2's clock because of clock skew.
+
+1. First, an INSERT of token 29 is issued to YSQL on TSERVER 1.
```
ddhodge, line 51:
```suggestion
1. An INSERT of token 29 is issued to YSQL on TSERVER 1.
```

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,26 +15,68 @@ rightNav:
...
+* For records with a commit timestamp later than `T1` + `max_clock_skew`, the database can be sure that these records were written after the read was issued. But for records with a commit timestamp between `T1` and `T1` + `max_clock_skew`, the tablet servers can find themselves in an ambiguous situation, such as the following:
+
+  * A read operation must observe all the commits that were acknowledged before the read was issued, even when the read time is picked on a node whose clock runs behind the node that committed the prior write. NOTE: the commit happened before the read as perceived by the user and not necessarily according to the machine clocks.
+
+  * A read operation must return a consistent snapshot of data, i.e. the read cannot simply advance its read time on observing a record with higher timestamp since the read already returned records from an older snapshot thus far. Therefore, the read must be restarted from the beginning with the advanced timestamp. Thus, the name read restart error.
```
ddhodge, line 69:
```suggestion
  * A read operation must return a consistent snapshot of data. That is, the read cannot simply advance its read time on observing a record with higher timestamp since the read already returned records from an older snapshot thus far. Therefore, the read must be restarted from the beginning with the advanced timestamp. Thus, the name read restart error.
```

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,26 +15,68 @@ rightNav:
...
+
+  * A read operation is free to include or exclude records committed by concurrent transactions. Implying, whenever the commit timestamp is lower than the read timestamp, the read is allowed to observe writes from the committed transaction, even when the commit happened concurrently with the read. Therefore, there is no ambiguity when the commit time is lower.
```
ddhodge, line 71:
```suggestion
  * A read operation is free to include or exclude records committed by concurrent transactions. This implies that whenever the commit timestamp is lower than the read timestamp, the read is allowed to observe writes from the committed transaction, even when the commit happened concurrently with the read. Therefore, there is no ambiguity when the commit time is lower.
```

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,26 +15,68 @@ rightNav:
...
-* For writes with a write timestamp later than `T1` + `max_clock_skew`, the database can be sure that these writes were done after the read timestamp had been chosen. But for writes with a write timestamp between `T1` and `T1` + `max_clock_skew`, node `N2` can find itself in an ambiguous situation, such as the following:
+* Whenever a read operation finds records with timestamp in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier, and to read a consistent snapshot, the read operation raises a `Read restart` error to restart the read.
 
-  * It should return the data even if the client issued the read from a different node after writing the data, because the following guarantee needs to be maintained: the database should always return data that was committed in the past (past refers to the user-perceived past, and not based on machine clocks).
+{{<note title="Causal Reverse Anomaly">}}
```
ddhodge, line 75: What is a causal reverse anomaly? First mention. This seems out of place, why are you mentioning it here?
Or is this entire scenario a 'causal reverse anomaly', in which case why not call it that earlier?

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,26 +15,68 @@ rightNav:
...
-* Assuming that node `N1` reads from node `N2`, it is possible that data had already been written on node `N2` with a write timestamp `T2` (> `T1`) but it had been written before the read was issued. This could be caused by clock skew if the physical clock on node `N2` ran ahead of node `M`, resulting in the write done in the past still having a write timestamp later than `T1`.
+* Consider a scenario where tokens 17, 29 are inserted into an empty tokens table. Then, all the tokens from the table are retrieved.
+
+  SQL commands for the scenario:
+  ```sql
+  INSERT INTO tokens VALUES (17);
+  INSERT INTO tokens VALUES (29);
+  SELECT * FROM tokens;
+  ```
+* The SELECT must return both 17, and 29. However, the clock skew can cause one or more tokens from the list to be omitted.
```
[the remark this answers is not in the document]
pao214: Retaining `However, the clock skew can cause one or more tokens from the list to be omitted.`
I am open to rephrasing it but not removing it altogether.

docs/content/preview/architecture/transactions/read-restart-error.md
pao214: It is another anomaly that stems from clock skew. We do not handle this anomaly. This is a piece of info for those who are curious. But I do agree that there is no point in mentioning this since readers who know about this anomaly already know that YB does not handle it.

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,26 +15,68 @@ rightNav:
...
+Follow the diagram below for a deeper discussion on the order of operations leading to this anomaly.
+
+  ![Read Restart Error](/images/architecture/txn/read_restart_error.svg)
+
+  In the above setup, there are three tablet servers, namely, `TSERVER 1`, `TSERVER 2`, and `TSERVER 3`. The data for the tokens table is hosted on servers TSERVER 2, and TSERVER 3. The query layer on TSERVER 1 is serving the SQL requests.
+
+  Moreover, TSERVER 3's clock is running 5 units of time ahead of TSERVER 2's clock because of clock skew.
+
+1. First, an INSERT of token 29 is issued to YSQL on TSERVER 1.
+2. This INSERT is routed to the tablet hosted on TSERVER 3. The operation picks a commit time of `T2=103` even though the "true time" is `98` (step 2).
```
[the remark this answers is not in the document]
pao214: I could not make the arrows bigger. I zoomed into the picture instead. Let me know what you think.

docs/content/preview/architecture/transactions/read-restart-error.md
```diff
@@ -15,33 +15,71 @@ rightNav:
...
+8. Thus, token 29 is omitted from the result set returned.
 
-  * It should not return the data if the write is performed in the future (that is, after the read point had been chosen) **and** had a write timestamp later than the read point. Because if that is allowed, everything written in future would be trivially visible to the read. Note that the latter condition is important because it is okay to return data that was written after the read point was picked, if the write timestamp was earlier than the read point (this doesn't break any consistency or isolation guarantees).
+  More generally, this anomaly occurs whenever the commit time of a prior write operation `T2` is higher than the read time `T1` of a later read operation, thus violating (distributed) consistency.
 
-* If node `N2` finds writes in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier, and to avoid reading data with a later write timestamp which was also actually written after the read had been issued, node `N2` raises a `Read restart` error.
+How does YugabyteDB mitigate this clock skew anomaly?
```
ddhodge, line 62: Would move this part up to after "YugabyteDB has optimizations to resolve this ambiguity internally with best-effort." given this appears to be an explanation of YBDB optimizations

ddhodge (approved):

docs/content/preview/architecture/transactions/read-restart-error.md
pao214: This is not one of the optimizations. We never describe the optimizations, just that they are present. Perhaps, mitigate is not the correct word to use here? I changed it to prevent. This basically describes how we raise a read restart error whenever we detect the anomaly. I hope that makes it clearer.

docs/content/stable/architecture/transactions/read-restart-error.md
```diff
@@ -15,33 +15,69 @@ rightNav:
...
+* Whenever a read operation finds records with timestamp in the range `(T1, T1+max_clock_skew]`, to avoid breaking the strong guarantee that a reader should always be able to read what was committed earlier, and to read a consistent snapshot, the read operation raises a `Read restart` error to restart the read.
 
 ## Troubleshooting
 
 You can handle and mitigate read restart errors using the following techniques:
 
 - Implement retry logic in the application. Application retries can help mitigate read restart errors. Moreover, a statement or a transaction may fail in other ways such as transaction conflicts or infrastructure failures. Therefore, a retry mechanism is strongly recommended for a cloud-native, distributed database such as YugabyteDB.
-- Use SERIALIZABLE READ ONLY DEFERRABLE mode whenever possible. Read restart errors usually occur when the query is a SELECT statement with a large output footprint and there are concurrent writes that satisfy the SELECT statement.
+
+  While implementing application retries is the best long-term approach, there a few short-term solutions in the interim.
```
ddhodge, line 79:
```suggestion
  While implementing application retries is the best long-term approach, there are a few short-term solutions you can use in the interim.
```

ddhodge (approved):

PR #23283, opened 2024-07-25
[doc] Delphix note on script

ddhodge:
Delphix note on script deprecation
DOC-430

@netflix /preview/integrations/delphix/


aishwarya24 (approved):
LGTM

PR #23349, opened 2024-07-31
[DOC-431] Added a note for GKE cluster docs

aishwarya24:
Deploy preview: https://deploy-preview-23349--infallible-bardeen-164bc9.netlify.app/preview/deploy/kubernetes/single-zone/gke/statefulset-yaml/#create-a-gke-cluster


docs/content/preview/deploy/kubernetes/single-zone/gke/statefulset-yaml.md
```diff
@@ -64,12 +63,18 @@ Before starting deployment, perform the following:
 
 ## Create a GKE cluster
 
-Create a Kubernetes cluster by running the following command.
+Create a Kubernetes cluster using the following command:
 
 ```sh
 gcloud container clusters create yugabyte
```
ddhodge, line 69: Why not just use the secure command - why even provide the public example?

docs/content/preview/deploy/kubernetes/single-zone/gke/statefulset-yaml.md
```diff
@@ -64,12 +63,18 @@ Before starting deployment, perform the following:
...
 ```
 
+Note that as this command creates public kubernetes clusters, you can't change this configuration later. If you want to run YugabyteDB in a private kubernetes cluster, run the following command instead:
```
bharat-mukheja, line 72: I believe the Note should be in a warning box or color to emphasize.

docs/content/preview/deploy/kubernetes/single-zone/gke/statefulset-yaml.md
aishwarya24: cc @nbhatia-yb @Arpit-yb

Arpit-yb: I think instead of using "yugabyte", we should modify the command to - 
`gcloud container clusters create [CLUSTER_NAME]`

docs/content/preview/deploy/kubernetes/single-zone/gke/statefulset-yaml.md
```diff
@@ -64,12 +63,18 @@ Before starting deployment, perform the following:
...
+
+```sh
+gcloud container clusters create yugabyte --enable-private-nodes
```
Arpit-yb, line 75: Same as above

docs/content/preview/deploy/kubernetes/single-zone/gke/statefulset-yaml.md
ddhodge: Still think we should just drop this insecure example and only provide the secure one that follows it. So

`gcloud container clusters create cluster_name --enable-private-nodes`

ddhodge (approved):

PR #23354, opened 2024-08-01
[Docs] Remove 6sense domains from CSP headers

samiahmedsiddiqui:
We are no longer using 6sense tool. Removing their domains from the CSP headers.


ddhodge (approved):
LGTM

PR #23507, opened 2024-08-14
[doc] yugabyted configure flag

ddhodge:
update use of configure flag

@netlify /preview/reference/configuration/yugabyted/


nchandrappa (commented):
LGTM.

PR #23606, opened 2024-08-23
[DOC] Remove restart from replicated-migrate start

chidmuthu:
Don't need to restart YBA after replicated migrate


aishwarya24 (approved):
LGTM. Can you make the change to preview as well if applicable?

ddhodge:
@chidmuthu what versions does this apply to

chidmuthu:
This should apply to all versions @ddhodge

PR #23607, opened 2024-08-23
Change error to warn

premkumr:
- Don't error out on Flags XML fetch failure. 
- Or else deploy previews will fail before the release is pushed out.


aishwarya24 (approved):
LGTM

PR #23642, opened 2024-08-27
[docs] [TA] Added TA-23476: YCQL currenttimestamp() precision

aishwarya24:
@netlify /preview/releases/techadvisories/ta-23476/

### DIFF
--- a/docs/content/preview/releases/techadvisories/ta-23476.md
+++ b/docs/content/preview/releases/techadvisories/ta-23476.md
@@ -153,6 +153,15 @@ This can create issues in certain scenarios as follows:
     ```
 
     ```cql
+    ycqlsh:temp> SELECT * FROM test WHERE v1 = 5;
+    ```
+
+    ```output
+     id | ts | v1 | v2
+    ----+----+----+----
+    ```
+
+    ```cql
     ycqlsh:temp> SELECT id, ts FROM test WHERE v1 = 5;
     ```
 
--- a/docs/content/preview/releases/techadvisories/ta-23476.md
+++ b/docs/content/preview/releases/techadvisories/ta-23476.md
@@ -1,0 +1,169 @@
+---
+title: TA-23476
+headerTitle: YCQL currenttimestamp() precision
+headcontent: 01 Jan 2024
+type: docs
+showRightNav: true
+cascade:
+  unversioned: true
+menu:
+  preview_releases:
+    identifier: ta-23476
+    weight: 1
+rightNav:
+  hideH2: true
+type: docs
+---
+
+|          Product           |  Affected Versions  |  Related Issues   | Fixed In |
+| :------------------------- | :------------------ | :---------------- | :------- |
+| {{<product "ycql">}}       | {{<release "All">}} | {{<issue 11052>}}, {{<issue 23476>}} | TBD |
+
+## Description
+
+In YCQL, timestamp data inserted with the `currenttimestamp()` function may not be returned for queries with an equality condition on the timestamp. If the timestamp is part of the primary key, this also affects index scans as the table lookup for an index scan internally uses an equality condition to lookup the corresponding table rows. The data can still be retrieved as expected by primary key or index-only scan queries with no condition or inequality conditions on the timestamp column.
+
+The root cause is that in YCQL, the timestamp type has millisecond precision, to align with vanilla Cassandra, ensuring compatibility with Cassandra drivers and tools. However, YCQL can internally store data with microsecond precision, adjusting the precision to milliseconds when data is sent or received through the YCQL API. The discrepancy between the wire protocol's millisecond precision and the storage system's microsecond precision can lead to unexpected behavior.
+
+## Mitigation
+
+There are two key aspects to mitigating the issues caused by the discrepancy between the YCQL wire protocol's millisecond precision and the storage system's microsecond precision:
+
+1. Ensure new writes use the correct precision:
+
+    1. Use totimestamp(now()) instead of currenttimestamp() to generate millisecond-precision timestamp values that are compatible with the wire protocol.
+
+1. Adjust existing data to millisecond precision:
+
+    1. If you have already inserted data with non-zero microsecond precision, contact {{% support-general %}} for assistance to run a script to read and re-insert the affected rows with millisecond precision.
+
+Additionally, if millisecond precision is not sufficient for your use case, there are two options:
+
+1. Switch from timestamp to timeuuid data type:
+
+    1. Timeuuid uses 100-nanosecond precision and includes a node-identifier component, ensuring values generated on different nodes can never collide.
+
+1. If millisecond-precision collisions are expected but rare:
+
+    1. Use IF NOT EXISTS on the data insert path and handle the existing row case in the application by retrying the insert. However, if collisions are common, this approach may result in too many retries.
+
+## Details
+
+The YCQL wire protocol uses millisecond precision for timestamps, while the YCQL storage system uses microsecond precision internally. This means that all timestamps that can be inserted are expected to have `000` for the microseconds component.
+
+However, the `currenttimestamp()` function currently generates values with microsecond precision, so the microseconds component is likely to be non-zero 99.9% of the time.
+
+When these non-zero microsecond timestamp values are stored, the microsecond information will be lost when the precision is adjusted to milliseconds for the wire protocol (for example, when responding to a user query). As a result, you will always see timestamp values with only millisecond precision.
+
+This can create issues in certain scenarios as follows:
+
+- Equality conditions: If you query for a timestamp value that was just read, the condition will not match, as the microsecond component will differ (0 for the query, non-zero for the stored value).
+
+- Index scans: If the primary key of the main table includes an affected timestamp column, the data in both the table and the index will be in microsecond precision, so they will match. However, when performing an index scan, the main-table read request uses the same protocol as the YCQL API, which adjusts the precision to milliseconds. This may cause the main-table read request to not match, and the index scan result to miss the affected rows with non-zero microsecond values.
+
+## Examples
+
+### Primary-key scan with equality condition
+
+1. Set up the keyspace and table as follows:
+
+    ```cql
+    CREATE KEYSPACE temp;
+    USE temp;
+    CREATE TABLE test(id int, ts timestamp, v1 int, v2 int, primary key( id, ts)) WITH transactions = {'enabled': 'true'};
+    ```
+
+1. Insert a row using `currenttimestamp()` as follows:
+
+    ```cql
+    INSERT INTO test(id, ts, v1, v2) VALUES (1, currenttimestamp(), 5, 10);
+    SELECT * FROM test;
+    ```
+
+    ```output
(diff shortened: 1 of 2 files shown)


docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+type: docs
+---
+
+|          Product           |  Affected Versions  |  Related Issues   | Fixed In |
+| :------------------------- | :------------------ | :---------------- | :------- |
+| {{<product "ycql">}}       | {{<release "All">}} | {{<issue 11052>}}, {{<issue 23476>}} | TBD |
+
+## Description
+
+In YCQL, timestamp data inserted with the `currenttimestamp()` function may not be returned for queries with an equality condition on the timestamp. If the timestamp is part of the primary key, this also affects index scans as the table lookup for an index scan internally uses an equality condition to lookup the corresponding table rows. The data can still be retrieved as expected by primary key or index-only scan queries with no condition or inequality conditions on the timestamp column.
```
ddhodge, line 24:
```suggestion
In YCQL, timestamp data inserted with the `currenttimestamp()` function may not be returned for queries with an equality condition on the timestamp. If the timestamp is part of the primary key, this also affects index scans as the table lookup for an index scan internally uses an equality condition to look up the corresponding table rows. The data can still be retrieved as expected by primary key or index-only scan queries with no condition or inequality conditions on the timestamp column.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+The root cause is that in YCQL, the timestamp type has millisecond precision, to align with vanilla Cassandra, ensuring compatibility with Cassandra drivers and tools. However, YCQL can internally store data with microsecond precision, adjusting the precision to milliseconds when data is sent or received through the YCQL API. The discrepancy between the wire protocol's millisecond precision and the storage system's microsecond precision can lead to unexpected behavior.
+
+## Mitigation
+
+There are two key aspects to mitigating the issues caused by the discrepancy between the YCQL wire protocol's millisecond precision and the storage system's microsecond precision:
+
+1. Ensure new writes use the correct precision:
```
ddhodge, line 32:
```suggestion
1. Ensure new writes use the correct precision.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+    1. Use totimestamp(now()) instead of currenttimestamp() to generate millisecond-precision timestamp values that are compatible with the wire protocol.
```
ddhodge, line 34:
```suggestion
    Use totimestamp(now()) instead of currenttimestamp() to generate millisecond-precision timestamp values that are compatible with the wire protocol.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+1. Adjust existing data to millisecond precision:
```
ddhodge, line 36:
```suggestion
1. Adjust existing data to millisecond precision.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+    1. If you have already inserted data with non-zero microsecond precision, contact {{% support-general %}} for assistance to run a script to read and re-insert the affected rows with millisecond precision.
```
ddhodge, line 38:
```suggestion
    If you have already inserted data with non-zero microsecond precision, contact {{% support-general %}} for help with running a script to read and re-insert the affected rows with millisecond precision.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+Additionally, if millisecond precision is not sufficient for your use case, there are two options:
+
+1. Switch from timestamp to timeuuid data type:
```
ddhodge, line 42:
```suggestion
1. Switch from timestamp to timeuuid data type.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+    1. Timeuuid uses 100-nanosecond precision and includes a node-identifier component, ensuring values generated on different nodes can never collide.
```
ddhodge, line 44:
```suggestion
    Timeuuid uses 100-nanosecond precision and includes a node-identifier component, ensuring values generated on different nodes can never collide.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+1. If millisecond-precision collisions are expected but rare:
```
ddhodge, line 46:
```suggestion
1. If millisecond-precision collisions are expected but rare, use IF NOT EXISTS.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+    1. Use IF NOT EXISTS on the data insert path and handle the existing row case in the application by retrying the insert. However, if collisions are common, this approach may result in too many retries.
```
ddhodge, line 48:
```suggestion
    Use IF NOT EXISTS on the data insert path and handle the existing row case in the application by retrying the insert. However, if collisions are common, this approach may result in too many retries.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+    SELECT * FROM test;
+    ```
+
+    ```output
+     id | ts                              | v1 | v2
+    ----+---------------------------------+----+----
+      1 | 2024-08-21 17:44:47.647000+0000 |  5 | 10
+    ```
+
+    As expected the row is present. However, most likely the internal row has a non-zero microsecond component, just that due to the YCQL API millisecond standard, it will always return `000` for the microsecond part.
```
ddhodge, line 89:
```suggestion
    As expected, the row is present. However, most likely the internal row has a non-zero microsecond component, just that due to the YCQL API millisecond standard, it will always return `000` for the microsecond part.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+    SELECT * FROM test WHERE id = 1 AND ts >= '2024-08-21 17:44:47.647000+0000' and ts < '2024-08-21 17:44:47.648000+0000';
+    ```
+
+    ```output
+      id | ts                              | v1 | v2
+     ----+---------------------------------+----+----
+       1 | 2024-08-21 17:44:47.647000+0000 |  5 | 10
+    ```
+
+    A simple query with equality on the same (id and) timestamp value returned before does not match. However, as you can see changing this to a range query search for all timestamps within that millisecond does return the row, confirming that the row is there but the microsecond value is non-zero. Note that for the second query the upper bound is the next millisecond value (648 rather than 647) so it will match everything in [647000, 647999] range from microsecond perspective.
```
ddhodge, line 112:
```suggestion
    A simple query with equality on the same (id and) timestamp value returned before does not match. However, changing this to a range query search for all timestamps within that millisecond does return the row, confirming that the row is there but the microsecond value is non-zero. Note that for the second query the upper bound is the next millisecond value (648 rather than 647) so it will match everything in [647000, 647999] range from the microsecond perspective.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+
+    ```cql
+    CREATE KEYSPACE temp;
+    USE temp;
+    CREATE TABLE test(id int, ts timestamp, v1 int, v2 int, primary key( id, ts)) WITH transactions = {'enabled': 'true'};
+    CREATE INDEX ON test(v1);
+    ```
+
+    This example includes the same keyspace and table as the preceding primary-key example, along with adding an index.
+    Note that if you continue the preceding example just create the index and skip steps 1 and 2.
```
ddhodge, line 126:
```suggestion
   (To continue the preceding example, just create the index and skip steps 1 and 2.)
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+1. Set up the keyspace and table as follows:
+
+    ```cql
+    CREATE KEYSPACE temp;
+    USE temp;
+    CREATE TABLE test(id int, ts timestamp, v1 int, v2 int, primary key( id, ts)) WITH transactions = {'enabled': 'true'};
+    CREATE INDEX ON test(v1);
+    ```
+
+    This example includes the same keyspace and table as the preceding primary-key example, along with adding an index.
```
ddhodge, line 125:
```suggestion
    This example includes the same keyspace and table as the preceding primary-key example, and adds an index.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+    SELECT * FROM test;
+    ```
+
+    ```output
+     id | ts                              | v1 | v2
+    ----+---------------------------------+----+----
+      1 | 2024-08-21 17:44:47.647000+0000 |  5 | 10
+    ```
+
+    Same as above, a row is present but likely with non-zero microsecond component internally.
```
ddhodge, line 141:
```suggestion
    A row is present, but likely with a non-zero microsecond component internally.
```

docs/content/preview/releases/techadvisories/ta-23476.md
```diff
@@ -0,0 +1,169 @@
...
+    ycqlsh:temp> SELECT id, ts FROM test WHERE v1 = 5;
+    ```
+
+    ```output
+     id | ts
+    ----+---------------------------------
+      1 | 2024-08-21 17:44:47.647000+0000
+    ```
+
+    The above scenario shows that the row exists and can be retrieved with both a primary key scan on id (first query) and an index-only scan on v1 (third query). The second query (unlike the third) selects all columns including v2 which is not in the index. So, it has to be an index scan, that is, query the table row to read all the target columns. Therefore it hits the index scan issue described above and misses returning the affected row(s).
```
ddhodge, line 165:
```suggestion
    This shows that the row exists and can be retrieved with both a primary key scan on id (first query) and an index-only scan on v1 (third query). The second query (unlike the third) selects all columns including v2, which is not in the index, which means it has to be an index scan; that is, query the table row to read all the target columns. As a result, it hits the index scan issue and misses returning the affected row(s).
```

ddhodge, line 165: I see 2 queries, not 3

aishwarya24: added the query

ddhodge (approved):

PR #23802, opened 2024-09-05
[Docs] Remove Drift chat bot

samiahmedsiddiqui:


docs/netlify.toml
```diff
@@ -1382,8 +1382,6 @@
             https://www.google.com/recaptcha/
             https://www.googletagmanager.com/
             https://www.gstatic.com/recaptcha/
-            https://x.clearbitjs.com
```
premkumr, line 1385: @samiahmedsiddiqui - why has the clearbitjs been removed ?

samiahmedsiddiqui: @premkumr We are no longer using Clearbit directly. We integrated it from the different way by which we don't need this URL in our security policy headers.

I found few more instances of clearbit in security headers, if you like I can create a separate PR for that.

samiahmedsiddiqui: @premkumr clearbitjs added back in this PR. I am creating a separate PR to remove clearbit headers.

premkumr (approved):
LGTM

PR #23944, opened 2024-09-16
[docs] Manual fix for a yb-version not rendering correctly.

aishwarya24:


ddhodge (approved):
why doesn't this work

aishwarya24:
> why doesn't this work

not sure, this looked like the only instance of being linked from includeMarkdown shortcode. Somehow this combination doesn’t work. Requires some debugging.

samiahmedsiddiqui:
@ddhodge I think we should make it configurable otherwise we need to manually update this version every time when the new preview version release.

ddhodge:
> @ddhodge I think we should make it configurable otherwise we need to manually update this version every time when the new preview version release.

Yes - but how. It wasn't working, but don't know how to fix.

PR #23945, opened 2024-09-16
[doc] Fix BNL flag defaults

ddhodge:
Fix BNL flag defaults
DOC-471

@netlify /preview/explore/ysql-language-features/postgresql-compatibility


aishwarya24 (commented):
Should this section in 2.18 also be tagged EA - https://docs.yugabyte.com/v2.18/explore/ysql-language-features/join-strategies/#batched-nested-loop-join-bnl

PR #24201, opened 2024-09-30
[Docs] Fix Code snippet

samiahmedsiddiqui:
Code snippet style inside the call out section is broken. Missing all paddings [View the issue](https://docs.yugabyte.com/stable/explore/going-beyond-sql/#cluster-topology).


premkumr:
LGTM

PR #24241, opened 2024-10-02
JDBC Driver version update

ashetkar:
- JDBC Driver version update to `42.3.5-yb-8`


ashetkar:
Going ahead with the merge since it is only a JDBC driver version change to `42.3.5-yb-8` and it needs to be updated urgently.

PR #24379, opened 2024-10-10
[Docs] Update Tip callout

samiahmedsiddiqui:
On this type of call out (Tip)

- Light bulb icon container size should be  24px. 
- Icon color should be #097345


ddhodge (approved):
LGTM

Issue #21157, opened 2024-02-22
[DocDB] Fix load_balancer_max_over_replicated_tablets to reflect per-table status


SrivastavaAnubhav:
### Description

`load_balancer_max_over_replicated_tablets` limits the number of over-replicated tablets **per table**, but this fact isn't clear from the name of the flag or any of the descriptions associated with it. Either update the flag name and descriptions, or change it to be a global limit.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

jsochava:
Hi! I’d like to work on this issue.
Plan: keep the current behavior (per-table cap) and update the flag’s description in the source so it explicitly states “per table,” then align the public configuration reference accordingly.
QA: build locally, verify the updated description is shown consistently wherever flag metadata is surfaced, and include before/after screenshots in the PR.
I’ll open a Draft PR shortly and iterate there; please let me know if this scope looks good (and assign me if possible).

jsochava:
@SrivastavaAnubhav is the referenced commit([eeb6b72](https://github.com/michaelzheng67/yugabyte-db/commit/eeb6b72ccc00b460f483f44585ac21aa3bc04561)) just a not-completed version of this task?

PR #24406, opened 2024-10-11
[#21157][DocDb]: Fix load_balancer_max_over_replicated_tablets description to reflect per-table status

michaelzheng67:
Currently, the flag '--load_balancer_max_over_replicated_tablets' does not accurately reflect the fact that the setting is meant to be applicable per table.

This commit looks to update the description of the flag to explicitly state that this settings is on a per-table basis (note: customers may already have flag overrides so naively changing flag name outright is not ideal)

Test Plan:
./yb_release works

Reviewers:

Reviewed By:

Subscribers:

Tags:

Differential Revision:


docs/content/preview/reference/configuration/yb-master.md
```diff
@@ -522,7 +522,7 @@ Default: `2`
 
 ##### --load_balancer_max_over_replicated_tablets
 
-Specifies the maximum number of running tablet replicas that are allowed to be over the configured replication factor.
+Specifies the maximum number of running tablet replicas that are allowed to be over the configured replication factor per table.
```
ddhodge, line 525:
```suggestion
Specifies the maximum number of running tablet replicas per table that are allowed beyond the configured replication factor. For example, if RF=3 and this flag is set to 3, you can have 6 tablet replicas per table.
```

src/yb/master/cluster_balance.cc
```diff
@@ -68,7 +68,7 @@ DEFINE_RUNTIME_int32(load_balancer_max_concurrent_tablet_remote_bootstraps_per_t
 
 DEFINE_RUNTIME_int32(load_balancer_max_over_replicated_tablets, 1,
     "Maximum number of running tablet replicas that are allowed to be over the configured "
```
ddhodge, line 70:
```suggestion
    "Maximum number of running tablet replicas per table allowed in addition to the configured "
```

src/yb/master/cluster_balance.cc
```diff
@@ -68,7 +68,7 @@ DEFINE_RUNTIME_int32(load_balancer_max_concurrent_tablet_remote_bootstraps_per_t
...
-    "replication factor.");
+    "replication factor per table.");
```
ddhodge, line 71:
```suggestion
    "replication factor.");
```

src/yb/master/cluster_balance_util.h
```diff
@@ -215,7 +215,7 @@ struct Options {
   // time.
   bool kAllowLimitOverReplicatedTablets = true;
 
-  // Max number of running tablet replicas that are over the configured limit.
+  // Max number of running tablet replicas that are over the configured limit per table.
```
ddhodge, line 218:
```suggestion
  // Max number of running tablet replicas per table allowed over the configured RF.
```

ddhodge:
@SrivastavaAnubhav Can you TAL

SrivastavaAnubhav:
I'm talking to Michael about this on community slack. The best fix would actually be to do the following: introduce a new flag called `load_balancer_max_over_replicated_tablets_per_table` and use that instead of `load_balancer_max_over_replicated_tablets` if set (otherwise default to `load_balancer_max_over_replicated_tablets`) in cluster_balance_util.h.

PR #24708, opened 2024-10-29
[DOC-535] [YBA] [2024.2] Support bundle phase 3

aishwarya24:
@netlify /preview/yugabyte-platform/troubleshoot/universe-issues/#use-support-bundles

Todo:
Copy to 2024.1 as well


asharma-yb (approved):
LGTM!

docs/content/preview/yugabyte-platform/troubleshoot/universe-issues.md
```diff
@@ -234,6 +244,22 @@ Note that YBA will only collect files if you have sufficient permissions to requ
   - Statefulsets.yaml
   - Persistentvolumeclaims.yaml
 
+#### Prometheus metrics
+
+This component collects Prometheus metrics (Promdump collection) for a given universe. You can select the duration of collection and the metrics to collect from the Support Bundle UI for Prometheus metrics.
+
+A brief description of each metric is as follows:
```
ddhodge, line 251: The Prometheus metrics for the selected universe. You can customize the timeframe and the metrics to collect.

You can include the following metrics with the support bundle:

- Master Export - YB-Master metrics.
- Node Export - `node_exporter` OS level metrics.
- Platform - YugabyteDB Anywhere metrics.
- Prometheus - YugabyteDB Anywhere Prometheus metrics.
- TServer Export - YB-TServer metrics; these are scraped from the TServer processes, and mostly cover the storage layer.
- YSQL Export - YSQL query layer metrics.
- YCQL Export - YCQL query layer metrics.

The Prometheus metrics are stored in JSON files in the `support_bundle/YBA/promdump` directory.

docs/content/preview/yugabyte-platform/troubleshoot/universe-issues.md
```diff
@@ -234,6 +244,22 @@ Note that YBA will only collect files if you have sufficient permissions to requ
...
+
+- Master Export: Enables or disables collection of the metrics for the Yugabyte Master process.
```
ddhodge, line 253: - Why do these options all have 'Export' in the name? Seems redundant
- `node_exporter` OS level metrics - what does this mean
- Platform - why is this named Platform, that name is obsolete
- Platform metrics produced by YBA vs Prometheus metrics produced by the YBA - what is the difference?
- these 2 YBA metrics ought to be separate from the node/universe metrics rather than mixed in like this

aishwarya24: cc @asharma-yb

asharma-yb: > Why do these options all have 'Export' in the name? Seems redundant

These metrics are labelled with the `export_type` label in Prometheus and these are possible values for the label. 

>node_exporter OS level metrics - what does this mean

Node Exporter is a tool used in Prometheus to collect system-level metrics from various hardware and OS parameters, such as CPU, memory, disk, and network usage.

>Platform - why is this named Platform, that name is obsolete

Again, the name of label in prometheus. But you are right, maybe we should change it in the UI to a newer name. 

>Platform metrics produced by YBA vs Prometheus metrics produced by the YBA - what is the difference? All these metrics are supposedly Prometheus

Yes we should word this better, but one is the metrics for YBA while the other is metrics for prometheus itself. 

>these 2 YBA metrics ought to be separate from the node/universe metrics rather than mixed in like this

We don't have a way today to collect metrics outside of the universe so they are present here. 
Another reason begin, customers don't like creating support bundle too often, so it would be a hassle to first create a universe bundle and then collect YBA/Prometheus metrics. Here we get both in one go. 

Also CCing @ionthegeek incase I missed anything and also the feature is based on one of the support tools: https://github.com/yugabyte/prometheus-remote-backfill?tab=readme-ov-file#exporters

docs/content/stable/yugabyte-platform/troubleshoot/universe-issues.md
```diff
@@ -231,6 +238,22 @@ Note that YBA will only collect files if you have sufficient permissions to requ
...
   - Persistentvolumeclaims.yaml
 
+#### Prometheus metrics
+
+The Prometheus metrics for the selected universe. You can customize the timeframe and the metrics to collect.
+
+You can include the following metrics with the support bundle:
+
+- Master Export - YB-Master metrics.
+- Node Export - node_exporter OS level metrics.
```
ddhodge, line 248:
```suggestion
- Node Export - System-level metrics for various hardware and OS parameters, such as CPU, memory, disk, and network usage, collected by Prometheus Node Exporter.
```

docs/content/stable/yugabyte-platform/troubleshoot/universe-issues.md
```diff
@@ -231,6 +238,22 @@ Note that YBA will only collect files if you have sufficient permissions to requ
...
+- Platform - YugabyteDB Anywhere metrics.
+- Prometheus - YugabyteDB Anywhere Prometheus metrics.
```
ddhodge, line 250:
```suggestion
- Prometheus - metrics for Prometheus.
```

PR #24881, opened 2024-11-12
[docs] Added release version convention for older format

aishwarya24:
@netlify /preview/releases/versioning/#preview-release-convention-prior-to-v2024-1


ddhodge:
Addressed in another PR

PR #25512, opened 2025-01-07
[doc][2.25] Update docs to remove limitation of tablespaces for colocated tables

utkarsh-um-yb:
Tablespaces support for colocated tables and databases was introduced in 2024.2.0
Copy to preview and add section to Colocation page


ddhodge (approved):
LGTM

PR #25532, opened 2025-01-08
[DOC-624][doc]Add caution about using explicit PREPARE/EXECUTE call

sudhanshu456:
@netlify /stable/develop/best-practices-ysql/#re-use-query-plans-with-prepared-statements


docs/content/stable/develop/best-practices-ysql.md
```diff
@@ -199,6 +199,18 @@ Whenever possible, use [prepared statements](../../api/ysql/the-sql-language/sta
 For more details, see [Prepared statements in PL/pgSQL](https://dev.to/aws-heroes/postgresql-prepared-statements-in-pl-pgsql-jl3).
 {{</lead>}}
 
+{{<warning title="Caution">}}
```
ddhodge, line 202:
```suggestion
{{<warning title="Avoid explicit PREPARE or EXECUTE">}}
```

docs/content/stable/develop/best-practices-ysql.md
```diff
@@ -199,6 +199,18 @@ Whenever possible, use [prepared statements](../../api/ysql/the-sql-language/sta
...
+
+When using server-side pooling, avoid explicit `PREPARE`/`EXECUTE` calls and use protocol-level prepared statements instead. Explicit prepare/execute calls can make connections sticky, preventing you from utilizing the benefits of the connection manager or server-side pooling.
```
ddhodge, line 204:
```suggestion
When using server-side pooling, avoid explicit PREPARE and EXECUTE calls and use protocol-level prepared statements instead. Explicit prepare/execute calls can make connections sticky, which prevents you from realizing the benefits of using YSQL Connection Manager and server-side pooling.
```

docs/content/stable/develop/best-practices-ysql.md
```diff
@@ -199,6 +199,18 @@ Whenever possible, use [prepared statements](../../api/ysql/the-sql-language/sta
...
+
+For example, with the NPGSQL driver, add the following to your connection string:
```
ddhodge, line 206:
```suggestion
For example, with the Npgsql driver, add the following to your connection string:

```

docs/content/stable/develop/best-practices-ysql.md
```diff
@@ -199,6 +199,18 @@ Whenever possible, use [prepared statements](../../api/ysql/the-sql-language/sta
...
+`Max Auto Prepare=100;Auto Prepare Min Usages=5;`.
```
ddhodge, line 207: make this a proper code block. no period.

docs/content/stable/develop/best-practices-ysql.md
ddhodge, line 206: Why? What does this do? I don't understand the link between not using explicit PREPARE and doing this.

markpyb:
the information is correct

docs/content/stable/develop/best-practices-ysql.md
markpyb: I have found a doc where some engineer has explained it some time back

https://docs.google.com/document/d/17J8RI97z0nTW8sx_XDLJUQYm57BFsCInR2TDLjL7jdg/edit?tab=t.0#heading=h.32man0onzr5u

markpyb: "For protocol level prepared statements the location of phase type (prepare/bind/execute), ppstmt name and query is fixed and  known to the odyssey. So odyssey is easily able to identify and modify the ppstmt name or prepare a query on the connection. But this is not true in case we want to use a PREPARE query (by either protocol)."

markpyb:
it looks like i cant see the preview version until @ddhodge suggestions are commited

docs/content/stable/develop/best-practices-ysql.md
ddhodge: I still don't understand the link with setting Max Auto Prepare and Auto Prepare Min Usages in the connections string. What do these options do/how do they help.

ddhodge: @markpyb Does every driver have some equivalent setting to turn on prepared statements? So would something like this work -

"Depending on your driver, you may have to set some parameters to leverage prepared statements. For example, Npgsql supports automatic preparation using the Max Auto Prepare and Auto Prepare Min Usages connection parameters, which you add to your connection string as follows:

Max Auto Prepare=100;Auto Prepare Min Usages=5;

Consult your driver documentation."

premkumr (approved):
LGTM

docs/content/stable/develop/best-practices-ysql.md
markpyb: Hi @ddhodge yes every driver will have a different way to do it (if they support it) yes consult your driver documentation is good

docs/content/stable/develop/best-practices-ysql.md
markpyb: we can add the consult your driver documentation bit

markpyb (approved):

docs/content/stable/develop/best-practices-ysql.md
sudhanshu456: @ddhodge I've updated the PR, please have a look.

docs/content/stable/develop/best-practices-ysql.md
```diff
@@ -199,6 +199,21 @@ Whenever possible, use [prepared statements](../../api/ysql/the-sql-language/sta
 For more details, see [Prepared statements in PL/pgSQL](https://dev.to/aws-heroes/postgresql-prepared-statements-in-pl-pgsql-jl3).
 {{</lead>}}
 
+{{<warning title="Avoid explicit PREPARE or EXECUTE">}}
+
```
ddhodge, line 203:
```suggestion

When using server-side pooling, avoid explicit PREPARE and EXECUTE calls and use protocol-level prepared statements instead. Explicit prepare/execute calls can make connections sticky, which prevents you from realizing the benefits of using YSQL Connection Manager and server-side pooling.

```

docs/content/stable/develop/best-practices-ysql.md
```diff
@@ -199,6 +199,21 @@ Whenever possible, use [prepared statements](../../api/ysql/the-sql-language/sta
 For more details, see [Prepared statements in PL/pgSQL](https://dev.to/aws-heroes/postgresql-prepared-statements-in-pl-pgsql-jl3).
 {{</lead>}}
 
+{{<warning title="Avoid explicit PREPARE or EXECUTE">}}
```
ddhodge, line 202: put the new stuff before the lead link

PR #25560, opened 2025-01-09
Improving Azure backup config docs

chidmuthu:
Also want to add this image into docs to show where container URL can be found. 
Screenshot 2025-01-09 at 12 05 05 PM


chidmuthu:
LGTM!

PR #25664, opened 2025-01-16
[DOC-610] Clarify the meaning of clientcert=1 in TLS docs

iSignal:
`md5 clientcert=1` in the hba conf file was documented to mean a combination of md5 and cert - however, there is a slight distinction. The cert method verifies the client cert and authenticates the user using the CN of the client cert. clientcert=1 only verifies the client cert.

### DIFF
--- a/docs/content/preview/secure/tls-encryption/tls-authentication.md
+++ b/docs/content/preview/secure/tls-encryption/tls-authentication.md
@@ -25,10 +25,9 @@ The default (auto-generated) configuration in the `ysql_hba.conf` file depends o
 The four default cases are shown in the following table.
 
 | | Auth disabled | Auth enabled |
-| ---|---|---|
+| :--- | :--- | :--- |
 | TLS disabled | `host all all all trust`</br>(no ssl, no password) | `host all all all md5`</br>(no ssl, password required) |
 | TLS enabled | `hostssl all all all trust`</br>(require ssl, no password) | `hostssl all all all md5`</br>(require ssl and password) |
-
 
 Additionally, `ysql_hba_conf_csv` can be used to manually configure a custom HBA configuration.
 
@@ -37,6 +36,10 @@ For instance, to use TLS with both password authentication and client certificat
 ```sh
 hostssl all all all md5 clientcert=verify-full
 ```
+
+{{< note title="Note" >}}
+To use the client certificate only for verification (signed by the CA) but not for authentication, you can set `clientcert` to verify-ca.
+{{< /note >}}
 
 The `ysql_hba_conf_csv` rules are added above the auto-generated rules in the `ysql_hba.conf` file, so if they do not match the connection type, database, user, or host, then the auto-generated rules (that is, from the table above) may still be used.
 
@@ -208,7 +211,3 @@ ysqlsh (11.2-YB-{{<yb-version version="preview" format="build">}})
 SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
 Type "help" for help.
 ```
-
-{{< note title="Note" >}}
-To use the client certificate only for verification (signed by the CA) but not for authentication, the clientcert parameter can be set to verify-ca.
-{{< /note >}}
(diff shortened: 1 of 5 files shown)


docs/content/preview/secure/tls-encryption/tls-authentication.md
```diff
@@ -30,15 +30,15 @@ The four default cases are shown in the following table.
 | TLS enabled | `hostssl all all all trust`</br>(require ssl, no password) | `hostssl all all all md5`</br>(require ssl and password) |
 
 {{< note title="Note" >}}
-Before version 2.5.2, when TLS was enabled the default was to use the more strict `cert` option when auth was disabled, and `md5 clientcert=1` (effectively md5 + cert) when auth was enabled.
+Before version 2.5.2, when TLS was enabled the default was to use the more strict `cert` option when auth was disabled, and `md5 clientcert=verify-ca` (password auth + cert verification) when auth was enabled.
```
ddhodge, line 33:
```suggestion
Before YugabyteDB v2.5.2, when TLS was enabled the default was to use the more strict `cert` option when auth was disabled, and `md5 clientcert=verify-ca` (password auth + cert verification) when auth was enabled.
```

docs/content/stable/secure/tls-encryption/tls-authentication.md
```diff
@@ -30,12 +30,12 @@ The four default cases are shown in the following table.
 | TLS enabled | `hostssl all all all trust`</br>(require ssl, no password) | `hostssl all all all md5`</br>(require ssl and password) |
 
 {{< note title="Note" >}}
-Before version 2.5.2, when TLS was enabled the default was to use the more strict `cert` option when auth was disabled, and `md5 clientcert=1` (effectively md5 + cert) when auth was enabled.
+Before version 2.5.2, when TLS was enabled the default was to use the more strict `cert` option when auth was disabled, and `md5 clientcert=1` (md5 auth + cert verification) when auth was enabled.
```
ddhodge, line 33:
```suggestion
Before YugabyteDB v2.5.2, when TLS was enabled the default was to use the more strict `cert` option when auth was disabled, and `md5 clientcert=1` (md5 auth + cert verification) when auth was enabled.
```

docs/content/preview/secure/tls-encryption/tls-authentication.md
```diff
@@ -30,15 +30,15 @@ The four default cases are shown in the following table.
...
 {{< /note >}}
 
 Additionally, `ysql_hba_conf_csv` can be used to manually configure a custom HBA configuration.
 
 For instance, to use TLS with both `md5` and `cert` authentication, you can set the `ysql_hba_conf_csv` flag as follows:
```
ddhodge, line 38:
```suggestion
For instance, to use TLS with both password authentication and client certificate verification, you can set the `ysql_hba_conf_csv` flag as follows:
```

docs/content/stable/secure/tls-encryption/tls-authentication.md
```diff
@@ -178,9 +178,9 @@ SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 25
 Type "help" for help.
 ```
 
-### TLS with password and certificate authentication
+### TLS with password authentication and certificate verification
 
-This configuration requires the client to use client-to-server encryption and authenticate with both the appropriate certificate and the password to connect.
+This configuration requires the client to use client-to-server encryption, supplying a signed client certificate, while using a password to authenticate. Note that the server verifies that the client certificate is signed by the configured CA but it does not use the CN (Common Name) of the client certificate to authenticate the user. It is not possible to configure the server to require authentication via both password and client certificate.
```
ddhodge, line 183: should these changes also go in preview

iSignal: Good point. In preview, I will add a note talking about verify-ca, just in case someone is interested. In general, I suspect people use TLS + password + no client certs or  TLS + client certs + no password so we don't need a separate section for clientcert=verify-ca

PR #25769, opened 2025-01-27
Add alter column actions

premkumr:


jasonyb:
ALTER COLUMN SET/DROP DEFAULT is supported by 0294a9c70c116aec6d705412948d31717a0df586.  The rest, I'm not sure about.  Please find the commit that enables them and also test that they work.

ddorian (approved):
LGTM (assuming you've tried the new options). 
I think we need to backport (but excluding "add/set generated" since it wasn't in PG 11).

premkumr:
Tested on 2.25 & 2.23.1.0

PR #25771, opened 2025-01-28
[doc] Memory default details

ddhodge:
Details on memory default flag

@netlify /preview/develop/best-practices-ycql/#memory-and-tablet-limits

### DIFF
--- a/docs/content/preview/develop/best-practices-ycql.md
+++ b/docs/content/preview/develop/best-practices-ycql.md
@@ -101,3 +101,9 @@ If your collections are immutable, or you update the whole collection in full, c
 ## TRUNCATE tables instead of DELETE
 
 [TRUNCATE](../../api/ycql/dml_truncate/) deletes the database files that store the table and is much faster than [DELETE](../../api/ycql/dml_delete/) which inserts a _delete marker_ for each row in transactions and they are removed from storage when a compaction runs.
+
+## Memory and tablet limits
+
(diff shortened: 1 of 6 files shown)


mdbridge:
cluster->universe

 universe is the correct term; YBM is just stuck with a historical mistake.  :-(

(in the general case, one universe contains multiple clusters --  the read replicas are called a cluster sadly)

mdbridge (changes requested):
> If you are not using YSQL, ensure the [use_memory_defaults_optimized_for_ysql](../../reference/configuration/yb-master/#use-memory-defaults-optimized-for-ysql) flag is set to false. Although the default setting is false, when creating a new cluster using yugabyted or YugabyteDB Anywhere, the flag is set to true, unless you explicitly set it to false.

This is what to do but doesn't explain why. 

> See [Tablet limits](../../architecture/docdb-sharding/tablet-splitting/#tablet-limits) for more information.

This doesn't really help because it's about the tablet limits not the memory usage.


maybe add something like:

This flag optimizes YugabyteDB's memory setup for YSQL, reserving a considerable amount of memory for Postgres; if you are not using YSQL then that memory is wasted when it could be helping improve performance by allowing more data to be cached.

If you want to be more specific, change considerable amount to ~25%.

I think I would use https://deploy-preview-25771--infallible-bardeen-164bc9.netlify.app/preview/reference/configuration/yb-tserver/#memory-division-flags as the more information link.

I would be tempted to leave out the part about tablet limits here.  It's not really a best practice one way or the other for YCQL.

PR #26080, opened 2025-02-17
[doc][voy] Fix download URLs

ddhodge:
Fix download URLs to correct format

@netlify /preview/yugabyte-voyager/install-yb-voyager/


ShivanshGahlot (approved):
LGTM

PR #26213, opened 2025-02-26
Typo fixed in doc

emin100:


ddhodge (approved):
LGTM. Thanks for the fix. Sorry for the delay, without the documentation tag I didn't notice it

PR #26226, opened 2025-02-27
[TA-14696][doc][yba] Add Tech advisory TA-14696 

sudhanshu456:
DOC-658 - DDL Atomicity Health Check Alert is Wrong

@netlify /preview/releases/techadvisories/ta-14696/


premkumr:
@eugeneckim , @dukedyer  - Could we wrap this up ?

eugeneckim (approved):
LGTM!

dukedyer (changes requested):
There is a discrepancy between the versions we list up at the top in "Fixed In" vs the versions we list in "The following versions have a fixed Health Check template:" 

These should be brought into alignment

dukedyer (approved):

dukedyer:
I'm good too. Added the review

PR #26341, opened 2025-03-11
[doc] Update required CMake version to 3.31.0

es1024:
Also update the instructions for installing CMake, since package manager versions are too old. Context: #26346
@netlify /preview/contribute/core-database/build-from-src-almalinux/


ddhodge (approved):
Looks like it could all go in cmake.md include? Otherwise LGTM

PR #26404, opened 2025-03-17
[#26073][2.25.1] Doc: pg_export_snapshot document changes

utkarsh-um-yb:
DOC-663

### DIFF
--- a/docs/content/preview/explore/ysql-language-features/advanced-features/snapshot-synchronization.md
+++ b/docs/content/preview/explore/ysql-language-features/advanced-features/snapshot-synchronization.md
@@ -12,13 +12,13 @@ menu:
 type: docs
 ---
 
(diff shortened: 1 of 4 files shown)


docs/content/preview/explore/ysql-language-features/advanced-features/snapshot-synchronization.md
```diff
@@ -0,0 +1,137 @@
...
+
+Note that a snapshot ID becomes invalid as soon as the exporting transaction ends. Therefore, a snapshot can only be set if the exporting transaction is still active.
+
+## Example
+
+{{% explore-setup-single %}}
+
+The following example demonstrates how to use snapshot export and import across two sessions to maintain a consistent view of the data.
+
+<table>
```
ddhodge, line 64: Because the sessions don't really intersect (that is, session 2 follows session 1, without ever switching back) and the commands are wide and therefore don't display very well, I would suggest not putting them in a table and simply have the one follow the other. You couls even leave it in the table but have a single column.

ddhodge (approved):

Issue #26448, opened 2025-03-19
[Docs] Azure default image won't work in airgapped environments for 2025.1


chidmuthu:
### Description

Azure default image for Almalinux 9 which we switched to for 2025.1+ does not have polkit installed so in airgapped modes we won't be able to configure systemd services. This causes universe deployment to work. We should call this out on Azure page, probably https://docs.yugabyte.com/preview/yugabyte-platform/configure-yugabyte-platform/azure/#linux-version-catalog

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

PR #26450, opened 2025-03-19
[DOC-264][#26448] Add airgapped cloud provider node requirements

chidmuthu:
Adding different doc updates regarding airgapped environments.
1. Requiring polkit to be installed on any airgapped machine
2. Specifying that Azure default image in 2025.1+ is not suitable for airgapped and customer must provide their own image.


ddhodge (approved):
LGTM

PR #26524, opened 2025-03-24
Fix incorrect deleted row count

### LINKED ISSUE #25305: [YSQL] Deleted row count reported is incorrect when using DELETE with USING clause
### Description

**Test Case:**
```
create extension if not exists pgcrypto;
drop table if exists funds;

-- creating a table with duplicate values for k.
create table funds(k integer, u uuid default gen_random_uuid(), v integer);

insert into funds(k, v) values (1, 1);
insert into funds(k, v) values (1, 1);
insert into funds(k, v) values (1, 1);

-- deduplicate rows with same k. Keep only the one with largest value of u.
delete from funds f1
      using funds f2
   where f1.u < f2.u and f1.k = f2.k;

select * from funds;
```
We would expect the DELETE to delete two rows, but it seems to be reporting an incorrect number of rows (3).

```
delete from funds f1
      using funds f2
   where f1.u < f2.u and f1.k = f2.k;
DELETE 3
``` 

It seems to be just a reporting mistake. Because table does seem to have 1 remaining row left and only 2 rows seem to have gotten deleted.

```
select * from funds;
 k |                  u                   | v
---+--------------------------------------+---
 1 | bb7e5ed3-2cf8-45be-8c3e-e568c7e8b4ba | 1
```

### Issue Type

kind/bug

Francisco-Morao:
The deleted row count was incorrect when using the `USING` clause. The problem was caused by the `target_tuple_fetched` variable always being set to true, preventing `rows_affected_count` from updating correctly and skipping the break statement. As a result, `es_processed` was not incremented as expected.

The fix involves introducing an auxiliary variable (`rows_affected_aux`) to store the number of rows affected by `YBCExecWriteStmt`. Then, this value is accumulated into `rows_affected_count`, ensuring accurate tracking.

Fix #25305

### DIFF
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgDelete.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgDelete.java
@@ -238,4 +238,28 @@ public class TestPgDelete extends BasePgSQLTest {
       assertEquals(expectedRows, getSortedRowList(returning));
     }
   }
+
+  @Test
+  public void testBasicDelete3() throws SQLException {
+    String tableName = "test_basic_del3";
+    createSimpleTable(tableName);
+
+    // Fill in the table:
+    List<Row> expectedRows = new ArrayList<>();
+    try (Statement insert_stmt = connection.createStatement()) {
+      String insert_format = "INSERT INTO %s(h, r, vi) VALUES(%d, %d, %d)";
+      for (long h = 0; h < 3; h++) {
+        String insert_text = String.format(insert_format, tableName, 1, h, h);
+        insert_stmt.execute(insert_text);
+      }
+    }
+
+    // Verify RETURNING clause.
+    try (Statement statement = connection.createStatement()) {
+      int deletedRowCount  = statement.executeUpdate("DELETE FROM test_basic_del3 f1 "
+          + "USING test_basic_del3 f2 WHERE f1.r < f2.r and f1.h = f2.h;");
+
+      assertEquals(2, deletedRowCount);
+    }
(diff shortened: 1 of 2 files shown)


karthik-ramanathan-3006:
@Francisco-Morao - Thank you for contributing to the yugabyte-db repository!

`UPDATE` statements with a `FROM clause` also suffer from the same problem - returning inflated counts when a JOIN producing duplicate rows is used to feed the Update/Delete node. 

Ideally, we'd like to use the same fix in both cases. Could you please try your fix with UPDATEs?
I suspect that this might not work as `rows_affected_count` would no longer return 0 for the duplicate row.

Francisco-Morao:
> @Francisco-Morao - Thank you for contributing to the yugabyte-db repository!
> 
> `UPDATE` statements with a `FROM clause` also suffer from the same problem - returning inflated counts when a JOIN producing duplicate rows is used to feed the Update/Delete node.
> 
> Ideally, we'd like to use the same fix in both cases. Could you please try your fix with UPDATEs? I suspect that this might not work as `rows_affected_count` would no longer return 0 for the duplicate row.

I just tried to make the change, but it did not work. The problem is most likely in `YBCExecuteUpdate`, as `row_found` always returns 1 so `true` regardless of the conditions. Investigating and solving this bug would require more time. Maybe we should open a new issue for this if one does not already exist.

Francisco-Morao:
@sushantrmishra

Francisco-Morao:
@sushantrmishra @karthik-ramanathan-3006, Is there anything I should change or improve?

PR #26542, opened 2025-03-25
[doc] Update Transparent hugepage guidance

ddhodge:
Update Transparent hugepage guidance

@netlify /preview/deploy/manual-deployment/system-config/#enable-transparent-hugepages

### DIFF
--- a/docs/content/preview/deploy/manual-deployment/system-config.md
+++ b/docs/content/preview/deploy/manual-deployment/system-config.md
@@ -266,7 +266,17 @@ It is generally not necessary to adjust the kernel command line if the output is
 
 However, if the value is set to "madvise" or "never", you should modify your kernel command line to set transparent hugepages to "always".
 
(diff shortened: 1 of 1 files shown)


docs/content/preview/deploy/manual-deployment/system-config.md
```diff
@@ -266,7 +266,17 @@ It is generally not necessary to adjust the kernel command line if the output is
 
 However, if the value is set to "madvise" or "never", you should modify your kernel command line to set transparent hugepages to "always".
 
-You should consult your operating system documentation to determine the best way to modify a kernel command line argument for your operating system.
+In addition, you should use the following settings:
```
lingamsandeep, line 269: The line above "It is generally not necessary to adjust the kernel command line if the output is as follows:" should be reworded to indicate that even if the output is [always], the should still verify and update the max_ptes_none and defrag settings.

lingamsandeep (approved):

PR #26756, opened 2025-04-09
[doc][ybm] Anomalies and Insights

ddhodge:
DOC-687
DOC-700

@netlify /yugabyte-cloud/cloud-monitor/monitor-anomalies/


docs/content/preview/yugabyte-cloud/cloud-monitor/monitor-anomalies.md
```diff
@@ -0,0 +1,163 @@
...
+    parent: cloud-monitor
+    weight: 420
+type: docs
+---
+
+Use Anomalies to monitor your cluster for anomalies in performance - whether with the database or applications.
+
+To detect anomalies, click the cluster **Perf Advisor** tab, choose **Anomalies**, and click **Run Detection**.
+
+For meaningful results, run your workload for at least an hour before running the scan.
```
kylehailey, line 21: Anomalies is usable almost in real time because of the cluster load chart.
I think we should rephrase this. Running the scan or waiting an hour is the exception and not the rule IMO.
I'd put this info lower down
Plus I want to verify that it's really an hour between anomaly checks

ddhodge:
Replaced by #30914

PR #26767, opened 2025-04-10
Remove stale content in docs. 

rthallamko3:
Couple of changes:

1. Remove stale notes for stream_compression_algo  gflag, regarding upgrade from 2.4 as it is not longer needed. Most customers are on 2.18+ versions.
2. Added additional justification for the default value of stream_compression_algo.


ddhodge (approved):
LGTM

Issue #26761, opened 2025-04-09
[Docs] Remove clone with sequences limitation from the Docs


yamen-haddad:
### Description

Now that cloning with sequences is supported on 2024.2 as of https://github.com/yugabyte/yugabyte-db/issues/21467 , We should remove the limitation from the Docs in both 2024.2 and preview release.

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

PR #26794, opened 2025-04-10
[#26761][Docs] Remove clone with sequences limitation from the Docs

yamen-haddad:
Now that cloning with sequences is supported on 2024.2 as of https://github.com/yugabyte/yugabyte-db/issues/21467 , We should remove the limitation from the Docs in both 2024.2 and preview release.
Also removed some old limitations in the stable release as they have been addressd.


ddhodge (approved):
LGTM, but not sure what is going on with the go files, maybe revert those

Issue #12449, opened 2022-05-10
[YSQL][LST] ERROR:  duplicate key value violates unique constraint "pg_attribute_relid_attnum_index"


def-:
### Description

This failure happened using the [Long-running System Test](https://github.com/yugabyte/yb-long-system-test/) framework I'm currently working on. Yugabyte-DB is on code state 817b50bd83a8590474b86741b8592504dc6edd7c, Release build on Almalinux 8:
```
$ git checkout 92ea26cdb35dad8c962aba03495c8353a28a24fa && python3.9 ./long_system_test.py --nodes=127.0.0.1:5433,127.0.0.2:5433,127.0.0.3:5433 --threads=8 --seed=813346 --complexity=full
[...]
2022-05-10 16:32:07,443 worker_4   ERROR    Unexpected query failure: UniqueViolation
Query: ALTER TABLE tg4_1 ADD c3_text TEXT;
  values: None
  runtime: 2022-05-10 16:32:07.409 - 2022-05-10 16:32:07.442
  supports explain: False
  supports rollback: False
  affected rows: None
Action: AddColumnAction
Error class: UniqueViolation
Error code: 23505
Error message: ERROR:  duplicate key value violates unique constraint "pg_attribute_relid_attnum_index"
CONTEXT:  Catalog Version Mismatch: A DDL occurred while processing this query. Try again.
Transaction isolation level: repeatable read
DB Node: host: 127.0.0.2, port: 5433
DB Backend PID: 127648
```
[lst_2022-05-10_16:31:09_813346.zip](https://github.com/yugabyte/yugabyte-db/files/8663485/lst_2022-05-10_16.31.09_813346.zip)

def-:
Also happens with other pg attributes:
```
2022-06-29 10:30:40,061 worker_14  ERROR    Unexpected query failure: UniqueViolation
Query: ALTER TABLE t3 ADD c33_numrange NUMRANGE DEFAULT '(60.7317313429661,85.58827923606171]'::NUMRANGE;
  values: None
  runtime: 2022-06-29 10:30:39.891 - 2022-06-29 10:30:40.061
  supports explain: False
  supports rollback: False
  affected rows: None
Action: AddColumnAction
Error class: UniqueViolation
Error code: 23505
Error message: ERROR:  duplicate key value violates unique constraint "pg_attrdef_adrelid_adnum_index"
    @           0x997be8  errmsg
[...]

PR #26900, opened 2025-04-21
[#12449] Recommend using a single conn for DDL scripts & migrations

iSignal:
See https://github.com/yugabyte/yugabyte-db/issues/12449 for more information on the original context.

### DIFF
--- a/docs/content/stable/develop/best-practices-ysql.md
+++ b/docs/content/stable/develop/best-practices-ysql.md
@@ -186,9 +186,9 @@ Set up different pools with different load balancing policies as needed for your
 For more information, see [Connection pooling](../../drivers-orms/smart-drivers/#connection-pooling).
 {{</lead>}}
 
-### DDLs and connection pools
+### Database migrations and connection pools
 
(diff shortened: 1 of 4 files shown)


docs/content/stable/develop/best-practices-ysql.md
```diff
@@ -186,6 +186,10 @@ Set up different pools with different load balancing policies as needed for your
 For more information, see [Connection pooling](../../drivers-orms/smart-drivers/#connection-pooling).
 {{</lead>}}
 
+### Database migrations and connection pools
+
+Note that, in some special cases, connection pools may trigger unexpected errors while running a sequence of database migrations or other DDL operations. Due to the distributed nature of YugabyteDB, it may take a while for the result of a DDL to fully propagate to all caches on all nodes in the cluster. After a DDL statement completes, the next DDL statement that runs right afterwards on a different postgres connection may, in rare cases, see errors such as `duplicate key value violates unique constraint "pg_attribute_relid_attnum_index"`, as described [here](https://github.com/yugabyte/yugabyte-db/issues/12449). It is recommended to use a single connection while running a sequence of DDL operations as is common with application migration scripts with tools such as Flyway or Active Record Migrations.
```
ddhodge, line 191:
```suggestion
In some special cases, connection pools may trigger unexpected errors while running a sequence of database migrations or other DDL operations. Due to the distributed nature of YugabyteDB, it can take a while for the result of a DDL to fully propagate to all caches on all nodes in a cluster. As a result, after a DDL statement completes, the next DDL statement that runs right afterwards on a different PostgreSQL connection may, in rare cases, see errors such as `duplicate key value violates unique constraint "pg_attribute_relid_attnum_index"` (see issue {{<issue 12449>}}). It is recommended to use a single connection while running a sequence of DDL operations, as is common with application migration scripts with tools such as Flyway or Active Record.
```

Issue #11605, opened 2022-02-28
[Docs] Port 7000 Must Be Allowed From Master Leader to Master Follower for Admin UI to Work


shawnweeks:
### Description

We should explicitly say that you have to allow port 7000 from a Master Follower to the Master Leader if you want the Admin UI to respond on all the Master Nodes. The documentation just says port 7100 and 9100 must be allowed between nodes but 7000 needs to be allowed too. If you don't do this the non leader just never responds on port 7000 and the logs record a timeout trying to connect to the leader.

https://docs.yugabyte.com/latest/deploy/checklist/#network

PR #26919, opened 2025-04-23
[doc] Port 7000 on master nodes

ddhodge:
Clarify port 7000 on master nodes needs to be accessible to other nodes
#11605 

@netlify /preview/reference/configuration/default-ports/


ddhodge:
Replacing with fixed base

PR #26921, opened 2025-04-23
[DOC-712][2024.2.3][yba] prepare to upgrade yba

ddhodge:
DOC-712
DOC-679
DOC-409

@netlify /preview/yugabyte-platform/upgrade/prepare-to-upgrade/


nkhogen:
Node agent and systemd are two different things. Legacy universe services like master, tserver may be managed by cron. We want to upgrade them to use systemd by manually running Upgrade Systemd from the actions drop-down.

Node agent is another new service like master or tserver. By default, it is managed by systemd. As it's a new service, we need to backfill into those existing deployments. This backfill process is done automatically in the background. The requirement is that the universes should have already migrated to use systemd for other services.

PR #27000, opened 2025-04-30
[DOC-746] add postgresql schema workaround in voyager docs

sudhanshu456:
@netlify /preview/yugabyte-voyager/known-issues/postgresql/#redundant-indexes

### DIFF
--- a/docs/content/preview/yugabyte-voyager/known-issues/postgresql.md
+++ b/docs/content/preview/yugabyte-voyager/known-issues/postgresql.md
@@ -57,6 +57,7 @@ Review limitations and implement suggested workarounds to successfully migrate d
 - [Two-Phase Commit](#two-phase-commit)
 - [DDL operations within the Transaction](#ddl-operations-within-the-transaction)
 - [Hotspots with range-sharded timestamp/date indexes](#hotspots-with-range-sharded-timestamp-date-indexes)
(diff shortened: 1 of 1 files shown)


docs/content/preview/yugabyte-voyager/known-issues/postgresql.md
```diff
@@ -1601,3 +1602,30 @@ CREATE INDEX idx_orders_created ON orders(shard_id HASH, created_at DESC);
 
 SELECT * FROM orders WHERE shard_id IN (0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15) AND created_at >= NOW() - INTERVAL '1 month'; -- for fetching orders of last one month
 ```
+
+### Redundant indexes
+
+**Description**: Redundant indexes are those that are fully covered by a stronger index, one that includes the same leading key columns in the same order and may also contain additional columns. The presence of such a stronger index makes the redundant one unnecessary, as it serves the same purpose and more.
```
ddhodge, line 1608:
```suggestion
**Description**: A redundant index is an index that duplicates the functionality of another index or is unnecessary because the database can use an existing index to achieve the same result. This happens when multiple indexes cover the same columns or when a subset of columns in one index is already covered by another. 
```

ddhodge (approved):
1 comment

PR #27092, opened 2025-05-07
[DOC-755] Multi cluster smart driver wording

ddhodge:
Multi cluster smart driver wording

@netlify /preview/drivers-orms/smart-drivers/#connect-to-multiple-clusters


jasonriddell (approved):
Yes, I think this covers it nicely.

PR #27197, opened 2025-05-15
[PLAT-17169][yba] Transparent hugepages

ddhodge:
Transparent hugepages config for legacy provisioning

@netlify /preview/yugabyte-platform/prepare/server-nodes-software/


nbhatia-yb:
Thanks Dwight. This looks great! Can we please also get this reviewed by Eugene and Yu-Shen? Thanks

PR #27225, opened 2025-05-16
[Docs] Autoscroll right nav where the right nav is a very long one

samiahmedsiddiqui:
@netlify /preview/reference/configuration/yb-tserver/


ddhodge (approved):
LGTM

PR #27493, opened 2025-06-04
Latest JDBC smart driver works with Java 8+

ashetkar:
- The removed line is no longer valid with the latest JDBC Smart driver.
- The HikariCP and driver versions are already updated.


aishwarya24 (approved):
LGTM

PR #27524, opened 2025-06-06
[DOC-796] Add ysql connection manager limitations in doc

manav-yb:
This diff adds 2 additional limitations of ysql connection manager in the public docs.


ddhodge (approved):
LGTM

PR #27713, opened 2025-06-20
[DOC-709] Updated 2.25 Performance numbers

aishwarya24:


aishwarya24:
Will re-open this when we revamp Benchmark section.

PR #27994, opened 2025-07-17
[DOC-783][yba] Airgap and Linux versions

ddhodge:
Airgap requires custom Linux versions

@netlify /preview/yugabyte-platform/create-deployments/create-universe-multi-zone/


Jethro-M (approved):
LGTM

PR #28091, opened 2025-07-27
[DOC-893] Remove Remove option from Tables tab

ddhodge:
No longer remove tables from replication via the Tables list
2024.2/2025.1/2.25.1

@netlify /preview/yugabyte-platform/manage-deployments/xcluster-replication/xcluster-replication-setup/#tables


Jethro-M (approved):
LG, thanks!

PR #28162, opened 2025-08-05
[DOC-899] Add xrefs to Replica identity

ddhodge:
Add xrefs for Replica identity

@netlify /preview/reference/configuration/yb-tserver/#ysql-yb-default-replica-identity

### DIFF
--- a/docs/content/preview/develop/change-data-capture/using-logical-replication/key-concepts.md
+++ b/docs/content/preview/develop/change-data-capture/using-logical-replication/key-concepts.md
@@ -99,7 +99,7 @@ Replica identity is a table-level parameter that controls the amount of informat
 - FULL
 - NOTHING
 
-The replica identity `INDEX` is not supported in YugabyteDB.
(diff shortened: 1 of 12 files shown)


docs/content/preview/develop/change-data-capture/using-logical-replication/key-concepts.md
```diff
@@ -99,17 +99,19 @@ Replica identity is a table-level parameter that controls the amount of informat
 - FULL
 - NOTHING
```
aishwarya24, line 101:
```suggestion

For more information , refer to [Replica Identity](../yugabytedb-connector/#replica-identity).
```

aishwarya24: I suggest moving the replica identity link here for details on the 4 values here rather than further down ..

vaibhav-yb (commented):
Content wise the changes look good to me.

PR #28241, opened 2025-08-11
Unify yb tools argument syntax to double dash

SrivastavaAnubhav:
We use a mix of single dash and double dash flags in our docs. This creates a bit of confusion. Single dash is usually used for "short" names, so this diff changes all the long-form flags (like `--master_addresses`) to use two dashes.
DOC-910


ddhodge (approved):
LGTM

PR #28286, opened 2025-08-14
[DOC-915] Removed /feature endpoint as its deprecated.

### LINKED ISSUE #28281: [Docs] Documentation is out of date for YBA. It refers to /features endpoint which has been deprecated in favor of runtime config
### Description

On this page: https://docs.yugabyte.com/stable/yugabyte-platform/back-up-restore-universes/configure-backup-storage/#amazon-s3

This is out of date. We no longer use the /features endpoint. We should be directing users to modify the runtime config via Admin -> Advanced 

The option is called: "Enable Path Access Style for Amazon S3"
There may be other pages that refer to /features as well -- those should also be remediated.

aishwarya24:
fixes #28281 
@netlify /preview/yugabyte-platform/back-up-restore-universes/configure-backup-storage/


ddhodge (commented):
LG but it may only apply to preview and stable; cohesity page also refers to it

PR #28289, opened 2025-08-14
[DOC-894] Renamed Network wait event type to RPCWait

aishwarya24:
@netlify /stable/explore/observability/active-session-history/


ddhodge (approved):
LGTM. what versions

PR #28298, opened 2025-08-16
[DOC-904] memberof not a valid column

ddhodge:
Remove memberof from Role examples

@netlify /preview/secure/authorization/create-roles/


docs/content/preview/secure/authorization/ysql-create-roles.md
aishwarya24, line 1: Add an alias to this file

PR #28319, opened 2025-08-19
[DOC-923] Clarify the import releases page

ddhodge:
Clarify the import releases page

@netlify /preview/yugabyte-platform/manage-deployments/ybdb-releases/

### DIFF
--- a/docs/content/preview/yugabyte-platform/manage-deployments/ybdb-releases.md
+++ b/docs/content/preview/yugabyte-platform/manage-deployments/ybdb-releases.md
@@ -16,6 +16,8 @@ type: docs
 
 Before you can upgrade your universe to a specific version of YugabyteDB, verify that the release is available in YugabyteDB Anywhere and, if necessary, import it.
 
+## View YugabyteDB releases
+
 To view the YugabyteDB releases that are available in your YugabyteDB Anywhere instance, do the following:
 
 - Click the user profile icon and choose **Releases**.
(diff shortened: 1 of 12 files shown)


docs/content/preview/yugabyte-platform/manage-deployments/ybdb-releases.md
```diff
@@ -28,25 +28,38 @@ To view the release details, select a release in the list.
...
 
+## Import a YugabyteDB release
+
+If the YugabyteDB release you want to use for your universe is not available in your YugabyteDB Anywhere instance, you must import it.
+
+If you are running YugabyteDB Anywhere airgapped, download the YugabyteDB Linux release package and copy it to the VM where your YugabyteDB Anywhere instance is running.
+
+If your YugabyteDB Anywhere instance has internet access, you can import the release from a bucket or an internal server using a URL.
+
+You can obtain YugabyteDB releases by navigating to the release from the [Releases](../../../releases/ybdb-releases/) page.
```
aishwarya24: Question: Are all the OS types we have in the YugabyteDB Releases supported- https://docs.yugabyte.com/preview/releases/ybdb-releases/v2025.1/#downloads?

ag26jan: @aishwarya24 Yes, so Linux x86 & Linux ARM is what enterprise customer needs based on machine architecture. rest all is MacOS based i.e for local yugabyte testing/playing etc.  

Regarding the comment:

> If you are running YugabyteDB Anywhere airgapped, download the YugabyteDB Linux release package and copy it to the VM where your YugabyteDB Anywhere instance is running.

We don't need to copy in YBA VM. When they use this upload option from UI, this means they are uploading a tar file from local browser and then YBA will start putting that file on node where needed, so YBA is actually downloading the file from you (in simple term). So we need to correct the wordings there. 

CC: @nkhogen

PR #28332, opened 2025-08-20
[doc] Correct units for wait queue metric

ddhodge:
Correct units for wait queue metric

@netlify /preview/architecture/transactions/concurrency-control/#metrics


pao214 (approved):
LGTM

PR #28434, opened 2025-08-29
Doc: Remove references to `local all all trust` hba conf setting

hari90:
From version 2025.1 the DB automatically adds `local all yugabyte trust` to the pg hba conf, so this is not necessary for the user to do anymore.


ddhodge (approved):
LGTM

yamen-haddad (approved):

PR #28503, opened 2025-09-05
build(postgres): add parallelization

HasithDeAlwis:
## Summary of PostgreSQL Build Parallelization Changes

1. **Added parallel processing support** by importing `concurrent.futures` and implementing a `ThreadPoolExecutor` to build multiple PostgreSQL directories concurrently instead of sequentially.
2. **Refactored build logic** by extracting the directory build process into a reusable `_build_directory_with_make()` method and creating a `build_directory_worker()` function for parallel execution.


python/yugabyte/build_postgres.py
```diff
@@ -814,54 +844,101 @@ def make_postgres(self) -> None:
...
+                return self._build_directory_with_make(
+                    work_dir, make_cmd, make_cmd_suffix)
 
-                if self.build_type != 'compilecmds' or work_dir == self.pg_build_root:
-                    self.run_make_install(make_cmd, make_cmd_suffix)
-                else:
-                    logging.info(
-                            "Not running 'make install' in the %s directory since we are only "
-                            "generating the compilation database", work_dir)
+        with TimestampSaver(self.pg_prefix, file_suffix='.h') as _:
```
HasithDeAlwis, line 866: Use TimestampSaver to wrap the entire parallel operation, this avoids conflicts from multiple threads scanning the same directory

python/yugabyte/build_postgres.py
```diff
@@ -18,6 +18,7 @@
 """
 
 import logging
+import concurrent.futures
```
jasonyb: sort

python/yugabyte/build_postgres.py
```diff
@@ -504,9 +505,9 @@ def set_env_vars(self, step: str) -> None:
 
         for var_name in COMPILER_AND_LINKER_FLAG_ENV_VAR_NAMES:
             os.environ[var_name] = adjust_compiler_or_linker_flags(
-                    os.environ.get(var_name, '') + ' ' + ' '.join(additional_c_cxx_flags),
-                    step,
-                    flag_var_name=var_name)
+                os.environ.get(var_name, '') + ' ' + ' '.join(additional_c_cxx_flags),
+                step,
+                flag_var_name=var_name)
```
jasonyb: undo unrelated style changes. same elsewhere.  if it bothers you, make a separate PR fixing them

python/yugabyte/build_postgres.py
```diff
@@ -814,54 +844,101 @@ def make_postgres(self) -> None:
...
+                    work_dir, make_cmd, make_cmd_suffix)
 
-                if self.build_type != 'compilecmds' or work_dir == self.pg_build_root:
-                    self.run_make_install(make_cmd, make_cmd_suffix)
-                else:
-                    logging.info(
-                            "Not running 'make install' in the %s directory since we are only "
-                            "generating the compilation database", work_dir)
+        with TimestampSaver(self.pg_prefix, file_suffix='.h') as _:
+            max_workers = self.get_make_parallelism() or min(len(work_dirs), 4)
```
jasonyb: get_make_parallelism will never return falsey

python/yugabyte/build_postgres.py
```diff
@@ -784,17 +785,46 @@ def run_make_install(self, make_cmd: List[str], make_cmd_suffix: List[str]) -> N
...
+            stdout_stderr_prefix='make_install',
+            cwd=work_dir,
+            error_ok=True,
+            # TODO: get rid of shell=True.
+            shell=True,
+        ).print_output_and_raise_error_if_failed()
+        logging.info("Successfully ran 'make install' in the %s directory in %.2f sec",
+                     work_dir, time.time() - start_time_sec)
+
+    def _build_directory_with_make(self, work_dir: str, make_cmd: List[str], make_cmd_suffix: List[str]) -> Optional[str]:
```
jasonyb: line too long (>100 chars)

jasonyb (changes requested):
Thanks.  logic looks good.  left nits.  it's still unclear whether this results in faster build.  testing should remove pg_parquet from build as that skews build times:

```diff
diff --git i/src/postgres/third-party-extensions/Makefile w/src/postgres/third-party-extensions/Makefile
index c383c0400e..6a5f0b80cc 100644
--- i/src/postgres/third-party-extensions/Makefile
+++ w/src/postgres/third-party-extensions/Makefile
@@ -11,7 +11,6 @@ DIRS = \
     pg_partman \
     wal2json \
     pgtap \
-    pg_parquet \
 
 ifneq ($(findstring YB_ENABLE_YSQL_DOCUMENTDB_EXT, $(CXXFLAGS)),)
     DIRS += documentdb
```

jasonyb:
Got these compilation errors (dependency issue?): https://gist.github.com/jasonyb/476f7a4449b7a04af189c6c6a1212e33

python/yugabyte/build_postgres.py
```diff
@@ -248,6 +249,167 @@ def SavedEnviron(*remove_vars: str, **update_vars: str) -> Generator:
...
+        compile_commands_path = _build_directory_with_make_standalone(
+            work_dir, make_cmd, final_make_cmd_suffix, build_type, build_root, export_compile_commands)
+        
+        return (work_dir, compile_commands_path, None)
+        
+    except Exception as exc:
+        return (work_dir, None, exc)
+
+
+def _build_directory_with_make_standalone(
```
HasithDeAlwis: These `standalone` functions are neccessary at the moment since a proccess won't have access to the `self`. I'll work tomorrow to refactor the logic here to reuse code and not just copy and paste it

python/yugabyte/build_postgres.py
```diff
@@ -248,6 +249,167 @@ def SavedEnviron(*remove_vars: str, **update_vars: str) -> Generator:
                 env.pop(var, None)
 
 
+def _build_directory_worker_process(args: Tuple[str, List[str], List[str], str, str, str, bool, List[str], Dict[str, Any]]) -> Tuple[str, Optional[str], Optional[Exception]]:
```
HasithDeAlwis: will reduce the complexity of this, a temporary function needed to act as a wrapper for the process

HasithDeAlwis:
> > I switched from using threads to processes
> 
> This should only be done if there's a CPU bottleneck on the python process (which shouldn't be in this case since the cpu stuff is happening outside it). I don't understand what `true isolation` means when you're just running a command?

The builds would never pass on the CI (but would on my local machine), I believe this was because for each `make install` we run, we do an `os.chdir()`. And since we're on threads, the builds would fail since they share the same env. With processes, we avoid this entirely. I think it simplifies the entire process, but let me know if this is a bad approach.

python/yugabyte/build_postgres.py
```diff
@@ -780,21 +781,53 @@ def write_debug_scripts(self, env_script_content: str) -> None:
 
         run_program(['chmod', 'u+x', make_script_path])
 
-    def run_make_install(self, make_cmd: List[str], make_cmd_suffix: List[str]) -> None:
-        work_dir = os.getcwd()
+    def run_make_install(self, make_cmd: List[str], make_cmd_suffix: List[str], cwd: str) -> None:
```
jasonyb, line 784:
```suggestion
    def run_make_install(self, make_cmd: List[str], make_cmd_suffix: List[str], work_dir: str) -> None:
```

python/yugabyte/build_postgres.py
```diff
@@ -780,21 +781,53 @@ def write_debug_scripts(self, env_script_content: str) -> None:
...
+        run_program(
+            shlex_join(complete_make_install_cmd),
+            stdout_stderr_prefix='make_install',
+            cwd=cwd,
+            error_ok=True,
+            # TODO: get rid of shell=True.
+            shell=True,
+        ).print_output_and_raise_error_if_failed()
+        logging.info("Successfully ran 'make install' in the %s directory in %.2f sec",
+                     cwd, time.time() - start_time_sec)
```
jasonyb, line 796:
```suggestion
            cwd=work_dir,
            error_ok=True,
            # TODO: get rid of shell=True.
            shell=True,
        ).print_output_and_raise_error_if_failed()
        logging.info("Successfully ran 'make install' in the %s directory in %.2f sec",
                     work_dir, time.time() - start_time_sec)
```

python/yugabyte/build_postgres.py
```diff
@@ -814,54 +847,64 @@ def make_postgres(self) -> None:
...
+                if main_compile_commands_path:
+                    pg_compile_commands_paths.append(main_compile_commands_path)
+
+            self.write_debug_scripts(env_script_content)
+
+            pg_config_dir = os.path.dirname(self.pg_config_path)
+
+            # Temporarily add the directory with pg_config to PATH so that
+            # contrib/ and other makefiles can find it.
+            # necessary for MacOs arm64 since pg_config is in /opt/homebrew/bin
```
jasonyb, line 884: Is this truly only a mac issue?  Previously highlighted linux failures here: https://gist.github.com/jasonyb/476f7a4449b7a04af189c6c6a1212e33

python/yugabyte/build_postgres.py
```diff
@@ -814,54 +847,64 @@ def make_postgres(self) -> None:
...
+            with SavedEnviron("MAKELEVEL", PATH=f"{pg_config_dir}:{os.environ.get('PATH', '')}", YB_PG_SKIP_CONFIG_STATUS='1'):
+                with concurrent.futures.ThreadPoolExecutor(max_workers=min(make_parallelism, len(work_dirs))) as executor:
```
jasonyb, line 886: line too long

python/yugabyte/build_postgres.py
```diff
@@ -814,54 +847,64 @@ def make_postgres(self) -> None:
...
+                    pg_compile_commands_paths.append(main_compile_commands_path)
+
+            self.write_debug_scripts(env_script_content)
+
+            pg_config_dir = os.path.dirname(self.pg_config_path)
+
+            # Temporarily add the directory with pg_config to PATH so that
+            # contrib/ and other makefiles can find it.
+            # necessary for MacOs arm64 since pg_config is in /opt/homebrew/bin
+            with SavedEnviron("MAKELEVEL", PATH=f"{pg_config_dir}:{os.environ.get('PATH', '')}", YB_PG_SKIP_CONFIG_STATUS='1'):
```
jasonyb, line 885: PATH ending in `:` doesn't sound like a good idea to me

python/yugabyte/build_postgres.py
```diff
@@ -814,54 +847,64 @@ def make_postgres(self) -> None:
...
-                    if not os.path.exists(compile_commands_path):
-                        raise RuntimeError("Failed to generate compilation database at: %s" %
-                                           compile_commands_path)
-                    pg_compile_commands_paths.append(compile_commands_path)
+        with TimestampSaver(self.pg_prefix, file_suffix='.h') as _:
+            # Postgresql requires MAKELEVEL to be 0 or non-set when calling its make.
+            # But in the case where the YB project is built with make,
+            # MAKELEVEL is not 0 at this point. We temporarily unset MAKELEVEL to
+            # deal with this
+            with SavedEnviron('MAKELEVEL', YB_PG_SKIP_CONFIG_STATUS='1'):
```
jasonyb, line 871: why are you setting `YB_PG_SKIP_CONFIG_STATUS='1'`?  This does not match the old code.

jasonyb, line 871: setting environment variables is not thread-safe. You could pass YB_PG_SKIP_CONFIG_STATUS directly to `run_program` using the [env param of Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen), which `run_program` uses under the hood.

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment. This mapping can be str to str on any platform or bytes to bytes on POSIX platforms much like [os.environ](https://docs.python.org/3/library/os.html#os.environ) or [os.environb](https://docs.python.org/3/library/os.html#os.environb).

python/yugabyte/build_postgres.py
jasonyb, line 885: ditto why are you setting YB_PG_SKIP_CONFIG_STATUS

jasonyb, line 885: Why is this PATH manipulation needed now and not without your parallelization changes?  What is the connection?  Give me a reason I can understand.

python/yugabyte/build_postgres.py
jasonyb, line 886: number of parallel threads and make parallelism sharing the same variable `make_parallelism` isn't actually intuitive.  You are potentially making the parallelism n^2.

PR #28641, opened 2025-09-19
[DOC-980] Recommended python version

ddhodge:
Recommended python version is 3.9

@netlify /preview/deploy/manual-deployment/install-software/


docs/content/preview/quick-start/include-prerequisites-linux.md
```diff
@@ -9,16 +9,12 @@ Before installing YugabyteDB, ensure that you have the following available:
 
 1. One of the [supported operating systems](/preview/reference/configuration/operating-systems/).
 
-1. Python 3. To check the version, execute the following command:
+1. Python 3. Version 3.9 is recommended. To check the version, execute the following command:
```
aishwarya24: I think we should be specific to the minimum Python 3 version that works than mentioning the recommendation.

PR #28763, opened 2025-10-01
[#21603] Add missing --read-time help text to ysql_dump

kenju:
close https://github.com/yugabyte/yugabyte-db/issues/21603

## Changes Made:

Add missing `--read-time` option to `ysql_dump` scriptl.

## Validated Via:

Validated that the flag help text is shown in a locally built binaries:

```sh
$ ./yb_build.sh --ninja release --target postgres
$ ./build/release-clang-dynamic-arm64-ninja/postgres/bin/ysql_dump --help | grep "read-time"
  --read-time=TIMESTAMP    read data as of specific time (Unix timestamp in microseconds)
```


rthallamko3:
@OlegLoginov , Can you help with the review?

yamen-haddad (changes requested):
Thanks @kenju for looking into this.
Indeed, this option was missing on purpose as it will be used to expose a new feature where we support taking backups even during DDL operations.
We already have a [diff](https://phorge.dev.yugabyte.com/D43598) that adds this option and will land after enough testing for the new feature has been added.

rthallamko3:
@kenju , @yamen-haddad  has the same change in review as part of #20480  and is performing internal testing before landing the change. Can you abandon this PR in favor of the other one that is already in flight.

Issue #17375, opened 2023-05-16
[DocDB] Add metrics around MemTracker::GcTcmalloc


bmatican:
### Description

We should add some metrics around how many times we call it and how much memory we release at a time.
This is primarily useful for the tserver, but could also be valuable for PG, as we now use a MemTracker in PG and we've observed, in some benchmarks, that the PG time spent on the MemTracker can add up to 5% total time to the scan queries.
cc @spolitov @timothy-e 

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

albertchokor:
Hi @bmatican, I will have pinged you on Slack for this issue first. I am just formally taking up this issue here with @eafdeafd for our UT Austin Distributed Computing Project. If this issue is still up for grabs, please let us know about any pointers/where to start.

Thank you,
Albert

albertchokor:
Hi @bmatican @rthallamko3, I will have pinged you on Slack for this issue first. I am just formally taking up this issue here with @eafdeafd for our UT Austin Distributed Computing Project. If this issue is still up for grabs, please let us know about any pointers/where to start.

Thank you,
Albert

albertchokor:
Hi @bmatican @rthallamko3 @timothy-e, I will have pinged you on Slack for this issue first. I am just formally taking up this issue here with @eafdeafd for our UT Austin Distributed Computing Project. If this issue is still up for grabs, please let us know about any pointers/where to start.

Thank you,
Albert

PR #28777, opened 2025-10-02
[#17375] Add metrics around MemTracker::GcTcmalloc

kenju:
Resolves https://github.com/yugabyte/yugabyte-db/issues/17375

## Changes Made:

Adds three metrics to track MemTracker::GcTcmallocIfNeeded() performance, addressing the "5% total time" impact on scan queries mentioned in #17375.

1. GC Call Frequency
```cpp
METRIC_DEFINE_counter(server, mem_tracker_gc_tcmalloc_calls,
                      "MemTracker GC TCMalloc Calls",
                      yb::MetricUnit::kOperations,
                      "Number of times MemTracker::GcTcmallocIfNeeded() has been called");
```

~~2. Total Memory Released~~
(Edited) this was dropped as `mem_tracker_gc_tcmalloc_bytes_per_gc_sum`, which actually reports the value with `mem_tracker_gc_tcmalloc_bytes_per_gc_sum`. See https://github.com/yugabyte/yugabyte-db/pull/28777#issuecomment-3407836439 for more details.

<details>
<summary>Original proposal for `mem_tracker_gc_tcmalloc_bytes_released`</summary>

```cpp
METRIC_DEFINE_counter(server, mem_tracker_gc_tcmalloc_bytes_released,
                      "MemTracker GC TCMalloc Bytes Released",
                      yb::MetricUnit::kBytes,
                      "Total number of bytes released by MemTracker::GcTcmallocIfNeeded()");
```
</details>

3. Distribution of Memory Released Per GC
```cpp
METRIC_DEFINE_histogram(server, mem_tracker_gc_tcmalloc_bytes_per_gc,
                        "MemTracker GC TCMalloc Bytes Per Call",
                        yb::MetricUnit::kBytes,
                        "Histogram of bytes released per call to MemTracker::GcTcmallocIfNeeded()",
                        60000000LU, 2);
```

## Test Verified:

Verified in the following set up:

- Multi-node clusters started 
- Prometheus/Grafana set up to scrape from `/prometheud-metrics` endpoint
- Created an ad-hod dashboard
- Doing a couple of load testing over `bin/ysqlsh` doing INSERT/SELECT queries against large dataset

Here is a screenshot:

memtracker-gc-metrics

### DIFF
--- a/src/yb/util/mem_tracker.cc
+++ b/src/yb/util/mem_tracker.cc
@@ -882,7 +882,8 @@ scoped_refptr<Histogram> MemTracker::gc_tcmalloc_bytes_per_gc_metric_;
 
 void MemTracker::InitializeGcMetrics(const scoped_refptr<MetricEntity>& metric_entity) {
   gc_tcmalloc_calls_metric_ = METRIC_mem_tracker_gc_tcmalloc_calls.Instantiate(metric_entity);
-  gc_tcmalloc_bytes_per_gc_metric_ = METRIC_mem_tracker_gc_tcmalloc_bytes_per_gc.Instantiate(metric_entity);
+  gc_tcmalloc_bytes_per_gc_metric_ =
+      METRIC_mem_tracker_gc_tcmalloc_bytes_per_gc.Instantiate(metric_entity);
 }
 
 scoped_refptr<MetricEntity> MemTracker::metric_entity() const {


src/yb/util/mem_tracker.cc
```diff
@@ -119,6 +119,22 @@ DEFINE_RUNTIME_int64(mem_tracker_tcmalloc_gc_release_bytes, -1,
...
+METRIC_DEFINE_counter(server, mem_tracker_gc_tcmalloc_bytes_released,
+                      "MemTracker GC TCMalloc Bytes Released",
+                      yb::MetricUnit::kBytes,
+                      "Total number of bytes released by MemTracker::GcTcmallocIfNeeded()");
+
+METRIC_DEFINE_histogram(server, mem_tracker_gc_tcmalloc_bytes_per_call,
+                        "MemTracker GC TCMalloc Bytes Per Call",
+                        yb::MetricUnit::kBytes,
+                        "Histogram of bytes released per call to MemTracker::GcTcmallocIfNeeded()",
+                        60000000 /* 60MB as max value */, 2 /* 2 digits precision */);
```
kenju: ☝🏻 just a hypothetical number which is large enough to accommodate most of real world use cases, would love to hear your feedback if this needs to be updated

src/yb/yql/pggate/pggate.cc
```diff
@@ -673,6 +673,8 @@ PgApiImpl::PgApiImpl(
       ybctid_reader_provider_(pg_session_),
       fk_reference_cache_(ybctid_reader_provider_, buffering_settings_, tablespace_map_),
       explicit_row_lock_buffer_(ybctid_reader_provider_, tablespace_map_) {
+  MemTracker::InitializeGcMetrics(metric_entity_);
```
kenju, line 677: https://github.com/yugabyte/yugabyte-db/issues/17375#issue-1712663516
> This is primarily useful for the tserver, **but could also be valuable for PG,**

If I understand it correctly, we'd like to initialise metrics here as well for PG

kenju:
@rthallamko3 Another PR for improving metrics - can I get a reviewer for this PR please?

src/yb/util/mem_tracker.cc
yusong-yan: Let's use `60000000LU`

src/yb/util/mem_tracker.cc
```diff
@@ -762,6 +784,19 @@ void MemTracker::GcTcmallocIfNeeded() {
...
     }
+
+    int64_t final_overhead = GetTCMallocPageHeapFreeBytes();
+    int64_t bytes_released = initial_overhead - final_overhead;
+    if (gc_tcmalloc_bytes_released_metric_ && bytes_released > 0) {
+      gc_tcmalloc_bytes_released_metric_->IncrementBy(bytes_released);
+    }
+    if (gc_tcmalloc_bytes_per_call_metric_ && bytes_released > 0) {
+      gc_tcmalloc_bytes_per_call_metric_->Increment(bytes_released);
+    }
```
yusong-yan: `bytes_released` could theoretically be negative if memory increased during GC (unlikely but possible due to concurrent allocations). The > 0 check prevents recording negative values, which is good.
I think we should add a comment to indicate that the metric doesn't perfectly record number of byte is released(as well as the metric description)

src/yb/util/mem_tracker.cc
```diff
@@ -762,6 +784,19 @@ void MemTracker::GcTcmallocIfNeeded() {
...
+  } else {
+    if (gc_tcmalloc_bytes_per_call_metric_) {
+      gc_tcmalloc_bytes_per_call_metric_->Increment(0);
+    }
```
yusong-yan: Do you think it might make more sense to change the metric semantics to represent “bytes released per GC” instead of “bytes released per call”?
Since we record 0 even when no GC happens, if there are many such calls, it could skew the histogram (especially the average and percentiles) toward 0, making it harder to interpret the real GC behavior.

src/yb/util/mem_tracker.cc
kenju: Good catch thanks ✅ https://github.com/yugabyte/yugabyte-db/pull/28777/commits/6108969b25172d3cf12089e4f574548242e17171

src/yb/util/mem_tracker.cc
kenju: > (unlikely but possible due to concurrent allocations)

Good point. I can naively come up with following scenarios:

```
Time T0: initial_overhead = 100MB (measured before GC)
Time T1: GC starts, releases 50MB from page heap
Time T2: Other threads allocate 80MB
Time T3: final_overhead = 130MB (measured after GC)
(Results) bytes_released = 100MB - 130MB = -30MB
```

I believe the current approach still give us some pictures around GC than having no metrics at all.

Added code comments & improve metrics description ✅ https://github.com/yugabyte/yugabyte-db/pull/28777/commits/04ca6f7af34187d8d4b684cd0051cd7baf000eea

src/yb/util/mem_tracker.cc
kenju: > it might make more sense to change the metric semantics to represent “bytes released per GC” instead of “bytes released per call”?

Stepping back a little bit, I think that's right - we're interested in GC behaviour and not particularly in `GcTcmallocIfNeeded()` calls anyway. Thanks for the feedback!

Stopped recording 0 to avoid skewing the aggregation result ✅ https://github.com/yugabyte/yugabyte-db/pull/28777/commits/9f6c439a5d6b0053b616e2b6170f7dbf60800896

PR #28805, opened 2025-10-04
[DOC-994] Add note on legacy provisioning

ddhodge:
Add note that legacy provisioning in YBA will be removed in 2025.2

@netlify /preview/releases/yba-releases/v2025.1/#highlights


aishwarya24 (approved):
LGTM

PR #28854, opened 2025-10-09
[DOC-1002] Warning for Azure VM image upgrade

aishwarya24:
@netlify /preview/yugabyte-platform/manage-deployments/upgrade-nodes-csp/

### DIFF
--- a/docs/content/preview/yugabyte-platform/manage-deployments/upgrade-nodes-csp.md
+++ b/docs/content/preview/yugabyte-platform/manage-deployments/upgrade-nodes-csp.md
@@ -53,7 +53,7 @@ Before patching or upgrading the operating system, verify that the Linux version
 
 {{< warning title="Azure SSH user restriction" >}}
 
-For Azure deployments, changing the SSH user from the original image bundle is currently not supported and will cause VM image upgrade to fail. Ensure that the SSH user configured in your Azure provider matches the default SSH user of the target Linux image.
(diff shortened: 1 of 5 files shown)


docs/content/preview/yugabyte-platform/manage-deployments/upgrade-nodes-csp.md
```diff
@@ -51,6 +51,12 @@ If your universe uses a custom Linux version and you want to upgrade to a new cu
 
 Before patching or upgrading the operating system, verify that the Linux version image that you intend to use can successfully run on a standalone virtual machine (VM) with the same instance type as the nodes you are patching. You can do this by creating a new VM with the same instance type as your universe and using the selected Linux version for the boot disk and verifying that the VM boots successfully.
 
+{{< warning title="Azure SSH user restriction" >}}
+
+For Azure deployments, changing the SSH user from the original image bundle is currently not supported and will cause VM image upgrade to fail. Ensure that the SSH user configured in your Azure provider matches the default SSH user of the target Linux image.
```
muthu-kroger: It's not about the default SSH user of linux image that is not a concept for Azure. In YBA they would have already created or are using some image bundle. The new image bundle has to match the user of the existing image bundle. Unfortunately there is not a way in the UI to see this until https://yugabyte.atlassian.net/browse/PLAT-18642

PR #29225, opened 2025-11-04
Clarify that Azure is not supported when using clockbound

pao214:


docs/content/preview/deploy/manual-deployment/system-config.md
```diff
@@ -51,6 +51,8 @@ sudo bash ./bin/configure_ptp.sh
 
 Although optional, configuring ClockBound improves clock accuracy by several orders of magnitude. ClockBound requires chrony and can be used in conjunction with PTP.
 
+Clockbound is supported on AWS and GCP. Azure is not supported.
```
aishwarya24, line 54: Can we just not state that "Clockbound is supported on AWS and GCP only." Should we even mention that Azure is not supported?

pao214: Depends. If we only care about optics then yes omitting this makes sense. However, if we want to make it clear to the users that Azure is not supported, then this is useful. People search using Azure, etc. And some have tried using Azure and got burned (figuratively). So, my preference is to keep it. However, I am fine with removing it as well.

ddhodge (approved):

docs/content/preview/deploy/manual-deployment/system-config.md
aishwarya24: Makes sense. Lets keep it then. Thanks!

aishwarya24:
LGTM

Issue #29451, opened 2025-11-21
[YSQL] JDBC API statement.getMoreResults() reads could case inconsistent reads


qvad:
Jira Link: [DB-19247](https://yugabyte.atlassian.net/browse/DB-19247)
### Description

Recently we added usage of `getMoreResults` JDBC method to our consistency validation workload
This cause inconsistent reads:
```
2025-11-20 23:12:05,335 [Thread-180] ERROR ExceptionsTracker - Failed to compare sums: Expected 1000000 - actual 999999, SeqScan,  WHERE id >= 0 OR id IS NULL, sum pipeline: true
2025-11-20 23:12:05,339 [Thread-115] ERROR ExceptionsTracker - Failed to compare sums: Expected 1000000 - actual 999999, IndexScan,  WHERE id >= 0 OR id IS NULL, sum pipeline: true
2025-11-20 23:12:05,341 [Thread-110] ERROR ExceptionsTracker - Failed to compare sums: Expected 1000000 - actual 1000006, IndexOnlyScan, , sum pipeline: true
2025-11-20 23:12:05,342 [Thread-101] ERROR ExceptionsTracker - Failed to compare sums: Expected 1000000 - actual 1000004, SeqScan, , sum pipeline: true
```
Looks like we always missing reads for only few keys.

Same workload pass in Postgres
```
        while (hasResult) {
            try (ResultSet rs = statement.getResultSet()) {
                if (rs != null && rs.next()) {
                    sum += rs.getInt(1);
                }
            }
            hasResult = statement.getMoreResults();
        }
        return sum;
```

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-19247]: https://yugabyte.atlassian.net/browse/DB-19247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pao214:
Potential root cause: #24463

PR #29455, opened 2025-11-23
Fix: Disable partial retries + correct RpcContext handling to resolve inconsistent reads when using statement.getMoreResults() (#29451)

Adityakk9031:
**Jira Issue:** [DB-19247]
**Issue Type:** Bug
**Area:** YSQL

---

## 🧩 Issue Summary (from #29451)

This PR fixes inconsistent reads caused when using the JDBC `getMoreResults()` API in YSQL workloads. The issue resulted in missing or duplicated row results during multi-resultset queries inside the consistency validation workload.

The same workload passes in PostgreSQL but fails in YugabyteDB with off-by-one mismatches such as:

```
Expected 1000000 - actual 999999
Expected 1000000 - actual 1000006
Expected 1000000 - actual 1000004
```

This happens because:

* YB-TServer **replayed partial batches** during tablet leader restarts.
* `RpcContext` was passed incorrectly by value, causing dropped results during multi-result RPC execution.

---

## 🔧 Root Cause

1. **Server-side transparent retries** re-executed only parts of a batched Perform RPC.
2. JDBC client using `getMoreResults()` expects all resultsets to remain intact.
3. Replays caused the first resultset to be lost, leading to missing rows.
4. Additionally, `PgClientServiceImpl::Perform()` incorrectly passed `RpcContext` by value, causing the context to go out of scope while async work was still using it.

---

## ✅ Fix Summary

This PR contains two critical fixes:

### **1. Disable Partial-Restart Retries for Multi-Result or Read Batches**

Added logic in `PgSession::Perform()`:

* Detect if the batch contains a read operation or more than one operation.
* Mark the request as `disable_retries_on_restarts = true`.

This prevents tserver from replaying only a subset of the statements, ensuring all resultsets remain consistent across retries.

### **2. Correct `RpcContext` Passing in `PgClientServiceImpl::Perform()`**

Updated function signature to pass `rpc::RpcContext*` instead of a temporary stack value.

This ensures the RPC context remains valid for the duration of async execution.

---

## 📝 Files Changed

* `src/yb/yql/pggate/pg_session.cc` — Added detection logic + new Perform option flag.
* `src/yb/yql/pggate/pg_client_service.cc` — Changed `Perform()` to correctly pass `RpcContext`.

---

## 🧪 Testing

* Re-ran the YSQL consistency validation workload with JDBC `getMoreResults()`.
* Verified that all resultsets are preserved and sum mismatch no longer occurs.
* Compared behavior with PostgreSQL — now identical.

[DB-19247]: https://yugabyte.atlassian.net/browse/DB-19247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/yb/tserver/pg_client_service.cc
+++ b/src/yb/tserver/pg_client_service.cc
@@ -2853,8 +2853,10 @@ PgClientServiceImpl::PgClientServiceImpl(
 PgClientServiceImpl::~PgClientServiceImpl() = default;
 
 void PgClientServiceImpl::Perform(
-    const PgPerformRequestPB* req, PgPerformResponsePB* resp, rpc::RpcContext context) {
-  impl_->Perform(const_cast<PgPerformRequestPB*>(req), resp, &context);
+    const PgPerformRequestPB* req,
+    PgPerformResponsePB* resp,
+    rpc::RpcContext* context) {
+  impl_->Perform(const_cast<PgPerformRequestPB*>(req), resp, context);
 }
 
 void PgClientServiceImpl::InvalidateTableCache() {
--- a/src/yb/yql/pggate/pg_session.cc
+++ b/src/yb/yql/pggate/pg_session.cc
@@ -976,43 +976,37 @@ Result<PerformFuture> PgSession::Perform(BufferableOperations&& ops, PerformOpti
     }
   }
 
-  // If all operations belong to the same database then set the namespace.
-  // System database template1 is ignored as we may read global system catalog like tablespaces
-  // in the same batch.
-  if (!ops.Empty()) {
-    auto database_oid = kPgInvalidOid;
-    for (const auto& relation : ops.relations()) {
-      if (relation.database_oid == kTemplate1Oid) {
-        continue;
-      }
-
-      if (PREDICT_FALSE(database_oid != kPgInvalidOid && database_oid != relation.database_oid)) {
-        // We do not expect this to be true. Adding a log to catch violation just in case.
-        YB_LOG_EVERY_N_SECS(WARNING, 60) << Format(
-            "Operations from multiple databases ('$0', '$1') found in a single Perform step",
-            database_oid, relation.database_oid);
-        database_oid = kPgInvalidOid;
-        break;
-      }
-
-      database_oid = relation.database_oid;
-    }
-
-    if (database_oid != kPgInvalidOid) {
-      options.set_namespace_id(GetPgsqlNamespaceId(database_oid));
-    }
+  // -----------------------------------------------------------------------
+  // NEW: Prevent partial replay for batches that may produce multiple resultsets
+  // -----------------------------------------------------------------------
+  //
+  // Problem: server-side transparent per-statement retry may replay only the
+  // failing statement within a batched request. If the client is iterating pages
+  // (getMoreResults) or expecting multiple resultsets, replaying only one
+  // statement can make earlier statements' results invisible -> missing rows.
+  //
+  // Fix: mark Perform RPCs that contain reads or multiple operations as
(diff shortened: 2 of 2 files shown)


pao214:
@Adityakk9031 I see that we have an ongoing PR: https://github.com/yugabyte/yugabyte-db/pull/28299

Do we need the same changes in this PR as well? i.e. why is Fix for Inconsistent Reads with JDBC getMoreResults() (DB-19247) dependent on PR 28299?

src/yb/tserver/pg_client_service.cc
```diff
@@ -2853,8 +2853,10 @@ PgClientServiceImpl::PgClientServiceImpl(
 PgClientServiceImpl::~PgClientServiceImpl() = default;
 
 void PgClientServiceImpl::Perform(
-    const PgPerformRequestPB* req, PgPerformResponsePB* resp, rpc::RpcContext context) {
-  impl_->Perform(const_cast<PgPerformRequestPB*>(req), resp, &context);
+    const PgPerformRequestPB* req,
```
pao214, line 2856: this is a purely stylistic change. we do not need it. please remove.

src/yb/yql/pggate/pg_session.cc
```diff
@@ -976,43 +976,37 @@ Result<PerformFuture> PgSession::Perform(BufferableOperations&& ops, PerformOpti
     }
   }
 
-  // If all operations belong to the same database then set the namespace.
```
pao214, line 886: why is this logic removed?

src/yb/yql/pggate/pg_session.cc
```diff
@@ -976,43 +976,37 @@ Result<PerformFuture> PgSession::Perform(BufferableOperations&& ops, PerformOpti
...
+  // non-retriable on the server-side. The tserver must honor
+  // `disable_retries_on_restarts` (proto + server change necessary).
+  //
+  // Heuristic: if the batch contains any read or has size > 1, mark it.
+  bool contains_read_or_multi = false;
+  try {
+    contains_read_or_multi = std::ranges::any_of(ops.operations(), [](const auto& op) {
+      return op->is_read();
+    }) || (std::distance(ops.operations().begin(), ops.operations().end()) > 1);
+  } catch (...) {
```
pao214, line 998: what exception are we expecting?

src/yb/yql/pggate/pg_session.cc
```diff
@@ -976,43 +976,37 @@ Result<PerformFuture> PgSession::Perform(BufferableOperations&& ops, PerformOpti
...
-
-    if (database_oid != kPgInvalidOid) {
-      options.set_namespace_id(GetPgsqlNamespaceId(database_oid));
-    }
+  // -----------------------------------------------------------------------
+  // NEW: Prevent partial replay for batches that may produce multiple resultsets
+  // -----------------------------------------------------------------------
+  //
+  // Problem: server-side transparent per-statement retry may replay only the
+  // failing statement within a batched request. If the client is iterating pages
```
pao214, line 984: how does getMoreResults mean from a postgres perspective? what network messages are sent? please suggest a scenario. add atleast one testcase.

this is not the appropriate solution because we only want to disable retries in the getMoreResults case.

pao214:
```
Re-ran the YSQL consistency validation workload with JDBC getMoreResults().
Verified that all resultsets are preserved and sum mismatch no longer occurs.
Compared behavior with PostgreSQL — now identical.
```

Please add an automated testcase for the same. You can add a cpp test case (see pg_txn-test) or a java test case (see TestPgBatch.java) for examples.

Issue #26138, opened 2025-02-20
[DocDB] PG does not start when mutual TLS for RPC node to node encryption is enabled


iSignal:
Jira Link: [DB-15470](https://yugabyte.atlassian.net/browse/DB-15470)
### Description

The combination of `--node_to_node_encryption_use_client_certificates=true` and `--enable_ysql=true` does not work due to issues in the way the pg -> tserver messenger client is setup.

Steps
1. Start a secure cluster with `--enable_ysql=true, --node_to_node_encryption_use_client_certificates=true, --certs_dir=certs --allow_insecure_connections=false --use_node_to_node_encryption=true --use_client_to_server_encryption=true --certs_for_client_dir=certs --certs_for_cdc_dir=certs/xcluster`
3. Attempt to connect via ysqlsh, PG fails to start a backend

```
$ bin/ysqlsh -h 127.0.0.3
ysqlsh: error: connection to server at "127.0.0.3", port 5433 failed: FATAL:  Failed to use certificate: passed a null parameter
connection to server at "127.0.0.3", port 5433 failed: FATAL:  Failed to use certificate: passed a null parameter
```

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-15470]: https://yugabyte.atlassian.net/browse/DB-15470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #7335, opened 2021-02-22
[YSQL][DocDB] Postgres fails when using `node_to_node_encryption_use_client_certificates`


Arnav15:
The following error happens when using client verification for TLS.
This is the error that is generated in the postgres log.
```
"000000010000300080000000000004ec", num_attempts: 1) failed: Invalid argument (yb/rpc/secure_stream.cc:761): Failed to use private key: passed a null parameter
2021-02-19 13:47:57.044 UTC [22985] FATAL:  Not found: Error loading table with oid 1260 in database with oid 1: Failed to use private key: passed a null parameter
```

m-iancu:
From @iSignal on Slack:

> it seems like when we call CreateSecureContext in pggate:
> https://github.com/yugabyte/yugabyte-db/blob/238099e717e51fb5c3a14080e825bcaa4ab06c18/src/yb/yql/pggate/pggate.cc
> we are not setting the third parameter (node name), so we don't end up loading the node certs/keys in the pg client: 
> https://github.com/yugabyte/yugabyte-db/blob/e092e3f47af47672235ec7d43ccc3fc47cac929d/ent/src/yb/server/secure.cc#L133

tedyu:
Maybe we can do the following:
```
diff --git a/src/yb/yql/pggate/pggate.cc b/src/yb/yql/pggate/pggate.cc
index d5f2d50bd..67d681f63 100644
--- a/src/yb/yql/pggate/pggate.cc
+++ b/src/yb/yql/pggate/pggate.cc
@@ -34,6 +34,7 @@
 #include "yb/yql/pggate/ybc_pggate.h"

 #include "yb/util/flag_tags.h"
+#include "yb/util/net/net_util.h"
 #include "yb/client/client_fwd.h"
 #include "yb/client/client_utils.h"
 #include "yb/client/table.h"
@@ -80,9 +81,12 @@ Result<PgApiImpl::MessengerHolder> BuildMessenger(
     const std::shared_ptr<MemTracker>& parent_mem_tracker) {
   std::unique_ptr<rpc::SecureContext> secure_context;
   if (FLAGS_use_node_to_node_encryption) {
+    string hostname;
+    RETURN_NOT_OK(GetHostname(&hostname));
     secure_context = VERIFY_RESULT(server::CreateSecureContext(
         FLAGS_certs_dir,
-        server::UseClientCerts(FLAGS_node_to_node_encryption_use_client_certificates)));
[...]

PR #29521, opened 2025-11-27
[#7335][#26138] YSQL: Fix postgres to work with node_to_node_encryption_use_client_certificates

the-glu:
This PR fix an issue with postgres not working when using the node_to_node_encryption_use_client_certificates option.

As described in https://github.com/yugabyte/yugabyte-db/issues/7335, this fix it by passing the node name correctly, assuring the client certificate is used.

I had to move around some flags in common to have then in pggate.

Tested in a local cluster, without any issue. I was not able to create some tests for that, if needed I will need some pointers ^^'

This should fix at least #7335 and #26138

### DIFF
--- a/bin/yugabyted
+++ b/bin/yugabyted
@@ -9500,8 +9500,9 @@ class ControlScript(object):
 
         # Initialize the binary path of ybadmin
         # TODO(Sanket): Clean up and refactor this file
-        YBAdminProxy.set_certs_dir(self.configs.saved_data.get("master_flags"), \
-            self.configs.saved_data.get("secure"), self.configs.saved_data.get("certs_dir"))
+        YBAdminProxy.set_certs_dir(
+            self.configs.saved_data.get("master_flags"), self.configs.saved_data.get("secure"),
+            self.configs.saved_data.get("certs_dir"), self.find_hostname_of_node())
 
         YBTableServerCLIProxy.set_certs_dir(self.configs.saved_data.get("master_flags"), \
             self.configs.saved_data.get("secure"), self.configs.saved_data.get("certs_dir"))
@@ -10399,22 +10400,31 @@ class YBAdminProxy(object):
         YBAdminProxy.cmd_args.append(find_binary_location("yb-admin"))
 
     @staticmethod
-    def set_certs_dir(master_flags, secure, certs_dir):
+    def set_certs_dir(master_flags, secure, certs_dir, hostname):
         # If the user is attempting to use TLS, let's point yb-admin to
         # the same certs dir as the master
+
         if secure:
             YBAdminProxy.cmd_args.append("--certs_dir_name={}".format(certs_dir))
-        elif master_flags:
+
+        if master_flags:
+
             flags_list = master_flags.split(",")
-            if 'use_node_to_node_encryption=true' not in flags_list:
-                return
-            certs_dir_name = [y for y in
-                [re.match('certs_dir=(.*)', x) for x in flags_list]
-                if y is not None]
-            if not certs_dir_name[0]:
-                raise RuntimeError("use_node_to_node_encryption=true must "
-                                "be accompanied by a certs_dir setting")
-            YBAdminProxy.cmd_args.append('--certs_dir_name={}'.format(certs_dir_name[0].group(1)))
+
+            if not secure:
+                if 'use_node_to_node_encryption=true' not in flags_list:
+                    return
+                certs_dir_name = [y for y in
+                    [re.match('certs_dir=(.*)', x) for x in flags_list]
+                    if y is not None]
+                if not certs_dir_name[0]:
+                    raise RuntimeError("use_node_to_node_encryption=true must "
+                                    "be accompanied by a certs_dir setting")
+                YBAdminProxy.cmd_args.append(
+                  '--certs_dir_name={}'.format(certs_dir_name[0].group(1)))
+
+            if 'node_to_node_encryption_use_client_certificates=true' in flags_list:
+                YBAdminProxy.cmd_args.append('--client_node_name={}'.format(hostname))
 
     @staticmethod
     def add_master(master_addrs, new_master_ip, new_master_rpc_port, timeout=30):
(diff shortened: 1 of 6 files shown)


iSignal:
@the-glu : thanks for proposing this fix. At a high level, it makes sense. Will take a closer look.

@es1024 - can you please review this change?

We should be able to enable_ysql = true for the test in https://github.com/yugabyte/yugabyte-db/blob/24777c06c0166f27b0c8a4d5878aa59f04c3593e/src/yb/integration-tests/external_mini_cluster_secure_test.cc#L220 so that we don't regress this code path.

bin/yugabyted
```diff
@@ -9501,7 +9501,7 @@ class ControlScript(object):
         # Initialize the binary path of ybadmin
         # TODO(Sanket): Clean up and refactor this file
         YBAdminProxy.set_certs_dir(self.configs.saved_data.get("master_flags"), \
-            self.configs.saved_data.get("secure"), self.configs.saved_data.get("certs_dir"))
+            self.configs.saved_data.get("secure"), self.configs.saved_data.get("certs_dir"), self.find_hostname_of_node())
```
es1024: lint: keep line under 100 chars

src/yb/yql/pggate/pggate.cc
```diff
@@ -124,10 +127,28 @@ Result<PgApiImpl::MessengerHolder> BuildMessenger(
     const scoped_refptr<MetricEntity>& metric_entity,
     const std::shared_ptr<MemTracker>& parent_mem_tracker) {
   std::unique_ptr<rpc::SecureContext> secure_context;
+
   if (FLAGS_use_node_to_node_encryption) {
+
+	auto& node_name = FLAGS_cert_node_filename;
```
es1024: lint: use spaces instead of tabs for c++ files

src/yb/common/common_flags.cc
```diff
@@ -169,6 +169,13 @@ DEFINE_test_flag(bool, check_catalog_version_overflow, false,
 DEFINE_RUNTIME_PG_FLAG(bool, yb_enable_invalidation_messages, true,
     "True to enable invalidation messages");
 
+DEFINE_UNKNOWN_string(rpc_bind_addresses, "0.0.0.0",
+              "Comma-separated list of addresses to bind to for RPC connections. "
+              "Currently, ephemeral ports (i.e. port 0) are not allowed.");
+TAG_FLAG(rpc_bind_addresses, stable);
+
+DEFINE_UNKNOWN_string(server_broadcast_addresses, "", "Broadcast addresses for this server.");
+
```
es1024: These should not be `common` flags: `common` flags are also used by tools like `yb-admin` and `yb-ts-cli`, where these flags make no sense.

The `cert_node_filename` -> `server_broadcast_addresses` -> `rpc_bind_addresses` logic is already handled by src/yb/yql/process_wrapper/common_config.cc as well, and is stored in the `cert_base_name` field of `PgProcessConf` in pgwrapper (the tserver-side code that manages the postgres process). 

We don't actually pass this value directly to postgres today (https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/yql/pgwrapper/pg_wrapper.cc#L650,L655 just uses it to specify cert/key files), but it seems reasonable to pass this value down to postgres via a new `pggate_cert_node_name` gflag, and use that instead of moving these flags and redoing the calculation.

the-glu: Ho, I did try to do it like that but I wasn't able to find exactly how data was passed around, thanks for the pointers.

I updated the PR with the new method and it seems to works as well, it's also way simpler now :)

bin/yugabyted
```diff
@@ -10415,6 +10417,12 @@ class YBAdminProxy(object):
                 raise RuntimeError("use_node_to_node_encryption=true must "
                                 "be accompanied by a certs_dir setting")
             YBAdminProxy.cmd_args.append('--certs_dir_name={}'.format(certs_dir_name[0].group(1)))
+            secure = True
+
+        if secure and master_flags:
+            flags_list = master_flags.split(",")
+            if 'node_to_node_encryption_use_client_certificates=true' in flags_list:
+                YBAdminProxy.cmd_args.append('--client_node_name={}'.format(hostname))
```
es1024, line 10427: I think this could be simplified to just:
```
if secure:
    ...
if master_flags:
    flags_list = ...
    if not secure:
        ...
        YBAdminProxy.cmd_args.append('--certs_dir_name={}'.format(certs_dir_name[0].group(1)))
    
    if 'node_to_node_encryption_use_client_certificates=true' in flags_list:
        ...
```
since the above if is the `master_flags and not secure` case and sets `secure = True` anyways

src/yb/yql/pggate/pggate_flags.cc
```diff
@@ -35,6 +35,9 @@ DEFINE_UNKNOWN_int32(pggate_ybclient_reactor_threads, 2,
 DEFINE_UNKNOWN_string(pggate_master_addresses, "",
               "Addresses of the master servers to which the PostgreSQL proxy server connects.");
 
+DEFINE_UNKNOWN_string(pggate_cert_base_name, "",
+              "Certificate base name computed by pg_wrapper");
+
```
es1024, line 40:
```suggestion
DEFINE_NON_RUNTIME_string(pggate_cert_base_name, "",
              "Certificate base name computed by pg_wrapper");

```
New flags need to be labeled RUNTIME or NON_RUNTIME, UNKNOWN ones are just uncategorized flags introduced before we started labeling them.

bin/yugabyted
the-glu: Yes, reworked :)

bin/yugabyted
```diff
@@ -10399,22 +10400,30 @@ class YBAdminProxy(object):
...
+            if not secure:
+                if 'use_node_to_node_encryption=true' not in flags_list:
+                    return
+                certs_dir_name = [y for y in
+                    [re.match('certs_dir=(.*)', x) for x in flags_list]
+                    if y is not None]
+                if not certs_dir_name[0]:
+                    raise RuntimeError("use_node_to_node_encryption=true must "
+                                    "be accompanied by a certs_dir setting")
+                YBAdminProxy.cmd_args.append('--certs_dir_name={}'.format(certs_dir_name[0].group(1)))
```
es1024: lint: line over 100 characters

the-glu: Ha sorry, fixed. Automatic linting would be great ^^'

es1024 (approved):
Thanks for making these changes!

PR #29676, opened 2025-12-11
`yugabyted`: stop child processes with `SIGTERM` instead of `SIGKILL`

Rylan12:
Currently, `yugabyted` kills child processes with `SIGKILL`, which does not allow them to perform any cleanup. For `tserver`, this means not cleaning up shared memory segments that were created. The current approach to cleaning these up is to use the previous `tserver` UUID from the configuration files in the base directory to determine the previous segment name and remove it. However, if the base directory no longer exists, this cleanup cannot be performed. As a consequence, the shared memory segments build up over time. On my machine, the limit for shared memory identifiers is 32, which means it doesn't take too long to encounter an issue with this.

This PR proposes a way to cleanup these shared segments at termination without replacing the `SIGKILL` signal. It doesn't replace the previous approach, so anything that slips through here should still be cleaned up on the next startup. However, I think this may be a slightly cleaner way to handle this.

When `yugabyted` is cleaning up child processes, it now does the following:

1. Terminate the `tserver` process using the pid stored in `self.processes`
    - This means the process will only be killed if it was actually started by `yugabyted`, which should preserve existing behavior
2. Wait 0.5 seconds for the process to clean up
3. Cleanup the shared memory segment using the `tserver` UUID from `self.configs` to generate the segment name
4. Kill the remaining child processes by stopping the entire process group, as it currently done

This should be a relatively minimal change that fails gracefully if any of the required information isn't available, but may help to avoid building up orphaned shared segments in certain cases.

### DIFF
--- a/bin/yugabyted
+++ b/bin/yugabyted
@@ -3605,14 +3605,18 @@ class ControlScript(object):
             os._exit(os.EX_OK)
         self.set_signals(SIG_DFL)
 
-        for p in self.processes.values():
-            p.delete_pidfile()
-        self.script.delete_pidfile()
-
         try:
+            for p in self.processes.values():
+                p.kill()
+
+            for p in self.processes.values():
+                p.wait_until_stop()
+
+            self.script.delete_pidfile()
+
             # Kill process group instead of self.processes to ensure
-            # any spawned child processes are killed. Use SIGKILL because YugaWare
-            # requires KILL signal to terminate and nodes currently do not gracefully terminate.
+            # any spawned child processes are killed.
+            # This will terminate any processes that didn't respond to SIGTERM.
             os.killpg(pgid, SIGKILL)
             Output.log(
                 "{} may not have terminated properly... "
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -165,3 +165,4 @@ This is a list of people who have contributed code to the [YugabyteDB](https://g
 * [utkarsh-um-yb](https://github.com/utkarsh-um-yb)
 * [aashir24-yb](https://github.com/aashir24-yb)
 * [keisku](https://github.com/keisku)
+* [Rylan12](https://github.com/Rylan12)


hari90:
@nmalladi can we make yugabyted use SIGTERM instead of SIGKILL now that the DB supports graceful shutdown?
cc @druzac

bin/yugabyted
```diff
@@ -3191,6 +3212,17 @@ class ControlScript(object):
             os._exit(os.EX_OK)
         self.set_signals(SIG_DFL)
 
+        # Kill tserver first to clean up its shared memory segments
+        if "tserver" in self.processes:
+            tserver_process = self.processes["tserver"].process
+            if tserver_process:
+                try:
+                    os.kill(tserver_process.pid, SIGKILL)
```
nchandrappa: @Rylan12 The approach to kill the tserver using `os.kill` wont kill the child processes as tserver can spawn others processes like pg, yb-controller etc. We will have to kill process group how it was originally implemented using `os.killpg(pgid, SIGKILL)` in `kill_children()` function.

Can we just invoke `cleanup_shared_memory()` before exiting in `kill_children` at 
```
        # exit no matter what
        os._exit(os.EX_OK)
```

Also, as @hari90 mentioned t-server now handles SIGTERM signal, we can update the process group kill signal to use SIGTERM instead of SIGKILL.

bin/yugabyted
```diff
@@ -3197,9 +3197,8 @@ class ControlScript(object):
 
         try:
             # Kill process group instead of self.processes to ensure
-            # any spawned child processes are killed. Use SIGKILL because YugaWare
-            # requires KILL signal to terminate and nodes currently do not gracefully terminate.
-            os.killpg(pgid, SIGKILL)
+            # any spawned child processes are killed.
+            os.killpg(pgid, SIGTERM)
```
druzac: This might leak YB processes. Graceful shutdown is not bulletproof, unfortunately, and everywhere we use `SIGTERM` we have a wait/poll loop that eventually sends `SIGKILL` if necessary.
Some examples:
Our C++ test harness which uses external processes to run yb daemons:
https://github.com/yugabyte/yugabyte-db/blob/16cdd44b872b7f596a68663f1b129ab40e2d1f83/src/yb/integration-tests/external_daemon.cc#L488-L517

Our java test harness:
https://github.com/yugabyte/yugabyte-db/blob/16cdd44b872b7f596a68663f1b129ab40e2d1f83/java/yb-client/src/test/java/org/yb/minicluster/MiniYBCluster.java#L1232-L1245

Rylan12: @druzac can you give some guidance for implementing that here? I'm not sure the best way to reliably find processes that are a part of the process group, since I think we'd need to kill them first. The outline of what I'm trying to do is something like this, but I'm not familiar with the Python invocations needed to make it work:

```python
for pid in process_group:
    # can't kill the current process until we've cleaned up everything else
    if pid != cur_pid:
        os.kill(pid, SIGTERM)

# wait some amount of time for processes to terminate on their own
wait()

# kill anything that remains including the current process
os.killpg(pgid, SIGKILL)
```

Is `self.processes` a reliable source for processes in the current process group?

Rylan12: I came up with a slightly different solution, let me know if this is now satisfied

bin/yugabyted
```diff
@@ -3191,14 +3191,15 @@ class ControlScript(object):
...
 
-        for p in self.processes.values():
-            p.delete_pidfile()
-        self.script.delete_pidfile()
-
         try:
+            for p in self.processes.values():
+                p.kill()
+            self.script.delete_pidfile()
+            time.sleep(0.5)
```
druzac: Looks good. I'd just ask you to bump the wait time, 0.5 is extremely short. Let's do 10 seconds.

druzac: Ah, a 10 second blind wait would be very annoying. Alright, 2 alternatives:

1. Keep the blind wait, and change the sleep to 3 seconds
2. Remove the blind wait - poll process status and only wait as long as any of the signalled processes are still alive.

Rylan12: Gotcha, makes sense. I added a `wait_until_stop` calls for each process, which looks like it should do the trick? Although this might just check the pidfile presence which isn't really the same as checking whether the process is actually running. Can you advise whether this approach is sufficient?

druzac: `wait_until_stop` looks fine to me. The `ProcessManager` runs a `ps`. I suppose you just went by the comment?

https://github.com/yugabyte/yugabyte-db/blob/1adabec5545c7055f817f7658957ec97309c2b07/bin/yugabyted#L10647-L10652

https://github.com/yugabyte/yugabyte-db/blob/1adabec5545c7055f817f7658957ec97309c2b07/bin/yugabyted#L10641-L10645

https://github.com/yugabyte/yugabyte-db/blob/1adabec5545c7055f817f7658957ec97309c2b07/bin/yugabyted#L10715-L10720

Rylan12: Ah yes, I see. I misunderstoof the `or` condition in `is_proc_running`. Agreed, this seems like it should work correctly, then

PR #29715, opened 2025-12-17
Update maximize-build-space action version and options

jharveysmith:
Updated maximize-build-space action to use version v10 and added additional removals.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,6 +76,7 @@ jobs:
           remove-android: 'true'
           remove-haskell: 'true'
           remove-codeql: 'true'
+          remove-docker-images: 'true'
           overprovision-lvm: 'true'
 
       - uses: actions/checkout@v4


.github/workflows/build.yml
```diff
@@ -70,9 +70,12 @@ jobs:
     steps:
       - name: Maximize build space
         # Use a specific commit for predictability and security.
-        uses: easimon/maximize-build-space@6ae56c86ea8db291ae39f62352a412c36ab8179b
+        uses: easimon/maximize-build-space@v10
```
jasonyb, line 73: looks like latest https://github.com/easimon/maximize-build-space/commit/c28619d8999a147d5e09c1199f84ff6af6ad5794 is just two commits ahead of v10 and makes logging slightly better.  No strong opinion on which to take.

.github/workflows/build.yml
```diff
@@ -70,9 +70,12 @@ jobs:
...
         with:
           remove-dotnet: 'true'
+          remove-android: 'true'
+          remove-haskell: 'true'
+          remove-codeql: 'true'
```
jasonyb, line 78: Curious what `remove-docker-images` actually means.  Since YB uses docker, sounds like it shouldn't be removed, but the description says "Removes cached Docker images", so there is a chance it is desirable to be removed.

.github/workflows/build.yml
jharveysmith: Yeah, I thought about grabbing that, but the build status for the latest commit was red for some reason (the logs are gone).

.github/workflows/build.yml
jharveysmith: I can try it.  It looks like github is using my action changes in this PR.  With the changes so far we are at 50% disk usage.

Issue #29906, opened 2026-01-08
[YSQL] Index build reads whole row from table


jasonyb:
Jira Link: [DB-19755](https://yugabyte.atlassian.net/browse/DB-19755)
### Description

At least concurrent index build (did not test nonconcurrent) appears to read whole rows from the table in order to write to the index.  This is wasteful in YugabyteDB as fetching unnecessary columns is expensive.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-19755]: https://yugabyte.atlassian.net/browse/DB-19755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #29928, opened 2026-01-09
[#29906] YSQL: Index backfill optimization to only read columns present in index definition

austenLacy:
## TLDR

Related issue: https://github.com/yugabyte/yugabyte-db/issues/29906

This PR introduces an index backfill optimization to only read columns present in index definition different from the default behavior which is to read all columns. This optimization should help significantly for wide (i.e. many columns) tables, tables with large blob type columns (e.g. jsonb, text, etc) and reduces network i/o by only transmitting data required to create the index.

Introduces a new GUC `yb_enable_index_backfill_column_projection` with a default value of `false`.

## Details

Adds a new specialized scan function `ybc_heap_beginscan_for_index_build()` that uses `IndexInfo` to determine exactly which columns are needed and only requests those from DocDB.

The function identifies required columns from:
1. **Direct index columns** (`ii_IndexAttrNumbers`) - columns directly referenced in the index key
2. **Expression index columns** (`ii_Expressions`) - columns used in index expressions like `(col1 + col2)`
3. **Partial index predicates** (`ii_Predicate`) - columns used in WHERE clauses like `WHERE col1 > 50`
4. **System columns** - always includes `ybctid` (needed for index entry construction)

The feature is gated behind a new GUC `yb_enable_index_backfill_column_projection`.

### Building and testing locally

_Compile just postgres changes (much faster than full build)_

```bash
./yb_build.sh release --target postgres --skip-java
```

_Full build to run tests_

```bash
./yb_build.sh release --skip-java
```

_Test against a locally running cluster with SQL and expected output_

```bash
./build/latest/bin/yb-master \
  --fs_data_dirs=/tmp/ybdata/disk1 \
  --webserver_interface=127.0.0.1 \
  --webserver_port=7001 \
  --rpc_bind_addresses=127.0.0.1:7100 \
  --master_addresses=127.0.0.1:7100 \
  --replication_factor=1 \
  --enable_ysql=true \
  >/tmp/ybdata/master.out 2>/tmp/ybdata/master.err &

# Important PG vars
# yb_enable_pg_stat_statements_rpc_stats=true
# yb_fetch_size_limit=1000
# yb_enable_index_backfill_column_projection=false
./build/latest/bin/yb-tserver \
  --fs_data_dirs=/tmp/ybdata/disk1 \
  --webserver_interface=127.0.0.1 \
  --webserver_port=9001 \
  --rpc_bind_addresses=127.0.0.1:9100 \
  --tserver_master_addrs=127.0.0.1:7100 \
  --enable_ysql=true \
  --pgsql_proxy_bind_address=127.0.0.1:5433 \
  --ysql_pg_conf_csv="yb_enable_pg_stat_statements_rpc_stats=true,yb_fetch_size_limit=1000,yb_enable_index_backfill_column_projection=false" \
  >/tmp/ybdata/tserver.out 2>/tmp/ybdata/tserver.err &
```

verify with 

```sq
-- create a wide table with large columns (~5KB per row)
CREATE TABLE wide_test (
    id SERIAL PRIMARY KEY,
    indexed_col INT,
    large_col1 TEXT,
    large_col2 TEXT,
    large_col3 TEXT,
    large_col4 TEXT,
    large_col5 TEXT
);

-- insert 500 rows with large data
INSERT INTO wide_test (indexed_col, large_col1, large_col2, large_col3, large_col4, large_col5)
SELECT i, repeat('a', 1000), repeat('b', 1000), repeat('c',1000), repeat('d', 1000), repeat('e', 1000)
FROM generate_series(1, 500) AS i;

-- reset stats before index creation
SELECT pg_stat_statements_reset();

-- create the index to trigger backfill
CREATE INDEX idx_test ON wide_test(indexed_col);

-- significantly less than with the optimization disabled (e.g. ~97% less)
yugabyte=# SELECT sum(docdb_read_rpcs) as rpcs
yugabyte-# FROM pg_stat_statements(true)
yugabyte-# WHERE query LIKE 'BACKFILL%';
 rpcs
------
   13
(1 row)
```

Run the unit tests with

```bash
./yb_build.sh release --cxx-test pgwrapper_pg_index_backfill-test --gtest_filter="*ColumnProjection*" --skip-java --no-odyssey --skip-pg-parquet
```

### DIFF
--- a/src/postgres/src/backend/access/yb_access/yb_scan.c
+++ b/src/postgres/src/backend/access/yb_access/yb_scan.c
@@ -3837,12 +3837,13 @@ ybcBuildScanPlanForIndexBuild(Relation relation, IndexInfo *indexInfo)
 	int			i;
 	int			idx;
 	int			resno = 1;
-	/* Use varno=1 since this is always scanning the base relation.
+	/*
+	 * Use varno=1 since this is always scanning the base relation.
 	 * Concurrent index creation in postgres is restricted to one index
 	 * per table/statement (unlike the non-concurrent index creation process).
 	 * As a result, we're guaranteed that only one table is involved in the process,
 	 * and said table is opened for inspection first, leading to it being varno=1.
-	*/
+	 */
 	const Index varno = 1;
 
 	/* This function is only for Yugabyte relations */
--- a/src/yb/yql/pgwrapper/pg_index_backfill-test.cc
+++ b/src/yb/yql/pgwrapper/pg_index_backfill-test.cc
@@ -2981,7 +2981,8 @@ TEST_P(PgIndexBackfillColumnProjectionTest, SingleColumnPartialDifferentColumn) 
 }
 
 // Multi column index with HASH and ASC
-TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnHashAsc) {
+TEST_P(PgIndexBackfillColumnProjectionTest,
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnHashAsc)) {
   ASSERT_OK(CreateWideTable(kTableName));
   ASSERT_OK(InsertTestData(kTableName));
   auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
@@ -2990,7 +2991,8 @@ TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnHashAsc) {
 }
 
 // Multi column index with compound hash key
-TEST_P(PgIndexBackfillColumnProjectionTest, YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnCompoundHash)) {
+TEST_P(PgIndexBackfillColumnProjectionTest,
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnCompoundHash)) {
   ASSERT_OK(CreateWideTable(kTableName));
   ASSERT_OK(InsertTestData(kTableName));
   auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
@@ -2999,7 +3001,8 @@ TEST_P(PgIndexBackfillColumnProjectionTest, YB_DISABLE_TEST_IN_SANITIZERS(MultiC
 }
 
 // Multi column index where col2 appears in both compound hash and range - fetched only once
-TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnDuplicateColumn) {
+TEST_P(PgIndexBackfillColumnProjectionTest,
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnDuplicateColumn)) {
   ASSERT_OK(CreateWideTable(kTableName));
   ASSERT_OK(InsertTestData(kTableName));
   auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
@@ -3008,7 +3011,8 @@ TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnDuplicateColumn) {
 }
 
 // Index columns in different order than table definition (table: col1, col2, col3)
-TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnOutOfOrder) {
+TEST_P(PgIndexBackfillColumnProjectionTest,
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnOutOfOrder)) {
   ASSERT_OK(CreateWideTable(kTableName));
   ASSERT_OK(InsertTestData(kTableName));
   auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
@@ -3017,7 +3021,8 @@ TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnOutOfOrder) {
 }
 
 // Index on (col1, col2) where PK is (col2, col3, col1) - tests column order independence
-TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnNonDefaultPkOrder) {
+TEST_P(PgIndexBackfillColumnProjectionTest,
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnNonDefaultPkOrder)) {
   ASSERT_OK(CreateWideTable(kTableName, "col2, col3, col1"));
   ASSERT_OK(InsertTestData(kTableName));
   auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs("CREATE INDEX idx ON t (col1, col2)"));
@@ -3025,7 +3030,8 @@ TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnNonDefaultPkOrder) {
 }
 
 // Multi column expression index - f(col1, col2) and g(col2), col2 fetched once
-TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnExpression) {
+TEST_P(PgIndexBackfillColumnProjectionTest,
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnExpression)) {
   ASSERT_OK(CreateWideTable(kTableName));
   ASSERT_OK(InsertTestData(kTableName));
   auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
@@ -3034,7 +3040,8 @@ TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnExpression) {
 }
 
 // Multi column expression index with partial predicate using col2
-TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnExpressionPartial) {
+TEST_P(PgIndexBackfillColumnProjectionTest,
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnExpressionPartial)) {
   ASSERT_OK(CreateWideTable(kTableName));
   ASSERT_OK(InsertTestData(kTableName));
   auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(


src/postgres/src/backend/access/heap/heapam_handler.c
```diff
@@ -1261,16 +1262,28 @@ heapam_index_build_range_scan(Relation heapRelation,
...
+		if (IsYBRelation(heapRelation) &&
+			yb_enable_index_backfill_column_projection)
 		{
+			/*
+			 * For YB relations, use the optimized scan function that only
+			 * fetches columns needed by the index. This avoids reading the
+			 * entire row from DocDB during index build/backfill.
+			 */
+			uint32		flags = SO_TYPE_SEQSCAN | SO_ALLOW_PAGEMODE |
+								SO_ALLOW_STRAT;
```
austenLacy: This logic was copied from `tableam.h#table_beginscan_strat`

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3820,6 +3820,148 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+ *   - Columns referenced in ii_Expressions (expression index columns)
+ *   - Columns referenced in ii_Predicate (partial index predicate)
+ *   - ybctid (always needed for index entry construction)
+ *
+ * This optimization significantly reduces the amount of data read from
+ * DocDB during concurrent index creation (backfill), especially for tables
+ * with many columns where only a few are indexed.
+ */
+TableScanDesc
+ybc_heap_beginscan_for_index_build(Relation relation,
```
jasonyb: this duplicates code from ybcBeginScan

src/postgres/src/include/pg_yb_utils.h
```diff
@@ -547,6 +547,14 @@ extern bool yb_enable_base_scans_cost_model;
  */
 extern bool yb_enable_update_reltuples_after_create_index;
 
+/*
+ * Enables index backfill column projection optimization.
+ * If true, index build/backfill only reads columns needed for the index,
+ * rather than all columns from the base table.
+ * Default is false (beta feature).
+ */
+extern bool yb_enable_index_backfill_column_projection;
```
karthik-ramanathan-3006, line 555: This flag can be defined as a GUC.
As an example, please take a look at `yb_enable_inplace_index_update` in guc.c (ref: [guc.c](https://github.com/yugabyte/yugabyte-db/blob/master/src/postgres/src/backend/utils/misc/guc.c#L3305))

The GUC can also be marked true, by default, for now. 
When we get closed to merging the PR, we can make a decision on its default value.

src/postgres/src/test/regress/sql/yb.orig.index_backfill_column_projection.sql
```diff
@@ -0,0 +1,133 @@
+--
```
karthik-ramanathan-3006: Thank you for adding tests. 
The way I am thinking about it, this optimization has two aspects -- performance and correctness:
 - Correctness: We want to ensure that we're fetching the correct columns as part of the scan
 - Performance: We want to validate that the data returned as part of the heap scan is minimized (when compared to the case where all columns are fetched by the scan).

Could you give a shot at implementing the latter?
The index backfill C++ test might be a good place for it (Ref: [pg_index_backfill-test.cc](https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/yql/pgwrapper/pg_index_backfill-test.cc))

This GUC might be useful to implement these tests: yb_fetch_size_limit
This GUC controls how much data is returned by a single scan. If we define the all the columns of the table being indexed to be of a fixed size, then we can calculate how many rows would be returned by the scan (with/without the optimization) and therefore how many round trips (RPCs) would be required to fetch all the matching rows from the table. The test could then assert that the number of RPCs has reduced by the expected number when the optimization is turned on.

Feel free to propose alternatives.

src/postgres/src/backend/access/heap/heapam_handler.c
```diff
@@ -1261,16 +1262,28 @@ heapam_index_build_range_scan(Relation heapRelation,
...
+				flags |= SO_ALLOW_SYNC;
+
+			scan = ybc_heap_beginscan_for_index_build(heapRelation,
+													  snapshot,
+													  0, /* number of keys */
+													  NULL, /* scan key */
+													  flags,
+													  indexInfo);
+
 			if (bfinfo)
```
karthik-ramanathan-3006, line 1287: This logic is applicable when `IsYBRelation() = true` and  `yb_enable_index_backfill_column_projection` = false. So the structure you'd want is:
```
if (IsYBRelation(heapRelation) && yb_enable_index_backfill_column_projection)
    scan = ybc_heap_beginscan_for_index_build(...)
else
    scan = table_beginscan_strat(...)

if (IsYBRelation(heapRelation) && bfinfo)
{
   // Populate exec_params
}
```

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3820,6 +3820,148 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
 }
 
+
+/*
+ * ybc_heap_beginscan_for_index_build
+ *		Begin a heap scan specifically for index build/backfill operations.
+ *
+ * Unlike the regular heap scan which fetches all columns, this function
+ * only requests the columns needed for the index:
+ *   - Columns referenced in ii_IndexAttrNumbers (direct index columns)
```
karthik-ramanathan-3006: It would be good to update this comment indicating that `ii_IndexAttrNumbers` contains both the index key columns as well as non-key columns included in the index:
```
CREATE INDEX ON t (a, b) INCLUDES (c, d);
```

`ii_IndexAttrNumbers` will contain attr numbers for [a, b, c, d]

src/postgres/src/backend/access/yb_access/yb_scan.c
karthik-ramanathan-3006: +1 to Jason's comment.

It would be good to call into `ybcBeginScan` directly, if possible.
`ybcBeginScan` populates the required_attrs for the scan from the `targetlist` of the param `Scan *pg_scan_plan`. Please check if constructing a pg_scan_plan object is feasible.
The target list can be populated using the code that you have below.

karthik-ramanathan-3006 (changes requested):
Thanks for working on this optimization.

I took an initial pass at the PR, and have left some comments/suggestions.
A good way to test your code would be to run the following collection of unit tests:
```
./yb_build.sh --cxx-test pgwrapper_pg_index_backfill-test
```

You can run the above command both with and without your optimizations in order to test for regressions.

src/postgres/src/test/regress/sql/yb.orig.index_backfill_column_projection.sql
austenLacy: That makes a lot of sense. I can look into implementing a test like that.

src/postgres/src/backend/access/heap/heapam_handler.c
austenLacy: Good catch. Thank you

austenLacy: Fixed in https://github.com/yugabyte/yugabyte-db/pull/29928/commits/8be00023c410a466981c557928d2ca7de5de9d3c

src/postgres/src/backend/access/yb_access/yb_scan.c
austenLacy: I refactored both functions to use some shared logic helpers in https://github.com/yugabyte/yugabyte-db/pull/29928/commits/a8937a137d9f41bac5281dc68da5c6a8e535557a. Let me know what you think.

src/postgres/src/include/pg_yb_utils.h
austenLacy: Added in https://github.com/yugabyte/yugabyte-db/pull/29928/commits/b40157a3fd6564b1d283b6f2cafcca662063f006

karthik-ramanathan-3006:
Thanks for updating the PR @austenLacy! Taking another pass today.

src/postgres/src/backend/access/heap/heapam_handler.c
karthik-ramanathan-3006: `SO_ALLOW_PAGEMODE`,  `SO_ALLOW_STRAT` and `SO_ALLOW_SYNC` are all postgres-storage specific options. Since we use RocksDB (DocDB) for storage, these aren't particularly relevant to us.

Only `SO_TYPE_SEQSCAN` needs to be set. The others can be skipped.

src/postgres/src/backend/access/yb_access/yb_scan.c
karthik-ramanathan-3006: `ybcBeginScan` is sort of an entry point function that most (if not all) of our scans call into. It accepts postgres scan parameters and lays it out in a format that our storage layer can understand. So it would be good to have the backfill scan also call into `ybcBeginScan` directly (rather than breaking down it into reusable components).

Looking at the params:
 - most are optional and can be set to NULL or 0 --> [rel_pushdown, idx_pushdown, aggrefs, distinct_prefixlen (0)]
 - is_internal_scan: likely needs to be false
 - xs_want_itup: false, we are not fetching index tuples (itup) 
 - fetch_ybctids_only: false, we need columns in addition to ybctids
 - pg_scan_plan: this needs to be populated as this contains the list of columns to be fetched
 - exec_params: we already have this
 
Let me know if you run into problems with this.

austenLacy: I reverted the reusable functions commit and will give this a shot.

src/postgres/src/test/regress/sql/yb.orig.index_backfill_column_projection.sql
austenLacy: I wrote `PgIndexBackfillColumnProjectionTest` in https://github.com/yugabyte/yugabyte-db/pull/29928/changes/a047226f9526a30feacaab9ae72429ff15c1bb41 which should mimic this idea.

I've been struggling to get the tests to run locally because the cluster started by `./yb_build.sh --cxx-test pgwrapper_pg_index_backfill-test` keeps dying and I haven't been able to figure out exactly why yet. Would you be able to run tests on the branch?

src/postgres/src/backend/access/yb_access/yb_scan.c
austenLacy: @karthik-ramanathan-3006 check out https://github.com/yugabyte/yugabyte-db/pull/29928/changes/c4af7c69cb7c491ac64d01091cd0850a708b21a4. I essentially build the scan plan with the columns included in the index and then call into `ybcBeginScan`. Let me know what you think.

src/postgres/src/test/regress/sql/yb.orig.index_backfill_column_projection.sql
karthik-ramanathan-3006: Thank you for taking the time to write this test!
I spent some time running it locally and analyzing the results. First up, the test failure:
```
I0114 11:24:28.557978 4210772992 pg_index_backfill-test.cc:2924] Backfill RPCs with column projection: 24
I0114 11:24:28.557992 4210772992 pg_index_backfill-test.cc:2929] RPC reduction ratio: 1
../../src/yb/yql/pgwrapper/pg_index_backfill-test.cc:2934: Failure
Expected: (rpcs_with_projection) < (rpcs_without_projection), actual: 24 vs 24
Test failure stack trace:
    @        0x1023fe2ac  testing::internal::TestEventRepeater::OnTestPartResult()
    @        0x1023e967c  testing::UnitTest::AddTestPartResult()
    @        0x1023e4f68  testing::internal::AssertHelper::operator=()
    @        0x100840318  yb::pgwrapper::PgIndexBackfillColumnProjectionTest_ReducedRpcsWithProjection_Test::TestBody()
    @        0x1023f772c  testing::internal::HandleExceptionsInMethodIfSupported<>()
    @        0x1023f7688  testing::Test::Run()
    @        0x1023f8d88  testing::TestInfo::Run()
    @        0x1023f9b78  testing::TestSuite::Run()
    @        0x102409e44  testing::internal::UnitTestImpl::RunAllTests()
    @        0x10240961c  testing::internal::HandleExceptionsInMethodIfSupported<>()
    @        0x102409598  testing::UnitTest::Run()
    @        0x101599580  RUN_ALL_TESTS()
    @        0x1015991cc  main
    @        0x192df20e0  start
```

So essentially, the test fails `rpcs_with_projection` and `rpcs_without_projection` return the same number of RPCs. Drilling deeper, the metric returned by `TotalBackfillRpcCalls()` might be the problem.
It returns `handler_latency_yb_tserver_TabletServerAdminService_BackfillIndex` which is a verbose of saying that it counts the number of control plane messages (RPCs).
With your optimization, we expect that the number of control plane messages ("create backfilled index x of table t") to remain the same. However, we expect the number of data plane messages to go down ("backfill chunk of length l offset o from table t"). 
`TotalBackfillRpcCalls()` might not be a good metric to use.

I'll post the alternative in a follow-up comment.

karthik-ramanathan-3006: We recently developed a new feature to publish RPC stats (ie. data plane messages) to [pg_stat_statements](https://docs.yugabyte.com/stable/launch-and-manage/monitor-and-alert/query-tuning/pg-stat-statements/). Using this feature, we can compare and contrast the stats with/out your optimization using the following query:
```
SELECT sum(docdb_read_rpcs) AS rpcs, sum(docdb_rows_scanned) AS rows FROM pg_stat_statements(true) WHERE query LIKE 'BACKFILL%';
```

With the optimization turned OFF and fetch size limit restricted, this should return something like:
```
yugabyte=# SELECT sum(docdb_read_rpcs), sum(docdb_rows_scanned) FROM pg_stat_statements(true) WHERE query LIKE 'BACKFILL%';
 sum | sum
-----+-----
   1 | 500
(1 row)
```

With the optimization turned ON and fetch size limit restricted, this should return:
```
yugabyte=# SELECT sum(docdb_read_rpcs), sum(docdb_rows_scanned) FROM pg_stat_statements(true) WHERE query LIKE 'BACKFILL%';
 sum | sum
-----+-----
  56 | 555
(1 row)
```

We mainly care about the first column.

src/postgres/src/test/regress/sql/yb.orig.index_backfill_column_projection.sql
```diff
@@ -0,0 +1,133 @@
...
+-- Tests for index backfill column projection optimization.
+-- This test verifies that concurrent index creation only reads columns
```
karthik-ramanathan-3006: To produce the above results, configure the following cluster-level flag:
```
ysql_pg_conf_csv='yb_enable_pg_stat_statements_rpc_stats=true,yb_fetch_size_limit=<value>,yb_enable_index_backfill_column_projection=<value>'
```

Reference this test as an example of how to set this flag: [link](https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/yql/pgwrapper/pg_libpq-test.cc#L4412)

This flag is just a programmatic way of setting postgres GUCs for every new backend in the cluster. The index backfill process spawns a bunch of worker postgres backends, which backfill the index in parallel. By setting the GUCs at a cluster level, all such workers inherit the GUCs.

 - yb_enable_pg_stat_statements_rpc_stats=true: This enables the feature I mentioned above. As this is a new feature, it is turned off by default.
 - yb_fetch_size_limit=\<value\>:  We need each backend to apply the fetch size limit individually.
 - yb_enable_index_backfill_column_projection=\<value\>: Similar to the above.

Another trick that you might find useful is to reset stat collection just before triggering `CREATE INDEX`:
```
SELECT pg_stat_statements_reset();
```

karthik-ramanathan-3006: I understand that this might be a lot to take in. Please take your time and feel free to reach out with any questions.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,181 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+			Form_pg_attribute attr = TupleDescAttr(tupdesc, attnum - 1);
+
+			var = makeVar(1,	/* varno = scanrelid */
+						  attnum,
+						  attr->atttypid,
+						  attr->atttypmod,
+						  attr->attcollation,
+						  0);	/* varlevelsup */
+		}
+		else
```
karthik-ramanathan-3006: An index cannot contain system columns other than ybctid (`YBTupleIdAttributeNumber`). This is because system columns are managed by the database system, and can change in unpredictable ways (for example ctids in postgres).

To enforce this:
 - We can assert that idx is not a system column.
 - `YBTupleIdAttributeNumber` need not be explicitly added to required_attrs. We can just make a targetlist entry out of it directly.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,181 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+											attnum -
+											FirstLowInvalidHeapAttributeNumber);
+	}
+
+	/*
+	 * Add columns referenced in index expressions.
+	 * Use varno=1 since this is always scanning the base relation.
+	 */
+	if (indexInfo->ii_Expressions != NIL)
+		pull_varattnos((Node *) indexInfo->ii_Expressions, 1, &required_attrs);
```
karthik-ramanathan-3006: Let's define `varno` as a const variable and move the comment to the declaration.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,181 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+ * index backfill. This creates Var nodes for each column needed by the index,
+ * allowing ybcBeginScan to use its standard column projection logic.
+ */
+static Scan *
+ybcBuildScanPlanForIndexBuild(Relation relation, IndexInfo *indexInfo)
+{
+	Scan	   *scan_plan;
+	TupleDesc	tupdesc = RelationGetDescr(relation);
+	List	   *targetlist = NIL;
+	Bitmapset  *required_attrs = NULL;
```
karthik-ramanathan-3006: Consider using `YbAttnumBmsState` and its helper functions defined at the top of this file.
Internally, a table may have a different number of system columns depending on whether it is backed by postgres storage or yugabyte storage. This manifests itself when adding an attribute number to bitmapsets which requires the use of a 'min_attr' offset.
In your case, most of these considerations aren't relevant but it would still be good to use the helper functions -- helps future maintainers.

In this function (or its caller), you can `Assert(IsYBRelation(relation)` and create a const min_attr variable setting it to `YBFirstLowInvalidAttributeNumber`.

Making the change suggested in this comment can sometimes be tricky due to "off-by-one" errors. Let me know if it creates a lot of trouble for you.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,181 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+	{
+		AttrNumber	attnum = idx + FirstLowInvalidHeapAttributeNumber;
+		Var		   *var;
+		TargetEntry *tle;
+
+		if (attnum > 0)
+		{
+			/* Regular column - verify it exists and is not dropped */
+			if (attnum > tupdesc->natts ||
+				TupleDescAttr(tupdesc, attnum - 1)->attisdropped)
```
karthik-ramanathan-3006: Do you happen to know what are cases where `attisdropped` may evaluate to true for a `CREATE INDEX` command?
I feel that if an index is requested to be create on a column that is invisible, that might defeat the purpose of the index. Worse, if the index definition contains no other columns, we end up with a phantom index. I don't know if `attisdropped` is checked upstream and returns an error.
Here, it would be good to either Assert that `attisdropped` is false or throw an error. Let me know your thoughts.

src/postgres/src/backend/access/heap/heapam_handler.c
```diff
@@ -1261,27 +1262,43 @@ heapam_index_build_range_scan(Relation heapRelation,
...
 
-		scan = table_beginscan_strat(heapRelation,	/* relation */
-									 snapshot,	/* snapshot */
-									 0, /* number of keys */
-									 NULL,	/* scan key */
-									 true,	/* buffer access strategy OK */
-									 allow_sync);	/* syncscan OK? */
+		if (IsYBRelation(heapRelation) && yb_enable_index_backfill_column_projection)
+		{
+			uint32 flags = SO_TYPE_SEQSCAN;
```
karthik-ramanathan-3006: Let's move `flags` into `ybc_heap_beginscan_for_index_build` and skip passing it as a param.

src/postgres/src/backend/access/heap/heapam_handler.c
```diff
@@ -1261,27 +1262,43 @@ heapam_index_build_range_scan(Relation heapRelation,
 		else
 			snapshot = SnapshotAny;
 
-		scan = table_beginscan_strat(heapRelation,	/* relation */
-									 snapshot,	/* snapshot */
-									 0, /* number of keys */
-									 NULL,	/* scan key */
-									 true,	/* buffer access strategy OK */
-									 allow_sync);	/* syncscan OK? */
+		if (IsYBRelation(heapRelation) && yb_enable_index_backfill_column_projection)
```
karthik-ramanathan-3006: When running the PR locally, I ran into a crash when backfilling a catalog table with the optimization turned on. It seems reasonable to skip the optimization for catalog tables (as this would largely occur during bootstrapping or during major upgrades on tables that are not very large).

We can use `is_system_catalog` (defined above) to identify catalog tables. 

Posting the crash logs for posterity:
```
2026-01-15 13:55:15.409 EST [34867] FATAL:  The catalog snapshot used for this transaction has been invalidated: expected: 1, got: 0: MISMATCHED_SCHEMA
```

src/postgres/src/backend/access/yb_access/yb_scan.c
austenLacy: Good point. I don't believe `attisdropped` would ever evaluate to true at this point as the parser would have already raised an error due to the missing column. `indexInfo->ii_IndexAttrNumbers` is populated by `DefineIndex → ComputeIndexAttrs`  which validates that each column exists in the relation. Invalid column references would have already failed there too. 

The simple assert is a good addition just in case though.

src/postgres/src/test/regress/sql/yb.orig.index_backfill_column_projection.sql
austenLacy: Thank you! Implemented in https://github.com/yugabyte/yugabyte-db/pull/29928/changes/241ec05088858d8e8a10954baebe4e0e8986a2e7

src/postgres/src/test/regress/sql/yb.orig.index_backfill_column_projection.sql
austenLacy: Implemented in https://github.com/yugabyte/yugabyte-db/pull/29928/changes/241ec05088858d8e8a10954baebe4e0e8986a2e7

src/postgres/src/backend/access/yb_access/yb_scan.c
austenLacy: I didn't notice those helper functions earlier. [This commit](https://github.com/yugabyte/yugabyte-db/pull/29928/changes/241ec05088858d8e8a10954baebe4e0e8986a2e7) uses them and cleans this code up quite a bit.

src/postgres/src/backend/access/yb_access/yb_scan.c
austenLacy: Good suggestion. Added here: https://github.com/yugabyte/yugabyte-db/pull/29928/changes/4bebabf2e5c2f3729e064673e01229d20978feba

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,129 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+//
+// We create a wide table with large columns (~5KB per row) and index just one
+// small INT column. The test parameter controls whether projection is enabled:
+// - With projection OFF: each row is ~5KB, fetch limit is 1KB, so ~1 row per RPC
+// - With projection ON: each row is ~few bytes, fetch limit is 1KB, so many rows per RPC
+//
+// Note: This test measures data plane RPCs (docdb_read_rpcs) not control plane
+// RPCs (BackfillIndex admin messages). The control plane RPCs stay the same,
+// but data plane RPCs decrease with projection because more rows fit per RPC.
+TEST_P(PgIndexBackfillColumnProjectionTest, ReducedRpcsWithProjection) {
```
karthik-ramanathan-3006: Thanks for adding this test! 

I was hoping to test a few different types of indices using the same workflow. So, let's modularize the test into helper functions and call this workflow with a few different index types:

1. Single column, simple index: `CREATE INDEX ON t (col1 HASH)`
2. Single column, simple index: `CREATE INDEX ON t (col1 ASC)`
   - to test that the storage strategy is irrelevant to this optimization
4. Single column, expression index: `CREATE INDEX ON t (f(col1))` 
    - where `f` is a function. This could be something simple like an identity function.
5. Single column, partial index: `CREATE INDEX ON t (col1) WHERE col1 > constant`    - to test that col1 is fetched only once. 
6. Single column, partial index: `CREATE INDEX ON t (col1) WHERE col2 > constant`
    - to test both col1, col2 are fetched.
7. Multi column, simple index: `CREATE INDEX ON t (col1 HASH, col2 ASC)`
8. Multi column, simple index: `CREATE INDEX ON t ((col1, col2) HASH, col3 ASC)`
9. Multi column, simple index: `CREATE INDEX ON t ((col1, col2) HASH, col2 ASC)`
    - to test that col2 is fetched only once.
10. Multi column, out of order index: `CREATE INDEX ON t (col2 HASH, col1 ASC, col3 ASC)`
    - where the columns in the table are defined in the order `[col1, col2, col3]`.
11. Multi column, out of order index: `CREATE INDEX ON t (col1, col2)`
    - where the columns in the table are defined in the order `[col1, col2]` and the primary key is defined as `[col2, col3, col1]`
12. Multi column, expression index: `CREATE INDEX ON t (f(col1, col2) HASH, g(col2) ASC)`
    - where `f`, `g` are functions. 
13. Multi column, partial index: `CREATE INDEX ON t (f(col1, col2) HASH, g(col2) ASC) WHERE col2 > constant`
14. Multi column, covering index: `CREATE INDEX ON t (f(col1, col2) HASH, g(col2) ASC) INCLUDE (col3, col4) WHERE col2 > constant`
15. Multi column index on a partitioned table 
16. Multi column index on a temp tableIt would be good for the columns in the multi column tests to be of different data types (string, int, bool, etc). Feel free to propose variations of the above.

karthik-ramanathan-3006: Unfortunately, indexes span a wide set of functionality that need to be validated. Thanks for being patient with this optimization. I may still be missing a couple of fundamental cases -- I'll continue thinking.

When implemented, I believe this test should cover a super set of the functionality tested in the regress test (`index_backfill_column_projection.sql`). So, we can probably get rid of that.

austenLacy: That makes perfect sense. I can definitely modularize and validate the logic for all the permutations we can.

austenLacy: Take a look in https://github.com/yugabyte/yugabyte-db/pull/29928/commits/6585f4d56edc148af92adfc4c54a2220f8ba503d. I'm working to get an environment locally that I can run these myself to verify.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,192 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+ *		Begin a heap scan specifically for index build/backfill operations.
+ *
+ * Unlike the regular heap scan which fetches all columns, this function
+ * only requests the columns needed for the index:
+ *   - Columns referenced in ii_IndexAttrNumbers (index key and non-key columns)
+ *   - Columns referenced in ii_Expressions (expression index columns)
+ *   - Columns referenced in ii_Predicate (partial index predicate)
+ *   - ybctid (always needed for index entry construction)
+ *
+ * This optimization significantly reduces the amount of data read from
```
karthik-ramanathan-3006, line 3964: nit: `concurrent index creation (backfill)` can be changed to `index backfill`
reason: this optimization is orthogonal to concurrency in index creation. It should work with non-concurrent index creation as well.

nit: `especially for tables with many columns where only a few are indexed.` can be changed to `especially for tables with many columns where a given index references only a few.` 
reason: a table can have a number of indexes which when combined can index a large subset of the columns. Here, we're concerned with only a specific index.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,192 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+						  0,		  /* distinct_prefixlen */
+						  NULL,		/* exec_params */
+						  false,	/* is_internal_scan */
+						  false);	/* fetch_ybctids_only */
+
+	/* Set additional TableScanDesc fields not handled by ybcBeginScan */
+	TableScanDesc tsdesc = (TableScanDesc) ybScan;
+
+	tsdesc->rs_snapshot = snapshot;
+	tsdesc->rs_flags = flags;
```
karthik-ramanathan-3006: nit: we can just assign `SO_TYPE_SEQSCAN` here, rather than declaring a variable.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,192 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+								   int nkeys,
+								   ScanKey key,
+								   IndexInfo *indexInfo)
+{
+	YbScanDesc	ybScan;
+	Scan	   *pg_scan_plan;
+
+	uint32 flags = SO_TYPE_SEQSCAN;
+
+	/*
```
karthik-ramanathan-3006: opinion: I personally feel that you can omit the comments in the body of this function as they're calling out fairly obvious things. I leave it up to you.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,288 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
 
 INSTANTIATE_TEST_CASE_P(, PgIndexBackfillIgnoreApplyTest, ::testing::Bool());
 
+// Tests index backfill column projection optimization across various index types.
+//
+// The optimization reduces data plane RPCs by fetching only columns needed for the index
```
karthik-ramanathan-3006: nit: Can omit `data plane`

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,288 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+    options->extra_tserver_flags.push_back(Format(
+        "--ysql_pg_conf_csv=yb_enable_pg_stat_statements_rpc_stats=true,"
+        "yb_fetch_size_limit=$0,"
+        "yb_enable_index_backfill_column_projection=$1",
+        kFetchSizeLimit, projection_enabled ? "true" : "false"));
+  }
+
+  void SetUp() override {
+    PgIndexBackfillTest::SetUp();
+    ASSERT_OK(conn_->Execute("CREATE EXTENSION IF NOT EXISTS pg_stat_statements"));
```
karthik-ramanathan-3006: This line can be skipped.
pg_stat_statements is a "core" extension that gets bundled with YugabyteDB and is enabled by default.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,288 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+        "       repeat('a', $1), repeat('b', $1), repeat('c', $1), "
+        "       repeat('d', $1), repeat('e', $1) "
+        "FROM generate_series(1, $2) AS i",
+        table_name, kLargeColSize, num_rows);
+  }
+
+  Result<int64_t> TestIndexBackfillProjection(
+      const std::string& create_index_sql,
+      const std::string& verify_query,
+      const std::string& test_name) {
```
karthik-ramanathan-3006: We can skip the `test_name` param. The test runner should make that obvious.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,288 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+  static constexpr int kLargeColSize = 1000;
+  static constexpr int kMaxProjectedRpcs = 100;
+  static constexpr int kMinUnprojectedRpcs = kNumRows / 2;
+};
+
+INSTANTIATE_TEST_CASE_P(, PgIndexBackfillColumnProjectionTest, ::testing::Bool());
+
+// Single column HASH index
+TEST_P(PgIndexBackfillColumnProjectionTest, SingleColumnHash) {
+  ASSERT_OK(CreateWideTable("t"));
```
karthik-ramanathan-3006: We can make the table name a static const in the test class.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,288 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+        "INSERT INTO $0 (col1, col2, col3, col4, "
+        "                large_col1, large_col2, large_col3, large_col4, large_col5) "
+        "SELECT i, 'text_' || i::text, (i % 2 = 0), i * 1.5, "
+        "       repeat('a', $1), repeat('b', $1), repeat('c', $1), "
+        "       repeat('d', $1), repeat('e', $1) "
+        "FROM generate_series(1, $2) AS i",
+        table_name, kLargeColSize, num_rows);
+  }
+
+  Result<int64_t> TestIndexBackfillProjection(
```
karthik-ramanathan-3006: convention: We can define this function as follows:
```
Status ValidateIndexBackfillProjection (...)
```
reason: We don't use the return value in any of the tests. We also generally avoid the prefix `Test` for helper functions.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,288 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+    LOG(INFO) << test_name << " [projection "
+              << (projection_enabled ? "ENABLED" : "DISABLED") << "]";
+
+    RETURN_NOT_OK(conn_->Execute("SELECT pg_stat_statements_reset()"));
+    RETURN_NOT_OK(conn_->Execute(create_index_sql));
+
+    auto rpcs = VERIFY_RESULT(GetBackfillReadRpcs());
+    LOG(INFO) << test_name << ": " << rpcs << " RPCs";
+
+    SCHECK_GT(rpcs, 0, RuntimeError, "Expected some RPCs during backfill");
```
karthik-ramanathan-3006: convention: Similarly, we also don't make assertions in helper functions. We generally return a non-OK() status to indicate that something is wrong and expect the caller to `ASSERT_OK(helperfunc())`

In this case, this can be something like:
```
if (rpc == 0) {
    return STATUS(RuntimeError, "Expected some RPCs during backfill");
}

File: src/yb/yql/pgwrapper/pg_index_backfill-test.cc [hunk 24]
```diff
@@ -2842,4 +2842,288 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+  Result<int64_t> GetBackfillReadRpcs() {
+    return conn_->FetchRow<int64_t>(
+        "SELECT COALESCE(sum(docdb_read_rpcs), 0) "
+        "FROM pg_stat_statements(true) "
+        "WHERE query LIKE 'BACKFILL%'");
+  }
+
+  Status CreateWideTable(const std::string& table_name, const std::string& pk_def = "id") {
+    return conn_->ExecuteFormat(
+        "CREATE TABLE $0 ("
```
karthik-ramanathan-3006, line 2886: Can we prune some of the columns in the table? For example, I don't see `col4` being used.
We could also merge all of the large columns into 1 or 2 columns for brevity.

austenLacy:
@karthik-ramanathan-3006 Thank you so much for the detailed review! I just committed the last set of nits and test convention improvements.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,233 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnExpressionPartial) {
+  ASSERT_OK(CreateWideTable(kTableName));
+  ASSERT_OK(InsertTestData(kTableName));
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetBackfillRpcs(
+      "CREATE INDEX idx ON t ((col1 + LENGTH(col2)) HASH, UPPER(col2) ASC) WHERE col2 > 'text_100'"));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Multi column index on partitioned table
+TEST_P(PgIndexBackfillColumnProjectionTest, PartitionedTable) {
```
karthik-ramanathan-3006, line 3054: This test fails because index backfilling across a partitioned table hierarchy uses the "non-concurrent" index backfill process. Non-concurrent indices don't issue `BACKFILL INDEX <oid>` queries, but still end up using the range scan optimization that you built.
For such non-concurrent indices, we can measure the improvement in `docdb_read_rpcs` for the `CREATE INDEX` query rather than the `BACKFILL INDEX` query.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,233 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+  ASSERT_OK(conn_->ExecuteFormat(
+      "CREATE TABLE $0_p2 PARTITION OF $0 FOR VALUES FROM (251) TO (501)", kTableName));
+
+  ASSERT_OK(InsertTestData(kTableName));
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetBackfillRpcs("CREATE INDEX idx ON t (col1, col2)"));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Multi column index on temp table
+TEST_P(PgIndexBackfillColumnProjectionTest, TempTable) {
```
karthik-ramanathan-3006: This test fails because temp tables are not backed by DocDB storage. So, `docdb_read_rpcs` always reads 0 for temp tables. We can skip the RPC validation for this test.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,233 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+        kFetchSizeLimit, projection_enabled ? "true" : "false"));
+  }
+
+  void SetUp() override {
+    PgIndexBackfillTest::SetUp();
+  }
+
+  Result<int64_t> GetBackfillReadRpcs() {
+    return conn_->FetchRow<int64_t>(
+        "SELECT COALESCE(sum(docdb_read_rpcs), 0) "
```
karthik-ramanathan-3006: `sum(docdb_read_rpcs)` returns a value of type 'numeric'. So, the return type must be cast to int64_t as follows: `sum(docdb_read_rpcs)::int8`.
Postgres' numeric type cannot be reinterpreted as int64.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,233 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+      }
+    }
+    return Status::OK();
+  }
+
+  static constexpr int kFetchSizeLimit = 1000;
+  static constexpr int kNumRows = 500;
+  static constexpr int kPaddingSize = 5000;
+  static constexpr int kMaxProjectedRpcs = 100;
+  static constexpr int kMinUnprojectedRpcs = kNumRows / 2;
```
karthik-ramanathan-3006: In some cases, I saw the number of RPCs corresponding to `yb_enable_index_backfill_column_projection=false` go as low as ~140, causing tests to fail. Maybe this value needs to be tweaked to `kNumRows / 4`? 
Feel free to take your call based on the column sizes. I didn't do the math behind calculating number of columns per RPC.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,233 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+  Status InsertTestData(const std::string& table_name, int num_rows = kNumRows) {
+    return conn_->ExecuteFormat(
+        "INSERT INTO $0 (col1, col2, col3, padding) "
+        "SELECT i, 'text_' || i::text, (i % 2 = 0), repeat('x', $1) "
+        "FROM generate_series(1, $2) AS i",
+        table_name, kPaddingSize, num_rows);
+  }
+
+  Result<int64_t> BuildIndexAndGetBackfillRpcs(const std::string& create_index_sql) {
+    RETURN_NOT_OK(conn_->Execute("SELECT pg_stat_statements_reset()"));
```
karthik-ramanathan-3006: `pg_stat_statements_reset()` can return rows. 
So `conn_->Execute()` needs to be changed to `conn_->Fetch()`

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,233 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnHashAsc) {
+  ASSERT_OK(CreateWideTable(kTableName));
+  ASSERT_OK(InsertTestData(kTableName));
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetBackfillRpcs(
+      "CREATE INDEX idx ON t (col1 HASH, col2 ASC)"));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Multi column index with compound hash key
+TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnCompoundHash) {
```
karthik-ramanathan-3006: Most of the MultiColumn* tests (except `MultiColumnHashAsc`) are getting tripped by our sanitizers (ASAN, TSAN) due to the following crash:
```
PgIndexBackfillColumnProjectionTest_MultiColumnCompoundHash__1.fatal_failure_details.ts-1.2026-01-31T05_32_28.pid229716.txt =====
F20260131 05:32:28 ../../src/yb/tserver/pg_client_session.cc:1252] Check failed: read_time.read >= ipair.first->second.read_time().read 
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/util/logging.cc:463:     @     0x7f863a9009b3  yb::LogFatalHandlerSink::send(int, char const*, char const*, int, tm const*, char const*, unsigned long)
    @     0x7f86392dd89f 
    @     0x7f86392de7ec 
    @     0x7f86392e27c9 
/opt/yb-build/thirdparty/yugabyte-db-thirdparty-v20251016191009-3fd3636b9a-almalinux8-x86_64-clang21/installed/tsan/libcxx/include/c++/v1/__ostream/basic_ostream.h:0:     @     0x7f8640ca5816  yb::tserver::(anonymous namespace)::ReadPointHistory::Save(yb::ConsistentReadPoint const&, unsigned long)
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/tserver/pg_client_session.cc:3225:     @     0x7f8640ca18d9  yb::tserver::PgClientSession::Impl::DoPerform(yb::tserver::PgTablesQueryResult const&, shared_ptr<yb::tserver::(anonymous namespace)::QueryData<yb::tserver::(anonymous namespace)::QueryTraits<yb::tserver::PgPerformRequestPB, yb::tserver::PgPerformResponsePB>>> const&, std::chrono::time_point<yb::CoarseMonoClock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>, yb::rpc::RpcContext*)
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/tserver/pg_client_session.cc:2627:     @     0x7f8640cc06b2  yb::tserver::PgClientSession::Impl::DoHandleSharedExchangeQuery(yb::LightweightFunction<yb::Status (unsigned long, std::chrono::time_point<yb::CoarseMonoClock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>)> const&, shared_ptr<yb::tserver::(anonymous namespace)::QueryData<yb::tserver::(anonymous namespace)::QueryTraits<yb::tserver::PgPerformRequestPB, yb::tserver::PgPerformResponsePB>>>&&, std::chrono::time_point<yb::CoarseMonoClock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>)
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/tserver/pg_client_session.cc:2648:     @     0x7f8640cc06b2  yb::Status yb::tserver::PgClientSession::Impl::DoHandleSharedExchangeQuery<yb::tserver::(anonymous namespace)::QueryTraits<yb::tserver::PgPerformRequestPB, yb::tserver::PgPerformResponsePB>, yb::tserver::PgTableCache&>(yb::LightweightFunction<yb::Status (unsigned long, std::chrono::time_point<yb::CoarseMonoClock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>)> const&, yb::tserver::(anonymous namespace)::SharedExchangeQuery<yb::tserver::(anonymous namespace)::QueryTraits<yb::tserver::PgPerformRequestPB, yb::tserver::PgPerformResponsePB>>&, unsigned char*, unsigned long, unsigned long, int, yb::tserver::PgTableCache&)
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/tserver/pg_client_session.cc:2657:     @     0x7f8640cc06b2  void yb::tserver::PgClientSession::Impl::HandleSharedExchangeQuery<yb::tserver::(anonymous namespace)::QueryTraits<yb::tserver::PgPerformRequestPB, yb::tserver::PgPerformResponsePB>, yb::tserver::PgTableCache&>(yb::LightweightFunction<yb::Status (unsigned long, std::chrono::time_point<yb::CoarseMonoClock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>)> const&, yb::tserver::SharedExchange&, unsigned char*, unsigned long, yb::tserver::PgTableCache&)
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/tserver/pg_client_session.cc:2674:     @     0x7f8640cc06b2  yb::tserver::PgClientSession::Impl::ProcessSharedRequest(unsigned long, yb::tserver::SharedExchange*, yb::LightweightFunction<yb::Status (unsigned long, std::chrono::time_point<yb::CoarseMonoClock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>)> const&)
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/tserver/pg_client_session.cc:4129:     @     0x7f8640c9a319  yb::tserver::PgClientSession::ProcessSharedRequest(unsigned long, yb::tserver::SharedExchange*, yb::LightweightFunction<yb::Status (unsigned long, std::chrono::time_point<yb::CoarseMonoClock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>)> const&)
/share/jenkins/workspace/yugabyte-db-phabricator/yugabyte-db/build/tsan-clang21-dynamic-ninja/../../src/yb/tserver/pg_client_service.cc:414:     @     0x7f8640c18988  yb::tserver::(anonymous namespace)::LockablePgClientSession::StartExchange(std::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, yb::YBThreadPool&)::'lambda'(unsigned long)::operator()(unsigned long) const
```

I haven't been able to resolve this one yet. I will get back to you on the fix for this.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,233 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+      "CREATE INDEX idx ON t ((col1 + LENGTH(col2)) HASH, UPPER(col2) ASC)"));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Multi column expression index with partial predicate using col2
+TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnExpressionPartial) {
+  ASSERT_OK(CreateWideTable(kTableName));
+  ASSERT_OK(InsertTestData(kTableName));
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetBackfillRpcs(
+      "CREATE INDEX idx ON t ((col1 + LENGTH(col2)) HASH, UPPER(col2) ASC) WHERE col2 > 'text_100'"));
```
karthik-ramanathan-3006: This line exceeds our 100 column policy by a couple of characters. Could you format it?

austenLacy:
@karthik-ramanathan-3006 https://github.com/yugabyte/yugabyte-db/pull/29928/commits/17960cba3701e8324e905270cc3e37f7567e24c7 includes the latest round of comments. I was also able to finally get the cxx tests to run locally and it looks like they were successful.

```
./yb_build.sh release \
  --cxx-test pgwrapper_pg_index_backfill-test 
  --gtest_filter="*ColumnProjection*" 
  --skip-java 
  --no-odyssey 
  --skip-pg-parquet

# ...

1/1 Test #532: pgwrapper_pg_index_backfill-test ...   Passed  471.31 sec

The following tests passed:
        pgwrapper_pg_index_backfill-test

100% tests passed, 0 tests failed out of 1

Total Test time (real) = 471.48 sec
```

karthik-ramanathan-3006:
Thanks @austenLacy! 
Could you try the same test with the build type as `asan` or `tsan`? Something like:
```
./yb_build.sh asan --cxx-test pgwrapper_pg_index_backfill-test --gtest_filter 'PgIndexBackfillColumnProjectionTest.MultiColumnCompoundHash/1' --skip-java --no-odyssey --skip-pg-parquet
```
(if you're running on a Mac, this may not work)
The issue seems to be specific to our sanitizer builds. Not sure what the connection is yet.

austenLacy:
@karthik-ramanathan-3006 

> Could you try the same test with the build type as asan or tsan?

I originally ran the tests on OSX and asan/tsan failed as you expected. I still can't seem to get the tests running on alma linux 8 due to initdb failures.

src/postgres/src/backend/access/heap/heapam_handler.c
```diff
@@ -1261,27 +1262,40 @@ heapam_index_build_range_scan(Relation heapRelation,
...
+										 allow_sync);	/* syncscan OK? */
+		}
+
 		if (IsYBRelation(heapRelation))
 		{
 			YbcPgExecParameters *exec_params = &estate->yb_exec_params;
 
 			if (bfinfo)
 			{
 				if (bfinfo->bfinstr)
```
karthik-ramanathan-3006, line 1289: Double checked this. The previous structure had poor readability but was intentional:


```suggestion
				if (bfinfo->bfinstr)
					exec_params->bfinstr = pstrdup(bfinfo->bfinstr);

				exec_params->backfill_read_time = bfinfo->read_time;
				exec_params->partition_key =
					pstrdup(bfinfo->row_bounds->partition_key);
				exec_params->out_param = bfresult;
				exec_params->is_index_backfill = true;
```

Let's add a newline after `exec_params->bfinstr = pstrdup(bfinfo->bfinstr);` for better readability.

src/postgres/src/backend/access/heap/heapam_handler.c
```diff
@@ -1261,27 +1262,40 @@ heapam_index_build_range_scan(Relation heapRelation,
...
-				exec_params->out_param = bfresult;
-				exec_params->is_index_backfill = true;
+					exec_params->backfill_read_time = bfinfo->read_time;
+					exec_params->partition_key =
+						pstrdup(bfinfo->row_bounds->partition_key);
+					exec_params->out_param = bfresult;
+					exec_params->is_index_backfill = true;
+				}
 			}
-
```
karthik-ramanathan-3006, line 1284: nit: Let's revert removing this line.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,244 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+
+  ASSERT_OK(InsertTestData(kTableName));
+  // Partitioned tables use non-concurrent index builds which don't issue BACKFILL queries.
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
+      "CREATE INDEX idx ON t (col1, col2)", /* use_backfill_rpcs */ false));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Multi column index on temp table
+TEST_P(PgIndexBackfillColumnProjectionTest, TempTable) {
```
karthik-ramanathan-3006: On second thoughts I think we can skip the TempTable test altogether because the entry point to our optimization explicitly excludes temp tables.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,179 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+	ybScan = ybcBeginScan(relation,
+						  NULL,		/* index */
+						  false,	/* xs_want_itup */
+						  nkeys,
+						  key,
+						  pg_scan_plan,
+						  NULL,		/* rel_pushdown */
+						  NULL,		/* idx_pushdown */
+						  NIL,		/* aggrefs */
+						  0,		  /* distinct_prefixlen */
```
karthik-ramanathan-3006: style: The spacing is slightly off in this line.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,179 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+	 * min_attr offset correctly for YugaByte relations which may have
+	 * different system columns than standard Postgres relations.
+	 * Initialize at declaration since min_attr is const.
+	 */
+	YbAttnumBmsState required_attrs = ybcAttnumBmsConstruct();
+
+	/*
+	 * Always need ybctid for index entry construction. Add it directly to
+	 * targetlist since it's the only system column allowed in indexes.
+	 * Other system columns (like ctid) are managed by the database and can
```
karthik-ramanathan-3006: nit: We can skip this sentence.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,179 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+	List	   *targetlist = NIL;
+	Var		   *ybctid_var;
+	TargetEntry *ybctid_tle;
+	int			i;
+	int			idx;
+	int			resno = 1;
+	/* Use varno=1 since this is always scanning the base relation */
+	const Index varno = 1;
+
+	/* This function is only for YugaByte relations */
```
karthik-ramanathan-3006: Please change `YugaByte` to either `YB` or `YugabyteDB`. Here and below. 
(We went through a rebranding exercise a few years back)

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,179 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+{
+	Scan	   *scan_plan;
+	TupleDesc	tupdesc = RelationGetDescr(relation);
+	List	   *targetlist = NIL;
+	Var		   *ybctid_var;
+	TargetEntry *ybctid_tle;
+	int			i;
+	int			idx;
+	int			resno = 1;
+	/* Use varno=1 since this is always scanning the base relation */
```
karthik-ramanathan-3006: It would be good to add the reason behind why `varno=1` always scans the base relation.
Concurrent index creation in postgres is restricted to one index per table/statement (unlike the non-concurrent index creation process). As a result, we're guaranteed that only one table is involved in the process, and said table is opened for inspection first, leading to it being varno=1.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,229 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnHashAsc) {
+  ASSERT_OK(CreateWideTable(kTableName));
+  ASSERT_OK(InsertTestData(kTableName));
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
+      "CREATE INDEX idx ON t (col1 HASH, col2 ASC)"));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Multi column index with compound hash key
+TEST_P(PgIndexBackfillColumnProjectionTest, YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnCompoundHash)) {
```
karthik-ramanathan-3006:
```suggestion
TEST_P(PgIndexBackfillColumnProjectionTest,
       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnCompoundHash)) {
```

style: The macro is tripping up the line limit lint rule.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,229 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+  ASSERT_OK(CreateWideTable(kTableName));
+  ASSERT_OK(InsertTestData(kTableName));
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
+      "CREATE INDEX idx ON t ((col1, col2) HASH, col3 ASC)"));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Multi column index where col2 appears in both compound hash and range - fetched only once
+TEST_P(PgIndexBackfillColumnProjectionTest, MultiColumnDuplicateColumn) {
```
karthik-ramanathan-3006: Could you add the `YB_DISABLE_TEST_IN_SANITIZERS` macro to all the MultiColumn* tests?
All of them were failing on sanitizer builds.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,182 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+{
+	Scan	   *scan_plan;
+	TupleDesc	tupdesc = RelationGetDescr(relation);
+	List	   *targetlist = NIL;
+	Var		   *ybctid_var;
+	TargetEntry *ybctid_tle;
+	int			i;
+	int			idx;
+	int			resno = 1;
+	/* Use varno=1 since this is always scanning the base relation.
```
karthik-ramanathan-3006:
```suggestion
	/*
	 * Use varno=1 since this is always scanning the base relation.
```

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2842,4 +2842,236 @@ TEST_P(PgIndexBackfillIgnoreApplyTest, Backward) {
...
+       YB_DISABLE_TEST_IN_SANITIZERS(MultiColumnDuplicateColumn)) {
+  ASSERT_OK(CreateWideTable(kTableName));
+  ASSERT_OK(InsertTestData(kTableName));
+  auto rpcs = ASSERT_RESULT(BuildIndexAndGetRpcs(
+      "CREATE INDEX idx ON t ((col1, col2) HASH, col2 ASC)"));
+  ASSERT_OK(ValidateRpcs(rpcs));
+}
+
+// Index columns in different order than table definition (table: col1, col2, col3)
+TEST_P(PgIndexBackfillColumnProjectionTest
```
karthik-ramanathan-3006:
```suggestion
TEST_P(PgIndexBackfillColumnProjectionTest,
```

This is missing a comma.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -3814,6 +3821,183 @@ ybc_heap_endscan(TableScanDesc tsdesc)
...
+	int			i;
+	int			idx;
+	int			resno = 1;
+	/*
+	 * Use varno=1 since this is always scanning the base relation.
+	 * Concurrent index creation in postgres is restricted to one index
+	 * per table/statement (unlike the non-concurrent index creation process).
+	 * As a result, we're guaranteed that only one table is involved in the process,
+	 * and said table is opened for inspection first, leading to it being varno=1.
+	*/
```
karthik-ramanathan-3006:
```suggestion
	 */
```

Sorry, I think I missed recommending an extra space to align the `*`s last time around.

karthik-ramanathan-3006:
```suggestion
	 */
```

Sorry, I think I missed recommending an extra space to align the `*`s last time around.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
karthik-ramanathan-3006:
```suggestion
TEST_P(PgIndexBackfillColumnProjectionTest,
```

This is missing a comma.

Issue #13182, opened 2022-07-07
YSQL: the ysql_dump utility does not honour PGHOST environment variable


fritshoogland-yugabyte:
Jira Link: [DB-2866](https://yugabyte.atlassian.net/browse/DB-2866)
The postgres pg_dump utility is a client utility that interacts with the CLI just like other postgres CLI tools such as psql.
One of the features that postgres CLI utilities have is to use environment variables for setting functionality, such as the host to connect to.
The YugabyteDB ysqlsh utility (the Yugabyte version of the psql utility) does honour this, and uses the PGHOST variable to set the host to connect to.

However, the ysql_dump utility doesn't seem to honour the PGHOST variable:
```
[yugabyte@ip-172-151-62-25 ~]$ env | grep PGHOST
PGHOST=ip-172-151-62-25.us-west-2.compute.internal
[yugabyte@ip-172-151-62-25 ~]$ ./tserver/postgres/bin/ysql_dump -d pixel_12
ysql_dump: [archiver (db)] connection to database "pixel_12" failed: could not connect to server: Connection refused
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5433?
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5433?
```
I set the PGHOST variable to the hostname, but when executing the ysql_dump utility, it shows it's trying to connect to localhost.

PR #29976, opened 2026-01-15
Respect `PGHOST` environment variable in `ysql_dump`

austenLacy:
Fixes https://github.com/yugabyte/yugabyte-db/issues/13182

Previously `ysql_dump` would ignore the PGHOST environment variable so this fixes that. The specificity is now:

- `-h` flag (highest priority)
- `PGHOST` environment variable
- localhost default (lowest priority)

_prefers `-h` if present regardless of PGHOST_

```bash
➜ PGHOST=invalid-test-host.example.com ./build/release-clang-dynamic-arm64/postgres/bin/ysql_dump -d postgres -h another-host.foo.com
ysql_dump: error: connection to server at "another-host.foo.com" (50.16.218.27), port 5433 failed: Network is unreachable
        Is the server running on that host and accepting TCP/IP connections?
```

_respects PGHOST_

```bash
➜ PGHOST=invalid-test-host.example.com ./build/release-clang-dynamic-arm64/postgres/bin/ysql_dump -d postgres
ysql_dump: error: could not translate host name "invalid-test-host.example.com" to address: nodename nor servname provided, or not known
```

_defaults to localhost if not provided_

```bash
➜ ./build/release-clang-dynamic-arm64/postgres/bin/ysql_dump -d postgres               
ysql_dump: error: connection to server at "localhost" (::1), port 5433 failed: Connection refused
        Is the server running on that host and accepting TCP/IP connections?
connection to server at "localhost" (127.0.0.1), port 5433 failed: Connection refused
        Is the server running on that host and accepting TCP/IP connections?
```

### DIFF
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlDump.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlDump.java
@@ -15,8 +15,10 @@ package org.yb.pgsql;
 
 import static org.yb.AssertionWrappers.*;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -26,6 +28,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


src/postgres/src/bin/pg_dump/t/001_basic.pl
```diff
@@ -222,4 +222,13 @@
 	'pg_dumpall: option --exclude-database cannot be used together with -g/--globals-only'
 );
 
+# YB: Test that PGHOST environment variable is respected
+{
```
austenLacy: Couldn't figure out how to run this locally so will need help with that.

src/postgres/src/bin/pg_dump/pg_dump.c
```diff
@@ -920,9 +920,15 @@ main(int argc, char **argv)
 
 	fout->numWorkers = numWorkers;
 
-	/* YB */
+	/* YB: Use PGHOST environment variable if set, otherwise default to localhost */
```
iSignal: I think if you just removed the two YB lines here, fe-connect.c would use PGHOST as it does normally. You shouldn't need to set it explicitly. 

These lines were added in https://github.com/yugabyte/yugabyte-db/commit/ccd9a1106ade392e18cd46bd5d02c3889028e0d9#diff-adcfee270d91d3d60ed8e8985ad3fe2748383584be29992cbd7b5546703fb2c0 by @OlegLoginov, bu t I'm not sure why. 

I think the default should become `DefaultHost` even if we remove those lines, so that's what I would suggest unless @OlegLoginov has any objections.

src/postgres/src/bin/pg_dump/t/001_basic.pl
iSignal: We don't run these tests, so what I would suggest is picking a test which calls ysql_dump like pg_read_time-test.cc which runs `ysql_dump --host` and instead setting the same host arg via PGHOST env var. TestYsqlDump.java is another option.

iSignal (changes requested):
Thanks for the fix @austenLacy! Let me know what you think of the comments. I can run this through unit tests internally once are closer to the final version.

src/postgres/src/bin/pg_dump/t/001_basic.pl
OlegLoginov: Agree. We don't use the script `001_basic.pl`, so I prefer to keep it unchanged to prevent future merging.

src/postgres/src/bin/pg_dump/pg_dump.c
```diff
@@ -920,9 +920,15 @@ main(int argc, char **argv)
...
 	if (dopt.cparams.pghost == NULL || dopt.cparams.pghost[0] == '\0')
-		dopt.cparams.pghost = DefaultHost;
+	{
+		const char *env_host = getenv("PGHOST");
+		if (env_host != NULL && env_host[0] != '\0')
+			dopt.cparams.pghost = pg_strdup(env_host);
+		else
+			dopt.cparams.pghost = DefaultHost;
+	}
```
OlegLoginov: The changes look good for me.
I would only recheck how PGHOST is handled in source PG code.

src/postgres/src/bin/pg_dump/pg_dump.c
OlegLoginov: @iSignal The lines:
```
	if (dopt.pghost == NULL || dopt.pghost[0] == '\0')
		dopt.pghost = DefaultHost;
```
were added because I needed default IP to connect to the `yb-master` below:
```
	if (dopt.include_yb_metadata)
	{
		if (dopt.master_hosts)
			YBCSetMasterAddresses(dopt.master_hosts);
		else
			YBCSetMasterAddresses(dopt.pghost);              <<<<<<<<<<
```
To get `DefaultHost` available - I moved this definition from *.cc to another available *.h file.

And - yes - I did not care about $PGHOST env-var here.

iSignal: @OlegLoginov : Thanks, that explains it. But those lines don't exist any more and we don't need the pg_dump code to be aware of the master anymore. So can we just remove these lines and fallback to normal libfe-connect logic which determines the host for connection?

OlegLoginov: Agree. Let's remove the lines:
```
if (dopt.pghost == NULL || dopt.pghost[0] == '\0')
		dopt.pghost = DefaultHost;
```
(`YBCSetMasterAddresses` call was already removed.)

And `DefaultHost` also can be returned back to the original code point. Where it is in PG-15 sources - in `src/interfaces/libpq/fe-connect.c`
```
#define DefaultHost "localhost"
```

austenLacy: Thanks to both of you. Updated in https://github.com/yugabyte/yugabyte-db/pull/29976/changes/5d4ad5df7b790660d1a791c8ed65adcb9e056ead#diff-f6766b0615f61cc9ed9c0b73094b8646d7f00b09ac1bbd3284086ec21c1bedf5

java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlDump.java
```diff
@@ -472,6 +475,97 @@ public void ysqlDumpColocatedTablesWithTablespaces() throws Exception {
...
+    // Capture output (stdout + stderr merged)
+    BufferedReader reader = new BufferedReader(
+        new InputStreamReader(process.getInputStream()));
+    StringBuilder output = new StringBuilder();
+    String line;
+    while ((line = reader.readLine()) != null) {
+      output.append(line).append("\n");
+    }
+
+    process.waitFor(30, TimeUnit.SECONDS);
```
iSignal: minor: maybe wait for 60s in this case too

PR #30058, opened 2026-01-21
Avoid "unqualified" tables for streams with NOEXPORT_SNAPSHOT

bdbene:
## Context
Currently, YugabyteDB is opinionated about how to start a CDC stream (gRPC, not postgres) because it's handling WAL retention and other barriers. 

1. If a table that was created before the stream creation is not polled within cdcsdk_tablet_not_of_interest_timeout_secs (default 4 hrs), we declare it as not of interest and the table is marked unqualified for this stream.
2. If a table that was created before stream creation is polled for some time and then left unpolled by the stream for more than cdc_intent_retention_ms (default 8hrs) , then it is marked as unqualified for the stream.
3. A dynamic table is one which is created after stream creation. A dynamic table will be declared as unqualified only if it has been left unpolled by the stream for more than cdc_intent_retention_ms. 

## Problems
- This solution makes onboarding tables to an existing stream impossible after the 4 hour mark
- Dynamic tables (tables created after the stream was created) don't actually benefit at all from any of these optimizations, they'll automatically get a longer retention time instead of the default 15m WAL

## Proposal
This is a rough implementation of what a solution could look like, because I want to get the conversation going and get feedback. 

Typically, when someone decides to CDC a table, they don't particularly care about the exact history of changes within a 4 hour window. They will either take a full snapshot, or they will just stream from the point in time where they add their table to Debezium's list of tables. 

Therefore, I see a couple solutions ranging in complexity
1. For `NOEXPORT_SNAPSHOT` streams, just never use the `unqualified` table feature. This is super simple but then of course 100% percent of tables will have the longer wall, regardless of whether Debezium is actually polling for changes. This may be true anyway if most tables are dynamic tables. 
2. Same as above, but by default do not increase the WAL retention until `GetChanges()` is polled on a table. Then increase the retention once we know there is someone interested in the table. This PR explores that this solution could look like.
3. A more extreme solution would build on the previous ones, but completely remove the concept of `unqualified` or `dynamic` tables. Just increase WAL as needed instead of proactively trying to do it when the stream is created or as new tables are created.

## In this PR
I've organized the commits to make it easier to explore

1. Fixed some compiler errors that were on the master branch, not really relevant to this PR but I wanted it to build
2. Avoid marking tables in a `NOEXPORT_SNAPSHOT` stream's namespace as 'not of interest', avoiding the whole unqualified table problem
3. Removing the increase in WAL retention, which is the default behaviour when a stream is created or a "dynamic table" is created
4. Increasing WAL retention, and barriers, as consumers of the stream are detected when they call `GetChanges()`. This is currently implemented with the existing cdc_service background thread detecting the poll and signalling the increase in WAL.

### DIFF
--- a/src/yb/cdc/cdc_service.cc
+++ b/src/yb/cdc/cdc_service.cc
@@ -2608,7 +2608,8 @@ void CDCServiceImpl::ProcessEntryForCdcsdk(
     TabletIdCDCCheckpointMap& tablet_min_checkpoint_map,
     StreamIdSet* slot_entries_to_be_deleted,
     const std::unordered_map<NamespaceId, uint64_t>& namespace_to_min_record_id_commit_time,
-    TableIdToStreamIdMap* expired_tables_map) {
+    TableIdToStreamIdMap* expired_tables_map,
+    TableStreamPairSet* tables_needing_on_demand_retention) {
   const auto& stream_id = entry.key.stream_id;
   const auto& tablet_id = entry.key.tablet_id;
   const auto& checkpoint = *entry.checkpoint;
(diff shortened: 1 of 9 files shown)


bdbene:
Closing this as it was just a sample solution. :)

PR #30150, opened 2026-01-29
[Docs] Fix link symbol for the H3 on configuration pages

samiahmedsiddiqui:
The link symbol for the h3 doesn't show up when it should. This is the `yb-tserver` page
https://docs.yugabyte.com/stable/reference/configuration/yb-tserver/#allowed-preview-flags-csv


ddhodge (approved):
LGTM

PR #30155, opened 2026-01-29
[#29974] odyssey/postgres: remove IPV6_ONLY socket option

### LINKED ISSUE #29974: Support binding IPV4 and IPV6 for YSQL API with Postgres and/or Connection Manager
Jira Link: [DB-19838](https://yugabyte.atlassian.net/browse/DB-19838)

[DB-19838]: https://yugabyte.atlassian.net/browse/DB-19838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jameshartig:
odyssey and postgres both forced IPV6 sockets to IPv6-only with the IPV6_ONLY socket option. This meant that listening on `[::]` did not additionally listen on IPv4 addresses like CQL does.

I didn't want to comment out the code as that could cause merge conflicts later as things are brought in from upstream so I used #if 0 like I saw in other parts of the postgres codebase. I thought about plumbing down a configuration flag into that code but it would've been a lot larger of a change and fairly difficult.

I tested this locally by running:

```
bin/yugabyted start --tserver_flags 'pgsql_proxy_bind_address=[::]:5433,cql_proxy_bind_address=[::]:9042'
```

and then connecting with both `127.0.0.1` and `::1`.

I similarly tested odyssey by running:

```
bin/yugabyted start --tserver_flags 'pgsql_proxy_bind_address=[::]:5433,cql_proxy_bind_address=[::]:9042,enable_ysql_conn_mgr=true'
```

I would like some guidance on the unit test for Postgres. I tried to verify it locally but I keep getting an error saying there isn't enough tablet servers to make the various initdb tables during setup before it even gets to my code. Is there something I'm missing about the test harness that I need to do?

Fixes #29974

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -83,7 +83,7 @@ This is a list of people who have contributed code to the [YugabyteDB](https://g
 * [aegershman](https://github.com/aegershman)
 * [hasheddan](https://github.com/hasheddan)
 * [ameyb](https://github.com/ameyb)
-* [fastest963](https://github.com/fastest963)
+* [jameshartig](https://github.com/jameshartig)
 * [eliahburns](https://github.com/eliahburns)
 * [bhavin192](https://github.com/bhavin192)
 * [graffido](https://github.com/graffido)
--- a/src/odyssey/third_party/machinarium/sources/bind.c
+++ b/src/odyssey/third_party/machinarium/sources/bind.c
@@ -30,6 +30,10 @@ MACHINE_API int machine_bind(machine_io_t *obj, struct sockaddr *sa, int flags)
 		mm_errno_set(errno);
 		goto error;
 	}
+	/*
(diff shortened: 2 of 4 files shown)


CONTRIBUTORS.md
```diff
@@ -83,7 +83,7 @@ This is a list of people who have contributed code to the [YugabyteDB](https://g
 * [aegershman](https://github.com/aegershman)
 * [hasheddan](https://github.com/hasheddan)
 * [ameyb](https://github.com/ameyb)
-* [fastest963](https://github.com/fastest963)
+* [jameshartig](https://github.com/jameshartig)
```
jameshartig, line 86: To clarify, I renamed my GitHub username and that's the reason for the change here.

PR #30286, opened 2026-02-11
Fix an overflow in is_staleness_acceptable

### LINKED ISSUE #15011: [Bug] Allow  yb_follower_read_staleness_ms higher than 35 minutes (2147484 seconds)
### Description

Setting `yb_follower_read_staleness_ms` to more than 2147483  raises an error:
```
yugabyte=> set yb_read_from_followers=on;
SET

yugabyte=> set yb_follower_read_staleness_ms=2147483 ;
SET

yugabyte=> set yb_follower_read_staleness_ms=2147484 ;
ERROR:  cannot enable yb_read_from_followers with a staleness of less than 2 * (max_clock_skew = 500000 usec)
Time: 1018.724 ms (00:01.019)
```
The error message is not correct. The reason for the error is that the function [is_staleness_acceptable](https://github.com/yugabyte/yugabyte-db/blob/master/src/postgres/src/backend/commands/variable.c#L647) uses an `int32_t`
Note that this also means that we can set higher value (but will return no rows without error).

The Follower Read feature is for performance, and should not require such high staleness. But it can also be used to do some kind of Flashback Query as of a previous point in time. Example: https://dev.to/franckpachot/flashback-query-in-yugabytedb-k7o

ali-a-a:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/30286/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

Resolves https://github.com/yugabyte/yugabyte-db/issues/15011.

Right now, the maximum value for `yb_follower_read_staleness_ms` is set to `INT_MAX` ([source](https://github.com/yugabyte/yugabyte-db/blob/ed0ea8a65669af2df36fba7b8925ef1f20b7e543/src/postgres/src/backend/utils/misc/guc.c#L5527)), and the current code overflows when it gets multiplied by 1000 in `is_staleness_acceptable`. This PR fixes this problem by casting it into `int64_t` while doing the comparison.

### DIFF
--- a/src/postgres/src/backend/commands/variable.c
+++ b/src/postgres/src/backend/commands/variable.c
@@ -625,7 +625,7 @@ is_staleness_acceptable(int32_t staleness_ms)
 	int32_t		max_clock_skew_usec = YBGetMaxClockSkewUsec();
 	const int	kMargin = 2;
 
-	if (staleness_ms * 1000 < kMargin * max_clock_skew_usec)
+	if ((int64_t) staleness_ms * 1000 < (int64_t) kMargin * max_clock_skew_usec)
 	{
 		GUC_check_errcode(ERRCODE_FEATURE_NOT_SUPPORTED);
 		GUC_check_errmsg("cannot enable yb_read_from_followers with a staleness of less than "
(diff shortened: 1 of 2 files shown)


bmcnama-shopify:
trigger jenkins

Issue #29808, opened 2025-12-26
[YSQL] Consider automatically single-quoting ysql_ gflags which have spaces in the values


iSignal:
Jira Link: [DB-19639](https://yugabyte.atlassian.net/browse/DB-19639)
### Description

Setting ysql_default_transaction_isolation to `read committed` causes YSQL to crash and fail to startup, because PG conf expects this to look like `'read committed'`. While generating PG flags, tserver could consider always single quoting string values in the generated PG conf file in pg_wrapper.cc unless they are already bounded by single quotes.

@sushantrmishra 

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-19639]: https://yugabyte.atlassian.net/browse/DB-19639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #30324, opened 2026-02-14
Auto single-quote `ysql_` prefixed gflags in PG config

keisku:
Closes https://github.com/yugabyte/yugabyte-db/issues/29808

---

This PR resolves this issue. 

```bash
./bin/yb-ctl --rf 3 create --tserver_flags "ysql_default_transaction_isolation='read committed'"

tail -n 5 ~/yugabyte-data/node-1/disk-1/tserver.err
2026-02-14 01:54:23.090 GMT [34439] WARNING:  the parameter "yb_enable_advisory_locks" is deprecated, toggle the runtime flag "ysql_yb_enable_advisory_locks" instead.
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0214 10:54:23.090657 71117888 tcmalloc_util.cc:266] Setting TCMalloc profiler sampling period to 1048576 bytes
2026-02-14 01:54:23.091 GMT [34439] LOG:  syntax error in file "/Users/keisukeumegaki/yugabyte-data/node-1/disk-1/pg_data/ysql_pg.conf" line 834, near token "committed"
2026-02-14 01:54:23.091 GMT [34439] FATAL:  configuration file "/Users/keisukeumegaki/yugabyte-data/node-1/disk-1/pg_data/ysql_pg.conf" contains errors

./bin/ysqlsh -c "SHOW default_transaction_isolation;"
ysqlsh: error: connection to server at "localhost" (127.0.0.1), port 5433 failed: Connection refused
        Is the server running on that host and accepting TCP/IP connections?
connection to server at "localhost" (::1), port 5433 failed: Connection refused
        Is the server running on that host and accepting TCP/IP connections?
```

With this PR, we can see `read committed` as this command result.

```bash
./bin/yb-ctl --rf 3 create --tserver_flags "ysql_default_transaction_isolation='read committed'"

./bin/ysqlsh -c "SHOW default_transaction_isolation;"
 default_transaction_isolation
-------------------------------
 read committed
(1 row)
```

---

> [!NOTE]
> **Low Risk**
> Small, localized change to config-file generation plus added tests; main risk is unintended quoting changes for string flags, but behavior is constrained to `type == "string"` and preserves already-quoted values.
> 
> **Overview**
> Fixes Postgres startup failures caused by unquoted string-valued `ysql_` gflags (e.g. `read committed`, `ISO, YMD`) being written into `ysql_pg.conf` without quoting.
> 
> Adds `FormatPgGFlagValue` and uses it when emitting Pg-tagged gflags into the generated config: string types are single-quoted with internal quotes escaped, while non-string values pass through unchanged. Updates tests to cover quoting/escaping behavior and extends the flag override integration test to include `ysql_default_transaction_isolation` and `ysql_datestyle` values that require quoting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a578f21c9ff5cf6620ec8058a3ebf91ee247fa0c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/yb/yql/pgwrapper/pg_wrapper.cc
+++ b/src/yb/yql/pgwrapper/pg_wrapper.cc
@@ -445,6 +445,20 @@ using namespace std::literals;  // NOLINT
 namespace yb {
 namespace pgwrapper {
 
+string FormatPgGFlagValue(const string& value, const string& type) {
+  if (type == "string") {
+    auto trimmed = boost::algorithm::trim_copy(value);
+    if (trimmed.size() >= 2 && trimmed.front() == '\'' && trimmed.back() == '\'') {
+      // Already single-quoted — pass through unchanged.
+      return value;
+    }
+    // Add single quotes and escape internal single quotes.
+    string escaped_value = boost::replace_all_copy(value, "'", "''");
+    return Format("'$0'", escaped_value);
+  }
+  return value;
+}
+
 namespace {
 
 Status WriteConfigFile(const string& path, const vector<string>& lines) {
@@ -594,8 +608,8 @@ void AppendPgGFlags(vector<string>* lines) {
     }
 
     string pg_variable_name = flag.name.substr(pg_flag_prefix.length());
-    string escaped_value = boost::replace_all_copy(flag.current_value, "'", "''");
-    lines->push_back(Format("$0='$1'", pg_variable_name, escaped_value));
+    lines->push_back(
+        Format("$0=$1", pg_variable_name, FormatPgGFlagValue(flag.current_value, flag.type)));
   }
 
   // Special handling for deprecated ysql_enable_pg_export_snapshot flag.
--- a/src/yb/yql/pgwrapper/pg_wrapper-test.cc
+++ b/src/yb/yql/pgwrapper/pg_wrapper-test.cc
@@ -34,6 +34,7 @@
 #include "yb/client/client.h"
 #include "yb/client/table_info.h"
 
+#include "yb/yql/pgwrapper/pg_wrapper.h"
 #include "yb/yql/pgwrapper/pg_wrapper_test_base.h"
 #include "yb/yql/pgwrapper/libpq_utils.h"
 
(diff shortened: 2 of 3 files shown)


src/yb/yql/pgwrapper/pg_wrapper.cc
```diff
@@ -594,7 +594,8 @@ void AppendPgGFlags(vector<string>* lines) {
     }
 
     string pg_variable_name = flag.name.substr(pg_flag_prefix.length());
-    lines->push_back(Format("$0=$1", pg_variable_name, flag.current_value));
+    string escaped_value = boost::replace_all_copy(flag.current_value, "'", "''");
```
iSignal: As the cursor bot said, we don't want to disturb existing users who are already single quoting the value like `'read committed'`. So could we do something like

1. if it is a string type pg flag, and
2. it does not have single quotes at start and end (accounting for whitespace at start/end)
then,
we will add single quotes to the beginning and end and also escape any single quotes in value to two single quotes as you did already.

Does this seem reasonable?

Cases to consider for flag values (would be good to include this in the commit summary)
```
'read committed' -> unchanged
read committed -> fails before and works now as 'read committed'
foo'bar -> 'foo''bar'
foo'''bar -> this probably fails parsing today, would be good to confirm. so it makes sense for us to set this as 'foo''''bar' without breaking any existing users

```

Right now, the flags that we would be changing behavior would most likely be

```
--ysql_timezone -> does this fail today if we set to America/New_York or GMT+5 without single quotes?
--ysql_datestyle -> does this fail if we set to ISO, MDY without single quotes?
ysql_default_transaction_isolation -> read committed, we know this fails
```

Please include this list as well in the summary or any other common pg flags you think will be affected.

iSignal (changes requested):

src/yb/yql/pgwrapper/pg_wrapper.cc
keisku: @iSignal Thank you for reviewing this PR.

Updated to only auto-quote string-type flags, skip quoting if already single-quoted, and escape internal quotes by doubling. Checked string-type PG flags defined via `DEFINE_RUNTIME_PG_FLAG(string, ...)` in [yb/yql/pgwrapper/pg_wrapper.cc](https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/yql/pgwrapper/pg_wrapper.cc). `ysql_default_transaction_isolation` and `ysql_datestyle` have non-single-token defaults, which is why the unit test specifically covers those values.

---

This could be a separate issue. When a user passes `--ysql_default_transaction_isolation=read committed` without shell quoting, the shell splits it into two args. gflags parses `--ysql_default_transaction_isolation=read` and leaves `committed` as an unparsed positional argument.

```bash
./build/latest/bin/yb-tserver \
--fs_data_dirs ~/yugabyte-data/tserver \
--tserver_master_addrs=127.0.0.1:7100 \
--pgsql_proxy_bind_address=127.0.0.1:5433 \
--enable_ysql=true \
--ysql_default_transaction_isolation=read committed
usage: ./build/latest/bin/yb-tserver
Unparsed arguments (1):
  argv[1]: committed
Invalid argument (yb/tserver/server_main_util.cc:267): Error parsing command-line flags
```

I checked this behavior after using this patch locally.

```diff
diff --git a/src/yb/tserver/server_main_util.cc b/src/yb/tserver/server_main_util.cc
index 88ad7bac3d..a039b0e67d 100644
--- a/src/yb/tserver/server_main_util.cc
+++ b/src/yb/tserver/server_main_util.cc
@@ -260,6 +260,10 @@ Status MasterTServerParseFlagsAndInit(
   ParseCommandLineFlags(argc, argv, /* remove_flag= */ true);
   if (*argc != 1) {
     std::cerr << "usage: " << (*argv)[0] << std::endl;
+    std::cerr << "Unparsed arguments (" << (*argc - 1) << "):" << std::endl;
+    for (int i = 1; i < *argc; ++i) {
+      std::cerr << "  argv[" << i << "]: " << (*argv)[i] << std::endl;
+    }
     return STATUS(InvalidArgument, "Error parsing command-line flags");
   }
```

iSignal: @keisku : Update looks reasonable, let me run it through local unit tests at my end.

Regarding `--ysql_default_transaction_isolation=read committed` it is assumed that user would quote at a shell level, so something like `"--ysql_default_transaction_isolation=read committed"` or `--ysql_default_transaction_isolation="read committed"` would work at the shell level. In both cases, the flag value will be set to "read committed"

iSignal (approved):
Thanks for your contribution @keisku.  

DB unit tests look good (internal diff: D50747)

PR #30493, opened 2026-02-26
 Attach `xrepl_origin_id` to every DML CDC event

kgalieva:
## Summary

Attach `xrepl_origin_id` to every DML CDC event (INSERT/UPDATE/DELETE), not just COMMIT records.

## Motivation

`pg_replication_origin` sets an integer `origin_id` per session. Today, this value only appears on COMMIT `RowMessage` records in CDC output. Individual DML events carry `xrepl_origin_id = 0`. This forces CDC consumers to buffer all DML events, wait for the COMMIT to learn the origin, then retroactively attribute it — impractical when consuming from independent per-tablet streams.

With this change, each DML record carries the origin_id immediately, enabling CDC consumers to route, filter, or tag events on the fly without buffering.

## Changes

The `xrepl_origin_id` is **already extracted** from the WAL at every point where DML records are created — it just wasn't being set on the `RowMessage`. This PR adds that.

### `src/yb/cdc/cdcsdk_producer.cc`

**Single-shard path** (`PopulateCDCSDKWriteRecord`): After extracting `xrepl_origin_id` from `msg->write()`, set it on the DML `row_message` immediately.

**Multi-shard path** (`PopulateCDCSDKIntentRecord`): Added `xrepl_origin_id` parameter to the function signature. The value was already available in the caller (`ProcessIntents`) but wasn't passed through. Now each DML `row_message` created from intents carries the origin_id.

### `src/yb/cdc/cdc_service.proto`

Updated comment on `xrepl_origin_id` field from "Only set for COMMIT Ops" to "Set on DML and COMMIT Ops". No wire format change.

## Test plan

- [x] Existing test `TestOriginId` passes unchanged (the `get_xrepl_origin_id` lambda finds the first non-zero origin_id, which is now a DML record instead of COMMIT — same value)
- [x] New test `TestOriginIdOnDMLRecords` verifies:
  - Single-shard INSERT/UPDATE/DELETE records carry `xrepl_origin_id`
  - Multi-shard (explicit transaction) DML records carry `xrepl_origin_id`
  - COMMIT records still carry `xrepl_origin_id` (backwards compat)
  - Local writes (no origin) have origin_id 0/absent
- [x] End-to-end verified with CDC consumer app against a local YugabyteDB build — DML records show varying `xrepl_origin_id` per transaction

---

> [!NOTE]
> **Medium Risk**
> Changes CDC output semantics by adding an additional field on high-volume DML records, which could affect downstream consumers that assume it is only present on COMMIT, but the change is additive and tested.
> 
> **Overview**
> CDCSDK `RowMessage` records for INSERT/UPDATE/DELETE now populate `xrepl_origin_id` immediately, for both the single-shard write path and the multi-shard intent path (by threading `xrepl_origin_id` into `PopulateCDCSDKIntentRecord`).
> 
> Adds an integration test `TestOriginIdOnDMLRecords` covering single-shard and multi-shard transactions and verifying local writes omit/zero the field; updates the proto comment to reflect the expanded semantics (no wire change).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c63e4c20db34f97b9dd6f3927e107654f1de6c0e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/yb/integration-tests/cdcsdk_ysql-test.cc
+++ b/src/yb/integration-tests/cdcsdk_ysql-test.cc
@@ -12988,7 +12988,6 @@ TEST_F(CDCSDKYsqlTest, TestOriginId) {
 }
 
 TEST_F(CDCSDKYsqlTest, TestOriginIdOnDMLRecords) {
-  ANNOTATE_UNPROTECTED_WRITE(FLAGS_cdc_populate_end_markers_transactions) = true;
   ASSERT_OK(SetUpWithParams(1 /* rf */, 1 /* num_masters*/));
   const auto kOrigin1 = "origin1";
   auto conn = ASSERT_RESULT(test_cluster_.ConnectToDB(kNamespaceName));


src/yb/integration-tests/cdcsdk_ysql-test.cc
```diff
@@ -12987,6 +12987,99 @@ TEST_F(CDCSDKYsqlTest, TestOriginId) {
   cdc_sdk_checkpoint = change_resp.cdc_sdk_checkpoint();
 }
 
+TEST_F(CDCSDKYsqlTest, TestOriginIdOnDMLRecords) {
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_yb_enable_cdc_consistent_snapshot_streams) = true;
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_cdc_populate_end_markers_transactions) = true;
```
Sumukh-Phalgaonkar: This is not needed since the flag is true by default.

src/yb/integration-tests/cdcsdk_ysql-test.cc
```diff
@@ -12987,6 +12987,99 @@ TEST_F(CDCSDKYsqlTest, TestOriginId) {
...
+
+  ASSERT_OK(conn.FetchFormat("SELECT pg_replication_origin_create('$0');", kOrigin1));
+
+  auto table = ASSERT_RESULT(CreateTable(&test_cluster_, kNamespaceName, kTableName));
+  google::protobuf::RepeatedPtrField<master::TabletLocationsPB> tablets;
+  ASSERT_OK(test_client()->GetTablets(table, 0, &tablets, nullptr));
+  ASSERT_EQ(tablets.size(), 1);
+  auto stream_id = ASSERT_RESULT(CreateConsistentSnapshotStream());
+
+  // Consume the initial schema record.
```
Sumukh-Phalgaonkar: What do you mean by initial schema record. Shouldn't the get changes response below contain change records corresponding to the below insert?

Maybe, you can assert that these records don't contain any origin id.

src/yb/integration-tests/cdcsdk_ysql-test.cc
```diff
@@ -12987,6 +12987,99 @@ TEST_F(CDCSDKYsqlTest, TestOriginId) {
...
+        }
+      };
+
+  // --- Single-shard (autocommit) path ---
+  // INSERT with origin.
+  ASSERT_OK(conn.FetchFormat("SELECT pg_replication_origin_session_setup('$0');", kOrigin1));
+  ASSERT_OK(conn.ExecuteFormat("INSERT INTO $0 VALUES (1, 100)", kTableName));
+  ASSERT_OK(conn.Fetch("SELECT pg_replication_origin_session_reset()"));
+  change_resp = ASSERT_RESULT(GetChangesFromCDC(stream_id, tablets, &cdc_sdk_checkpoint));
+  LOG(INFO) << "Single-shard INSERT: " << change_resp.ShortDebugString();
```
Sumukh-Phalgaonkar: We generally do not keep any INFO level logs in the tests

src/yb/integration-tests/cdcsdk_ysql-test.cc
```diff
@@ -12845,6 +12987,95 @@ TEST_F(CDCSDKYsqlTest, TestOriginId) {
...
+          auto op = record.row_message().op();
+          if (op == RowMessage::BEGIN || op == RowMessage::INSERT || op == RowMessage::UPDATE ||
+              op == RowMessage::DELETE || op == RowMessage::COMMIT) {
+            if (expected_origin_id != 0) {
+              ASSERT_TRUE(record.row_message().has_xrepl_origin_id())
+                  << "Expected xrepl_origin_id on op=" << RowMessage::Op_Name(op);
+              ASSERT_EQ(record.row_message().xrepl_origin_id(), expected_origin_id)
+                  << "Wrong xrepl_origin_id on op=" << RowMessage::Op_Name(op);
+            } else {
+              // origin_id 0 means local — field should be absent or zero.
```
shishir2001-yb: Please replace the "—" with "-"

```
   Error  (TXT5) Bad Charset
    Source code should contain only ASCII bytes with ordinal decimal values
    between 32 and 126 inclusive, plus linefeed. Do not use UTF-8 or other
    multibyte charsets.

           13015               ASSERT_EQ(record.row_message().xrepl_origin_id(), expected_origin_id)
           13016                   << "Wrong xrepl_origin_id on op=" << RowMessage::Op_Name(op);
           13017             } else {
    >>>    13018               // origin_id 0 means local — field should be absent or zero.
           13019               ASSERT_TRUE(!record.row_message().has_xrepl_origin_id() ||
           13020                           record.row_message().xrepl_origin_id() == 0)
           13021                   << "Expected no xrepl_origin_id on op=" << RowMessage::Op_Name(op);
```

src/yb/integration-tests/cdcsdk_ysql-test.cc
```diff
@@ -12845,6 +12987,95 @@ TEST_F(CDCSDKYsqlTest, TestOriginId) {
...
+  ASSERT_OK(conn.ExecuteFormat(
+      "UPDATE $0 SET $1 = 300 WHERE $2 = 2", kTableName, kValueColumnName, kKeyColumnName));
+  ASSERT_OK(conn.ExecuteFormat("DELETE FROM $0 WHERE $1 = 0", kTableName, kKeyColumnName));
+  ASSERT_OK(conn.Execute("COMMIT"));
+  ASSERT_OK(conn.Fetch("SELECT pg_replication_origin_session_reset()"));
+  change_resp = ASSERT_RESULT(GetChangesFromCDC(stream_id, tablets, &cdc_sdk_checkpoint));
+  ASSERT_NO_FATAL_FAILURE(verify_origin_id_on_all_records(change_resp, 1));
+  cdc_sdk_checkpoint = change_resp.cdc_sdk_checkpoint();
+
+  // --- Local (no origin) path — verify origin_id is 0/absent ---
```
shishir2001-yb: Seme as above, replace it with "-"

src/yb/integration-tests/cdcsdk_ysql-test.cc
```diff
@@ -12845,6 +12987,95 @@ TEST_F(CDCSDKYsqlTest, TestOriginId) {
   cdc_sdk_checkpoint = change_resp.cdc_sdk_checkpoint();
 }
 
+TEST_F(CDCSDKYsqlTest, TestOriginIdOnDMLRecords) {
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_cdc_populate_end_markers_transactions) = true;
```
Sumukh-Phalgaonkar: The flag `cdc_populate_end_markers_transactions` is true by default, so this can be skipped.

PR #30583, opened 2026-03-04
[DocDB]: Swap the use of fdopen for open syscall in filesystem util to avoid FD limit errors on OSX

austenLacy:
# Summary

Fixes: https://github.com/yugabyte/yugabyte-db/issues/30601

On macOS, `fopen()` returns `EMFILE` once the process has ~32,767 open file descriptors, regardless of system limits. This is because Apple's libc stores the fd in a `short` field inside the `FILE` struct and [unconditionally rejects fds > SHRT_MAX](https://github.com/apple-open-source-mirror/Libc/blob/master/stdio/FreeBSD/fopen.c). The raw `open()` syscall has no such limit.

RocksDB's `NewSequentialFile` and `NewLogger` both use `fopen()`, which causes tserver tablet bootstrap failures when running multiple databases with many tables on macOS.

This PR replaces `fopen()` / `FILE*` with raw POSIX fd I/O in those two code paths, matching the pattern already used by `PosixRandomAccessFile` and `PosixWritableFile`.

## Steps to Reproduce

Requires macOS. Ensure system limits are raised first:

```bash
sudo sysctl -w kern.maxvnodes=1048576
sudo launchctl limit maxfiles 1048576 unlimited
ulimit -n 1048576
```

Start a single-node cluster with 1 tablet per table:

```bash
bin/yugabyted start \
  --base_dir /tmp/yb-emfile-repro \
  --listen 127.0.0.1 \
  --tserver_flags "ysql_num_shards_per_tserver=1,yb_num_shards_per_tserver=1,tablet_replicas_per_core_limit=0,tablet_replicas_per_gig_limit=0" \
  --master_flags "ysql_num_shards_per_tserver=1,yb_num_shards_per_tserver=1,replication_factor=1,tablet_replicas_per_core_limit=0,tablet_replicas_per_gig_limit=0"
```

Create 2 databases with 1,500 tables and 1 secondary index each:

```bash
YSQLSH="bin/ysqlsh -h 127.0.0.1 -U yugabyte"

for db in testdb1 testdb2; do
  $YSQLSH -d yugabyte -c "CREATE DATABASE $db;"
  for batch in $(seq 0 14); do
    SQL=""
    for i in $(seq $((batch*100+1)) $((batch*100+100))); do
      SQL="${SQL}CREATE TABLE t${i}(id INT PRIMARY KEY, val TEXT);"
      SQL="${SQL}CREATE INDEX t${i}_val_idx ON t${i}(val);"
    done
    $YSQLSH -d $db -c "$SQL"
  done
done
```

Before the change during creation of `testdb2`, the tserver log will show errors like:

```
E tablet.cc:1224] Failed to open a RocksDB database:
  IO error (env_posix.cc:592): .../CURRENT (num opened files 32770): Too many open files (system error 24)
```

Table creation will stall or fail with `Timed out waiting for table creation`.

After the change it worked and I was able to create 3-4 of these same size databases without issues.

---

> [!NOTE]
> **Medium Risk**
> Changes low-level file and logger I/O paths on macOS from `FILE*`/`fopen` to raw `open`/`read`/`write`, which could affect resource lifecycle and flushing behavior under load, though it is largely platform-gated.
> 
> **Overview**
> Fixes macOS "~32K `FILE*` stream" exhaustion by switching key RocksDB/DocDB paths from `fopen`/`FILE*` to raw file-descriptor I/O.
> 
> On **macOS only**, `NewLogger` and `NewSequentialFile` now use `open` and pass fds through to updated `PosixLogger` and `PosixSequentialFile` implementations that perform `read`/`write` loops (handling `EINTR`) and close fds directly. The build also adds `_DARWIN_UNLIMITED_STREAMS` and required includes to support the new fd-based code paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cea43c06762e58b54adc173dea0433f421a1fee9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/yb/rocksdb/util/posix_logger.h
+++ b/src/yb/rocksdb/util/posix_logger.h
@@ -54,6 +54,9 @@ const int kDebugLogChunkSize = 128 * 1024;
 
 class PosixLogger : public Logger {
  private:
+#if !defined(__APPLE__)
+  FILE* file_;
+#endif
   int fd_;
   uint64_t (*gettid_)();  // Return the thread id for the current thread
   std::atomic_size_t log_size_;
@@ -63,6 +66,7 @@ class PosixLogger : public Logger {
   std::atomic<bool> flush_pending_;
   std::string fname_;
  public:
+#if defined(__APPLE__)
   PosixLogger(const std::string& fname, int fd, uint64_t (*gettid)(), Env* env,
               const InfoLogLevel log_level = InfoLogLevel::ERROR_LEVEL)
       : Logger(log_level),
@@ -78,13 +82,38 @@ class PosixLogger : public Logger {
   }
   virtual void Flush() override {
     DEBUG_ONLY_TEST_SYNC_POINT_CALLBACK("PosixLogger::Flush:BeginCallback", nullptr);
-    // With raw write() syscalls, data goes directly to OS kernel buffers — there is no user-space
-    // stdio buffer to flush. We intentionally avoid fsync() here to prevent forcing data to disk
-    // every 5 seconds per logger, which would be a significant perf regression in high-fd-count
-    // scenarios (see also win_logger.cc which documents this same reasoning).
+    // raw write() bypasses stdio buffering, so data is already in OS kernel buffers - no flush
+    // needed. Just reset the pending flag.
     flush_pending_.store(false);
     last_flush_micros_ = env_->NowMicros();
   }
+#else
+  PosixLogger(const std::string& fname, FILE* f, uint64_t (*gettid)(), Env* env,
+              const InfoLogLevel log_level = InfoLogLevel::ERROR_LEVEL)
+      : Logger(log_level),
+        file_(f),
+        fd_(fileno(f)),
+        gettid_(gettid),
+        log_size_(0),
+        last_flush_micros_(0),
+        env_(env),
+        flush_pending_(false),
+        fname_(fname) {}
+  virtual ~PosixLogger() {
+    fclose(file_);
+  }
+  virtual void Flush() override {
+    DEBUG_ONLY_TEST_SYNC_POINT_CALLBACK("PosixLogger::Flush:BeginCallback", nullptr);
+    {
+      bool expected_flush_pending = true;
+      // TODO: use a weaker memory order?
+      if (flush_pending_.compare_exchange_strong(expected_flush_pending, false)) {
+        fflush(file_);
+      }
+    }
+    last_flush_micros_ = env_->NowMicros();
+  }
+#endif // defined(__APPLE__)
 
   using Logger::Logv;
   virtual void Logv(const char* format, va_list ap) override {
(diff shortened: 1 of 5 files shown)


src/yb/rocksdb/util/posix_logger.h
```diff
@@ -53,39 +54,35 @@ const int kDebugLogChunkSize = 128 * 1024;
...
   virtual void Flush() override {
     DEBUG_ONLY_TEST_SYNC_POINT_CALLBACK("PosixLogger::Flush:BeginCallback", nullptr);
-    {
-      bool expected_flush_pending = true;
-      // TODO: use a weaker memory order?
-      if (flush_pending_.compare_exchange_strong(expected_flush_pending, false)) {
-        fflush(file_);
-      }
-    }
+    // With raw write() syscalls, data goes directly to OS kernel buffers — there is no user-space
```
iSignal: Did not follow why we want to change this behavior here? and the logic on the "every 5 secs per logger" part - 
1. If it was working this way earlier we need to maintain the same behavior. If we want to change perf impact of this, that should be ideally a separate PR with separate discussion.
2. is posix logger used for any perf critical stuff like compaction etc? For smaller files like manifest etc, not sure if it matters.

CMakeLists.txt
```diff
@@ -454,6 +454,7 @@ endif()
 # distinguishes further into BSD vs everything else, but we probably do not care further.
 if(APPLE)
   ADD_CXX_FLAGS("-DOS_MACOSX")
+  ADD_CXX_FLAGS("-D_DARWIN_UNLIMITED_STREAMS")
```
iSignal, line 457: minor: I guess we don't read this?

austenLacy: It's respected in most parts of OSX, but not this case apparently 🙃. It's also a good practice and one we should probably apply to the YB third party libs as well.

src/yb/rocksdb/util/posix_logger.h
austenLacy: Updated the comment because it was misleading. Essentially we don't need to flush when using `write()`.

src/yb/rocksdb/util/posix_logger.h
```diff
@@ -53,23 +54,47 @@ const int kDebugLogChunkSize = 128 * 1024;
...
+        last_flush_micros_(0),
+        env_(env),
+        flush_pending_(false),
+        fname_(fname) {}
+  virtual ~PosixLogger() {
+    close(fd_);
+  }
+  virtual void Flush() override {
+    DEBUG_ONLY_TEST_SYNC_POINT_CALLBACK("PosixLogger::Flush:BeginCallback", nullptr);
+    // raw write() bypasses stdio buffering, so data is already in OS kernel buffers — no flush
```
iSignal: minor: this emdash here fails lint because it is not an ASCII char

austenLacy: Fixed

PR #30623, opened 2026-03-06
Cleanup orphaned cdc state entries

egladysh:
Problem
Orphaned cdc_state entries for deleted tablets were never cleaned up, causing persistent "Read invalid stream id" warnings. This may happen when tablets are deleted after splits (parent is deleted before cleanup completes) or table drops (?). 
Possible reasons for orphaned cdc entries need to be investigate further. This PR is to deal only when it does happen.

The current cleanup logic required a tablet to exist locally on a tserver to proceed:
```
  auto tablet_peer_result = GetServingTablet(tablet_id);
  if (!tablet_peer_result.ok()) {
    continue;
  }
  if ((*tablet_peer_result)->IsLeaderAndReady()) {
    // Delete cdc_state entry
  }
```

When a tablet is deleted from the entire cluster, no tserver has a replica anymore, so GetServingTablet() fails everywhere. This created a condition when cdc state entries could never be removed.

Changes
Use GetRemoteTablet() to query the Master and check if the tablet exists anywhere in the cluster. If it returns IsNotFound(), the tablet is truly gone and *any* tserver can safely delete the orphaned cdc_state entry.

Added logging for an edge case: when a tablet exists locally but is not the leader and can never elect one. Cannot delete CDC state table entry for tablet X and stream Y - tablet exists locally but this tserver is not leader. If this persists, the tablet may be permanently unable to elect a leader or some other bad state... The persistent log, `Cannot delete CDC state table entry for tablet` should be a call to action.

---

> [!NOTE]
> **Medium Risk**
> Changes background `cdc_state` cleanup to allow deletion based on master tablet lookups, which could affect replication metadata if the existence check is wrong or flaky. Added warnings help detection, but misclassification could prematurely delete state for a still-live tablet.
> 
> **Overview**
> **Orphaned `cdc_state` rows are now eligible for cleanup even when no TServer hosts the tablet.** When `DeleteCDCStateTableMetadata` can’t find a local peer, it now calls `GetRemoteTablet()` and, if the master returns `NotFound`, deletes the orphaned `{tablet_id, stream_id}` entry (with warning logs).
> 
> **Improved observability for non-leader cleanup skips.** If a tablet exists locally but this TServer isn’t leader, the code now emits a periodic warning indicating the entry can’t be deleted.
> 
> Adds an integration test (`TestCleanupOrphanedEntriesForNonExistentTablet`) that drops a table, recreates a lingering `cdc_state` row, and asserts the background thread eventually removes it.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4fa3e7705c16243d68913cd4a222c5b582d3d7e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/yb/cdc/cdc_service.cc
+++ b/src/yb/cdc/cdc_service.cc
@@ -3538,7 +3538,7 @@ Status CDCServiceImpl::DeleteCDCStateTableMetadata(
       // Log a warning in this case but skip deletion. Who knows if the leader 
       // comes back up and cleans up the entry later. If the warning persists
       // it may indicate a problem with the tablet that needs to be investigated.
-      LOG(WARNING)
+      YB_LOG_EVERY_N_SECS(WARNING, 10)
           << "Cannot delete CDC state table entry for tablet " << tablet_id
           << " and stream " << stream_id
           << " - tablet exists but this tserver is not leader.";


egladysh:
@asrinivasanyb do you have any comments? We are still seeing over 170K traces with `cdc_service.cc:3122] Read invalid stream id` per 5min.

asrinivasanyb:
> @asrinivasanyb do you have any comments? We are still seeing over 170K traces with `cdc_service.cc:3122] Read invalid stream id` per 5min.

As mentioned in the previous comment, we could introduce a yb-admin command to cleanup the stale entries. 
A cdc_state table entry can be considered to be stale if it satisfies any of the following conditions - 

1. If stream (corresponding to stream_id) is not present
2. if tablet is not present (corresponding to tablet_id) is not present
3. if table containing the tablet is not a regular table (i.e it is an index or MV)

Perhaps this PR itself can be modified to move over to this approach of adding support for this yb-admin command.
Also it would be good to have --dump option for this command where the stale entries are only reported but not cleaned up.

egladysh:
@asrinivasanyb I agree that yb_admin option is a good to have but I don't think it's ideal. BTW: I suggested a syntax on our shared slack channel in case you missed it: 
`yb-admin cleanup_orphaned_cdc_state_entries [<namespace>] [--dry_run]`

Going back to the PR. If understand you correctly you main concern is RTT's to the master to GetServingTablet and when it fails, all involved tservers my get in sync deleting cleaning the same stream at the same time? I don't see it as a huge problem because it only happens for stale entries:
`for (const auto& [tablet_id, stream_id] : cdc_state_entries_to_delete) {` 
and should not be happening otherwise?

Also (as you mentioned) with With https://github.com/yugabyte/yugabyte-db/commit/58f9ec18f4be457983ecde08f0801df1d02e7c5e and https://github.com/yugabyte/yugabyte-db/commit/1ef98f6fa4a9e4ff1e4f47218a36de78ae93a504, we do not expect to get stale entries going forward. Could you clarify? I'll DM you the slack thread for you to comment as well.

asrinivasanyb:
> I don't see it as a huge problem because it only happens for stale entries:

My issue is with the genuine cases (not a stale entry) - when a table is dropped or stream is dropped these would be part of _cdc_state_entries_to_delete_ -- for each such entry, each Tserver (except for those on which peers for that tablet exist, RF in number, but much smaller compared to overall number of tservers) would make a call to master and master would respond confirming the presence of the tablet and there would be no action required - but the cost of this has been a lot of traffic to master for nothing.

Also going forward, we do not expect stale entries to gather - bugs in this connection have been fixed. Hence, all the more the only cases that would go to master would be the genuine cases. That is why I dont feel it is appropriate to make this change in the repeatedly running background thread. Instead a one time cleanup of the stale entries in the state table (the yb-admin command) and subsequently moving ahead to a release that picks up the fixes in the commits mentioned earlier should resolve this problem.

egladysh:
Sorry my mistake, `GetServingTablet()` is already called for each entry in `cdc_state_entries_to_delete`. You concerned about a call to `GetRemoteTablet` during table/stream drops only and if `!tablet_peer_result.ok()`?

asrinivasanyb:
> Sorry my mistake, `GetServingTablet()` is already called for each entry in `cdc_state_entries_to_delete`. You concerned about a call to `GetRemoteTablet` during table/stream drops only and if `!tablet_peer_result.ok()`?

Yes. Once there are no stale entries (and the big fixes picked up) this will be the only result

egladysh:
Agreed. I'll close this PR for the big fixes.

PR #30651, opened 2026-03-10
[YSQL] Fix pagination over hash-sharded tables with yb_hash_code() = const

### LINKED ISSUE #30524: [YSQL] Performance issues with paginating over an index scan on yb_hash_code(hash_cols)
Jira Link: [DB-20398](https://yugabyte.atlassian.net/browse/DB-20398)
### Description

When filtering by `yb_hash_code(hash_cols) = const`, the planner does not recognize that the index can provide sorted output within the pinned hash bucket.
This forces an explicit Sort node even when the query includes `ORDER BY` on the index's range key columns, making `LIMIT`-based pagination over a hash
partition unusable.

## Reproduction

```sql
CREATE TABLE foo.orders (
    uuid        UUID,
    user_uuid   UUID,
    -- ... many columns ...
    PRIMARY KEY (user_uuid HASH, uuid ASC)
);

-- Insert enough data so that a single hash bucket has ~150k rows.

EXPLAIN (ANALYZE, DIST, COSTS OFF)
SELECT * FROM foo.orders
WHERE yb_hash_code(user_uuid) = 10
ORDER BY user_uuid, uuid
LIMIT 10000;
```

### Actual plan

```
Limit (actual time=790ms)
-> Sort (Sort Method: external merge  Disk: 65904kB)
    Sort Key: user_uuid, uuid
    -> Index Scan using orders_pkey on orders
            Index Cond: (yb_hash_code(user_uuid) = 10)
            Storage Table Rows Scanned: 179690
```

All ~180k rows in the hash bucket are fetched, written to a 65MB on-disk sort, and only then is the LIMIT applied.

 ### Expected plan

 ```
 Limit
   -> Index Scan using orders_pkey on orders (forward)
        Index Cond: (yb_hash_code(user_uuid) = 10)
 ```
No Sort node. The index scan returns rows in `(user_uuid, uuid)` order within the hash bucket. The LIMIT stops the scan after 10,000 rows.

## Impact

This blocks a common pattern for batch-processing large hash-sharded tables: iterating all rows in a hash partition using keyset pagination:

```sql
SELECT * FROM foo.orders
WHERE yb_hash_code(user_uuid) = 10
  AND (user_uuid, uuid) > (:last_user_uuid, :last_uuid)
ORDER BY user_uuid, uuid
LIMIT 10000;
```

Without index ordering, every page requires a full sort of the entire partition. On a 10B-row table with ~150k rows per hash bucket, this makes pagination impractical.

## Potential fix

I looked briefly at the source and this seems like a missing implementation rather than a design issue. The storage layer already returns rows in sorted order within a hash bucket — the gap is purely in the planner's pathkey generation.

The fix is in `build_index_pathkeys()` in `pathkeys.c`. Today, when it iterates over index columns, hash columns need a direct equality EC (e.g. `WHERE user_uuid = 'abc'`) to be "satisfied". Without one, it bails out and returns no pathkeys at all (line ~681: `if (i < index->nhashcolumns) return NULL`).

The change: when a hash column has no equality EC, check if there's a `yb_hash_code(hash_cols) = const` clause covering all hash columns of this index. If so, skip past the hash columns and continue generating sort pathkeys for the range columns. Most of the detection logic already exists in `indxpath.c` — `is_yb_hash_code_call()` (line ~2857) and `yb_hash_code_call_matches_indexcol()` (line ~2871) — it just needs to be reachable from `pathkeys.c`.

No executor changes needed. Once the planner emits pathkeys and sets `ForwardScanDirection` instead of `NoMovement`, the existing code in `nodeIndexscan.c` calls `YBCPgSetForwardScan(true)` and DocDB returns rows in key order.

One thing to watch: this should only apply to `yb_hash_code = const` (equality), not range conditions like `yb_hash_code < X`, since those can span multiple tablets.

### Issue Type

kind/bug

[DB-20398]: https://yugabyte.atlassian.net/browse/DB-20398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kirs:
When `yb_hash_code(hash_cols) = constant` pins a scan to a single hash bucket, DocDB stores rows in `(hash_cols, range_cols)` order within that bucket. Three independent code paths prevented exploiting this ordering for pagination:

1. `pathkeys.c` (`build_index_pathkeys`): Hash columns were never allowed to produce sort pathkeys, forcing a Sort node even when ORDER BY matched the index key order within a pinned bucket. Fix: when yb_hash_code equality is detected, pass is_hash_index=false so hash columns generate sort pathkeys.

2. `indxpath.c` (`match_rowcompare_to_indexcol`): ROW comparisons like `(h, r) > (?, ?)` were unconditionally rejected when the first column was a hash column. Fix: allow through when yb_hash_code equality pins the scan to a single bucket.

3. `yb_scan.c` (`YbBindRowComparisonKeys`): Two guards prevented binding ROW comparison keys to DocDB when any hash column was present in the index or the comparison subkeys. Fix: relax both guards when a yb_hash_code equality scan key is present.

Together these fixes enable efficient keyset pagination:

```
  SELECT * FROM t
  WHERE yb_hash_code(h) = 10 AND (h, r) > (:last_h, :last_r)
  ORDER BY h, r LIMIT 1000;
```

Previously this required a full sort of the hash bucket on every page and scanned from the start of the bucket. Now it seeks directly to the cursor position.

Fixes: #30524

Test Plan: New regression test yb.orig.hash_code_pagination covering:
- Sort elimination with yb_hash_code equality + ORDER BY
- ROW comparison pushdown as Index Cond
- Negative tests (range predicates, arity mismatches)
- Multi-column hash key support

---

> [!NOTE]
> **Medium Risk**
> Planner and scan-binding logic for YB hash-sharded primary keys is relaxed based on detecting `yb_hash_code(...) = const`, which can change query plans and index bound pushdown behavior and may impact correctness if the pinning detection misfires.
> 
> **Overview**
> Improves keyset pagination and ORDER BY planning on hash-sharded (LSM) primary keys when the scan is *pinned to a single hash bucket* via `yb_hash_code(hash_cols) = const`.
> 
> This teaches the optimizer to treat hash columns as providing valid ordering/pathkeys and allows `ROW(...)` range comparisons to match and be pushed down as index bounds in that pinned-bucket case, and updates `YbBindRowComparisonKeys` to permit binding row-comparison bounds involving hash columns while avoiding `IS NOT NULL` binds on hash/partition columns.
> 
> Adds a new regression test `yb.orig.hash_code_pagination` (and schedule entry) covering sort elimination, row-compare pushdown, multi-column hash keys, and negative cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 321541d4b51fc44ad54ee8cf8564d5584890d83c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/postgres/src/backend/access/yb_access/yb_scan.c
+++ b/src/postgres/src/backend/access/yb_access/yb_scan.c
@@ -1590,6 +1590,28 @@ YbCheckScanTypes(YbScanDesc ybScan, YbScanPlan scan_plan, int i)
 			IsPolymorphicType(valtypid));
 }
 
+/*
+ * yb_has_hash_code_equality_scan_key
+ *	  Return true if the scan has a yb_hash_code equality scan key, meaning
+ *	  the scan is pinned to a single hash bucket.  In that case, hash columns
+ *	  can participate in ROW comparison pushdown since rows within a single
+ *	  bucket are stored in (hash_cols, range_cols) order.
+ */
+static bool
+yb_has_hash_code_equality_scan_key(YbScanDesc ybScan)
+{
+	ListCell   *lc;
+
+	foreach(lc, ybScan->hash_code_keys)
+	{
+		ScanKey		key = (ScanKey) lfirst(lc);
+
+		if (key->sk_strategy == BTEqualStrategyNumber)
+			return true;
+	}
+	return false;
+}
+
 static bool
 YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 						int skey_index, bool is_not_null[],
@@ -1636,9 +1658,14 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 		}
 		last_att_no = key->sk_attno;
 
-		/* Make sure that there are no hash key columns. */
-		if (index->rd_indoption[key->sk_attno - 1]
-			& INDOPTION_HASH)
+		/*
+		 * Make sure that there are no hash key columns, unless the scan
+		 * is pinned to a single hash bucket via yb_hash_code() = const.
+		 * Within a single bucket rows are ordered by (hash_cols, range_cols)
+		 * so range comparisons on hash columns are valid.
+		 */
+		if ((index->rd_indoption[key->sk_attno - 1] & INDOPTION_HASH) &&
+			!yb_has_hash_code_equality_scan_key(ybScan))
 		{
 			can_pushdown = false;
 			break;
@@ -1646,14 +1673,41 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 	}
 
 	/*
-	 * Make sure that the primary key has no hash columns in order
-	 * to push down.
+	 * Make sure that the index has no hash columns in order to push down.
+	 * When pinned to a single hash bucket via yb_hash_code() = const, we can
+	 * relax this only if the ROW comparison subkeys actually include all hash
+	 * columns -- otherwise EncodeRowKeyForBound would receive NULL for
+	 * unspecified hash column positions, which is not supported.
+	 *
+	 * Hash columns always precede range columns in a YB index, so we can
+	 * stop as soon as we see a non-hash column.
 	 */
-
-	for (int i = 0; (i < index->rd_index->indnkeyatts) && can_pushdown; i++)
+	if (can_pushdown)
 	{
-		if (index->rd_indoption[i] & INDOPTION_HASH)
+		bool	has_hash_code_eq = yb_has_hash_code_equality_scan_key(ybScan);
+
+		for (int i = 0;
+			 i < index->rd_index->indnkeyatts &&
+			 (index->rd_indoption[i] & INDOPTION_HASH);
+			 i++)
+		{
+			if (has_hash_code_eq)
+			{
+				bool	found = false;
+				for (int j = 0; j < subkey_count; j++)
+				{
+					if (subkeys[j]->sk_attno - 1 == i)
+					{
+						found = true;
+						break;
+					}
+				}
+				if (found)
+					continue;
+			}
 			can_pushdown = false;
+			break;
+		}
 	}
 
 	bool		needs_recheck = true;
@@ -1756,8 +1810,22 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 				int			att_idx = YBAttnumToBmsIndex(((TableScanDesc) ybScan)->rs_rd,
 														 subkeys[subkey_index]->sk_attno);
 
-				/* Set the first column in this RC to not null. */
-				is_not_null[att_idx] |= subkey_index == 0;
+				/*
+				 * Synthesize IS NOT NULL on the leading RC column to
+				 * exclude NULL rows that might otherwise fall within
+				 * the DocDB row bound.
+				 *
+				 * Skip this for hash columns: pggate does not support
+				 * BindColumnCondIsNotNull on hash columns (it CHECKs
+				 * !col.is_partition()).  This is safe because the only
+				 * code path that reaches here with a hash column as the
+				 * leading subkey is the pinned-bucket case, where
+				 * yb_hash_code(hash_cols) = const already excludes
+				 * NULL rows (NULL values hash to a different bucket).
+				 */
+				is_not_null[att_idx] |= (subkey_index == 0 &&
+										 !(index->rd_indoption[subkeys[subkey_index]->sk_attno - 1] &
+										   INDOPTION_HASH));
 
 				subkey_index++;
 			}
--- a/src/postgres/src/backend/access/yb_access/yb_scan.c
+++ b/src/postgres/src/backend/access/yb_access/yb_scan.c
@@ -1816,12 +1816,19 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 				 * the DocDB row bound.
 				 *
 				 * Skip this for hash columns: pggate does not support
-				 * BindColumnCondIsNotNull on hash columns (it CHECKs
-				 * !col.is_partition()).  This is safe because the only
-				 * code path that reaches here with a hash column as the
-				 * leading subkey is the pinned-bucket case, where
-				 * yb_hash_code(hash_cols) = const already excludes
-				 * NULL rows (NULL values hash to a different bucket).
+				 * BindColumnCondIsNotNull on hash/partition columns
+				 * (it CHECKs !col.is_partition()).  This is safe
+				 * because:
+				 *
+				 * - Primary key hash columns cannot be NULL.
+				 * - Secondary index hash columns can be NULL, and
+				 *   yb_hash_code(NULL) may land in the pinned bucket.
+				 *   However, the ROW comparison is always rechecked at
+				 *   the SQL layer (needs_recheck = true when any
+				 *   column is unspecified or has mixed directionality),
+				 *   and SQL comparisons involving NULL yield NULL
+				 *   (treated as false), so NULL rows are correctly
+				 *   excluded.
 				 */
 				is_not_null[att_idx] |= (subkey_index == 0 &&
 										 !(index->rd_indoption[subkeys[subkey_index]->sk_attno - 1] &
--- a/src/postgres/src/backend/optimizer/path/indxpath.c
+++ b/src/postgres/src/backend/optimizer/path/indxpath.c
@@ -3660,7 +3660,16 @@ match_rowcompare_to_indexcol(PlannerInfo *root,
 	if (index->relam != BTREE_AM_OID && index->relam != LSM_AM_OID)
 		return NULL;
 
-	if (is_hash_column_in_lsm_index(index, indexcol))
+	/*
+	 * YB: Hash columns in an LSM index normally cannot use range comparisons
+	 * because the hash values are not ordered across buckets.  However, when
+	 * yb_hash_code(hash_cols) = constant pins the scan to a single hash
+	 * bucket, rows within that bucket ARE stored in (hash_cols, range_cols)
+	 * order, so range comparisons (including ROW comparisons used for keyset
+	 * pagination) on hash columns are valid.
+	 */
+	if (is_hash_column_in_lsm_index(index, indexcol) &&
+		!yb_has_hash_code_equality_qual(index))
 		return NULL;
 
 	index_relid = index->rel->relid;
--- a/src/postgres/src/backend/optimizer/path/pathkeys.c
+++ b/src/postgres/src/backend/optimizer/path/pathkeys.c
@@ -31,6 +31,7 @@
 /* YB includes */
 #include "access/sysattr.h"
 #include "optimizer/yb_saop_merge.h"
+#include "pg_yb_utils.h"
 
 
 static bool pathkey_is_redundant(PathKey *new_pathkey, List *pathkeys);
@@ -570,10 +571,18 @@ build_index_pathkeys(PlannerInfo *root,
 	int			i;
 	int			yb_distinct_prefixlen;
 	int			yb_saop_merge_cardinality = 1;
+	bool		yb_hash_bucket_pinned;
 
 	if (index->sortopfamily == NULL)
 		return NIL;				/* non-orderable index */
 
+	/*
+	 * YB: When yb_hash_code(hash_cols) = constant constrains the scan to a
+	 * single hash bucket, DocDB guarantees rows are stored in
+	 * (hash_cols, range_cols) order within that bucket, so hash columns
+	 * provide valid sort-order pathkeys.
+	 */
+	yb_hash_bucket_pinned = yb_has_hash_code_equality_qual(index);
 	/*
 	 * YB: Compute the set of pathkeys corresponding to the distinct index scan
 	 * prefix. 0 when the prefix is empty.
@@ -618,6 +627,10 @@ build_index_pathkeys(PlannerInfo *root,
 		/*
 		 * OK, try to make a canonical pathkey for this sort key.  Note we're
 		 * underneath any outer joins, so nullable_relids should be NULL.
+		 *
+		 * YB: When pinned to a single hash bucket, treat hash columns as
+		 * range columns for ordering purposes -- they provide valid sort
+		 * pathkeys within that bucket.
 		 */
 		cpathkey = make_pathkey_from_sortinfo(root,
 											  indexkey,
@@ -630,7 +643,8 @@ build_index_pathkeys(PlannerInfo *root,
 											  0,
 											  index->rel->relids,
 											  false,
-											  yb_is_hash_column);
+											  yb_is_hash_column &&
+											  !yb_hash_bucket_pinned);
 
 		/*
 		 * YB: Index keys part of scalar array operations may be eligible for
@@ -647,8 +661,14 @@ build_index_pathkeys(PlannerInfo *root,
 		{
 			/* Do nothing */
 		}
-		/* YB: For hash columns, sort pathkeys are invalid. */
+		/*
+		 * YB: For hash columns, sort pathkeys are normally invalid.
+		 * Exception: when pinned to a single hash bucket via
+		 * yb_hash_code(hash_cols) = const, hash columns provide valid
+		 * sort ordering within that bucket.
+		 */
 		else if (cpathkey && !(yb_is_hash_column &&
+							   !yb_hash_bucket_pinned &&
 							   cpathkey->pk_eclass->ec_sortref != 0))
 		{
 			/*
@@ -671,7 +691,7 @@ build_index_pathkeys(PlannerInfo *root,
 			 * keys won't be useful either.
 			 */
 			if (!indexcol_is_bool_constant_for_query(root, index, i) ||
-				yb_is_hash_column)
+				(yb_is_hash_column && !yb_hash_bucket_pinned))
 				break;
 		}
 
@@ -699,11 +719,15 @@ build_index_pathkeys(PlannerInfo *root,
 	 * We can request a distinct index scan on h1, h2 tuples but there may still
 	 * be some duplicate values of h1 in the result.
 	 */
-	if (i < index->nhashcolumns)
+	if (i < index->nhashcolumns && !yb_hash_bucket_pinned)
 	{
 		/*
 		 * All hash columns must have EQ pathkeys. Otherwise, we cannot use
-		 * the index
+		 * the index.
+		 *
+		 * YB: Exception when pinned to a single hash bucket via
+		 * yb_hash_code(hash_cols) = const -- hash columns provide valid
+		 * ordering within the bucket.
 		 */
 		return NULL;
 	}
--- a/src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
+++ b/src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
@@ -1,0 +1,312 @@
+--
+-- Test hash code pagination: yb_hash_code(hash_cols) = const should allow
+-- the planner to recognize index ordering within a single hash bucket and
+-- push down ROW comparison conditions for keyset pagination.
+-- See https://github.com/yugabyte/yugabyte-db/issues/30524
+--
+CREATE TABLE hk_pagination(
+    h int,
+    r1 int,
+    r2 int,
+    v text,
+    PRIMARY KEY(h HASH, r1 ASC, r2 ASC)
+);
+INSERT INTO hk_pagination SELECT i % 5, i, i * 10, 'val' || i FROM generate_series(1, 20) i;
+--
+-- Fix 1: Pathkey recognition -- no Sort when yb_hash_code = const + ORDER BY
+-- matches the full index key order.
+--
+-- Should NOT have a Sort node (single hash bucket, index provides ordering).
+EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) = 1000
+    ORDER BY h, r1, r2
+    LIMIT 10;
+                         QUERY PLAN                         
+------------------------------------------------------------
+ Limit
+   ->  Index Scan using hk_pagination_pkey on hk_pagination
+         Index Cond: (yb_hash_code(h) = 1000)
+(3 rows)
+
+-- Negative: range predicate on yb_hash_code should still Sort (multiple buckets).
+EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) < 1000
+    ORDER BY h, r1, r2
+    LIMIT 10;
+                            QUERY PLAN                            
+------------------------------------------------------------------
+ Limit
+   ->  Sort
+         Sort Key: h, r1, r2
+         ->  Index Scan using hk_pagination_pkey on hk_pagination
+               Index Cond: (yb_hash_code(h) < 1000)
+(5 rows)
+
+-- Negative: yb_hash_code equality but ORDER BY doesn't match leading key.
+EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) = 1000
+    ORDER BY r1, r2
+    LIMIT 10;
+                            QUERY PLAN                            
+------------------------------------------------------------------
+ Limit
+   ->  Sort
+         Sort Key: r1, r2
+         ->  Index Scan using hk_pagination_pkey on hk_pagination
+               Index Cond: (yb_hash_code(h) = 1000)
+(5 rows)
+
+--
+-- Fix 2+3: ROW comparison pushdown -- (h, r1) > (?, ?) should become an
+-- index bound (Index Cond), not a post-filter (Filter), when hash bucket
+-- is pinned.
+--
+-- ROW comparison on (h, r1) with yb_hash_code equality: should be Index Cond.
+EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) = 1000
+      AND (h, r1) > (2, 10)
+    LIMIT 10;
+                                  QUERY PLAN                                  
+------------------------------------------------------------------------------
+ Limit
+   ->  Index Scan using hk_pagination_pkey on hk_pagination
+         Index Cond: ((ROW(h, r1) > ROW(2, 10)) AND (yb_hash_code(h) = 1000))
+(3 rows)
+
+-- Full keyset pagination pattern: hash_code equality + ROW comparison + ORDER BY + LIMIT.
+EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) = 1000
+      AND (h, r1, r2) > (2, 10, 100)
+    ORDER BY h, r1, r2
+    LIMIT 10;
+                                      QUERY PLAN                                       
+---------------------------------------------------------------------------------------
+ Limit
+   ->  Index Scan using hk_pagination_pkey on hk_pagination
+         Index Cond: ((ROW(h, r1, r2) > ROW(2, 10, 100)) AND (yb_hash_code(h) = 1000))
+(3 rows)
+
+-- Negative: without yb_hash_code equality, ROW comparison on hash column
+-- should NOT become Index Cond (cannot seek across buckets).
+EXPLAIN (COSTS OFF) /*+IndexScan(hk_pagination)*/ SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) < 1000
+      AND (h, r1) > (2, 10)
+    LIMIT 10;
+                         QUERY PLAN                         
+------------------------------------------------------------
+ Limit
+   ->  Index Scan using hk_pagination_pkey on hk_pagination
+         Index Cond: (yb_hash_code(h) < 1000)
+         Filter: (ROW(h, r1) > ROW(2, 10))
+(4 rows)
+
+--
+-- Correctness: verify the queries return correct results.
+--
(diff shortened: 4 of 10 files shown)


src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1756,8 +1785,15 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 
-				/* Set the first column in this RC to not null. */
-				is_not_null[att_idx] |= subkey_index == 0;
+				/*
+				 * Set the first column in this RC to not null, except for
+				 * partition/hash columns. Pggate does not allow binding
+				 * IS NOT NULL on partition columns, and for primary-key hash
+				 * columns the null check is redundant anyway.
+				 */
+				is_not_null[att_idx] |= (subkey_index == 0 &&
```
kirs, line 1833: Manual repro context for this guard: after allowing ROW-comparison pushdown on hash columns in the pinned-bucket case, planning was fine but actual execution crashed in pggate. The RC bind path still synthesizes `IS NOT NULL` on the first RC column; for queries like `WHERE yb_hash_code(h) = 4624 AND (h, r1, r2) > (...) ORDER BY h, r1, r2`, that first column is now the hash/partition key `h`. Pggate rejects `BindColumnCondIsNotNull` on partition columns (`pg_dml_read.cc:765`, `CHECK(!col.is_partition())`), so skipping the synthetic null-check here for hash columns fixes the crash while keeping the row-bound pushdown.

src/postgres/src/backend/optimizer/path/pathkeys.c
```diff
@@ -521,6 +524,69 @@ get_cheapest_parallel_safe_total_inner(List *paths)
...
+			continue;
+
+		if (list_length(funcexpr->args) != index->nhashcolumns)
+			continue;
+
+		return true;
+	}
+
+	return false;
+}
```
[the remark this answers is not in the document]
mtakahar: Would be best combining those two versions and move it to `pg_yb_utils.h`/`pg_yb_utils.c`.

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -4828,6 +4838,61 @@ is_hash_column_in_lsm_index(const IndexOptInfo *index, int columnIndex)
...
+		if (is_yb_hash_code_call(lhs) && IsA(rhs, Const))
+			funcexpr = (FuncExpr *) lhs;
+		else if (is_yb_hash_code_call(rhs) && IsA(lhs, Const))
+			funcexpr = (FuncExpr *) rhs;
+		else
+			continue;
+
+		if (list_length(funcexpr->args) != index->nhashcolumns)
+			continue;
+
```
mtakahar: Since `index->rel->baserestrictinfo` contains all the predicates on the table, not only the ones relevant to the index, I think the `yb_hash_code` arguments need to match the ones in the expression based index key item expression.  Otherwise the code can mix up another index on `yb_hash_code` on a different column if such index exists.

Example:
```
CREATE TABLE hk_pagination(
    h int,
    r1 int,
    r2 int,
    v text,
    PRIMARY KEY(h HASH, r1 ASC, r2 ASC)
);
INSERT INTO hk_pagination SELECT i % 5, i, i * 10, 'val' || i FROM generate_series(1, 20) i;
create index hk_pagination_r1_r2_idx on hk_pagination (r1 hash, r2 asc);
/*+ IndexScan(hk_pagination hk_pagination_r1_r2_idx) */
EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
    WHERE yb_hash_code(h) = 49348
    ORDER BY r1, r2
    LIMIT 2;
                           QUERY PLAN                            
-----------------------------------------------------------------
 Limit
   ->  Index Scan using hk_pagination_r1_r2_idx on hk_pagination
         Filter: (yb_hash_code(h) = 49348)
(3 rows)

/*+ IndexScan(hk_pagination hk_pagination_r1_r2_idx) */
SELECT * FROM hk_pagination
    WHERE yb_hash_code(h) = 49348
    ORDER BY r1, r2
    LIMIT 2;
 h | r1 | r2  |   v   
---+----+-----+-------
 2 | 12 | 120 | val12
 2 |  7 |  70 | val7
(2 rows)

/*+ SeqScan(hk_pagination) */
EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
    WHERE yb_hash_code(h) = 49348
    ORDER BY r1, r2
    LIMIT 2;
                   QUERY PLAN                    
-------------------------------------------------
 Limit
   ->  Sort
         Sort Key: r1, r2
         ->  Seq Scan on hk_pagination
               Filter: (yb_hash_code(h) = 49348)
(5 rows)

/*+ SeqScan(hk_pagination) */
SELECT * FROM hk_pagination
    WHERE yb_hash_code(h) = 49348
    ORDER BY r1, r2
LIMIT 2;
 h | r1 | r2 |  v   
---+----+----+------
 2 |  2 | 20 | val2
 2 |  7 | 70 | val7
(2 rows)
```

mtakahar (changes requested):
Great work, and thank you so much for putting this together!

I've left a couple of inline comments, one correctness issue and another one, follow-up on what cursor bot pointed out (code duplication).

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1636,24 +1658,59 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 	 */
-
-	for (int i = 0; (i < index->rd_index->indnkeyatts) && can_pushdown; i++)
+	if (can_pushdown)
 	{
-		if (index->rd_indoption[i] & INDOPTION_HASH)
+		bool	has_hash_code_eq = yb_has_hash_code_equality_scan_key(ybScan);
+		int		hash_cols_covered = 0;
+
+		for (int i = 0; i < index->rd_index->indnkeyatts; i++)
```
egladysh: nit: really tiny optimization `i++`  to `++i` but maybe it just matches the rest of the code.

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -4828,6 +4838,92 @@ is_hash_column_in_lsm_index(const IndexOptInfo *index, int columnIndex)
...
+		if (!IsA(rinfo->clause, OpExpr))
+			continue;
+
+		opexpr = (OpExpr *) rinfo->clause;
+
+		if (opexpr->opno != Int4EqualOperator)
+			continue;
+
+		if (list_length(opexpr->args) != 2)
```
egladysh: nit: ideally this should be like an assert in debug build.

egladysh (approved):

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1636,24 +1658,59 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 		}
 	}
 
 	/*
-	 * Make sure that the primary key has no hash columns in order
-	 * to push down.
+	 * Make sure that the primary key has no hash columns in order to push
+	 * down.  When pinned to a single hash bucket via yb_hash_code() = const,
+	 * we can relax this only if the ROW comparison subkeys actually include
+	 * all hash columns — otherwise EncodeRowKeyForBound would receive NULL
```
jasonyb: Do not use non-ASCII characters.

src/postgres/src/backend/optimizer/path/indxpath.c
jasonyb, line 3840: note: it appears this planner code doesn't care about the order of the columns in the row compare, but I believe (could be wrong) the executor yb_scan.c code does care and does not support pushing down the filter in out-of-order cases (e.g. key (r1, r2) and row compare (r2, r1) op (a, b)).

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -213,6 +213,7 @@ static bool ec_member_matches_indexcol(PlannerInfo *root, RelOptInfo *rel,
 
 /* YB declarations */
 static bool is_hash_column_in_lsm_index(const IndexOptInfo *index, int columnIndex);
+/* yb_has_hash_code_equality_qual is declared in paths.h */
```
jasonyb: remove unnecessary comment

src/postgres/src/backend/optimizer/path/pathkeys.c
```diff
@@ -619,6 +627,11 @@ build_index_pathkeys(PlannerInfo *root,
 		 * OK, try to make a canonical pathkey for this sort key.  Note we're
 		 * underneath any outer joins, so nullable_relids should be NULL.
 		 */
+		/*
```
jasonyb:
```suggestion
		 *
```

src/postgres/src/backend/optimizer/path/pathkeys.c
```diff
@@ -671,7 +691,7 @@ build_index_pathkeys(PlannerInfo *root,
 			 * keys won't be useful either.
 			 */
 			if (!indexcol_is_bool_constant_for_query(root, index, i) ||
-				yb_is_hash_column)
+				(yb_is_hash_column && !yb_hash_bucket_pinned))
```
jasonyb, line 694: note: I don't think this whole clause is unnecessary, but better to take that investigation separately

src/postgres/src/backend/optimizer/path/pathkeys.c
```diff
@@ -699,11 +719,15 @@ build_index_pathkeys(PlannerInfo *root,
...
-	if (i < index->nhashcolumns)
+	if (i < index->nhashcolumns && !yb_hash_bucket_pinned)
 	{
 		/*
 		 * All hash columns must have EQ pathkeys. Otherwise, we cannot use
-		 * the index
+		 * the index.
+		 *
+		 * YB: Exception when pinned to a single hash bucket via
+		 * yb_hash_code(hash_cols) = const — hash columns provide valid
```
jasonyb: ASCII

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1590,6 +1590,28 @@ YbCheckScanTypes(YbScanDesc ybScan, YbScanPlan scan_plan, int i)
...
+static bool
+yb_has_hash_code_equality_scan_key(YbScanDesc ybScan)
+{
+	ListCell   *lc;
+
+	foreach(lc, ybScan->hash_code_keys)
+	{
+		ScanKey		key = (ScanKey) lfirst(lc);
+
+		if (key->sk_strategy == BTEqualStrategyNumber)
```
jasonyb, line 1609: note: it is possible to derive equality in case of hash_code > x and hash_code <= x + 1, for example, but that could be done separately (if worth doing at all)

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1636,24 +1658,59 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+						break;
+					}
+				}
+				if (found)
+				{
+					hash_cols_covered++;
+					continue;
+				}
+			}
+			/* Hash column not covered or no hash code equality → reject. */
```
jasonyb: ASCII

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1636,24 +1658,59 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+	if (can_pushdown)
 	{
-		if (index->rd_indoption[i] & INDOPTION_HASH)
+		bool	has_hash_code_eq = yb_has_hash_code_equality_scan_key(ybScan);
+		int		hash_cols_covered = 0;
+
+		for (int i = 0; i < index->rd_index->indnkeyatts; i++)
+		{
+			if (!(index->rd_indoption[i] & INDOPTION_HASH))
+				continue;
```
jasonyb: could break as it is not possible for there to be any more hash columns after the first range column.  could put this condition in the for loop

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1756,8 +1813,15 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 				int			att_idx = YBAttnumToBmsIndex(((TableScanDesc) ybScan)->rs_rd,
 														 subkeys[subkey_index]->sk_attno);
 
-				/* Set the first column in this RC to not null. */
-				is_not_null[att_idx] |= subkey_index == 0;
+				/*
+				 * Set the first column in this RC to not null, except for
+				 * partition/hash columns. Pggate does not allow binding
+				 * IS NOT NULL on partition columns, and for primary-key hash
```
jasonyb: what does "partition" mean?

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1756,8 +1813,15 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+				 * columns the null check is redundant anyway.
```
jasonyb: hash primary key cannot have null, but hash index key can

src/postgres/src/test/regress/sql/yb.orig.hash_code_pagination.sql
```diff
@@ -0,0 +1,109 @@
...
+    WHERE yb_hash_code(h1, h2) = 500
+    ORDER BY h1, h2, r
+    LIMIT 10;
+
+-- Negative: arity mismatch (partial hash key spec) should still Sort.
+EXPLAIN (COSTS OFF) SELECT * FROM hk_multi
+    WHERE yb_hash_code(h1) = 500
+    ORDER BY h1, h2, r
+    LIMIT 10;
+
```
jasonyb, line 106: there is a lack of tests involving

- secondary indexes
- nulls
- multi-column hash + row compare

jasonyb (changes requested):
Don't want to expand the scope too much, but do you mind checking what happens when `ybctid = const`?  That case is very similar to `yb_hash_code() = const`.

src/postgres/src/test/regress/sql/yb.orig.hash_code_pagination.sql
kirs: Added the tests.

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+ Limit
+   ->  Index Scan using hk_pagination_pkey on hk_pagination
+         Index Cond: (yb_hash_code(h) = 1000)
+(3 rows)
+
+-- Negative: range predicate on yb_hash_code should still Sort (multiple buckets).
+EXPLAIN (COSTS OFF) SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) < 1000
+    ORDER BY h, r1, r2
+    LIMIT 10;
```
jasonyb, line 35: note: this is not ready in the near future, but `yb_hash_code(h) IN (0, 1, ..., 999)` + `yb_max_saop_merge_streams = 1024` + `yb_enable_cbo` + pg_hint_plan hints to force SAOP merge + yb_hash_code IN-clause Index Cond support would be able to avoid Sort node

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+-- Verify ROW comparison pagination produces correct results.
+-- First page:
+SELECT * FROM hk_pagination
+    WHERE yb_hash_code(h) = 1000
+      AND (h, r1, r2) > (0, 0, 0)
+    ORDER BY h, r1, r2
+    LIMIT 5;
+ h | r1 | r2 | v 
+---+----+----+---
+(0 rows)
```
jasonyb, line 123: these are useless as no rows are returned.  pick better numbers or load more data

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+   ->  Sort
+         Sort Key: r1, r2
+         ->  Index Scan using hk_pagination_sec_idx on hk_pagination
+               Storage Filter: (yb_hash_code(h) = 1000)
+(5 rows)
+
+--
+-- ROW comparison on range-only columns of a hash-sharded index.
+-- The hash column is NOT in the ROW comparison, so pushdown must NOT
+-- happen (EncodeRowKeyForBound needs all hash columns in the bound).
```
jasonyb, line 235: so what happens if you omit a range column, such as r1?

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+------------------------------------------------------------
+ Limit
+   ->  Index Scan using hk_pagination_pkey on hk_pagination
+         Index Cond: (yb_hash_code(h) = 1000)
+         Filter: (ROW(r1, r2) > ROW(5, 50))
+(4 rows)
+
+--
+-- NULL handling tests.
+-- NULLs hash to a different bucket, so yb_hash_code(col) = const already
```
jasonyb, line 252: What is "different bucket"?  Isn't there at least one bucket they hash to?  Wouldn't it be best to test using that one specific const rather than others?  That same const should include rows with nulls and rows without nulls.

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+-- excludes them.  Verify queries work correctly with NULL data.
+--
+INSERT INTO hk_pagination VALUES (NULL, 100, 1000, 'null_h');
+INSERT INTO hk_pagination VALUES (1, NULL, 1000, 'null_r1');
+INSERT INTO hk_pagination VALUES (1, 100, NULL, 'null_r2');
+-- Verify NULLs do not appear in a hash-code-pinned scan.
```
jasonyb, line 258: misleading comment if you read it as a matter of fact, not as specific to this table, primary culprit being use of "a"

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+(4 rows)
+
+--
+-- NULL handling tests.
+-- NULLs hash to a different bucket, so yb_hash_code(col) = const already
+-- excludes them.  Verify queries work correctly with NULL data.
+--
+INSERT INTO hk_pagination VALUES (NULL, 100, 1000, 'null_h');
+INSERT INTO hk_pagination VALUES (1, NULL, 1000, 'null_r1');
+INSERT INTO hk_pagination VALUES (1, 100, NULL, 'null_r2');
```
jasonyb, line 257: shouldn't these fail because PKs cannot contain nulls?  did you run the test before pushing?

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+    ORDER BY h, r1, r2;
+ h | r1 | r2  |    v    
+---+----+-----+---------
+ 1 |  1 |  10 | val1
+ 1 |  6 |  60 | val6
+ 1 | 11 | 110 | val11
+ 1 | 16 | 160 | val16
+(4 rows)
+
+-- Secondary index with NULLs: secondary hash indexes allow NULL values.
```
jasonyb, line 285: > secondary hash indexes allow NULL values

then why did you have inserts of nulls to PKs in the above queries?

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+-- yb_hash_code(col) works when col is NULL (function is non-strict).
```
jasonyb, line 286: > yb_hash_code(col) works when col is NULL (function is non-strict).

I see no null involved in the following query

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
```diff
@@ -0,0 +1,358 @@
...
+ 1 | val11
+ 1 | val16
+ 1 | val6
+(6 rows)
+
+-- Secondary hash index with NULLs and ROW comparison pushdown.
+-- Secondary hash indexes allow NULL values, and yb_hash_code(NULL::type)
+-- returns a valid hash code, so NULL rows live in a real bucket.
+-- The IS NOT NULL guard is skipped for hash columns (pggate rejects it),
+-- but SQL-layer recheck excludes NULLs because NULL comparisons yield NULL.
```
jasonyb, line 314: seems like you have a lot of null tests above that are disorganized.  Can all tests be more sensibly organized as follows:

- pk_table (h1, h2, r1, r2, v), PRIMARY KEY ((h1, h2) HASH, r1 ASC, r2 ASC)
- ik_table (h1, h2, r1, r2, v), UNIQUE INDEX KEY ((h1, h2) HASH, r1 ASC, r2 ASC)
- Skip single-column tests and go straight to multi-column
- Incorporate NULLS at the very beginning of data load: pk_table and ik_table have the same data, but ik_table additionally has extra NULL key columns.  Ensure that for yb_hash_code = const values you test, there are both NULL and non-NULL rows matching that constant.
- For each query you want to test, you want to test both pk_table and ik_table.  No need for pg_hint_plan hints as both tables have only the choice between index/primary key scan and sequential scan.  You also want to show both explain plan and results.  I suggest you use `:explain1run1` on both table names like

```
\getenv abs_srcdir PG_ABS_SRCDIR                                                                                                                                                                                                                                                                                                                                             
\set filename :abs_srcdir '/yb_commands/explainrun.sql'                                                                                                                                                                                                                                                                                                           
\i :filename

\set explain 'EXPLAIN (COSTS OFF)'
\set query 'SELECT * FROM :table_name
    WHERE yb_hash_code(h1, h2) = <some_meaningful_constant>
    ORDER BY h1, h2;

\set table_name 'pk_table'
:explain1run1
\set table_name 'ik_table'
:explain1run1
```

Since I see this pattern come up a few times, it may be worth augmenting the explainrun framework to generalize the "hints" system to "params": define `\set explain1 ':explain :hint1 :query;'` instead as `\set explain1 ':explain :query;'` and make the responsibility of `query` to have hint as param at the front of the query.  I can take care of this, so you can in the meantime use the above slight duplication of back-and-forth set table_name.

src/postgres/src/backend/access/yb_access/yb_scan.c
jasonyb: commit cd0d8589a788fc79f78822cef5448ee330a80083 never explains why IS NOT NULL was not supported for hash.  Would have been good to know that.

jasonyb (commented):
I'm going to do another test run.  Hope you saw the previous test results I posted and fixed those issues.

```
>>> Lint for src/postgres/src/backend/access/yb_access/yb_scan.c:


   Error  (S&RX) bad_variable_declaration_spacing
    Variable declarations should align variable names to the 12 column mark

            1684         */
            1685        if (can_pushdown)
            1686        {
    >>>     1687                bool    has_hash_code_eq = yb_has_hash_code_equality_scan_key(ybScan);<TAB><TAB>bool<TAB>has_hash_code_eq = yb_has_hash_code_equality_scan_key(ybScan);
            1688 
            1689                for (int i = 0;
            1690                         i < index->rd_index->indnkeyatts &&

   Error  (S&RX) bad_variable_declaration_spacing
    Variable declarations should align variable names to the 12 column mark

            1693                {
            1694                        if (has_hash_code_eq)
            1695                        {
    >>>     1696                                bool    found = false;<TAB><TAB><TAB><TAB>bool<TAB>found = false;
            1697                                for (int j = 0; j < subkey_count; j++)
            1698                                {
            1699                                        if (subkeys[j]->sk_attno - 1 == i)
>>> Lint for src/postgres/src/backend/utils/misc/pg_yb_utils.c:


   Error  (S&RX) bad_variable_declaration_spacing
    Variable declarations should align variable names to the 12 column mark

            8922                col = 0;
            8923                foreach(arg_lc, funcexpr->args)
            8924                {
    >>>     8925                        Node   *arg = (Node *) lfirst(arg_lc);<TAB><TAB><TAB>Node   *arg = (Node *) lfirst(arg_lc);
            8926 
            8927                        if (IsA(arg, RelabelType))
            8928                                arg = (Node *) ((RelabelType *) arg)->arg;
 LINT ERRORS  Lint raised errors!
```

jasonyb:
Couple things:

- explainrun rewrite is in the works, but it shouldn't be a blocker as you can just duplicate queries for each table in the meantime.
- there is a lack of tests using inequality conditions on hash columns. Only ORDER BY is used. Is inequality not supported from planner side, or both ORDER BY and inequality are supported?
- Any tests for backwards scan? ORDER BY hash columns DESC.
- Some test results are concerning. for example,

```
 SELECT h, v FROM hk_pagination
     WHERE yb_hash_code(h) = yb_hash_code(1)
     ORDER BY h, v;
- h |    v    
----+---------
- 1 | null_r1
- 1 | null_r2
+ h |   v   
+---+-------
  1 | val1
  1 | val11
  1 | val16
  1 | val6
-(6 rows)
+(4 rows)
```

src/postgres/src/backend/access/yb_access/yb_scan.c
kirs: Do you see this as a requirement for this PR or as something that can be done as a follow-up? Sorry, I have very little context on https://github.com/yugabyte/yugabyte-db/commit/cd0d8589a788fc79f78822cef5448ee330a80083.

kirs:
@andrei-mart thanks for your feedback! Here's the main issue that I'm after:

```
SELECT * FROM foo.orders
WHERE yb_hash_code(user_uuid) = 10
  AND (user_uuid, uuid) > ('01961f1f-...', '01968344-...')
LIMIT 5000;
Condition lands in Filter (post-filter on the PG side, after rows are returned from storage):

Index Scan using orders_pkey on orders (actual time=256.437..269.259 rows=5000)
      Index Cond: (yb_hash_code(user_uuid) = 10)
      Filter: (ROW(user_uuid, uuid) > ROW('01961f1f-...'::uuid, '01968344-...'::uuid))
      Rows Removed by Filter: 90001
      Storage Table Read Requests: 93
      Storage Table Rows Scanned: 95232
Execution Time: 270.568 ms
```

Even without `ORDER BY`, this still has to scan 90k rows to get to 5k in the middle of the LSM, making pagination over large tables perform very poorly.

Sort elimination is nice to have but only as long as we have a way to paginate over a table with `yb_hash_code` efficiently.

Btw it's the same issue if I expand that into `WHERE yb_hash_code(user_uuid) = 10 AND (user_uuid > '01961f1f-...'  OR (user_uuid = '01961f1f-...' AND uuid > '01968344-...'))`. I've outlined this in more detail in https://github.com/yugabyte/yugabyte-db/issues/30524#issuecomment-4014889832.

It sounds like `ROW( yb_hash_code(h1), h1, r1) < ( yb_hash_code(constant1), constant2, constant3)` is exactly the feature we're after for that sort of pagination.

I'm all up for splitting work into multiple PRs. Let me know what you think is best given the context above.

src/postgres/src/test/regress/expected/yb.orig.hash_code_pagination.out
jasonyb: Commit d258233f964b5743dc59c46e4fa59fc0dab3b9ce lets you parameterize tablename

andrei-mart:
Welcome back @kirs!
In meantime I finished fixing PgDmlRead::EncodeRowKeyForBound, it is under review new, hopefully it will be merged soon. In meantime the patch I shared before should work to test Postgres side changes, if needed.

Yeah, you will be able to distribute the work between multiple routines.
You can assign one hash bucket at a time by adding a `yb_hash_code(h) = const` condition, or a range of hash buckets like `yb_hash_code(h) BETWEEN const1 AND const2`.
However, I have realized that we are missing a feature required for pagination to work properly: support for `ORDER BY yb_hash_code(h)`. It works correctly today, but planner does not understand that yb_hash_code(h) is the hash code, which is a valid index key. It is closely related to the problem you pointed out from the beginning:  `yb_hash_code(h) = const` condition should tell the planner that index is ordered by the hash key. The thing is, if planner knows, that the hash code represented by the `yb_hash_code(h)` expression is a sort key, and the hash columns follow it. With that knowledge Postgres would understand, that `ORDER BY yb_hash_code(h)` makes the hash order desired, and `WHERE yb_hash_code(h) = const` makes it redundant.

So I've started looking at that ORDER BY yb_hash_code(h). I've taken out the code to match yb_hash_code expression to an index from the match_index_to_operand() into a separate function, it is supposed to be used in multiple places. Then I took whole yb_hash_code handling out of the loop over the index columns in the match_clause_to_index(), as it is not related to the columns. I'm taking pause here to run the regression tests. I uploaded my WIP patch here: https://gist.github.com/andrei-mart/3836fc3ca0f3322be35537fd68a98da2 so you can take look. I think the yb_match_clause_to_index the patch adds is a good place to identify the "special" row condition. "Special" here should mean a row condition where the first element is a yb_hash_code matching the index. I can hand it over to you. But if we both are about to work on these items, we will need to synchronize somehow.

kirs:
Closing as the work is now happening in https://github.com/yugabyte/yugabyte-db/pull/31035

PR #30818, opened 2026-03-25
[DOC-1238] pg_stat_monitor and major upgrade

ddhodge:
@netlify /stable/manage/ysql-major-upgrade-yugabyted/

### DIFF
--- a/docs/content/v2025.1/manage/ysql-major-upgrade-local.md
+++ b/docs/content/v2025.1/manage/ysql-major-upgrade-local.md
@@ -60,6 +60,8 @@ Performing a YSQL major upgrade on a universe with [CDC with logical replication
 
 - If you have PITR enabled, delete the configuration before performing the upgrade. Recreate it only after the major upgrade is either finalized or rolled back.
 
+- Drop the `pg_stat_monitor` extension before upgrading (`DROP EXTENSION pg_stat_monitor;`) and re-enable it after the upgrade is finalized (`CREATE EXTENSION pg_stat_monitor;`).
+
 ### Precheck
 
 New PostgreSQL major versions add many new features and performance improvements, but also remove some older unsupported features and data types. You can only upgrade after you remove all deprecated features and data types from your databases.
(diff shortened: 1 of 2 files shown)


docs/content/stable/manage/ysql-major-upgrade-yugabyted.md
```diff
@@ -60,6 +60,8 @@ Performing a YSQL major upgrade on a universe with [CDC with logical replication
 
 - If you have PITR enabled, delete the configuration before performing the upgrade. Recreate it only after the major upgrade is either finalized or rolled back.
 
+- Drop the `pg_stat_monitor` extension before upgrading (`DROP EXTENSION pg_stat_monitor;`) and re-enable it after the upgrade is finalized (`CREATE EXTENSION pg_stat_monitor;`).
```
fizaaluthra, line 63: it feels a bit weird to call out `pg_stat_monitor` specifically imo

We have an extension upgrade check (as part of the precheck listed below) that checks for whitelisted extensions. Example:

```
./build/latest/postgres/bin/pg_upgrade --check  -U yugabyte --old-socketdir /tmp/.yb.127.0.0.1:5433 --old-port 5433 --old-datadir ~/yugabyte-data/node-1/disk-1/pg_data_11
Performing Consistency Checks on Old Live Server
------------------------------------------------
Checking cluster versions                                   ok
Checking attributes of the 'yugabyte' user                  ok
Checking for all 3 system databases                         ok
Checking database connection settings                       ok
Checking for system-defined composite types in user tables  ok
Checking for reg* data types in user tables                 ok
Checking for user-defined postfix operators                 ok
Checking for incompatible polymorphic functions             ok
Checking for invalid "sql_identifier" user columns          ok
Checking installed extensions                               fatal

In database: yugabyte
  pg_stat_monitor

Your installation contains extensions that are not compatible
with YSQL major version upgrade. Please uninstall the extensions
using DROP EXTENSION, and reinstall them after the upgrade. A list of
extensions with the problem is printed above and in the file:
    /Users/fizaaluthra/yugabyte-data/node-1/disk-1/pg_data_11/pg_upgrade_output.d/20250305T155850.551/unsupported_extensions.txt



Failure, exiting
```

Maybe we can make a more general point under the precheck section below that any incompatible extensions (example: pg_stat_monitor) should be dropped

PR #30907, opened 2026-04-01
[#32161] CDCSDK: Allow concurrent sessions to tag writes with same origin 

### LINKED ISSUE #32161: [CDCSDK] Allow concurrent sessions to tag writes with the same replication origin
Jira Link: [DB-21849](https://yugabyte.atlassian.net/browse/DB-21849)
### Description

CDC origin filtering needs a way for multiple concurrent writer sessions to tag their writes with the same replication origin ID.

Today, `pg_replication_origin_session_setup(text)` acquires an exclusive replication-origin progress slot for the session. That is appropriate for logical replication replay, where the session must maintain replication progress, but it is too restrictive for CDC write tagging. Multiple sessions may need to mark writes with the same `xrepl_origin_id` so downstream CDC consumers can identify and filter those writes, without requiring each session to own an exclusive progress-tracking slot.

This blocks concurrent CDC write-tagging workflows because only one session can use a given replication origin through the existing exclusive session setup path at a time.

Add shared replication-origin session helpers for the write-tagging use case:

- `yb_replication_origin_session_setup_shared(text)`
- `yb_replication_origin_session_reset_shared()`

Expected behavior:

- Multiple sessions can concurrently set the same replication origin for write tagging.
- Writes made while a shared origin is active carry the expected `xrepl_origin_id` in CDC records.
- Shared-origin sessions are for tagging only and do not update replication-origin progress.
- Shared setup/reset fail safely when the session already has incompatible or duplicate origin state.
- Access is restricted consistently with existing `pg_replication_origin_*` functions.

Related PR: https://github.com/yugabyte/yugabyte-db/pull/30907

### Source connector version

N/A

### Connector configuration

N/A

### YugabyteDB version

master

### Issue Type

kind/new-feature

[DB-21849]: https://yugabyte.atlassian.net/browse/DB-21849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kgalieva:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/30907/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

## Summary

Adds YB-specific shared replication-origin session helpers for CDCSDK write tagging:

- `yb_replication_origin_session_setup_shared(text)`
- `yb_replication_origin_session_reset_shared()`

These functions let multiple sessions tag writes with replication origins concurrently without acquiring exclusive replication-origin progress slots. The tagged writes carry `xrepl_origin_id` in CDC records, while replication-origin progress tracking remains exclusive-session behavior.

Fixes #32161.

## Problem

`pg_replication_origin_session_setup(text)` acquires an exclusive replication-origin progress slot in shared memory. That is required for logical-replication replay because the session advances replay progress, but it is too restrictive for CDCSDK write tagging.

For CDC origin filtering, multiple writer sessions may need to tag their writes with the same origin ID so downstream CDC consumers can identify and filter those writes. They do not need to own or advance the origin's replication progress.

With the existing exclusive setup path, only one session can use a given origin at a time. Other sessions fail with:

```text
ERROR: replication origin with ID <origin_id> is already active for PID <other_pid> (SQLSTATE 55006)
```

## Solution

Add shared-origin session helpers for the write-tagging use case:

### `yb_replication_origin_session_setup_shared(text)`

Sets the current session's replication origin for CDC write tagging without acquiring an exclusive progress slot.

Behavior:

- Validates that the origin exists.
- Sets the backend-local `replorigin_session_origin` used to stamp `xrepl_origin_id` on writes.
- Does not call `replorigin_session_setup()` and does not acquire a shared-memory progress slot.
- Allows multiple sessions to use the same origin concurrently.
- Rejects duplicate setup in the same session.
- Rejects setup if an exclusive replication-origin session is already active.
- Keeps YSQL connection-manager sessions sticky while shared-origin state is active.

### `yb_replication_origin_session_reset_shared()`

Clears shared-origin state for the current session without touching shared-memory progress slots.

Behavior:

- Rejects reset if no replication origin is active in the session.
- Rejects reset if an exclusive replication-origin session is active; callers must use `pg_replication_origin_session_reset()` for that case.
- Clears origin LSN/timestamp session state along with the origin ID.

## Guardrails

- Shared-origin behavior is gated by default-on GUC `yb_enable_replication_origin_shared`.
- When the GUC is disabled, both shared setup/reset functions error without changing session state.
- Existing `pg_replication_origin_session_setup()` / `pg_replication_origin_session_reset()` remain the exclusive progress-tracking APIs.
- Shared-to-exclusive, exclusive-to-shared, shared-to-shared, and reset mismatch cases fail cleanly instead of silently clobbering session state.

## Catalog And Upgrade

- Adds catalog entries for OIDs `8894` and `8895`.
- Adds `REVOKE EXECUTE` entries so the new functions are not executable by `PUBLIC`, matching existing `pg_replication_origin_*` functions.
- Adds YSQL migration `V104__30907__pg_replication_origin_shared.sql` so upgraded clusters get the new functions, descriptions, and initial privileges.
- Migration rows set `proisstrict = true`, matching fresh initdb catalog behavior.

## Usage

```sql
-- Create the origin once.
SELECT pg_replication_origin_create('my_origin');

-- On each writer session. Safe for multiple sessions concurrently.
SELECT yb_replication_origin_session_setup_shared('my_origin');

-- Subsequent writes on this session carry xrepl_origin_id.
INSERT INTO ...;

-- Clear the origin for this session.
SELECT yb_replication_origin_session_reset_shared();
```

### DIFF
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgReplicationSlot.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgReplicationSlot.java
@@ -5058,7 +5058,8 @@ public class TestPgReplicationSlot extends BasePgSQLTest {
     final String originId2 = "shared_origin2";
 
     Connection replConn = getConnectionBuilder().withTServer(0).replicationConnect();
-    PGReplicationConnection replConnection = replConn.unwrap(PGConnection.class).getReplicationAPI();
+    PGReplicationConnection replConnection =
+        replConn.unwrap(PGConnection.class).getReplicationAPI();
     Connection conn2 = getConnectionBuilder().withTServer(1).connect();
 
     try (Statement stmt1 = connection.createStatement();
@@ -5109,6 +5110,7 @@ public class TestPgReplicationSlot extends BasePgSQLTest {
 
       stmt1.execute("SELECT yb_replication_origin_session_reset_shared()");
       stmt2.execute("SELECT yb_replication_origin_session_reset_shared()");
+      stmt1.execute("INSERT INTO shared_origin_tbl1 values (3, 'local_after_reset')");
 
       stmt1.execute("SELECT pg_replication_origin_session_setup('" + originId1 + "')");
       runInvalidQuery(
@@ -5131,10 +5133,14 @@ public class TestPgReplicationSlot extends BasePgSQLTest {
                                      .withSlotOption("publication_names", "shared_origin_pub")
                                      .start();
 
-    List<PgOutputMessage> result = receiveMessage(stream, 10);
+    List<PgOutputMessage> result = receiveMessage(stream, 13);
+    int originMessages = 0;
     int origin1Messages = 0;
     int origin2Messages = 0;
     for (PgOutputMessage message : result) {
+      if (message.messageType() == PgOutputMessageType.ORIGIN) {
+        originMessages++;
+      }
       if (message.equals(PgOutputOriginMessage.Create(originId1))) {
         origin1Messages++;
       }
@@ -5142,6 +5148,7 @@ public class TestPgReplicationSlot extends BasePgSQLTest {
         origin2Messages++;
       }
     }
+    assertEquals(2, originMessages);
     assertEquals(1, origin1Messages);
     assertEquals(1, origin2Messages);
 
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgReplicationSlot.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgReplicationSlot.java
@@ -5052,6 +5052,126 @@ public class TestPgReplicationSlot extends BasePgSQLTest {
   }
 
   @Test
+  public void testPgoutputWithSharedOriginId() throws Exception {
+    final String slotName = "test_replication_with_shared_origin_id";
+    final String originId1 = "shared_origin1";
+    final String originId2 = "shared_origin2";
+
+    Connection replConn = getConnectionBuilder().withTServer(0).replicationConnect();
+    PGReplicationConnection replConnection = replConn.unwrap(PGConnection.class).getReplicationAPI();
+    Connection conn2 = getConnectionBuilder().withTServer(1).connect();
+
+    try (Statement stmt1 = connection.createStatement();
+         Statement stmt2 = conn2.createStatement()) {
+      stmt1.execute("DROP TABLE IF EXISTS shared_origin_tbl1");
+      stmt1.execute("DROP TABLE IF EXISTS shared_origin_tbl2");
+      stmt1.execute("CREATE TABLE shared_origin_tbl1 (a INT PRIMARY KEY, b TEXT)");
+      stmt1.execute("ALTER TABLE shared_origin_tbl1 REPLICA IDENTITY DEFAULT");
+      stmt1.execute("CREATE TABLE shared_origin_tbl2 (a INT PRIMARY KEY, b TEXT)");
+      stmt1.execute("ALTER TABLE shared_origin_tbl2 REPLICA IDENTITY DEFAULT");
+      stmt1.execute("CREATE PUBLICATION shared_origin_pub FOR ALL TABLES");
+
+      stmt1.execute("SELECT pg_replication_origin_create('" + originId1 + "')");
+      stmt1.execute("SELECT pg_replication_origin_create('" + originId2 + "')");
+      stmt1.execute("CREATE ROLE shared_origin_non_su LOGIN PASSWORD 'pwd'");
+      createSlot(replConnection, slotName, PG_OUTPUT_PLUGIN_NAME);
+
+      try (Connection nonSuConn = getConnectionBuilder()
+               .withUser("shared_origin_non_su").withPassword("pwd").connect();
+           Statement nonSuStmt = nonSuConn.createStatement()) {
+        runInvalidQuery(
+            nonSuStmt,
+            "SELECT yb_replication_origin_session_setup_shared('" + originId1 + "')",
+            "permission denied for function yb_replication_origin_session_setup_shared");
+        runInvalidQuery(
+            nonSuStmt,
+            "SELECT yb_replication_origin_session_reset_shared()",
+            "permission denied for function yb_replication_origin_session_reset_shared");
+      }
+
+      stmt1.execute("SELECT yb_replication_origin_session_setup_shared('" + originId1 + "')");
+      stmt2.execute("SELECT yb_replication_origin_session_setup_shared('" + originId2 + "')");
+
+      runInvalidQuery(
+          stmt1,
+          "SELECT pg_replication_origin_session_setup('" + originId2 + "')",
+          "replication origin session already setup");
+      runInvalidQuery(
+          stmt1,
+          "SELECT yb_replication_origin_session_setup_shared('" + originId2 + "')",
+          "replication origin session already setup");
+
+      stmt1.execute("CREATE TEMP TABLE shared_origin_temp(i int)");
+      stmt1.execute("INSERT INTO shared_origin_temp VALUES (1)");
+      stmt1.execute("DROP TABLE shared_origin_temp");
+      stmt1.execute("INSERT INTO shared_origin_tbl1 values (1, 'from_origin1')");
+      stmt2.execute("INSERT INTO shared_origin_tbl2 values (2, 'from_origin2')");
+
+      stmt1.execute("SELECT yb_replication_origin_session_reset_shared()");
+      stmt2.execute("SELECT yb_replication_origin_session_reset_shared()");
+
+      stmt1.execute("SELECT pg_replication_origin_session_setup('" + originId1 + "')");
+      runInvalidQuery(
+          stmt1,
+          "SELECT yb_replication_origin_session_setup_shared('" + originId2 + "')",
+          "cannot use shared replication origin session while an exclusive one is active");
+      runInvalidQuery(
+          stmt1,
+          "SELECT yb_replication_origin_session_reset_shared()",
+          "cannot reset shared replication origin session while an exclusive one is active");
+      stmt1.execute("SELECT pg_replication_origin_session_reset()");
+
+    }
+
+    PGReplicationStream stream = replConnection.replicationStream()
+                                     .logical()
+                                     .withSlotName(slotName)
+                                     .withStartPosition(LogSequenceNumber.valueOf(0L))
+                                     .withSlotOption("proto_version", 1)
+                                     .withSlotOption("publication_names", "shared_origin_pub")
+                                     .start();
+
+    List<PgOutputMessage> result = receiveMessage(stream, 10);
+    int origin1Messages = 0;
+    int origin2Messages = 0;
+    for (PgOutputMessage message : result) {
+      if (message.equals(PgOutputOriginMessage.Create(originId1))) {
+        origin1Messages++;
+      }
+      if (message.equals(PgOutputOriginMessage.Create(originId2))) {
+        origin2Messages++;
+      }
+    }
+    assertEquals(1, origin1Messages);
+    assertEquals(1, origin2Messages);
+
+    stream.close();
+    conn2.close();
+    replConn.close();
+  }
+
+  @Test
+  public void testSharedOriginIdDisabled() throws Exception {
+    Map<String, String> tserverFlags = getTServerFlags();
+    appendToYsqlPgConf(tserverFlags, "yb_enable_replication_origin_shared=false");
+    restartClusterWithFlags(getMasterFlags(), tserverFlags);
+
+    try (Statement stmt = connection.createStatement()) {
+      stmt.execute("SELECT pg_replication_origin_create('disabled_shared_origin')");
+      runInvalidQuery(
+          stmt,
+          "SELECT yb_replication_origin_session_setup_shared('disabled_shared_origin')",
+          "yb_replication_origin_session_setup_shared requires "
+              + "yb_enable_replication_origin_shared to be enabled");
+      runInvalidQuery(
+          stmt,
+          "SELECT yb_replication_origin_session_reset_shared()",
+          "yb_replication_origin_session_reset_shared requires "
+              + "yb_enable_replication_origin_shared to be enabled");
+    }
+  }
+
+  @Test
   public void testYboutputWithOriginIdCreatedAfterSlot() throws Exception {
     final String slotName = "test_replication_with_origin_id_post_slot";
     final String originId1 = "origin1_post";
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlUpgrade.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlUpgrade.java
@@ -1138,6 +1138,41 @@ public class TestYsqlUpgrade extends BasePgSQLTest {
     runMigrations(false /* useSingleConnection */);
   }
 
+  @SkipOnTSAN
+  @Test
+  public void sharedReplicationOriginMigration() throws Exception {
+    recreateWithYsqlVersion(YsqlSnapshotVersion.PG15_12);
+    createDbConnections();
+    runMigrations(false /* useSingleConnection */);
+
+    try (Connection conn = template1Cb.connect();
+         Statement stmt = conn.createStatement()) {
+      assertQuery(stmt,
+          "SELECT p.oid::int, p.proname, p.proisstrict, oidvectortypes(p.proargtypes), "
+              + "p.prosrc, p.proacl::text, d.description "
+              + "FROM pg_catalog.pg_proc p "
+              + "JOIN pg_catalog.pg_description d "
+              + "ON d.objoid = p.oid AND d.classoid = 'pg_proc'::regclass AND d.objsubid = 0 "
+              + "WHERE p.oid IN (8894, 8895) "
+              + "ORDER BY p.oid",
+          new Row(8894, "yb_replication_origin_session_setup_shared", true, "text",
+              "yb_replication_origin_session_setup_shared", "{postgres=X/postgres}",
+              "set session replication origin without exclusive lock (for concurrent write tagging)"),
+          new Row(8895, "yb_replication_origin_session_reset_shared", true, "",
+              "yb_replication_origin_session_reset_shared", "{postgres=X/postgres}",
+              "reset shared replication origin set by "
+                  + "yb_replication_origin_session_setup_shared"));
+
+      assertQuery(stmt,
+          "SELECT objoid::int, initprivs::text "
+              + "FROM pg_catalog.pg_init_privs "
+              + "WHERE objoid IN (8894, 8895) "
+              + "ORDER BY objoid",
+          new Row(8894, "{postgres=X/postgres}"),
+          new Row(8895, "{postgres=X/postgres}"));
+    }
+  }
+
   //
   // Helpers
   //
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlUpgrade.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlUpgrade.java
@@ -1138,41 +1138,6 @@ public class TestYsqlUpgrade extends BasePgSQLTest {
     runMigrations(false /* useSingleConnection */);
   }
 
-  @SkipOnTSAN
-  @Test
-  public void sharedReplicationOriginMigration() throws Exception {
-    recreateWithYsqlVersion(YsqlSnapshotVersion.PG15_12);
-    createDbConnections();
-    runMigrations(false /* useSingleConnection */);
-
-    try (Connection conn = template1Cb.connect();
-         Statement stmt = conn.createStatement()) {
-      assertQuery(stmt,
-          "SELECT p.oid::int, p.proname, p.proisstrict, oidvectortypes(p.proargtypes), "
-              + "p.prosrc, p.proacl::text, d.description "
-              + "FROM pg_catalog.pg_proc p "
-              + "JOIN pg_catalog.pg_description d "
-              + "ON d.objoid = p.oid AND d.classoid = 'pg_proc'::regclass AND d.objsubid = 0 "
-              + "WHERE p.oid IN (8894, 8895) "
-              + "ORDER BY p.oid",
-          new Row(8894, "yb_replication_origin_session_setup_shared", true, "text",
-              "yb_replication_origin_session_setup_shared", "{postgres=X/postgres}",
-              "set session replication origin without exclusive lock (for concurrent write tagging)"),
-          new Row(8895, "yb_replication_origin_session_reset_shared", true, "",
-              "yb_replication_origin_session_reset_shared", "{postgres=X/postgres}",
-              "reset shared replication origin set by "
-                  + "yb_replication_origin_session_setup_shared"));
-
-      assertQuery(stmt,
-          "SELECT objoid::int, initprivs::text "
-              + "FROM pg_catalog.pg_init_privs "
-              + "WHERE objoid IN (8894, 8895) "
-              + "ORDER BY objoid",
-          new Row(8894, "{postgres=X/postgres}"),
-          new Row(8895, "{postgres=X/postgres}"));
-    }
-  }
-
   //
   // Helpers
   //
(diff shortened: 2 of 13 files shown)


src/postgres/src/backend/replication/logical/origin.c
```diff
@@ -1376,6 +1376,42 @@ pg_replication_origin_session_setup(PG_FUNCTION_ARGS)
...
 }
 
+/*
+ * Setup a replication origin for this session without acquiring an exclusive
+ * slot. This allows multiple sessions to tag their writes with the same
+ * xrepl_origin_id concurrently. No progress tracking is performed — this is
+ * only suitable for write tagging (e.g., CDC origin filtering).
+ */
+Datum
+pg_replication_origin_session_setup_shared(PG_FUNCTION_ARGS)
```
shishir2001-yb: New functions are executable by PUBLIC. Every existing `pg_replication_origin_*` function is REVOKE EXECUTE ... FROM public in src/postgres/src/backend/catalog/system_functions.sql. The PR adds no REVOKE for the two new ones, so any unprivileged user can tag writes with an arbitrary origin

src/postgres/src/backend/replication/logical/origin.c
```diff
@@ -1393,6 +1429,32 @@ pg_replication_origin_session_reset(PG_FUNCTION_ARGS)
 	PG_RETURN_VOID();
 }
 
+/*
+ * Reset a shared replication origin previously set with
+ * pg_replication_origin_session_setup_shared(). Simply clears the
+ * session variable without touching shared-memory slots.
+ */
+Datum
+pg_replication_origin_session_reset_shared(PG_FUNCTION_ARGS)
```
shishir2001-yb: Add in `src/postgres/src/backend/catalog/system_functions.sql`

src/postgres/src/include/catalog/pg_proc.dat
```diff
@@ -11602,6 +11602,18 @@
   proparallel => 'u', prorettype => 'void', proargtypes => 'text',
   prosrc => 'pg_replication_origin_session_setup' },
 
+{ oid => '6015',
```
shishir2001-yb: OIDs 6015/6016. OIDs 6003–6014 are already used in the upstream PG range, so 6015/6016 are exactly what a future upstream patch will grab on the next PG rebase.

Replace with `8894`

src/postgres/src/include/catalog/pg_proc.dat
```diff
@@ -11602,6 +11602,18 @@
...
+  descr => 'set session replication origin without exclusive lock (for concurrent write tagging)',
+  proname => 'pg_replication_origin_session_setup_shared', provolatile => 'v',
+  proparallel => 'u', prorettype => 'void', proargtypes => 'text',
+  prosrc => 'pg_replication_origin_session_setup_shared' },
+
+{ oid => '6016',
```
shishir2001-yb: Replace with `8895`

src/postgres/contrib/test_decoding/sql/replorigin.sql
```diff
@@ -119,3 +119,34 @@ SELECT data FROM pg_logical_slot_get_changes('regression_slot_no_lsn', NULL, NUL
...
+
+-- setup again, then reset normally
+SELECT pg_replication_origin_session_setup_shared('regress_test_decoding: shared_origin');
+SELECT pg_replication_origin_session_reset_shared();
+
+-- session should no longer report origin as setup
+SELECT pg_replication_origin_session_is_setup();
+
+-- clean up
+SELECT pg_replication_origin_drop('regress_test_decoding: shared_origin');
```
shishir2001-yb: We also need to cover multi session. not a good fit for here.

src/postgres/src/backend/replication/logical/origin.c
```diff
@@ -1376,6 +1376,42 @@ pg_replication_origin_session_setup(PG_FUNCTION_ARGS)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_OBJECT_IN_USE),
+				 errmsg("replication origin session already setup")));
+
+	name = text_to_cstring((text *) DatumGetPointer(PG_GETARG_DATUM(0)));
+	origin = replorigin_by_name(name, false);
+
+	replorigin_session_origin = origin;
+	replorigin_session_origin_lsn = InvalidXLogRecPtr;
+	replorigin_session_origin_timestamp = 0;
```
shishir2001-yb: These two assignments are redundant and can be dropped. By the time we reach here, both are already InvalidXLogRecPtr/0. So whenever this function runs, the lsn/timestamp are already zero. It's also asymmetric with pg_replication_origin_session_setup(), which intentionally leaves them untouched

src/postgres/src/backend/replication/logical/origin.c
shishir2001-yb: We will need to consider this scenario and further explore. `pg_replication_origin_session_setup_shared` sets `replorigin_session_origin` without a slot, so a temp-table transaction's commit calls `replorigin_session_advance()` on a NULL `session_replication_state` (xact.c:1509 → origin.c:1236). Reachable via setup_shared('o') + temp-table insert. 


```
yugabyte=# SELECT pg_replication_origin_create('o');
 pg_replication_origin_create 
------------------------------
                            2
(1 row)

yugabyte=# SELECT pg_replication_origin_session_setup_shared('o');
 pg_replication_origin_session_setup_shared 
--------------------------------------------
 
(1 row)

yugabyte=# CREATE TEMP TABLE t(i int); INSERT INTO t VALUES (1);
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
The connection to the server was lost. Attempting reset: Failed.
You are currently not connected to a database.
```

src/postgres/src/include/catalog/pg_proc.dat
shishir2001-yb: Missing YSQL migration. Adding entries to `pg_proc.dat` only affects fresh initdb. Existing clusters which are upgraded. get nothing, the functions won't exist. YB requires a src/yb/yql/pgwrapper/ysql_migrations/V100__<issue>__...sql ) that INSERT … ON CONFLICT DO NOTHING into `pg_proc/pg_description` and sets proacl to match the REVOKE. 
Refer to: `src/yb/yql/pgwrapper/ysql_migrations/V95__29990__yb_get_local_tserver_uuid.sql`

src/postgres/contrib/test_decoding/expected/replorigin.out
```diff
@@ -252,8 +252,73 @@ SELECT pg_drop_replication_slot('regression_slot_no_lsn');
...
+ pg_replication_origin_session_is_setup
+----------------------------------------
+ f
+(1 row)
+
+-- clean up
+SELECT pg_replication_origin_drop('regress_test_decoding: shared_origin');
+ pg_replication_origin_drop
+----------------------------
+
```
shishir2001-yb: Have you verified this test is passing?

shishir2001-yb: I believe we need to add more tests to test concurrency. Add it in `src/yb/integration-tests/cdcsdk_ysql-test.cc`

src/postgres/src/include/catalog/pg_proc.dat
shishir2001-yb: When you re-build to verify your changes run  `./yb_build.sh initdb` otherwise your change won't reflect

shishir2001-yb (changes requested):
Please rebase before making changes

shishir2001-yb:
@kgalieva, is this ready for re-review? If so, could you please resolve the comments so it doesn't get confusing? Thanks.

src/postgres/src/backend/replication/logical/origin.c
kgalieva: Fixed. Added REVOKE EXECUTE entries for both pg_replication_origin_session_setup_shared(text) and pg_replication_origin_session_reset_shared() in src/postgres/src/backend/catalog/system_functions.sql.

Also added the YSQL migration with proacl/initprivs set to {postgres=X/postgres}, so upgraded clusters get the same restricted ACLs as fresh initdb clusters.

kgalieva: Fixed. replorigin_session_advance() now returns immediately when session_replication_state is NULL, which preserves shared-origin write tagging without attempting progress tracking through a non-existent slot.

I also added a temp-table write while a shared origin is active in CDCSDKYsqlTest.TestSharedOriginIdFromConcurrentSessions.

src/postgres/src/include/catalog/pg_proc.dat
kgalieva: Fixed. Added src/yb/yql/pgwrapper/ysql_migrations/V104__30907__pg_replication_origin_shared.sql.

The migration inserts both pg_proc rows with ON CONFLICT DO NOTHING, inserts pg_description rows, and inserts pg_init_privs rows. The proacl/initprivs values are {postgres=X/postgres} to match the REVOKE behavior for fresh initdb clusters.

src/postgres/contrib/test_decoding/expected/replorigin.out
kgalieva: I could not complete a direct contrib/test_decoding make check locally because the direct pg_regress harness fails during initdb in this checkout before running the tests.

I did verify the behavior through the targeted CDC integration test that now covers the shared-origin multi-session case and the temp-table write path

kgalieva: Done. Added CDCSDKYsqlTest.TestSharedOriginIdFromConcurrentSessions in src/yb/integration-tests/cdcsdk_ysql-test.cc.

The test sets up the same shared origin in two concurrent sessions, writes rows from both sessions, verifies both INSERT records carry the shared xrepl_origin_id, and also exercises a temp-table write while the shared origin is active.

kgalieva:
@shishir2001-yb I addressed the review threads, added the YSQL migration, restricted EXECUTE privileges for the new functions, and added CDC integration coverage for concurrent shared-origin sessions plus the temp-table write path. The targeted CDC test passed. Ready for another review 🙏

src/postgres/src/backend/replication/logical/origin.c
```diff
@@ -1233,6 +1233,10 @@ replorigin_session_reset(void)
 void
 replorigin_session_advance(XLogRecPtr remote_commit, XLogRecPtr local_commit)
 {
+	/* Shared origin sessions only tag writes and do not track progress. */
+	if (session_replication_state == NULL)
+		return;
+
 	Assert(session_replication_state != NULL);
```
shishir2001-yb, line 1240: This Assert is redundant now? Please remove it.

src/yb/yql/pgwrapper/ysql_migrations/V104__30907__pg_replication_origin_shared.sql
```diff
@@ -0,0 +1,33 @@
...
+  SET LOCAL yb_non_ddl_txn_for_sys_tables_allowed TO true;
+
+  INSERT INTO pg_catalog.pg_proc (
+    oid, proname, pronamespace, proowner, prolang, procost, prorows, provariadic,
+    prosupport, prokind, prosecdef, proleakproof, proisstrict, proretset,
+    provolatile, proparallel, pronargs, pronargdefaults, prorettype, proargtypes,
+    proallargtypes, proargmodes, proargnames, proargdefaults, protrftypes,
+    prosrc, probin, prosqlbody, proconfig, proacl) VALUES
+    (8894, 'pg_replication_origin_session_setup_shared', 11, 10, 12, 1, 0, 0,
+     '-', 'f', false, false, false, false, 'v', 'u', 1, 0, 2278, '25',
```
shishir2001-yb: The `proisstrict` value (13th column) is false for both rows, but fresh-initdb clusters get `proisstrict = true`,  `pg_proc.h` has `BKI_DEFAULT(t)` and the new pg_proc.dat entries don't override it. This causes
1.  a catalog mismatch between upgraded and fresh clusters (the TestYsqlUpgrade consistency check should catch this)
2. a backend crash on upgraded clusters: with strict=false, `SELECT pg_replication_origin_session_setup_shared(NULL)` calls the C function with a NULL datum and `text_to_cstring(NULL)` segfaults, the C code relies on strictness, same as the existing `pg_replication_origin_session_setup`. Please change both rows to ..., false, false, true, false, ... .

Please cover this in a test in  `TestYsqlUpgrade`

src/yb/yql/pgwrapper/ysql_migrations/V104__30907__pg_replication_origin_shared.sql
```diff
@@ -0,0 +1,33 @@
...
+     NULL, NULL, NULL, NULL, NULL,
+     'pg_replication_origin_session_setup_shared', NULL, NULL, NULL, '{postgres=X/postgres}'),
+    (8895, 'pg_replication_origin_session_reset_shared', 11, 10, 12, 1, 0, 0,
+     '-', 'f', false, false, false, false, 'v', 'u', 0, 0, 2278, '',
```
shishir2001-yb: Same comment as above

src/postgres/src/backend/replication/logical/origin.c
```diff
@@ -1393,6 +1431,32 @@ pg_replication_origin_session_reset(PG_FUNCTION_ARGS)
 	PG_RETURN_VOID();
 }
 
+/*
+ * Reset a shared replication origin previously set with
+ * pg_replication_origin_session_setup_shared(). Simply clears the
+ * session variable without touching shared-memory slots.
+ */
+Datum
+pg_replication_origin_session_reset_shared(PG_FUNCTION_ARGS)
```
shishir2001-yb: Given this is specific to YB, the function name should be `yb_replication_origin_session_reset_shared`. 
You will have to change it in the remaining places as well

src/postgres/src/backend/replication/logical/origin.c
```diff
@@ -1376,6 +1380,40 @@ pg_replication_origin_session_setup(PG_FUNCTION_ARGS)
...
 }
 
+/*
+ * Setup a replication origin for this session without acquiring an exclusive
+ * slot. This allows multiple sessions to tag their writes with the same
+ * xrepl_origin_id concurrently. No progress tracking is performed. This is
+ * only suitable for write tagging (e.g., CDC origin filtering).
+ */
+Datum
+pg_replication_origin_session_setup_shared(PG_FUNCTION_ARGS)
```
shishir2001-yb: Given this is specific to YB, the function name should be `yb_replication_origin_session_setup_shared`. 
You will have to change it in the remaining places as well

src/postgres/contrib/test_decoding/expected/replorigin.out
```diff
@@ -252,8 +252,73 @@ SELECT pg_drop_replication_slot('regression_slot_no_lsn');
 (1 row)
 
 SELECT pg_replication_origin_drop('regress_test_decoding: regression_slot_no_lsn');
- pg_replication_origin_drop 
+ pg_replication_origin_drop
 ----------------------------
- 
+
```
shishir2001-yb: YB doesn't run the `contrib/test_decoding` regress suite at all, our contrib testing only covers the 16 modules with a yb_schedule + TestPgRegressContrib* class, and test_decoding isn't one of them (we only use it as a slot output plugin in TestPgReplicationSlot). So these sql/ + expected/ additions can never execute in CI, can't be validated, and hand-editing the upstream-tracked `expected/replorigin.out` guarantees conflicts on the next PG rebase. Please drop the `contrib/test_decoding` changes from this PR. 

Instead add the remaining/required test in `TestPgReplicationSlot.java`

src/postgres/contrib/test_decoding/sql/replorigin.sql
```diff
@@ -119,3 +119,34 @@ SELECT data FROM pg_logical_slot_get_changes('regression_slot_no_lsn', NULL, NUL
 SELECT pg_replication_origin_session_reset();
 SELECT pg_drop_replication_slot('regression_slot_no_lsn');
 SELECT pg_replication_origin_drop('regress_test_decoding: regression_slot_no_lsn');
+
```
shishir2001-yb: Undo changes from this file, as discussed in above comment

src/postgres/contrib/test_decoding/expected/replorigin.out
shishir2001-yb: Explained above

src/postgres/src/backend/replication/logical/origin.c
```diff
@@ -1376,6 +1380,40 @@ pg_replication_origin_session_setup(PG_FUNCTION_ARGS)
...
+{
+	char	   *name;
+	RepOriginId origin;
+
+	replorigin_check_prerequisites(true, false);
+
+	if (session_replication_state != NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_OBJECT_IN_USE),
+				 errmsg("cannot use shared replication origin session while an exclusive one is active")));
```
shishir2001-yb, line 1411: What about the other way around i.e you have shared origin and you trying to get exclusive one?

shishir2001-yb: setup_shared('A') followed by the exclusive pg_replication_origin_session_setup('B') silently clobbers A. The exclusive path's only guard is `session_replication_state != NULL` (origin.c:1125), which a shared session leaves NULL, so it proceeds and overwrites replorigin_session_origin. The other three directions all error cleanly (shared→shared, exclusive→shared, and reset mismatches). For symmetry, add to pg_replication_origin_session_setup():
```
if (replorigin_session_origin != InvalidRepOriginId)
    ereport(ERROR,
            (errcode(ERRCODE_OBJECT_IN_USE),
             errmsg("replication origin session already setup")));
```

shishir2001-yb: Actually I feel like we should guard your origin.c change with a flag, we can make the default as true.
1. 
```
void
replorigin_session_advance(XLogRecPtr remote_commit, XLogRecPtr local_commit)
{
	/* Shared origin sessions only tag writes and do not track progress. */
	if (your_flag ==true && session_replication_state == NULL)
		return;
    else
	Assert(session_replication_state != NULL);
```

2. For the above comment
```
if (your_flag ==true && replorigin_session_origin != InvalidRepOriginId)
    ereport(ERROR,
            (errcode(ERRCODE_OBJECT_IN_USE),
             errmsg("replication origin session already setup")));
```

In remaining places in origin.c, check the flag value first if it's false, throw an error saying you need to enable this flag.....

shishir2001-yb (changes requested):
Additionally, we will also need test coverage 
1. Shared vs exclusive interactions
2. Multiple origins working in parallel with _shared
3. Test in TestPgReplicationSlot.java
4. Test in TestYsqlUpgrade

src/postgres/src/backend/replication/logical/origin.c
kgalieva: I kept the assert after adding the feature flag guard. The NULL-slot return is now conditional on yb_enable_replication_origin_shared; when the flag is disabled, the function keeps the original behavior and the assert remains meaningful. So this is no longer an unconditional redundant assert.

src/yb/yql/pgwrapper/ysql_migrations/V104__30907__pg_replication_origin_shared.sql
kgalieva: Fixed. The V104 migration now sets `proisstrict = true` for both `yb_replication_origin_session_setup_shared(text)` and `yb_replication_origin_session_reset_shared()`.

I also added TestYsqlUpgrade#sharedReplicationOriginMigration to verify the migrated pg_proc rows, pg_description rows, and pg_init_privs rows for OIDs 8894/8895, including proisstrict and ACL/initprivs values.

src/postgres/src/backend/replication/logical/origin.c
kgalieva: Fixed. pg_replication_origin_session_setup() now rejects the shared-to-exclusive case before it can clobber replorigin_session_origin.

I also added yb_enable_replication_origin_shared as a default-on postmaster GUC and gated the shared-origin behavior on it:

- shared setup/reset error when the flag is disabled
- replorigin_session_advance() only skips the NULL session_replication_state path when the flag is enabled

TestPgReplicationSlot now covers shared vs exclusive interactions, multiple shared origins in parallel, and disabled-flag behavior.

kgalieva:
Rebased onto latest master and addressed the follow-up coverage requests:

- shared vs exclusive interactions are covered in TestPgReplicationSlot#testPgoutputWithSharedOriginId
- multiple shared origins in parallel are covered in TestPgReplicationSlot#testPgoutputWithSharedOriginId
- TestPgReplicationSlot now also covers permissions and disabled flag behavior for the new yb_* functions
- TestYsqlUpgrade#sharedReplicationOriginMigration verifies the V104 migration catalog rows, including proisstrict, ACLs/initprivs, descriptions, and function metadata
- contrib/test_decoding changes were removed from the final diff

src/postgres/src/include/catalog/pg_yb_migration.dat
```diff
@@ -12,7 +12,7 @@
 [
 
 # For better version control conflict detection, list latest migration filename
-# here: V103__28937__yb_fix_catalog_version_table_perdb_only.sql
+# here: V104__30907__pg_replication_origin_shared.sql
 { major => '103', minor => '0', name => '<baseline>', time_applied => '_null_' }
```
shishir2001-yb: You need to update `major => '104'`

java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgReplicationSlot.java
```diff
@@ -5051,6 +5051,126 @@ public void testPgoutputWithOriginId() throws Exception {
...
+      stmt2.execute("SELECT yb_replication_origin_session_reset_shared()");
+
+      stmt1.execute("SELECT pg_replication_origin_session_setup('" + originId1 + "')");
+      runInvalidQuery(
+          stmt1,
+          "SELECT yb_replication_origin_session_setup_shared('" + originId2 + "')",
+          "cannot use shared replication origin session while an exclusive one is active");
+      runInvalidQuery(
+          stmt1,
+          "SELECT yb_replication_origin_session_reset_shared()",
```
shishir2001-yb, line 5122: After this insert row in this and verify it doesn't have xprel_origin_id

java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgReplicationSlot.java
```diff
@@ -5051,6 +5051,126 @@ public void testPgoutputWithOriginId() throws Exception {
...
   }
 
+  @Test
+  public void testPgoutputWithSharedOriginId() throws Exception {
+    final String slotName = "test_replication_with_shared_origin_id";
+    final String originId1 = "shared_origin1";
+    final String originId2 = "shared_origin2";
+
+    Connection replConn = getConnectionBuilder().withTServer(0).replicationConnect();
+    PGReplicationConnection replConnection = replConn.unwrap(PGConnection.class).getReplicationAPI();
```
shishir2001-yb: Lint error
```
   Warning  (TXT3) Line Too Long
    This line is 101 characters long, but the convention is 100 characters.
```

java/yb-pgsql/src/test/java/org/yb/pgsql/TestYsqlUpgrade.java
```diff
@@ -1138,6 +1138,41 @@ public void migrationInGeoPartitionedSetup() throws Exception {
     runMigrations(false /* useSingleConnection */);
   }
 
+  @SkipOnTSAN
+  @Test
+  public void sharedReplicationOriginMigration() throws Exception {
```
shishir2001-yb: Actually, other test already covers it, so you can remove

shishir2001-yb (changes requested):
Just those minor comments need to be addressed, please also update the PR Description and then we are good to merge.

kgalieva:
I addressed the latest comments and updated the description

shishir2001-yb:
Trigger Jenkins

shishir2001-yb (approved):

PR #30961, opened 2026-04-02
Fix DynamicMemoryUsageOf(std::string) on macOS

ellabaron-code:
The memory_usage-test is disabled on Linux because the build uses Google
tcmalloc, which does not support malloc hooks (YB_GPERFTOOLS_TCMALLOC is
not defined). However, on macOS, Google tcmalloc is not supported, so
the build uses gperftools tcmalloc instead (YB_GPERFTOOLS_TCMALLOC is
defined), which enables the test.

On macOS, gperftools tcmalloc is statically linked but does not replace
the system allocator (libsystem_malloc). It only provides malloc hooks
for observing allocations. The existing DynamicMemoryUsageOf formula for
clang used tcmalloc's 16-byte size class rounding ((capacity + 16) &
~0xf), which is incorrect on macOS where the system allocator allocates
exactly capacity + 1 bytes.

This caused memory_usage-test to fail on macOS, e.g. for a string with
capacity 24: DynamicMemoryUsageOf returned 32 but the actual allocation
was 25.

Split the clang path so that the tcmalloc rounding formula is only used
on Linux, where gperftools tcmalloc is the active allocator. On macOS
(and other non-Linux clang platforms), use capacity + 1, which matches
the system allocator behavior. The SSO capacity (22) is correct for
libc++ on both platforms.

---

Phorge: [D51803](https://phorge.dev.yugabyte.com/D51803)

### DIFF
--- a/src/yb/util/memory/memory_usage.h
+++ b/src/yb/util/memory/memory_usage.h
@@ -67,7 +67,8 @@ inline std::size_t DynamicMemoryUsageOf(const std::string& value) {
     return 0;
   } else {
 #if defined(__linux__)
-    // gperftools tcmalloc allocates 16*n bytes for std::string capacity from [16*(n - 1); 16*n - 1].
+    // gperftools tcmalloc allocates 16*n bytes for std::string capacity
+    // from [16*(n - 1); 16*n - 1].
     // 48 bytes for capacity in [32; 47], 64 bytes for capacity in [48; 63] and so on...
     return (capacity + 16) & ~(size_t(0xf));
 #else


src/yb/util/memory/memory_usage.h
```diff
@@ -87,6 +91,10 @@ inline std::size_t DynamicMemoryUsageOf(const std::string& value) {
 
 constexpr const auto kStdStringInternalCapacity = 15;
 
+#endif // defined(__clang__)
+
+#if !defined(__linux__) && !(__GNUC__ >= 9 && __GNUC__ < 11)
```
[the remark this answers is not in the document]
ellabaron-code: Fixed.

hari90:
✅ **Jenkins build for commit `016fde5b`: Pass**

**Passed:** 12

---
### 🔨 DB Build/Test Job Summary
| Build | Total | Passed | Failed | Failed After Retries |
|-------|------:|-------:|-------:|---------------------:|
| D51803-ubuntu22.04-clang21-debug | 2 | 2 | 0 | 0 |
| D51803-arm-alma8-clang21-release | 11135 | 10705 | **16** | 10 |
| D51803-arm-mac14-clang-release | 17 | 17 | 0 | 0 |
| D51803-mac14-clang-release | 2 | 2 | 0 | 0 |
| D51803-alma8-clang21-release | 11137 | 10709 | **14** | 7 |
| D51803-alma8-gcc12-fastdebug | 11876 | 11413 | **8** | 1 |
| D51803-alma8-clang21-tsan | 11665 | 10028 | **8** | 5 |
| D51803-alma9-clang21-asan | 11753 | 11017 | **12** | 11 |

[Full status](http://csiweb.dev.yugabyte.com/D51803)

---
*JenkinsBot*

hari90:
✅ **Jenkins build for commit `8909348f`: Pass**

**Passed:** 12

---
### 🔨 DB Build/Test Job Summary
| Build | Total | Passed | Failed | Failed After Retries |
|-------|------:|-------:|-------:|---------------------:|
| D51803-arm-alma8-clang21-release | 11146 | 10715 | **14** | 9 |
| D51803-arm-mac14-clang-release | 17 | 17 | 0 | 0 |
| D51803-alma8-clang21-release | 11148 | 10720 | **10** | 7 |
| D51803-mac14-clang-release | 2 | 2 | 0 | 0 |
| D51803-alma8-clang21-tsan | 11676 | 10037 | **5** | 1 |
| D51803-alma9-clang21-asan | 11764 | 11030 | **11** | 9 |
| D51803-ubuntu22.04-clang21-debug | 2 | 2 | 0 | 0 |
| D51803-alma8-gcc12-fastdebug | 11887 | 11424 | **11** | 4 |

[Full status](http://csiweb.dev.yugabyte.com/D51803)

---
*JenkinsBot*

src/yb/util/memory/memory_usage.h
```diff
@@ -66,9 +66,15 @@ inline std::size_t DynamicMemoryUsageOf(const std::string& value) {
   if (capacity <= kStdStringInternalCapacity) {
     return 0;
   } else {
-    // std::string allocates 16*n bytes for capacity from [16*(n - 1); 16*n - 1].
+#if defined(__linux__)
```
SrivastavaAnubhav: Can you use `#if YB_GPERFTOOLS_TCMALLOC` instead?

ellabaron-code: Fixed.

src/yb/util/memory/memory_usage.h
```diff
@@ -66,9 +66,15 @@ inline std::size_t DynamicMemoryUsageOf(const std::string& value) {
   if (capacity <= kStdStringInternalCapacity) {
     return 0;
   } else {
-    // std::string allocates 16*n bytes for capacity from [16*(n - 1); 16*n - 1].
+#if YB_GPERFTOOLS_TCMALLOC
```
SrivastavaAnubhav: This is still incorrect; YB_GPERFTOOLS_TCMALLOC is only set if we are using gperftools tcmalloc. So your statements are backwards

ellabaron-code: Sorry about the mix up.

I think that the comment is correct and we should revert back to distinguishing between Linux vs. macOS and not `YB_GPERFTOOLS_TCMALLOC`.

It looks like this `memory_usage-test` test that uses this function only runs when `YB_GPERFTOOLS_TCMALLOC` is set for normal release or debug build. On Linux we use google tcmalloc by default, on macOS we only use gperftools tcmalloc. The YB_GPERFTOOLS_TCMALLOC case definitely correct for macOS, but I do not know if it is correct for Linux because on Linux we only set YB_GPERFTOOLS_TCMALLOC for asan and tsan builds and `memory_usage-test` is disabled for those builds.

So, if you do not mind, I am going to revert the change you suggested.

ellabaron-code: I have removed my last commit. `YB_GPERFTOOLS_TCMALLOC` can't be used here as condition as both cases are for `YB_GPERFTOOLS_TCMALLOC` case.

ellabaron-code: The statements are not backwards. Try running this test on macOS and you will see that it fails. YB_GPERFTOOLS_TCMALLOC applies to both cases here, only the memory usage calculation on macOS is not the same as on Linux.

Issue #31007, opened 2026-04-07
[YSQL] Backward scan on a hash index skips data


andrei-mart:
### Description

We don't currently use scans with directions in YSQL, but plan to, and our unit tests can do it.
Attached patch reproduces the issue.

[test_hash_scan_backward.patch](https://github.com/user-attachments/files/26552324/test_hash_scan_backward.patch)

Apply attached patch and run the test
./yb_build.sh  --gtest_filter PggateTestSelect.HashBackwardScan

The scan works properly if the table is one tablet. From a 3-tablet table the test reads 4 rows out of 10.
It looks like the scan only reads the first tablet, not all of them in sequence from the last to the first. 

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

Issue #30927, opened 2026-04-01
[YSQL] Support ORDER BY yb_hash_code(hash_key)


andrei-mart:
Jira Link: [DB-20825](https://yugabyte.atlassian.net/browse/DB-20825)
### Description

The YB hash indexes are ordered by the hash code, and then by the hash key and range key.
That generally prevents usage of the hash indexes for ordering, unless all the hash key columns are equaled to constants. However, the natural ordering of the hash index can naturally be represented by this `ORDER BY   yb_hash_code(hash_key), hash_key, range_key`, and we can just iterate over the index to implement this ORDER BY or any its prefix. Planner should just identify this case, no executor changes are needed.
While ordering by the hash_code is not practical, there are many use cases, where not exactly the order of keys is important, but the fact that equal keys are grouped together, i.e. for group aggregation.
If the yb_hash_code(hash_key) expression is recognized as a sort key, a constant equality condition should make it redundant, in other words, `WHERE yb_hash_code(hash_key) = const ORDER BY hash_key, range_key` can be implemented without Sort. Similarly, `yb_hash_code(hash_key) IN (list)` should allow YB Merge Scan. 

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-20825]: https://yugabyte.atlassian.net/browse/DB-20825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #30900, opened 2026-03-31
[YSQL] Support in PgGate for ROW conditions on hash indexes


andrei-mart:
Jira Link: [DB-20797](https://yugabyte.atlassian.net/browse/DB-20797)
### Description

We have implemented limited support for ROW conditions on the range keys via document bounds.
Given a condition like `ROW(r1, r2, ...) OP (r1_value, r2_value, ...)` where OP is the one of (<, <=, =, >=, >), we build a DocKey with the provided values {r1_value, r2_value, ...} and make it an upper, or a lower bound, or both, depending on the comparison operator.
We want to support ROW conditions on the hash keys as well.

Currently the PgDmlRead::EncodeRowKeyForBound function has code for the hash key case, but it is unused and incorrect. The current implementation assumes condition like `ROW(<hash_keys>[, <range_keys>]) OP (<hash_key_values>[, <range_key_values>])` and calculates the DocKey bound as {yb_hash_code(<hash_key_values>), <hash_key_values>[, <range_key_values>]}. Such bound is incorrect, since the rows matching the condition may be in any hash bucket, not only the buckets below or above the yb_hash_code(<hash_key_values>). Specifically, strict inequality conditions may target the values starting from <hash_key_values>+1 or <hash_key_values>-1, which most likely lay in the different hash bucket than <hash_key_values>. But without hash code DocKey bounds are inefficient.

Though we can implement efficient DocKey bounds if the ROW condition has yb_hash_code(<hash_keys>) in it: `ROW(yb_hash_code(<hash_keys>), <hash_keys>[, <range_keys>]) OP (hash_code, <hash_key_values>[, <range_key_values>])`.
With explicitly specified hash code, the DocKey bound {hash_code, <hash_key_values>[, <range_key_values>]} correctly implements the ROW condition.
NB: with explicitly provided hash code the bound can be built, and be efficient even if not all the hash keys are provided.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-20797]: https://yugabyte.atlassian.net/browse/DB-20797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31035, opened 2026-04-09
[#31035] YSQL: recognize ROW(yb_hash_code(...), ...) as Index Cond on hash indexes

kirs:
Refs: #30524, #30651, @andrei-mart to review.

This teaches the planner to recognize `ROW(yb_hash_code(hash_cols), ...) > (...)` as a valid Index Cond on hash-sharded indexes.

Previously, all ROW comparisons involving hash columns were rejected by `match_rowcompare_to_indexcol`. This meant queries like:

  ```sql
SELECT * FROM orders
WHERE (yb_hash_code(user_uuid), user_uuid, uuid) > (10, '01961f1f-...', '01968344-...')
  AND yb_hash_code(user_uuid) < 11
LIMIT 5000;
```

would place the ROW comparison as a post-scan Filter, scanning from the start of the index and discarding rows until the cursor position.

With this change, the ROW comparison appears as Index Cond, enabling better plan selection. Actual DocDB bound pushdown (seeking directly to the cursor position) will be enabled in a follow-up once the pggate EncodeRowKeyForBound changes land (WIP patch by @andrei-mart
(https://gist.github.com/andrei-mart/2755d1d02002040b5ba99f599f2f303f)).

## Tests

New regression test yb.orig.yb_hash_code_row_compare added to yb_index_schedule covering:
- Full key ROW with leading yb_hash_code (single and multi-column hash) -> Index Cond
- Without leading yb_hash_code -> Seq Scan (negative)
- Combined with yb_hash_code upper bound -> both as Index Cond
- Arity mismatch -> Seq Scan (negative)
- Less-than direction -> Index Cond
- Execution correctness with pinned hash bucket

What this does NOT include

- pggate EncodeRowKeyForBound changes -- @andrei-mart is handling this separately. His patch makes col_values[0] accept a pre-computed hash code.
- ORDER BY yb_hash_code(h), h, r pathkeys support -- also being handled by @andrei-mart.
- Lossy/partial prefix ROW truncation -- disabled for now to avoid executor complexity; the ROW must cover all its elements as a valid index key prefix.

Once the pggate patch lands, a follow-up will enable actual DocDB bound pushdown (seek instead of scan + recheck).

### CODE REVIEW

File: src/postgres/src/backend/access/yb_access/yb_scan.c [hunk 1]
```diff
@@ -963,6 +963,18 @@ ybcSetupScanPlan(bool xs_want_itup, YbScanDesc ybScan, YbScanPlan scan_plan)
 	{
 		ScanKey		key = ybScan->keys[i];
 
+		/*
+		 * YB: A yb_hash_code subkey inside a ROW comparison has
+		 * sk_attno = InvalidAttrNumber.  Skip the normal attnum
+		 * mapping to avoid an out-of-bounds access on indkey.
+		 */
+		if (key->sk_attno == InvalidAttrNumber)
```
andrei-mart, line 971: If I remember correctly, the condition `key->sk_attno == InvalidAttrNumber` indicates that the scan key is on an expression, not on an attribute. Am I mistaken? The comment implies the case is only relevant to yb_hash_code subkey inside a ROW comparison, is it true, no other possibilities of InvalidAttrNumber in key->sk_attno?

Also, there's a special place for the yb_hash_code expressions: `ybScan->hash_code_keys`. Any pros and cons of ROW's yb_hash_codes not being there?

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1029,8 +1041,14 @@ YbIsHashCodeSearch(ScanKey key)
...
 	bool		is_hash_search = (key->sk_flags & YB_SK_SEARCHHASHCODE) != 0;
 
-	/* We currently don't support hash code search with any other flags */
-	Assert(!is_hash_search || key->sk_flags == YB_SK_SEARCHHASHCODE);
+	/*
+	 * We currently don't support hash code search with any other flags,
+	 * except SK_ROW_MEMBER which is set on yb_hash_code subkeys inside
+	 * a RowCompareExpr.
+	 */
+	Assert(!is_hash_search ||
```
andrei-mart, line 1049: I'd suggest to delete this assertion.
I'm not sure if there is any problem if YB_SK_SEARCHHASHCODE is combined with other flags.

andrei-mart:
Thanks for submitting this. I've taken a brief look.
I'm off until Monday, I'll give you detailed feedback when I'll be back.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1601,8 +1619,15 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 	 * We can only push down right now if the key columns are specified in the
 	 * correct order and the key has no hashed columns. We also need to ensure
```
andrei-mart, line 1620: My understanding is that with yb_hash_code support you just delete ` and the key has no hashed columns`. The rest of the statement stands, assuming `the correct order` means (hash_code, hash_keys, range_keys) for hash indexes and (range_keys) for range indexes.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1611,6 +1636,17 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 	{
 		ScanKey		key = subkeys[j];
 
+		/*
+		 * Skip the leading yb_hash_code subkey.  It has sk_attno =
+		 * InvalidAttrNumber and cannot be validated against rd_indoption.
+		 */
+		if (j == 0 && YbIsHashCodeSearch(key))
```
andrei-mart, line 1643: If the index is hash, the first scan key has to be a hash code, or can't pushdown and break.
We probably can't have a hash code scan key if the index isn't hash.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1611,6 +1636,17 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+		{
+			has_yb_hash_code_subkey = true;
+			last_att_no = key->sk_attno;
+			continue;
+		}
+
 		/* Make sure that the specified keys are in the right order. */
 		if (key->sk_attno <= last_att_no)
```
andrei-mart, line 1651: That does not look right. `if (key->sk_attno != last_att_no +1)` then can't pushdown and break.
Btw, what "can_pushdown" is supposed to mean here?
We can make a bound if we have one or more scan keys. It is slightly more complicated if the bound is "lossy".

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5083,3 +5108,346 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
...
+	List	   *non_var_args;
+	Node	   *first_var;
+	Oid			first_opno;
+	int			hc_strategy;
+	int			matching_cols;
+	int			indexcol;
+	bool		var_on_left;
+
+	/* Only btree/lsm indexes support row comparisons. */
+	if (index->relam != BTREE_AM_OID && index->relam != LSM_AM_OID)
```
andrei-mart, line 5253: BTREE is a Postgres index format, still possible when dealing with temp tables.
If we want to handle only the hash case here, the condition is redundant.
The next one `index->nhashcolumns == 0` is false iff the index is an LSM and a hash.

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5083,3 +5108,346 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
...
+		return NULL;
+
+	/* Only useful for hash indexes. */
+	if (index->nhashcolumns == 0)
+		return NULL;
+
+	/* Need at least yb_hash_code + one column. */
+	if (list_length(clause->largs) < 2)
```
andrei-mart, line 5261: Always true? RowCompareExpr would always have at least two.

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5083,3 +5108,346 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
...
+	hc_strategy = get_op_opfamily_strategy(first_opno, INTEGER_LSM_FAM_OID);
+	switch (hc_strategy)
+	{
+		case BTLessStrategyNumber:
+		case BTLessEqualStrategyNumber:
+		case BTGreaterEqualStrategyNumber:
+		case BTGreaterStrategyNumber:
+			break;
+		default:
+			return NULL;
```
andrei-mart, line 5316: Isn't it the same as
```
if (hc_strategy == BTEqualStrategyNumber)
    return NULL;
```
?

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5083,3 +5108,346 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
...
+	}
+
+	/*
+	 * Walk the remaining ROW elements.  Each must reference the next index
+	 * key column in order (forming a prefix), with a compatible operator.
+	 */
+	matching_cols = 1;
```
andrei-mart, line 5323: It looks like the logic below is similar to match_rowcompare_to_indexcol and expand_indexqual_rowcompare.
It may make sense to refactor and make some subroutines to reuse between here and there. 
I'd take some time to think about this.

andrei-mart (commented):
I do have couple pending changes so far on both sides, both with some limitations we need to discuss to integrate.

On execution side we expect efficient bounds set by YBCPgDmlAddRowLowerBound/ YBCPgDmlAddRowUpperBound while the Row condition matches the index prefix. Current implementation allows missing columns, and the pending change maintains that. Though the columns after the missing one won't participate in the bound, so it is recommended not to include them. When missing values are not allowed, it will be possible to use is_inclusive to implement non-lossy strict and non-strict inequalities. For now, the bounds are lossy, unless all the key values are provided.

On the planning side I've discovered that backward scan on the hash indexes does not work properly. So the pending change does not allow `ORDER BY yb_hash_code(h) DESC`, as well as `WHERE yb_hash_code(h) = const ORDER BY h DESC`. Not sure if that limitation affects the Row conditions case.

src/postgres/src/test/regress/sql/yb.orig.yb_hash_code_row_compare.sql
```diff
@@ -0,0 +1,88 @@
...
+    v text,
+    PRIMARY KEY((h1, h2) HASH, r ASC)
+);
+INSERT INTO hc_rc_t2
+    SELECT i % 3, i % 5, i, 'val' || i FROM generate_series(1, 30) i;
+
+EXPLAIN (COSTS OFF)
+SELECT * FROM hc_rc_t2
+    WHERE (yb_hash_code(h1), h1, h2, r) > (yb_hash_code(1), 1, 2, 7)
+    LIMIT 10;
```
DerekStride, line 83: I'm concerned about indexes with columns stored in `DESC` order.

The comparison would need to look different for those cases e.g.

```sql
SELECT * FROM hc_rc_t3
    WHERE (yb_hash_code(h), h) > (yb_hash_code(1), 1)
      AND r < 7
    LIMIT 10;
```

```sql
CREATE TABLE hc_rc_t3(
    h int,
    r int,
    v text,
    PRIMARY KEY((h) HASH, r DESC)
);
```

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5083,3 +5108,346 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
...
+									list_nth(clause->largs, i));
+				new_collids = lappend_oid(new_collids,
+										  list_nth_oid(clause->inputcollids, i));
+			}
+
+			rc = makeNode(RowCompareExpr);
+			rc->rctype = (RowCompareType) hc_strategy;
+			rc->opnos = new_ops;
+			rc->opfamilies = NIL;
+			rc->inputcollids = new_collids;
```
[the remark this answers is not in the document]
egladysh: fixed

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5083,3 +5108,346 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
...
+
+		if (yb_hash_code_match_index(leftop, index))
+		{
+			IndexClause *iclause = makeNode(IndexClause);
+
+			iclause->rinfo = rinfo;
+			iclause->indexquals = list_make1(rinfo);
+			iclause->lossy = false;
+			iclause->indexcol = 0;
+			iclause->indexcols = NIL;
```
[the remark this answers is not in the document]
egladysh: I tried the suggested change but it trips the existing assertion in yb_single_row_update_or_delete_path: `Assert(iclause->indexcols == NULL)`. Added an integration test, `PgIndexTest.YbHashCodeRowCompareExecutionAndPlannerInvariants` to verify it. Keeping it as is for now.

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5083,3 +5108,346 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
...
+				IndexClause *iclause;
+
+				commrinfo = commute_restrictinfo(rinfo, comm_op);
+				iclause = makeNode(IndexClause);
+				iclause->rinfo = rinfo;
+				iclause->indexquals = list_make1(commrinfo);
+				iclause->lossy = false;
+				iclause->indexcol = 0;
+				iclause->indexcols = NIL;
+				return iclause;
```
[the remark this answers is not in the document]
egladysh: same here

egladysh:
@andrei-mart I'll be helping with this PR for now. Could you update me on the state of the related pggate changes? Do we have an open PR? It seems like this PR would really need pggate/executor changes to be useful.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: The first part has been fixed. Now we only skip attnum mapping (InvalidAttrNumber) for specific cases of YB_SK_SEARCHHASHCODE or SK_ROW_HEADER. 

As for the second part. It seems like `hash_code_keys` is intended for standalone hash code bounds. I'm not sure if it fits for a lexicographic row comparison so it goes into ybScan->keys. Moving it to `hash_code_keys` would break the comparison apart I think.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Fixed the condition, renamed to `can_pushdown_bound`. It seems like the lossy-bound logic is the later part of YbBindRowComparisonKeys(). I don't think we are changing it yet.

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: It was changed to 
```
	if (hc_strategy == InvalidStrategy ||
		hc_strategy == BTEqualStrategyNumber)
		return NULL;

```

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: @andrei-mart Yeah, there is an overlap. I think one of the main differences is that one allows partial/lossy matches while another (YB path) requires full match of the supplied row. I'm not sure if we want to touch the generic path in this PR. I'll wait for your suggestions?

src/postgres/src/test/regress/sql/yb.orig.yb_hash_code_row_compare.sql
egladysh: Agreed DESC needs to be handled before this can do a DocDB row-bound pushdown. I think this PR only teaches the planner to recognize (yb_hash_code(...),...) as an index qual. `YbBindRowComparisonKeys()` seems to disable row bound pushdowns. 
```
if (has_yb_hash_code_subkey)
   can_pushdown_bound = false;
```
I think the test is fine for this PR as it is.

egladysh:
Note: new patch from @andrei-mart, https://gist.github.com/andrei-mart/5e3d9e7d5ac56440db9c216e4a95b5c4

rahulb-yb:
Dependent GH issues which are being worked upon by YB:
https://github.com/yugabyte/yugabyte-db/issues/31007
https://github.com/yugabyte/yugabyte-db/issues/30927
https://github.com/yugabyte/yugabyte-db/issues/30900

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -3047,24 +3050,22 @@ match_clause_to_index(PlannerInfo *root,
...
+	 * just a single column.  For example, the yb_hash_code function may match
+	 * the hash code of a hash index, and a ROW comparison may start with
+	 * yb_hash_code followed by key columns.  Handle such expressions here,
+	 * before the per-column loop.
 	 */
-	IndexClause *yb_iclause = yb_match_clause_to_index(root, rinfo, index);
-	if (yb_iclause)
 	{
-		/*
-		 * TODO it seems correct to associate the hash code clause with the
```
andrei-mart, line 3069: Please keep the TODO.
We associate the yb_iclause with the first column, while it is not necessarily correct.
We'd like to reconsider that point later.

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5268,10 +5269,9 @@ yb_truncate_embedded_index_pathkeys(PlannerInfo *root, RelOptInfo *rel,
 
 /*
  * yb_hash_code_match_index
- *
- * Check if the given expression matches the index's hash code.
- * That is the index is a hash index, and the expression is a yb_hash_code call
- * with arguments that match the hash key columns of the index.
+ *	  Check if the given expression is a yb_hash_code call matching the index.
+ *	  That is, the index is a hash index and the expression is a yb_hash_code
```
andrei-mart: Duplicate "expression is a yb_hash_code call"

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5307,15 +5307,16 @@ yb_hash_code_match_index(Node *expr, IndexOptInfo *index)
 			return false;
 
 		if (index->rel->relid != ((Var *) arg)->varno ||
-			index->indexkeys[indexcol++] != ((Var *) arg)->varattno)
+			index->indexkeys[indexcol] != ((Var *) arg)->varattno)
 			return false;
+
+		indexcol++;
```
andrei-mart: Is there a problem with indexcol incremented in the expression?
If standalone, we prefer the prefix form of the unary operator: `++indexcol;`

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5328,15 +5329,24 @@ yb_match_clause_to_index(PlannerInfo *root,
...
-		Node	   *rightop = (Node *) lsecond(clause->args);
+
+		leftop = (Node *) linitial(clause->args);
+		rightop = (Node *) lsecond(clause->args);
+
 		if (yb_hash_code_match_index(leftop, index))
 		{
 			IndexClause *iclause = makeNode(IndexClause);
+
+			if (!is_pseudo_constant_for_index(root, rightop, index))
```
andrei-mart, line 5362: It is better to move it to the outer if condition:
```
if (yb_hash_code_match_index(leftop, index) &&
    is_pseudo_constant_for_index(root, rightop, index))
{
...
```
It saves makeNode that way.

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5348,16 +5358,20 @@ yb_match_clause_to_index(PlannerInfo *root,
 		{
 			Oid			comm_op = get_commutator(clause->opno);
 
+			if (!is_pseudo_constant_for_index(root, leftop, index))
```
andrei-mart, line 5377: ditto

src/postgres/src/backend/optimizer/path/indxpath.c
```diff
@@ -5367,5 +5381,237 @@ yb_match_clause_to_index(PlannerInfo *root,
...
+	List	   *var_args;
+	List	   *non_var_args;
+	Node	   *first_var;
+	Oid			first_opno;
+	int			hc_strategy;
+	int			matching_cols;
+	int			indexcol;
+	bool		var_on_left;
+
+	/* Only useful for hash indexes. */
```
andrei-mart, line 5437: Not necessarily so.
Currently, while ROW expression is converted to a scan bound, row compare is not useful as an index condition unless it is matching an index prefix. That applies to both, hash and range indexes.
So it makes sense to use this function to check if the expression matches the index prefix, and remove LSM from the  match_rowcompare_to_indexcol().

src/postgres/src/backend/optimizer/path/indxpath.c
andrei-mart: So, as I commented above, YB makes row condition out of the row compare, so the row compare is useful in index scan if the prefix of the row compare matches the prefix of the index key.
Note the "prefix" on both sides. It is broader than the current implementation implies.
In particular, `(yb_hash_code(h1,h2), h1, r1) > (...)` isn't bad. The prefix `yb_hash_code(h1,h2), h1` would make valid lower bound. The presence of r1 in the same row condition would require Postgres to recheck, but it is still useful.
On the other hand, if you have a range key (r1, r2, r3, r4) the row compare like `(r2, r3, r4) > (...)` is not useful even though match_rowcompare_to_indexcol finds a match to r2 column.

andrei-mart (commented):
Review is still in progress. Submitting my comments so far (to indxpath.c).

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: Sorry, not sure how it got deleted.

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: In this specific case, incrementing indexcol in the expression is not wrong but the only thing that makes it "right" is the immediate `return`. I think that inline index increments in complex conditional expressions is not a very good idea in general? 
As for the unary operator, I do agree but I was not sure because it's the postfix operator everywhere across the file. Should we continue to follow the style?

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: Good point! However I'm not sure if moving it to the outer if condition is the correct fix because it seems to alter the short-circut if yb_hash_code_match_index is true but is_pseudo_constant_for_index is false. It's probably not a big deal but I'd suggesting fixing as the following instead:
```
			IndexClause *iclause;

			if (!is_pseudo_constant_for_index(root, rightop, index))
				return NULL;

			iclause = makeNode(IndexClause);
```

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: Same concern.

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: @andrei-mart before I change the generic path, I want to confirm I understood your suggestion correctly. Are you suggesting that we should remove LSM from the generic `match_rowcompare_to_indexcol()` path, and make `yb_match_rowcompare_to_index()` handle both hash and range indexes?
For example, `(yb_hash_code(h1,h2), h1, r1) > (...)` could use the prefix `(yb_hash_code(h1,h2), h1)` as the index condition, while the full row comparison is checked as a filter.

src/postgres/src/backend/optimizer/path/indxpath.c
andrei-mart: Actually, there's one in line 5265.
That's one of the few preferences where we have opinion strong enough to break consistency with Postgres' code.
I assume you are going to make changes to this file anyway, so not a big deal to move that ++?

src/postgres/src/backend/optimizer/path/indxpath.c
andrei-mart: Ah, if the yb_hash_code is found on the left, it does not make sense to go to else if and check for yb_hash_code on the right. Agreed.

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: No, it's not a big deal and a right thing to do I think. To confirm, we are changing it across the file?

src/postgres/src/backend/optimizer/path/indxpath.c
andrei-mart: Yes, that matches the executor implementation.
The match_rowcompare_to_indexcol() was once reused for LSM case because of some similarity.
Now we are introducing yb_match_rowcompare_to_index(), which is even more similar.
I mean, LSM range case is more similar to LSM hash implemented here, than to btree case implemented in match_rowcompare_to_indexcol(). The flow in yb_match_rowcompare_to_index() may look like this:
```
if (/*index is hash*/)
{
  /* if the first row element is a ab_hash_code(), process it move on, else quit */ 
}
// loop over both remaining row expressions and index columns until mismatch
```
So there's special part for hash and common loop over index columns.

On the other hand, while executor converts row iclauses to bounds, it is not helpful if the row expression matches to the second, or further column on, so there is no point in finding these matches.

There are interesting cases though. What if there is a standalone condition on the first column like `r1=1` and row condition like `(r2, r3, r4) > (2, 3, 4)`. Or equivalent hash case `yb_hash_code(h1, h2, h3) = 1 and (h1, h2, h3) > (2, 3, 4)`. In such cases found matches to subsequent index columns may produce helpful results, but not at the moment.

Finally, what I'm suggesting is a subject for discussion.
Am I making valid points?
Do we want to change this PR or plan for follow-up?

src/postgres/src/backend/optimizer/path/indxpath.c
andrei-mart: Not across the file, only lines you add/modify.
We don't touch Postgres code unless necessary to avoid problems when we rebase on newer Postgres versions.

egladysh: Got it.

src/postgres/src/backend/optimizer/path/indxpath.c
egladysh: Yeah, I think it makes a lot of sense but I'm not sure about the scope of this PR. Changing generic rowcompare planning for range indexes is broader and could affect existing plans/tests. The branch is about ROW(yb_hash_code(...), ...). I think I'd prefer to plan for a follow-up.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -963,6 +963,25 @@ ybcSetupScanPlan(bool xs_want_itup, YbScanDesc ybScan, YbScanPlan scan_plan)
...
+		if (key->sk_flags & YB_SK_SEARCHHASHCODE ||
+			((key->sk_flags & SK_ROW_HEADER) &&
+			 key->sk_attno == InvalidAttrNumber))
+		{
+			ybScan->target_key_attnums[i] = InvalidAttrNumber;
+			scan_plan->bind_key_attnums[i] = InvalidAttrNumber;
+			continue;
+		}
+
+		if (key->sk_attno == InvalidAttrNumber)
```
andrei-mart: It is preferable to keep the loop body structure consistent.
Either have one `if/else if/else` statement, or series of `if` with a `continue` ate the end of the block.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,11 +1678,20 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+	 * YB_SK_SEARCHHASHCODE flag) is allowed but not yet pushed down as a
+	 * DocDB bound.  Disable pushdown so the condition is rechecked at the
+	 * Postgres level.
 	 */
-	bool		can_pushdown = true;
+	bool		can_pushdown_bound = true;
+	bool		has_yb_hash_code_subkey = false;
+	bool		is_hash_index = (index->rd_indoption[0] & INDOPTION_HASH) != 0;
+	bool		have_last_attno = false;
+	AttrNumber	last_attno = InvalidAttrNumber;
```
andrei-mart: Redundant. You can track attno continuity by comparing key->sk_attno to j.
Continuous, if `key->sk_attno == j + 1` in range case, `key->sk_attno == j` in hash case.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1677,43 +1700,78 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
 	{
 		ScanKey		key = subkeys[j];
 
-		/* Make sure that the specified keys are in the right order. */
-		if (key->sk_attno <= last_att_no)
+		if (YbIsHashCodeSearch(key))
```
andrei-mart: It would be better to check if the 0-th subkey is a hash code before the loop.
```
int j = 0;
if (is_hash_index)
{
  if (YbIsHashCodeSearch(subkeys[0]))
    ++j;
  else
    return true /* need recheck */;
}
```
After that, no need to recheck if the index is hash, the hash code subkey encountered.
Also, if there's a hash code subkey in the loop, it is unconditional break.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,11 +1678,20 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 
 	/*
-	 * We can only push down right now if the key columns are specified in the
-	 * correct order and the key has no hashed columns. We also need to ensure
-	 * that the same comparison operation is done to all subkeys.
+	 * We can only push down right now if the key columns form a contiguous
+	 * index-key sequence and the same comparison operation is done to all
+	 * subkeys.
+	 *
+	 * YB: A leading yb_hash_code subkey (sk_attno = InvalidAttrNumber,
```
andrei-mart: Why not push it down? PgGate supports hash bounds already.
If there's a problem with that, please create a ticket, and put a TODO here, mentioning the ticket.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1677,43 +1700,78 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
-			can_pushdown = false;
+			can_pushdown_bound = false;
 			break;
 		}
-		last_att_no = key->sk_attno;
+		have_last_attno = true;
+		last_attno = key->sk_attno;
 
 		/* Make sure that there are no hash key columns. */
 		if (index->rd_indoption[key->sk_attno - 1]
```
andrei-mart: if (is_hash_index)

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1677,43 +1700,78 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+	 * The pggate EncodeRowKeyForBound API needs to be updated to accept
+	 * a pre-computed hash code before we can push these down as DocDB
+	 * row bounds.  The condition will be rechecked at the Postgres level.
+	 */
+	if (has_yb_hash_code_subkey)
+		can_pushdown_bound = false;
+
 	/* Make sure that there are no hash keys in order to push down. */
-	for (int i = 0; (i < index->rd_index->indnkeyatts) && can_pushdown; i++)
+	for (int i = 0; (i < index->rd_index->indnkeyatts) && can_pushdown_bound; i++)
```
andrei-mart: This loop is actually redundant.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1677,43 +1700,78 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+		}
+
+		if (is_hash_index && j == 0)
+		{
+			can_pushdown_bound = false;
+			break;
+		}
+
+		/* Make sure that the specified keys are contiguous. */
+		if (have_last_attno && key->sk_attno != last_attno + 1)
```
andrei-mart: Broken continuity does not mean the bound can't be pushed bound, it means the bound  should be shorter. Actually, this change introduces a performance regression. Before the change a row condition on columns (1,3,5) was pushed down, though only first value was effective. After the change condition is not pushed down at all, falling back to full scan with all filtering done at Postgres side.

src/postgres/src/test/regress/sql/yb.orig.yb_hash_code_row_compare.sql
```diff
@@ -0,0 +1,88 @@
...
+    PRIMARY KEY(h HASH, r1 ASC, r2 ASC)
+);
+INSERT INTO hc_rc_t1
+    SELECT i % 5, i, i * 10, 'val' || i FROM generate_series(1, 50) i;
+
+-------------------------------------------------------------------
+-- Test 1: Full key ROW comparison with leading yb_hash_code.
+-- Should show ROW(yb_hash_code...) as Index Cond, not Filter.
+-------------------------------------------------------------------
+EXPLAIN (COSTS OFF)
```
andrei-mart: You want to EXPLAIN ANALYZE these test queries.
EXPLAIN ANALYZE will show actual row count, chances to catch incorrect filtering.
Also "Rows Removed by Index Recheck" will show efficiency of the DocDB filter.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: If we refactor the code as you suggested to check `is_hash_key` pre-loop, we don't need any checks here unless there is a bug I think. Maybe an assert?
```
assert(!((index->rd_indoption[key->sk_attno - 1] & INDOPTION_HASH) && !has_yb_hash_code_subkey));
```

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Agreed.

src/postgres/src/backend/access/yb_access/yb_scan.c
andrei-mart: It is a bit of overlook on my side, the expression looked similar to one in
```
bool		is_hash_index = (index->rd_indoption[0] & INDOPTION_HASH) != 0;
```
but I did not notice different index in rd_indoption.

This code was meant to allow to pushdown row conditions on the range keys of a hash index, but actually such row conditions can't make scan bounds and useless.

Now if yb_hash_code presents, those keys are useful, if other conditions are met (i.e. continuity).

egladysh: Thanks, that makes sense. I’ll just remove the per-column hash key check here. For hash indexes, row bounds are only useful when led by yb_hash_code (we'll check it pre-loop). After that the continuity, etc. checks determine usable prefixes.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Are you suggesting to something like: 
```if (key->sk_flags & YB_SK_SEARCHHASHCODE ||
      ((key->sk_flags & SK_ROW_HEADER) &&
       key->sk_attno == InvalidAttrNumber))
  {
      ybScan->target_key_attnums[i] = InvalidAttrNumber;
      scan_plan->bind_key_attnums[i] = InvalidAttrNumber;
  }
  else if (key->sk_attno == InvalidAttrNumber)
  {
      ...
  }
  else if (!index)
  {
     ...
  }
```

I think it is harder to reason because we introduce behavioral dependancies between seemingly independent conditions. I'd refactor it as 
```
if (key->sk_flags & YB_SK_SEARCHHASHCODE ||
      ((key->sk_flags & SK_ROW_HEADER) &&
       key->sk_attno == InvalidAttrNumber))
  {
      ...
      continue;
  }

  if (key->sk_attno == InvalidAttrNumber) 
      elog(ERROR, "unexpected scan key without attribute number");

  if (!index)
  {
          ...
         continue;
  }

if (index->rd_index->indisprimary)
  {
      ...
      continue;
  }
...
```

Thoughts?

Also keeping in mind that  `if (key->sk_attno == InvalidAttrNumber) 
      elog(ERROR, "unexpected scan key without attribute number"); ` makes it correct only because elog doesn't return.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Good point. I think I missed an old comment.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Agreed.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Yeah, it is kind of related to your previous comments. I'll try to fix it.

egladysh:
@andrei-mart Question, we have this flag `bool		needs_recheck = true;` it seems to remain true always. Is it intentional? At the first glance it depends on the whether the prefix is lossy and can be initialized to something like: `pushdown_subkey_count != subkey_count`?

src/postgres/src/test/regress/sql/yb.orig.yb_hash_code_row_compare.sql
egladysh: Agreed. I'll change it and regenerate the expected output.

egladysh:
@andrei-mart I think I addressed your comments except some minor things that I asked you about? I kept `needs_recheck` always true but please let me know. Also I added a gate on `yb_allow_dockey_bounds`. Could you take a look when you have a chance?

src/postgres/src/backend/access/yb_access/yb_scan.c
andrei-mart: My complain was about inconsistency, I'd like to see either one `if/else if/else` or series of `if/continue`, either works. To me in both cases the logic is equally difficult to follow. In either case, the dependencies implying the condition order won't go away and worth to mention in a code comment. Though in many cases in series of checks order matters, perhaps for weaker reasons, like more probable go first. In this case sk_attno is used multiple times, it is clearly means a dependency.

src/postgres/src/backend/optimizer/path/indxpath.c
andrei-mart: OK, for the scope reasons, let's keep current logic intact.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1678,51 +1678,52 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 
 	int			strategy = header_key->sk_strategy;
 	int			subkey_count = length_of_key - 1;
+	int			pushdown_subkey_count = subkey_count;
+	int			first_key_subkey = 0;
 
-	for (int j = 0; j < subkey_count; j++)
+	if (is_hash_index)
+	{
+		if (subkey_count == 0 || !YbIsHashCodeSearch(subkeys[0]))
```
andrei-mart: I don't think subkey_count  can ever be 0.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1678,51 +1678,52 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+			return true;
+
+		/*
+		 * Hash row bounds are encoded as DocKey bounds, which can be
+		 * disabled by the AutoFlag-backed GUC.
+		 */
+		if (!yb_allow_dockey_bounds)
+			return true;
+
+		has_yb_hash_code_subkey = true;
```
andrei-mart: Redundant, from this point on, has_yb_hash_code_subkey equals to is_hash_index.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1678,51 +1678,52 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
-			has_yb_hash_code_subkey = true;
-			have_last_attno = true;
-			last_attno = InvalidAttrNumber;
-			continue;
-		}
-
-		if (is_hash_index && j == 0)
-		{
-			can_pushdown_bound = false;
+			pushdown_subkey_count = j;
```
andrei-mart: I think you can eliminate pushdown_subkey_count if you move `int j = first_key_subkey;` out of the loop.
When the loop ends, the j is the pushdown_subkey_count.
You can use more descriptive variable name than j if you want.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: I just noticed that `if/else` was there before the change. I think we'll continue with the same flow. Fixed.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1732,50 +1733,32 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 	{
+		int			yb_hash_code_offset = first_key_subkey;
+		int			n_bound_values =
+			index->rd_index->indnkeyatts + yb_hash_code_offset;
 
 		YbcPgExpr  *col_values = palloc(sizeof(YbcPgExpr) *
-										index->rd_index->indnkeyatts);
+										n_bound_values);
+
+		if (has_yb_hash_code_subkey && !is_for_precheck)
```
andrei-mart: First, I'll answer the question about `needs_recheck`.
I believe initially it was planned to change it to false, if the scan bound certainly matches the row condition. However, there are tricky things around NULLs, shorter keys, inclusivity, etc., including implementation bugs, so we ended up with it being constantly true.
You may change it to false in the cases you are confident about. If the number of subkeys equals to the number of index keys, no nulls then it may be safe. Fine to keep it always true for now.

So, if this function is called with `is_for_precheck=true`, it can bail out early, as it is known the recheck is needed and function does not need to do anything else. That would allow to skip checking  is_for_precheck value further on, including this line.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1732,50 +1733,32 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
-		if (index->rd_indoption[i] & INDOPTION_HASH)
-			can_pushdown_bound = false;
 	}
 
 	bool		needs_recheck = true;
 
-	if (can_pushdown_bound)
+	if (pushdown_subkey_count > first_key_subkey)
 	{
+		int			yb_hash_code_offset = first_key_subkey;
```
andrei-mart: Neither yb_hash_code_offset nor first_key_subkey are modified below, why don't you just use  first_key_subkey?

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1732,50 +1733,32 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+		int			n_bound_values =
+			index->rd_index->indnkeyatts + yb_hash_code_offset;
 
 		YbcPgExpr  *col_values = palloc(sizeof(YbcPgExpr) *
-										index->rd_index->indnkeyatts);
+										n_bound_values);
```
andrei-mart: Actual number of values to push down is `pushdown_subkey_count`

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1805,25 +1788,29 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 
 		/* Whether or not the RHS values make up a DocDB upper bound */
 		bool		is_upper_bound = gt ^ is_direction_asc;
-		size_t		subkey_index = 0;
+		size_t		subkey_index = yb_hash_code_offset;
 
 		for (int j = 0; j < index->rd_index->indnkeyatts; j++)
 		{
-			bool		is_column_specified = (subkey_index < subkey_count &&
+			bool		is_column_specified = (subkey_index < pushdown_subkey_count &&
```
andrei-mart: Why didn't we check if the column is specified in the previous loop?
Same for direction mismatch.
Mismatched or missed column values are not to be pushed down, as well as the subsequent.

I thought the first loop is to figure out how many values we are going to push down, next, if we have enough, we create the constants and bind them as bound(s).

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Overly defensive way to protect subkeys access. I'm fine with removing it.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: I'll change j to `pushdown_subkey_count`.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: I think the idea was to reuse the existing code as much as possible... Same applies to your previous comment about `n_bound_values` that's not less then `pushdown_subkey_count`. Let me try to refactor it.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Sounds good. We'll keep needs_recheck as is. As for `is_for_precheck`, are you suggesting to exit immediately? I am concerned about potentially loosing some test coverage for example.
```
 YbBindRowComparisonKeys(...)
 {
  	if (is_for_precheck)
  		return true;

  	...
  }
```

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: @andrei-mart I refactored the whole function quite a bit. I might have messed up the formatting or missed something but could you take a look when you have a chance? One thing I kept the `is_for_precheck` semantic about the way it was before but I'd be fine if you think that we can just stop immediately or at some other point when it's true?

egladysh:
Asked AI to write additional test cases:

- partial key prefix: (yb_hash_code(h), h, r1) > (...)
- commuted compare: constants on the left
- mismatch: (yb_hash_code(h1,h2), h1, r) stays non-indexed
- mixed ASC/DESC range keys

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,182 +1677,186 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
 
-	/* Make sure that there are no hash keys in order to push down. */
-	for (int i = 0; (i < index->rd_index->indnkeyatts) && can_pushdown; i++)
-	{
-		if (index->rd_indoption[i] & INDOPTION_HASH)
-			can_pushdown = false;
-	}
+	bool		is_direction_asc =
+		is_hash_index ||
+		!(index->rd_indoption[subkeys[first_key_subkey]->sk_attno - 1] &
```
andrei-mart: Nit: first_key_subkey is always 0 in this line

andrei-mart: Also note that `subkey_count > 0` is not asserted on the execution path to that line.

Same is in "previous" code (lines 1750-1751). So, as I mentioned, it is safe to assume that subkey_count can't be zero.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,182 +1677,186 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
-									strategy != BTLessStrategyNumber);
+			if (asc != is_direction_asc)
+				direction_mismatch_subkey = pushdown_subkey_count;
+		}
+	}
 
-		bool		is_point_scan = ((subkey_count == index->rd_index->indnatts) &&
-									 (strategy == BTEqualStrategyNumber));
+	bool		needs_recheck = true;
+	bool		is_point_scan = ((pushdown_subkey_count ==
```
andrei-mart: This is old code, but do you have an idea, why the numbers of subkeys and index keys matter for direction mismatch? I feel like `strategy == BTEqualStrategyNumber` is sufficient. If my feeling is correct, we don't need direction_mismatch_subkey variable, in line 1742 above `if (strategy != BTEqualStrategyNumber)` and just break in the case of the direction mismatch.

If you are not sure, it is OK to keep it as is.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,182 +1677,186 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+		if (!is_for_precheck)
 		{
-			HandleYBStatus(YBCPgDmlAddRowUpperBound(ybScan->handle,
-													index->rd_index->indnkeyatts,
-													col_values,
-													is_inclusive));
+			AttrNumber	attnum =
+				scan_plan->bind_key_attnums[skey_index + 1 + subkey_index];
+
+			col_values[j + first_key_subkey] =
```
andrei-mart: col_values[subkey_index] =

We have already checked that the scan keys being pushed down match the index keys.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,182 +1677,186 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
-					scan_plan->bind_key_attnums[skey_index + 1 + subkey_index];
-
-				col_values[j] = YBCNewConstant(ybScan->handle,
-											   ybc_get_atttypid(scan_plan->bind_desc,
-																attnum),
-											   current->sk_collation,
-											   current->sk_argument,
-											   false);
-			}
+		col_values[0] = YBCNewConstant(ybScan->handle,
```
andrei-mart, line 1763: This constant creation matches the one in the loop below, except for explicit data type here.
I'm not sure if the scan key representing yb_hash_code would always have int4 type. The PgGate definitely expects an int4 constant for hash code.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,182 +1677,186 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
+							   current->sk_argument,
+							   false);
 		}
 
-		if (!is_upper_bound || strategy == BTEqualStrategyNumber)
+		/*
+		 * PgGate rejects IS NOT NULL binds on partition columns, and
+		 * the full row comparison remains rechecked by Postgres.
+		 */
+		if (!is_hash_index &&
```
andrei-mart: It can be as simple as `if (subkey_index == 0)`.

src/postgres/src/backend/access/yb_access/yb_scan.c
```diff
@@ -1664,182 +1677,186 @@ YbBindRowComparisonKeys(YbScanDesc ybScan, YbScanPlan scan_plan,
...
-											   (subkeys[subkey_index]->sk_attno - 1) == j);
+	if (pushdown_subkey_count <= first_key_subkey)
+		return needs_recheck;
 
-			/*
-			 * Is the current column stored in ascending order in the
-			 * underlying index?
-			 */
-			bool		asc = (index->rd_indoption[j] & INDOPTION_DESC) == 0;
+	YbcPgExpr  *col_values = palloc(sizeof(YbcPgExpr) *
```
andrei-mart, line 1756:
```
if (is_for_precheck)
	return needs_recheck;
```	
can be here, just before this line.
The needs_recheck value is finalized, the only thing left to do is to create and bind the constants, which is skipped for precheck.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Yeah, cut and paste.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Isn't it the same? You are only asking to replace `subkeys[first_key_subkey]` with `subkeys[0]`?

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: I'm not sure. Maybe it was intended for distinguishing cases with full-key equality regardless of the direction/strategy. I think it's worth simplifying it even for efficiency reasons.

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: Agreed

src/postgres/src/backend/access/yb_access/yb_scan.c
egladysh: I think we should be good. The actual datum should be int4. I verified `prorettype => int4` for yb_hash_code in `include/catalog/pg_proc.dat`

andrei-mart:
Trigger Jenkins

andrei-mart (approved):
LGTM, merge pending the test results

egladysh:
@andrei-mart Just in case, I don't have access to the jenkins logs.

andrei-mart:
> @andrei-mart Just in case, I don't have access to the jenkins logs.

Some jobs are still running.
I'll let you know the details if any tests require attentions.

andrei-mart:
Failing tests: 
- org.yb.pgsql.TestPgRegressYbBitmapScans#testPgRegressYbBitmapScans
To rerun it locally:
./yb_build.sh --java-test 'org.yb.pgsql.TestPgRegressYbBitmapScans#testPgRegressYbBitmapScans'
[yb.orig.bitmap_scans, yb.orig.bitmap_scans_system] mismatch between expected and actual results
As far as I can see, the yb.orig.bitmap_scans shows less rows removed by recheck, likely due to better filter pushed down.
The yb.orig.bitmap_scans_system shows different number of actual rows in some queries, possible correctness issue, needs deeper look.

egladysh:
I'll try to repro it locally.

egladysh:
I can repro it locally. As you said, there is definitely an issue with expected output due to better pushdown. It's set as: 
``` 
Rows Removed by Index Recheck: 9
  ->  BitmapOr (actual rows=10 loops=1)
        ->  Bitmap Index Scan ... (actual rows=10 loops=1)
```

but actual:

```
  Rows Removed by Index Recheck: 1
  ->  BitmapOr (actual rows=2 loops=1)
        ->  Bitmap Index Scan ... (actual rows=2 loops=1)
```

Correctness seems to be fine in this case. I'll try to verify/regenerate all expected outputs before pushing a fix.

andrei-mart:
I meant yb.orig.bitmap_scans is good, only update of the expected results is needed.
The yb.orig.bitmap_scans_system is suspicious one.

egladysh:
Regenerated the expected output for the failing test, ran other tests with `Rows Removed` and `Index Scan` in the outputs that seems to pass fine. After some back and forth with AI on `bitmap_scans_system`:
```
Yes, I agree it’s suspicious as a test, but not as a branch diff.

  yb.orig.bitmap_scans_system depends on exact contents of pg_authid, including roles like yb_global_views_user, and exact EXPLAIN ANALYZE row counts from system catalogs. That makes it sensitive to stale initdb state. After ./yb_build.sh debug reinitdb, the generated
  result matches the checked-in expected file byte-for-byte.
```
I'm not sure if it has anything to do with this PR and it seems to pass after reinitdb?

andrei-mart:
Oh, right, yb.orig.bitmap_scans_system was my local failure, it did not fail in Jenkins.
I jump from branch to branch, and our scripts do not run reinitdb automatically.

I'll kick off the tests again.

andrei-mart:
Trigger Jenkins

PR #31058, opened 2026-04-13
Filter cdc_state table scan by stream ID in GetTabletIdsToPoll

jtanza:
## Problem

GetTabletIdsToPoll() calls cdc_state_table_->GetTableRange() which performs an unfiltered scan of the entire cdc_state table, then filters by stream_id in application code:

 ```cpp
   if (entry.key.stream_id != stream_id) {
       continue;
   }
 ```

On clusters with many CDC streams and tables, the cdc_state table can have tens of thousands of rows. Tserver logs confirm the scale of the problem:

 ```
   I0403 14:41:43.584196 cdc_service.cc:3195] Read 87128 records from cdc_state
 ```

 ```
   W0403 14:48:09.320588 yb_rpc.cc:383] Call yb.cdc.CDCService.GetTabletListToPollForCDC ... took 56875ms
 ```

Every GetTabletListToPollForCDC RPC scans all 87K+ rows to find the handful belonging to the requested stream, resulting in 30-57 second latencies per call.

## Changes

 - CDCStateTable::GetTableRange() — New overload accepts an xrepl::StreamId and constructs a TableFilter that pushes a range condition (>= streamId AND < streamId~) down to the storage layer via TableIteratorOptions::filter. The range handles both non-colocated rows (where the column stores a plain stream UUID) and colocated rows (where it stores streamId_tableId).
 - CDCStateTableRange — Constructor now accepts an optional client::TableFilter, passed through to TableIteratorOptions. Default is empty (no filter), preserving existing behavior for all other callers.
 - CDCServiceImpl::GetTabletIdsToPoll() — Uses the filtered overload and removes the now-redundant in-memory stream_id checks from both iteration loops.

No changes to the unfiltered GetTableRange signature — all other callers (UpdatePeersAndMetrics, etc.) are unaffected

### DIFF
--- a/src/yb/cdc/cdc_service.cc
+++ b/src/yb/cdc/cdc_service.cc
@@ -3252,7 +3252,7 @@ Status CDCServiceImpl::GetTabletIdsToPoll(
   Status iteration_status;
   // Filter the cdc_state scan by stream_id to avoid reading all rows.
   // Without this filter, the full table (87K+ rows on large clusters) is scanned
-  // on every call, causing 30-60s latencies.
+  // on every call.
   auto entries = VERIFY_RESULT(cdc_state_table_->GetTableRange(
       CDCStateTableEntrySelector()
           .IncludeCheckpoint()
--- a/src/yb/cdc/cdc_service.cc
+++ b/src/yb/cdc/cdc_service.cc
@@ -3250,12 +3250,16 @@ Status CDCServiceImpl::GetTabletIdsToPoll(
   std::set<TabletId> polled_tablets;
 
   Status iteration_status;
+  // Filter the cdc_state scan by stream_id to avoid reading all rows.
+  // Without this filter, the full table (87K+ rows on large clusters) is scanned
+  // on every call, causing 30-60s latencies.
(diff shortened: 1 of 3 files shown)


asrinivasanyb:
> Every GetTabletListToPollForCDC RPC scans all 87K+ rows to find the handful belonging to the requested stream, resulting in 30-57 second latencies per call.

With this change what are the latency numbers ?

asrinivasanyb:
Given that tablet_id is the hash primary key, it would be the primary sort key in the SST file. Thus, even though stream_id filter is pushed down to the storage layer, my expectation was that all tablets will still need to be looked up. Hence, not sure how significant the performance gain will be to have this.

asrinivasanyb:
The motivation for this change is to reduce the latency of the GetTabletListToPollForCDC RPC - this RPC currently takes 1 table as input - and for each table, the entire cdc_state table is scanned once. This RPC is called repeatedly for each table that is to be polled. 

A better approach would be to support a batch version of this RPC where all the tables of interest are passed in the request and there is just 1 cdc_state table scan across all the tables

PR #31120, opened 2026-04-15
[DOC-1262] DDL operation best practices

ddhodge:
DOC-1291

### DIFF
--- a/docs/content/stable/best-practices-operations/administration.md
+++ b/docs/content/stable/best-practices-operations/administration.md
@@ -12,6 +12,8 @@ menu:
 aliases:
   - /stable/develop/best-practices/administration/
 type: docs
+rightNav:
+  hideH3: true
 ---
 
 Database administrators can fine-tune YugabyteDB deployments for better reliability, performance, and operational efficiency by following targeted best practices. This guide outlines key recommendations for configuring single-AZ environments, optimizing memory use, accelerating CI/CD tests, and safely managing concurrent DML and DDL operations. These tips are designed to help DBAs maintain stable, scalable YSQL clusters in real-world and test scenarios alike.
@@ -64,7 +66,7 @@ Most DDL statements complete quickly, so this is typically not a significant iss
 
 Concurrent Data Definition Language (DDL) operations can cause inconsistencies or temporary unavailability. Potential outcomes include catalog inconsistencies, tables not registered in the schema, issues with restoring from backups taken during an inconsistent state, and occasional column modification issues.
 
-YugabyteDB provides [table-level locking](../../architecture/transactions/concurrency-control/#table-level-locks) {{<tags/feature/ea idea="1114">}} for DDL operations, which queues conflicting DDL instead of allowing concurrent execution. However, if table-level locking is not enabled (the default), concurrent DDL statements can proceed simultaneously and race against each other, potentially leaving the two metadata layers out of sync. As a result, all DDL serialization must be enforced at the application and operational level.
+YugabyteDB provides [table-level locking](../../architecture/transactions/concurrency-control/#table-level-locks) {{<tags/feature/ea idea="1114">}} for DDL operations, which queues conflicting DDL instead of allowing concurrent execution. However, if table-level locking is not enabled (the default), concurrent DDL statements can proceed simultaneously and race against each other, potentially leaving the YSQL system catalog and the YB-Master catalog out of sync. As a result, all DDL serialization must be enforced at the application and operational level.
 
 All DDL statements targeting the same database must be executed sequentially, one at a time, from a single database connection. DDL statements that operate on shared objects (roles, tablespaces) affect all databases in the cluster and must also be serialized. DDL statements that affect entities in different databases can be run concurrently.
 
@@ -77,6 +79,7 @@ All DDL statements targeting the same database must be executed sequentially, on
 - Run DDL and backup jobs separately. DDL verification states can block backup and restore operations. 
 - Monitor YB-Master logs (or universe Health in YugabyteDB Anywhere) for DDL verification states and DDL atomicity alerts.
 - Enable Point-in-Time Recovery (PITR) as a safety net before performing schema changes, to allow [rollback if DDL operations](../../explore/cluster-management/point-in-time-recovery-ysql/#undo-metadata-changes) leave the database in an inconsistent state.
+- ANALYZE is treated as a DDL operation in YugabyteDB and will conflict with other concurrent DDLs. Run ANALYZE on individual tables.
 
 ## Preload PostgreSQL system catalog entries into the local catalog cache
 
--- a/docs/content/stable/best-practices-operations/administration.md
+++ b/docs/content/stable/best-practices-operations/administration.md
@@ -62,9 +62,21 @@ Most DDL statements complete quickly, so this is typically not a significant iss
 
 ## Concurrent DDL during a DDL operation
 
-DDL statements that affect entities in different databases can be run concurrently. However, for DDL statements that impact the same database, it is recommended to execute them sequentially.
+Concurrent Data Definition Language (DDL) operations can cause inconsistencies or temporary unavailability. Potential outcomes include catalog inconsistencies, tables not registered in the schema, issues with restoring from backups taken during an inconsistent state, and occasional column modification issues.
 
-DDL statements that relate to shared objects, such as roles or tablespaces, are considered as affecting all databases in the cluster, so they should also be run sequentially.
+YugabyteDB provides [table-level locking](../../architecture/transactions/concurrency-control/#table-level-locks) {{<tags/feature/ea idea="1114">}} for DDL operations, which queues conflicting DDL instead of allowing concurrent execution. However, if table-level locking is not enabled (the default), concurrent DDL statements can proceed simultaneously and race against each other, potentially leaving the two metadata layers out of sync. As a result, all DDL serialization must be enforced at the application and operational level.
+
+All DDL statements targeting the same database must be executed sequentially, one at a time, from a single database connection. DDL statements that operate on shared objects (roles, tablespaces) affect all databases in the cluster and must also be serialized. DDL statements that affect entities in different databases can be run concurrently.
+
+### Best practices
+
+- Execute all DDLs sequentially from a single connection. Use a dedicated, non-pooled connection for schema migrations.
+- Wait for each DDL to fully complete before issuing the next statement.
+- Implement client-side retry logic for schema mismatch and catalog version mismatch errors in any [DML that may overlap with DDL windows](#concurrent-dml-during-a-ddl-operation).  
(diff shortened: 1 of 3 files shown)


docs/content/stable/best-practices-operations/administration.md
```diff
@@ -50,21 +52,30 @@ You can set certain flags to increase performance using YugabyteDB in CI and CD
...
 
 ## Concurrent DDL during a DDL operation
 
-DDL statements that affect entities in different databases can be run concurrently. However, for DDL statements that impact the same database, it is recommended to execute them sequentially.
+Concurrent Data Definition Language (DDL) operations are currently unsupported. All DDL statements targeting the same database must be executed sequentially, one at a time, from a single database connection. DDL statements that operate on shared objects (roles, tablespaces) affect all databases in the cluster and must also be serialized. DDL statements that affect entities in different databases can be run concurrently.
 
-DDL statements that relate to shared objects, such as roles or tablespaces, are considered as affecting all databases in the cluster, so they should also be run sequentially.
+To have YugabyteDB queue conflicting DDL operations automatically, enable [table-level locking](../../architecture/transactions/concurrency-control/#table-level-locks) {{<tags/feature/ea idea="1114">}}.
```
iSignal: Just enabling object locks is not enough for running concurrent DDL, so we can't say this.

docs/content/stable/best-practices-operations/administration.md
```diff
@@ -50,21 +52,30 @@ You can set certain flags to increase performance using YugabyteDB in CI and CD
...
 
 - Operate with the old schema prior to the DDL.
 - Operate with the new schema after the DDL completes.
-- Encounter temporary errors such as `schema mismatch errors` or `catalog version mismatch`. It is recommended for the client to [retry such operations](https://www.yugabyte.com/blog/retry-mechanism-spring-boot-app/) whenever possible.
+- Encounter temporary errors such as `schema mismatch errors` or `catalog version mismatch`. The client should [retry such operations](https://www.yugabyte.com/blog/retry-mechanism-spring-boot-app/) whenever possible.
 
 Most DDL statements complete quickly, so this is typically not a significant issue in practice. However, [certain kinds of ALTER TABLE DDL statements](../../api/ysql/the-sql-language/statements/ddl_alter_table/#alter-table-operations-that-involve-a-table-rewrite) involve making a full copy of the table(s) whose schema is being modified. For these operations, it is not recommended to run any concurrent DML statements on the table being modified by the `ALTER TABLE`, as the effect of such concurrent DML may not be reflected in the table copy.
```
iSignal: We should also call out `create index nonconcurrently` here as not safe to do or link to the create index docs where we call this out.

iSignal (changes requested):

docs/content/stable/best-practices-operations/administration.md
```diff
@@ -50,21 +52,32 @@ You can set certain flags to increase performance using YugabyteDB in CI and CD
 
 ## Concurrent DML during a DDL operation
 
-In YugabyteDB, DML is allowed to execute while a DDL statement modifies the schema that is accessed by the DML statement. For example, an `ALTER TABLE <table> .. ADD COLUMN` DDL statement may add a new column while a `SELECT * from <table>` executes concurrently on the same relation. In PostgreSQL, this is typically not allowed because such DDL statements take a table-level exclusive lock that prevents concurrent DML from executing. (Support for similar behavior in YugabyteDB is being tracked in issue {{<issue 11571>}}.)
+To prevent DML and DDL operations executing concurrently, enable [table-level locking](../../architecture/transactions/concurrency-control/#table-level-locks) {{<tags/feature/ea idea="1114">}}.
```
iSignal: It may be better to start by explaining the default situation.

By default, YBDB doesn't restrict DML and DDL concurrency, so this means DML can use old or new schema and cause the following problems
1. Errors like "schema version mismatch" / "catalog version mismatch". Applications must be prepared to retry such errors.
2. Be unsafe (results in inconsistencies) in cases like alter tables which cause rewrites and create index nonconcurrently. 

To avoid this, users can manually enable table locks.

docs/content/stable/best-practices-operations/administration.md
iSignal: We can update this doc separately at the end of 2025.2.3 when concurrent DDL feature is shipped. 

For now, maybe it is enough to use the same text as the previous one and then say something like 
It is recommended to run DDL serially from a single (non-pooled) connection.

docs/content/stable/best-practices-operations/administration.md
```diff
@@ -50,21 +52,32 @@ You can set certain flags to increase performance using YugabyteDB in CI and CD
...
+
+Enforce DDL serialization at the application and operational level:
 
-DDL statements that relate to shared objects, such as roles or tablespaces, are considered as affecting all databases in the cluster, so they should also be run sequentially.
+- Execute all DDLs sequentially from a single connection. Use a dedicated, non-pooled connection for schema migrations.
+- Wait for each DDL to fully complete before issuing the next statement.
+- Implement client-side retry logic for schema mismatch and catalog version mismatch errors in any [DML that may overlap with DDL windows](#concurrent-dml-during-a-ddl-operation).  
+- Schedule DDL during maintenance windows to minimize overlap with application DML traffic, backup jobs, and other administrative operations.
+- Run DDL and backup jobs separately. DDL verification states can block backup and restore operations. 
+- Enable Point-in-Time Recovery (PITR) as a safety net before performing schema changes, to allow [rollback](../../explore/cluster-management/point-in-time-recovery-ysql/#undo-metadata-changes) if DDL operations leave the database in an inconsistent state.
```
iSignal: The PITR part doesn't seem to belong here. In general, PITR is a good safety net but if users follow the advice here, PITR is not relevant.

docs/content/stable/best-practices-operations/administration.md
```diff
@@ -50,21 +52,32 @@ You can set certain flags to increase performance using YugabyteDB in CI and CD
...
+To have YugabyteDB queue conflicting DDL operations automatically, enable [table-level locking](../../architecture/transactions/concurrency-control/#table-level-locks) {{<tags/feature/ea idea="1114">}}.
+
+Enforce DDL serialization at the application and operational level:
 
-DDL statements that relate to shared objects, such as roles or tablespaces, are considered as affecting all databases in the cluster, so they should also be run sequentially.
+- Execute all DDLs sequentially from a single connection. Use a dedicated, non-pooled connection for schema migrations.
+- Wait for each DDL to fully complete before issuing the next statement.
+- Implement client-side retry logic for schema mismatch and catalog version mismatch errors in any [DML that may overlap with DDL windows](#concurrent-dml-during-a-ddl-operation).  
+- Schedule DDL during maintenance windows to minimize overlap with application DML traffic, backup jobs, and other administrative operations.
+- Run DDL and backup jobs separately. DDL verification states can block backup and restore operations.
```
iSignal: Given we also document this feature https://docs.yugabyte.com/stable/manage/backup-restore/#:~:text=In%20v2025.2.1%20and%20later%2C%20taking%20YSQL%20backups%20during%20DDL%20operations%20is%20supported%20by%20default
should we make this advice "run ddl and backup separately" only relevant to older versions?

iSignal (changes requested):

docs/content/stable/best-practices-operations/administration.md
```diff
@@ -50,21 +52,28 @@ You can set certain flags to increase performance using YugabyteDB in CI and CD
...
-- Encounter temporary errors such as `schema mismatch errors` or `catalog version mismatch`. It is recommended for the client to [retry such operations](https://www.yugabyte.com/blog/retry-mechanism-spring-boot-app/) whenever possible.
+To avoid this, you can manually enable [table-level locking](../../architecture/transactions/concurrency-control/#table-level-locks) {{<tags/feature/ea idea="1114">}}.
 
 Most DDL statements complete quickly, so this is typically not a significant issue in practice. However, [certain kinds of ALTER TABLE DDL statements](../../api/ysql/the-sql-language/statements/ddl_alter_table/#alter-table-operations-that-involve-a-table-rewrite) involve making a full copy of the table(s) whose schema is being modified. For these operations, it is not recommended to run any concurrent DML statements on the table being modified by the `ALTER TABLE`, as the effect of such concurrent DML may not be reflected in the table copy.
 
+Nonconcurrent index builds are not safe to perform while there are ongoing changes to the main table; for more information, see [Concurrent index creation](../../api/ysql/the-sql-language/statements/ddl_create_index/#semantics).
```
iSignal: We need to expand this to a few other statements like


CREATE TABLE ... PARTITION OF
ALTER TABLE ... ATTACH PARTITION 
ALTER TABLE ... ADD CONSTRAINT

and possibly also document on the respective pages just like we do for create index

docs/content/stable/best-practices-operations/administration.md
```diff
@@ -50,21 +52,28 @@ You can set certain flags to increase performance using YugabyteDB in CI and CD
...
+By default, YugabyteDB doesn't restrict DML and DDL concurrency. As a result, a DML statement can potentially operate using the old (prior to the DDL) or new schema; for example, an `ALTER TABLE <table> .. ADD COLUMN` DDL statement may add a new column while a `SELECT * from <table>` executes concurrently on the same relation. This can cause the following problems:
 
-In YugabyteDB, when a DDL modifies the schema of tables that are accessed by concurrent DML statements, the DML statement may do one of the following:
+- Errors such as `schema mismatch errors` or `catalog version mismatch`. The client should [retry such operations](https://www.yugabyte.com/blog/retry-mechanism-spring-boot-app/) whenever possible.
+- Inconsistencies. For example, cases such as alter tables that cause rewrites, or creating indexes nonconcurrently.
```
iSignal: Might be worth expanding to include the following

1. A table rewrite like `ALTER TABLE ADD COLUMN c int DEFAULT random()` may create a rewritten table that does not include some concurrently written rows in the old table. 
2. `CREATE INDEX NONCONCURRENTLY` may create an index that does not include some concurrently written rows from the old table.
3. `ALTER TABLE ADD CONSTRAINT c NOT NULL` may still have some NULL entries for c if they were written concurrent to this DDL. 
4. Concurrent writes during `ALTER TABLE partition_parent ATTACH child` or `CREATE TABLE partition_child PARTITION OF parent` for the newly attached partition column range may end up in the default partition.

To avoid this, you can either
(a) manually pause writes during the DDL workload, or
(b) enable table level locking.

For specific DDL like `ALTER TABLE ADD CONSTRAINT`, an alternate workaround is to 
```
ALTER TABLE ADD CONSTRAINT ... INVALID
ALTER TABLE ... VALIDATE CONSTRAINT
```
in two steps. This should be safe to do even without table locking enabled.

For `ALTER TABLE partition_parent ATTACH child`, you can first add CHECK constraints to the existing partitions that exclude the newly attached range of partition columns from the existing partitions. This can guarantee that no rows are present in or can be inserted for this range to the existing partitions before adding the new partition.

PR #31126, opened 2026-04-16
[DOC-1248] Time travel queries do not work on temp tables

ddhodge:
@netlify /stable/manage/backup-restore/time-travel-query/


aishwarya24 (approved):
LGTM

yamen-haddad (changes requested):
I would prefer creating a limitation section just above the example (or even below it at the end of the page).
I will then add this statement as a limitation.
We are working with QA on identifying these limitations, and I expect we will add more in the future.

Issue #30946, opened 2026-04-02
[YSQL][DistributedTracing] Add parsing for comments at the end of the queries


IshanChhangani:
Jira Link: [DB-20846](https://yugabyte.atlassian.net/browse/DB-20846)
### Description

Currently:
`/*traceparent='00-...-01'*/ SELECT 1; ` queries with comment at the start are traced but
` SELECT 1 /*traceparent='00-...-01'*/  ` queries with comment at the end are not. Add support for this.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-20846]: https://yugabyte.atlassian.net/browse/DB-20846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

rahulb-yb:
@egladysh mentioned he would be responsible for the code that fixes this issue - we can have Ishan review his changes when the PR is ready!

khosrow:
PR supporting traceparent comment at the query start and end: https://github.com/yugabyte/yugabyte-db/pull/31146

PR #31145, opened 2026-04-17
[#30946] YSQL: Support traceparent comment anywhere in query

egladysh:
Currently, OpenTelemetry tracing via SQL comment (/*traceparent='...'*/) only works when the comment appears at the very start of the query. This change scans all comment blocks in the query, so trailing and mid-query comments should work equally well. 
The first comment containing a valid traceparent field wins. Also adds TestTraceparentCommentAppended unit test for various cases.


egladysh:
We don't need to support comments anywhere only at the start and at the end as was the original intention. I'll open another PR.

PR #31344, opened 2026-04-29
Prototype batching lock intent for skip locked

emilienoel:
# Batched `SKIP LOCKED` Walkthrough

This document walks through the rebased change and identifies whether each step happens in the **YSQL** layer or on the **tserver** side.

Definitions:

- **YSQL**: PostgreSQL/YSQL layer, including executor and PgGate client-side API.
- **tserver**: tablet server / DocDB side.

---

## 1. Enable batching via GUC

**Where:** **YSQL**

Files:

```text
src/postgres/src/backend/utils/misc/guc.c
src/postgres/src/include/utils/guc.h
```

Adds:

```c
yb_skip_locked_batch_size
```

Default:

```c
32
```

Meaning:

- `1` disables the optimization.
- `>1` allows YSQL executor to prefetch multiple candidate rows for `SKIP LOCKED`.

---

## 2. Detect eligible `SKIP LOCKED` query

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/executor/nodeLockRows.c
```

In `ExecLockRows`, the batch path is used only when:

```c
node->yb_are_row_marks_for_yb_rels &&
yb_skip_locked_batch_size > 1 &&
list_length(node->lr_arowMarks) == 1 &&
erm->waitPolicy == LockWaitSkip
```

So YSQL decides whether to use the optimization.

This means the query must be a YB relation with one row mark and `SKIP LOCKED`.

---

## 3. Prefetch candidate rows from the child plan

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/executor/nodeLockRows.c
```

Function:

```c
ExecLockRowsBatchSkipLocked(...)
```

YSQL pulls up to `yb_skip_locked_batch_size` rows from the child plan:

```c
slot = ExecProcNode(outerPlan);
```

For each candidate row, YSQL extracts the `ybctid`:

```c
datum = ExecGetJunkAttribute(slot, aerm->ctidAttNo, &isNull);
```

Then stores:

```c
batch_ybctids[batch_count] = datumCopy(...);
batch_tuples[batch_count] = ExecCopySlotHeapTuple(slot);
```

At this point, YSQL has a local batch of candidate rows.

---

## 4. Send candidate `ybctid`s to PgGate

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/access/yb_access/yb_scan.c
```

Function:

```c
YBCLockTupleBatch(...)
```

YSQL creates a new select statement:

```c
YbcPgStatement ybc_stmt = YbNewSelect(...);
```

Then it adds each candidate `ybctid`:

```c
YBCPgDmlAddBatchYbctidArg(ybc_stmt, data, len);
```

This is still YSQL-side code preparing the request.

---

## 5. PgGate appends batch arguments to the read request

**Where:** **YSQL / PgGate client side**

Files:

```text
src/yb/yql/pggate/ybc_pggate.cc
src/yb/yql/pggate/pggate.cc
src/yb/yql/pggate/pg_dml_read.cc
```

Call chain:

```c
YBCPgDmlAddBatchYbctidArg(...)
```

calls:

```cpp
PgApiImpl::DmlAddBatchYbctidArg(...)
```

which calls:

```cpp
PgDmlRead::AddBatchYbctidArg(...)
```

That appends each candidate to:

```cpp
read_req_->add_batch_arguments();
```

So the request sent to the tserver contains:

```text
batch_arguments = [ybctid0, ybctid1, ybctid2, ...]
```

This is PgGate building the tserver request, but it is on the YSQL side of the boundary.

---

## 6. Execute/fetch the YSQL statement, causing RPC to tserver

**Where:** starts in **YSQL**, crosses to **tserver**

File:

```text
src/postgres/src/backend/access/yb_access/yb_scan.c
```

YSQL calls:

```c
YBCPgExecSelect(ybc_stmt, &exec_params);
YBCPgDmlFetch(ybc_stmt, 0, NULL, NULL, NULL, &has_data);
```

The fetch is what forces the PgGate operation to perform the remote read RPC.

Boundary:

```text
YSQL/PgGate  --->  tserver
```

---

## 7. tserver detects special batch `SKIP LOCKED` request

**Where:** **tserver**

File:

```text
src/yb/tserver/read_query.cc
```

In `ReadQuery::DoPerform`, tserver checks:

```cpp
has_row_mark &&
!serializable_isolation &&
req_->pgsql_batch_size() == 1 &&
pgsql_read.wait_policy() == WAIT_SKIP &&
pgsql_read.batch_arguments_size() > 1
```

If true, tserver chooses the new path:

```cpp
TryLockBatchArg(0, ...);
```

So the tserver, not YSQL, decides how to process the batched lock request internally.

---

## 8. tserver tries to lock one batch argument at a time

**Where:** **tserver**

File:

```text
src/yb/tserver/read_query.cc
```

Function:

```cpp
ReadQuery::TryLockBatchArg(...)
```

For each candidate index, tserver builds a write operation that creates read-lock intents for only that candidate.

It calls:

```cpp
tablet_ptr->CreateReadIntentForBatchArg(
    isolation_level,
    pgsql_read,
    batch_arg_index,
    &write_batch);
```

Then it submits:

```cpp
peer->WriteAsync(std::move(query));
```

This is tserver-side async write/conflict resolution.

---

## 9. Tablet builds intents for exactly one candidate

**Where:** **tserver**

Files:

```text
src/yb/tablet/tablet.cc
src/yb/tablet/tablet.h
src/yb/docdb/pgsql_operation.cc
src/yb/docdb/pgsql_operation.h
```

Call chain:

```cpp
Tablet::CreateReadIntentForBatchArg(...)
```

calls:

```cpp
docdb::GetIntentsForBatchArg(...)
```

That function picks one candidate:

```cpp
const auto& batch_argument = request.batch_arguments(batch_arg_index);
```

and creates lock intents only for that candidate.

This is important: tserver does **not** lock all batch arguments at once.

---

## 10. tserver handles lock success or conflict

**Where:** **tserver**

File:

```text
src/yb/tserver/read_query.cc
```

In the callback from `WriteAsync`:

### If the lock succeeds

```cpp
self->first_locked_batch_arg_index_ = batch_arg_index;
peer->Enqueue(self.get());
```

The tserver records the winning index and proceeds to read the row.

### If the lock conflicts / transaction error occurs

```cpp
TransactionError(status).value() != TransactionErrorCode::kNone
```

Then because this is `SKIP LOCKED`, tserver skips this candidate and tries the next one:

```cpp
self->TryLockBatchArg(batch_arg_index + 1, ...);
```

### If all candidates conflict

Eventually:

```cpp
batch_arg_index >= total_batch_args
```

Then tserver sets:

```cpp
first_locked_batch_arg_index_ = -1;
```

and proceeds to read phase with no winner.

---

## 11. tserver reads only the winning candidate

**Where:** **tserver**

File:

```text
src/yb/tserver/read_query.cc
```

Function:

```cpp
ReadQuery::DoReadImpl()
```

The original request still contains all candidates:

```text
batch_arguments = [row0, row1, row2, row3]
```

But after locking, tserver creates a modified effective request.

If there is a winner:

```cpp
modified_req->clear_batch_arguments();
*modified_req->add_batch_arguments() = batch_argument_for_winner;
effective_req = modified_req;
```

So the actual read phase reads only the winning row.

If no candidate was locked:

```cpp
modified_req->clear_batch_arguments();
effective_req = modified_req;
```

So the read returns zero rows.

---

## 12. tserver populates response metadata

**Where:** **tserver**

File:

```text
src/yb/tserver/read_query.cc
```

After reading, tserver sets:

```cpp
result.response->set_batch_arg_count(pgsql_read_req.batch_arguments_size());
```

That tells PgGate:

> All batch arguments were consumed/tried.

If there was a winner, tserver also sets:

```cpp
result.response->set_first_locked_batch_arg_index(first_locked_batch_arg_index_);
```

This field was added in:

```text
src/yb/common/pgsql_protocol.proto
```

as:

```proto
optional int32 first_locked_batch_arg_index = 22;
```

So the tserver response carries the winning candidate index back to YSQL.

Boundary:

```text
tserver  --->  YSQL/PgGate
```

---

## 13. PgGate reads the winner index from response

**Where:** **YSQL / PgGate client side**

Files:

```text
src/yb/yql/pggate/pg_doc_op.cc
src/yb/yql/pggate/pg_dml_read.cc
src/yb/yql/pggate/pggate.cc
src/yb/yql/pggate/ybc_pggate.cc
```

Call chain:

```c
YBCPgDmlGetFirstLockedBatchArgIndex(...)
```

calls:

```cpp
PgApiImpl::DmlGetFirstLockedBatchArgIndex(...)
```

then:

```cpp
PgDmlRead::GetFirstLockedBatchArgIndex()
```

then:

```cpp
PgDocReadOp::GetFirstLockedBatchArgIndex()
```

which reads:

```cpp
resp->first_locked_batch_arg_index()
```

If absent, it returns:

```cpp
-1
```

---

## 14. YSQL maps the winner index to a PostgreSQL lock result

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/access/yb_access/yb_scan.c
```

Back in:

```c
YBCLockTupleBatch(...)
```

YSQL gets:

```c
winner
```

Then:

```c
if (winner >= 0 && winner < count)
{
    *locked_index = winner;
    res = TM_Ok;
}
else
{
    *locked_index = -1;
    res = TM_WouldBlock;
}
```

So YSQL converts tserver's response into PostgreSQL executor semantics:

- winner found: `TM_Ok`
- no winner: `TM_WouldBlock`

---

## 15. YSQL returns the winning tuple

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/executor/nodeLockRows.c
```

Back in:

```c
ExecLockRowsBatchSkipLocked(...)
```

If the batch lock returned `TM_Ok`, YSQL restores the winning tuple into the result slot:

```c
ExecForceStoreHeapTuple(batch_tuples[locked_index], result_slot, true);
```

Then it returns that tuple to the upper executor nodes.

So if the query was:

```sql
SELECT * FROM jobs FOR UPDATE SKIP LOCKED LIMIT 1;
```

this is the point where the selected row goes back up the executor tree.

---

## 16. YSQL saves candidates after the winner as leftovers

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/executor/nodeLockRows.c
```

Suppose the batch was:

```text
[row0, row1, row2, row3]
```

and tserver locked:

```text
row1
```

Then:

- `row0` was tried and skipped.
- `row1` is returned.
- `row2`, `row3` were prefetched but not tried.

YSQL stores candidates after the winner:

```c
node->yb_batch_leftover_tuples
node->yb_batch_leftover_ybctids
```

These are used on later calls to `ExecLockRows`.

---

## 17. YSQL tries leftovers before scanning more rows

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/executor/nodeLockRows.c
```

At the top of `ExecLockRows`, before fetching a new row from the child plan, YSQL checks:

```c
if (node->yb_batch_leftover_count > node->yb_batch_leftover_idx)
```

If leftovers exist, it calls:

```c
ExecLockRowsTryLeftover(...)
```

That tries each leftover using the existing single-row path:

```c
YBCLockTuple(...)
```

The leftover entries themselves do not store a table id or relation key. They only store parallel
arrays of:

```text
ybctid
tuple
```

The table is recovered from the same `LockRowsState` row mark:

```c
ExecAuxRowMark *aerm = (ExecAuxRowMark *) linitial(node->lr_arowMarks);
ExecRowMark *erm = aerm->rowmark;
```

and the lock call uses:

```c
YBCLockTuple(erm->relation, ybctid, erm->markType, LockWaitSkip, estate);
```

This is safe because the batch optimization is only enabled when:

```c
list_length(node->lr_arowMarks) == 1
```

So every prefetched candidate and every leftover belongs to the single row-marked YB relation for
this `LockRowsState`. For multi-row-mark queries, such as joins with multiple locked tables, the
batch path is not used because a bare leftover `ybctid` would not be enough to identify which
relation to lock.

So after the first batch winner, later prefetched-but-untried rows are not lost.

---

## 18. YSQL cleans up leftovers

**Where:** **YSQL**

File:

```text
src/postgres/src/backend/executor/nodeLockRows.c
```

In:

```c
ExecEndLockRows(...)
```

YSQL frees any unconsumed leftover tuples and ybctids.

---

# Compact end-to-end sequence

| Step | Layer | What happens |
|---:|---|---|
| 1 | **YSQL** | GUC `yb_skip_locked_batch_size` controls batch size |
| 2 | **YSQL** | `ExecLockRows` detects eligible `FOR UPDATE SKIP LOCKED` |
| 3 | **YSQL** | Executor prefetches candidate rows and extracts ybctids |
| 4 | **YSQL** | `YBCLockTupleBatch` builds a select request |
| 5 | **YSQL/PgGate** | PgGate adds ybctids as `batch_arguments` |
| 6 | **YSQL -> tserver** | Fetch triggers RPC |
| 7 | **tserver** | `ReadQuery::DoPerform` detects batched `SKIP LOCKED` |
| 8 | **tserver** | `TryLockBatchArg` tries candidate 0 |
| 9 | **tserver** | Tablet/DocDB creates intent for only that candidate |
| 10 | **tserver** | On conflict, try next candidate; on success, record winner |
| 11 | **tserver** | Read phase reads only winning candidate |
| 12 | **tserver** | Response includes `first_locked_batch_arg_index` |
| 13 | **tserver -> YSQL** | Response returns to PgGate |
| 14 | **YSQL/PgGate** | PgGate exposes winner index |
| 15 | **YSQL** | `YBCLockTupleBatch` maps winner to `TM_Ok` / `TM_WouldBlock` |
| 16 | **YSQL** | Executor returns winning tuple |
| 17 | **YSQL** | Executor saves later prefetched candidates as leftovers |
| 18 | **YSQL** | Future calls try leftovers before scanning more rows |

### DIFF
--- a/src/postgres/src/backend/executor/nodeLockRows.c
+++ b/src/postgres/src/backend/executor/nodeLockRows.c
@@ -31,7 +31,203 @@
 
 /* YB includes */
 #include "access/yb_scan.h"
+#include "utils/datum.h"
+#include "utils/guc.h"
 
+
+/*
+ * ExecLockRowsBatchSkipLocked
+ *
+ * Optimized batch path for YB SKIP LOCKED.  Pre-fetches up to batch_size
+ * candidate tuples from the scan, sends all their ybctids in a single RPC,
+ * and the tserver locks the first unlocked one.
+ *
+ * Returns the winning slot (with the locked tuple restored) or NULL if all
+ * candidates in this batch were skipped (caller should retry) or the scan
+ * is exhausted.
+ *
+ * Candidates after the winner are saved into node->yb_batch_leftover_* so
+ * subsequent ExecLockRows calls can try them via the per-row path without
+ * losing rows consumed from the scan.
+ *
+ * *scan_exhausted is set to true when ExecProcNode returned NULL, meaning
+ * there are no more rows to try.
+ */
+static TupleTableSlot *
+ExecLockRowsBatchSkipLocked(LockRowsState *node,
+							TupleTableSlot *first_slot,
+							ExecAuxRowMark *aerm,
+							ExecRowMark *erm,
+							EState *estate,
+							PlanState *outerPlan,
+							bool *scan_exhausted)
+{
+	int				batch_size = yb_skip_locked_batch_size;
+	Datum		   *batch_ybctids;
+	HeapTuple	   *batch_tuples;
+	int				batch_count = 0;
+	TupleTableSlot *slot = first_slot;
+	int				locked_index;
+	TM_Result		result;
+
+	*scan_exhausted = false;
+
+	batch_ybctids = (Datum *) palloc(sizeof(Datum) * batch_size);
+	batch_tuples = (HeapTuple *) palloc(sizeof(HeapTuple) * batch_size);
+
+	/*
+	 * Collect candidate ybctids.  The first slot was already fetched by the
+	 * caller; fetch up to batch_size - 1 more.
+	 */
+	for (;;)
+	{
+		Datum	datum;
+		bool	isNull;
+
+		datum = ExecGetJunkAttribute(slot, aerm->ctidAttNo, &isNull);
+		if (isNull)
+			elog(ERROR, "ctid/ybctid is NULL");
+
+		batch_ybctids[batch_count] = datumCopy(datum, false, -1);
+		batch_tuples[batch_count] = ExecCopySlotHeapTuple(slot);
+		batch_count++;
+
+		if (batch_count >= batch_size)
+			break;
+
+		slot = ExecProcNode(outerPlan);
+		if (TupIsNull(slot))
+		{
+			*scan_exhausted = true;
+			break;
+		}
+	}
+
+	if (batch_count == 0)
+	{
+		pfree(batch_ybctids);
+		pfree(batch_tuples);
+		return NULL;
+	}
+
+	/* Single batch RPC to lock the first available candidate. */
+	result = YBCLockTupleBatch(erm->relation, batch_ybctids, batch_count,
+							   erm->markType, estate, &locked_index);
+
+	slot = NULL;
+
+	if (result == TM_Ok && locked_index >= 0 && locked_index < batch_count)
+	{
+		/*
+		 * Winner found.  Restore the winning tuple into the outer plan's
+		 * result slot so the caller can return it.
+		 */
+		TupleTableSlot *result_slot = first_slot;
+
+		ExecForceStoreHeapTuple(batch_tuples[locked_index], result_slot, true);
+		slot = result_slot;
+
+		/* Free candidates before the winner (they were skipped by tserver). */
+		for (int i = 0; i < locked_index; i++)
+		{
+			pfree(DatumGetPointer(batch_ybctids[i]));
+			heap_freetuple(batch_tuples[i]);
+		}
+		/* Winner ybctid datum no longer needed (tuple is in the slot). */
+		pfree(DatumGetPointer(batch_ybctids[locked_index]));
+
+		/*
+		 * Save candidates after the winner as leftovers.  These were NOT tried
+		 * by the tserver (it stops at the first success), so subsequent
+		 * ExecLockRows calls will try them via the per-row path.
+		 */
+		int leftover_count = batch_count - locked_index - 1;
+		if (leftover_count > 0)
+		{
+			node->yb_batch_leftover_tuples = (HeapTuple *)
+				palloc(sizeof(HeapTuple) * leftover_count);
+			node->yb_batch_leftover_ybctids = (Datum *)
+				palloc(sizeof(Datum) * leftover_count);
+			for (int i = 0; i < leftover_count; i++)
+			{
+				int src = locked_index + 1 + i;
+				node->yb_batch_leftover_tuples[i] = batch_tuples[src];
+				node->yb_batch_leftover_ybctids[i] = batch_ybctids[src];
+			}
+			node->yb_batch_leftover_count = leftover_count;
+			node->yb_batch_leftover_idx = 0;
+		}
+	}
+	else
+	{
+		/* All candidates were skipped. Free everything. */
+		for (int i = 0; i < batch_count; i++)
+		{
+			pfree(DatumGetPointer(batch_ybctids[i]));
+			heap_freetuple(batch_tuples[i]);
+		}
+	}
+
+	pfree(batch_ybctids);
+	pfree(batch_tuples);
+	return slot;
+}
+
+/*
+ * ExecLockRowsTryLeftover
+ *
+ * Try to lock the next leftover candidate from a previous batch RPC using
+ * the per-row YBCLockTuple path.  Returns a slot if a row was locked, or
+ * NULL if the leftover was skipped (caller should try the next one or
+ * fall through to the scan).
+ */
+static TupleTableSlot *
+ExecLockRowsTryLeftover(LockRowsState *node,
+						ExecRowMark *erm,
+						EState *estate,
+						TupleTableSlot *result_slot)
+{
+	while (node->yb_batch_leftover_idx < node->yb_batch_leftover_count)
+	{
+		int		idx = node->yb_batch_leftover_idx++;
+		Datum	ybctid = node->yb_batch_leftover_ybctids[idx];
+		HeapTuple tup = node->yb_batch_leftover_tuples[idx];
+		TM_Result res;
+
+		res = YBCLockTuple(erm->relation, ybctid, erm->markType,
+						   LockWaitSkip, estate);
+
+		pfree(DatumGetPointer(ybctid));
+
+		if (res == TM_Ok)
+		{
+			ExecForceStoreHeapTuple(tup, result_slot, true);
+
+			/* Free remaining leftovers if this is the last one we need. */
+			/* (They'll be freed when the next call consumes or exhausts them.) */
+			return result_slot;
+		}
+
+		/* Skipped — free this tuple and try the next leftover. */
+		heap_freetuple(tup);
+	}
+
+	/* All leftovers exhausted. Clean up. */
+	if (node->yb_batch_leftover_tuples)
+	{
+		pfree(node->yb_batch_leftover_tuples);
+		node->yb_batch_leftover_tuples = NULL;
+	}
+	if (node->yb_batch_leftover_ybctids)
+	{
+		pfree(node->yb_batch_leftover_ybctids);
+		node->yb_batch_leftover_ybctids = NULL;
+	}
+	node->yb_batch_leftover_count = 0;
+	node->yb_batch_leftover_idx = 0;
+
+	return NULL;
+}
 
 /* ----------------------------------------------------------------
  *		ExecLockRows
@@ -59,6 +255,26 @@ ExecLockRows(PlanState *pstate)
 	 * Get next tuple from subplan, if any.
 	 */
 lnext:
+	/*
+	 * If we have leftover candidates from a previous batch SKIP LOCKED RPC,
+	 * try to lock them one-by-one before pulling new rows from the scan.
+	 * This is needed for LIMIT > 1: the batch path only returns the first
+	 * winner, and leftover candidates may still be lockable.
+	 */
+	if (node->yb_batch_leftover_count > node->yb_batch_leftover_idx)
+	{
+		ExecAuxRowMark *aerm = (ExecAuxRowMark *) linitial(node->lr_arowMarks);
+		ExecRowMark *erm = aerm->rowmark;
+		TupleTableSlot *leftover_result;
+
+		/* Use the outer plan's result slot — same slot the batch path uses. */
+		leftover_result = ExecLockRowsTryLeftover(node, erm, estate,
+												  outerPlan->ps_ResultTupleSlot);
+		if (leftover_result != NULL)
+			return leftover_result;
+		/* All leftovers exhausted, fall through to scan for more rows. */
+	}
+
 	slot = ExecProcNode(outerPlan);
 
 	if (node->yb_are_row_marks_for_yb_rels &&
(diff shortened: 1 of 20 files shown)


src/postgres/src/backend/executor/nodeLockRows.c
```diff
@@ -31,8 +31,204 @@
...
+			elog(ERROR, "ctid/ybctid is NULL");
+
+		batch_ybctids[batch_count] = datumCopy(datum, false, -1);
+		batch_tuples[batch_count] = ExecCopySlotHeapTuple(slot);
+		batch_count++;
+
+		if (batch_count >= batch_size)
+			break;
+
+		slot = ExecProcNode(outerPlan);
```
cuivienor, line 98: 🐛 **Bug**: The new path calls `ExecProcNode(outerPlan)` repeatedly until it has `yb_skip_locked_batch_size` candidates, which defaults to 32 and can be raised to 1024. For the common `Limit -> LockRows -> ordered/index scan` shape behind `SELECT ... FOR UPDATE SKIP LOCKED LIMIT 1`, the upper `LIMIT` asks for one row but this node can scan and copy dozens of full tuples before knowing whether the first row is lockable. That is a batch over-materialization risk: default batching can increase scanned rows, tuple materialization, and storage work even when the query only needs a single row.

**Suggestion:** Make batching adaptive or opt-in rather than default-on for every `SKIP LOCKED` query, or cap prefetching to the known remaining limit when available. Before widening the default, include measurements comparing scanned rows/storage RPCs for batch sizes 1, 32, and 1024 on contended and uncontended large-table queues.

---
_Binks review on behalf of [@cuivienor](https://github.com/cuivienor)._

src/postgres/src/backend/executor/nodeLockRows.c
```diff
@@ -31,8 +31,204 @@
...
+	if (batch_count == 0)
+	{
+		pfree(batch_ybctids);
+		pfree(batch_tuples);
+		return NULL;
+	}
+
+	/* Single batch RPC to lock the first available candidate. */
+	result = YBCLockTupleBatch(erm->relation, batch_ybctids, batch_count,
+							   erm->markType, estate, &locked_index);
```
cuivienor, line 115: 🐛 **Bug**: `ExecLockRowsBatchSkipLocked` can reach this call with `batch_count == 1` when the first candidate is the only row left in the child plan. A one-argument request does not take the server's new `batch_arguments_size() > 1` path, so no `first_locked_batch_arg_index` is set; `YBCLockTupleBatch` then treats the absent field as `TM_WouldBlock`. With the default batch size, `SELECT ... FOR UPDATE SKIP LOCKED` on a one-row table, or on the last remaining candidate, returns no row even though the row is lockable.

**Suggestion:** Bypass `YBCLockTupleBatch` when `batch_count == 1` and use the existing `YBCLockTuple` result, or have the server/client protocol explicitly report success for single-argument requests before enabling the batch path.

---
_Binks review on behalf of [@cuivienor](https://github.com/cuivienor)._

src/postgres/src/backend/executor/nodeLockRows.c
```diff
@@ -59,6 +255,26 @@ ExecLockRows(PlanState *pstate)
 	 * Get next tuple from subplan, if any.
 	 */
 lnext:
+	/*
+	 * If we have leftover candidates from a previous batch SKIP LOCKED RPC,
+	 * try to lock them one-by-one before pulling new rows from the scan.
+	 * This is needed for LIMIT > 1: the batch path only returns the first
+	 * winner, and leftover candidates may still be lockable.
+	 */
+	if (node->yb_batch_leftover_count > node->yb_batch_leftover_idx)
```
cuivienor, line 264: 🐛 **Bug**: Leftovers are consumed before fetching from the child plan, but `ExecReScanLockRows` does not clear the new `yb_batch_leftover_*` state. In a parameterized rescan, such as a `LATERAL` subquery selecting one `SKIP LOCKED` job per outer row with `LIMIT 1`, tuples prefetched for the previous parameter can be locked and returned before the rescanned child applies the new quals, producing rows for the wrong outer parameter.

**Suggestion:** Factor the leftover cleanup into a helper and call it from `ExecReScanLockRows` before rescanning the child; also reset the counters before any new batch overwrites the leftover arrays.

---
_Binks review on behalf of [@cuivienor](https://github.com/cuivienor)._

src/yb/tserver/read_query.cc
```diff
@@ -734,11 +773,39 @@ Result<ReadQuery::ReadRestartInfo> ReadQuery::DoReadImpl() {
...
+            if (arg_index++ == first_locked_batch_arg_index_) {
+              *modified_req->add_batch_arguments() = batch_argument;
+              break;
+            }
+          }
+          effective_req = modified_req;
+        } else if (first_locked_batch_arg_index_ == -1) {
+          // All candidates were skipped -- return empty result.
+          modified_req = req_->arena().NewArenaObject<PgsqlReadRequestMsg>(pgsql_read_req);
+          modified_req->clear_batch_arguments();
```
cuivienor, line 800: 🐛 **Bug**: In the no-winner case this clears only `batch_arguments` before calling `HandlePgsqlReadRequest`. The request still carries the scalar `ybctid_column_value` populated by the batch builder, so the storage layer falls through to the scalar ybctid read path and performs a point read for a row that the lock phase already determined should be skipped. On a hot `SKIP LOCKED` queue where every candidate is contended, that adds a storage read to every fully skipped batch.

**Suggestion:** Short-circuit the no-winner path by synthesizing an empty Pgsql response with `batch_arg_count` set, instead of invoking `HandlePgsqlReadRequest`. Avoid simply clearing the scalar ybctid and falling through, since that could turn the request into an unbounded scan.

---
_Binks review on behalf of [@cuivienor](https://github.com/cuivienor)._

src/yb/tserver/read_query.cc
```diff
@@ -767,6 +845,85 @@ Result<ReadQuery::ReadRestartInfo> ReadQuery::DoReadImpl() {
...
+      self->first_locked_batch_arg_index_ = batch_arg_index;
+      self->retained_self_ = self;
+      peer->Enqueue(self.get());
+    } else if (TransactionError(status).value() != TransactionErrorCode::kNone) {
+      // For SKIP LOCKED, any transaction-level failure to lock a row (kSkipLocking,
+      // kConflict, kAborted, kDeadlock, etc.) means we cannot acquire this row.
+      // Skip it and try the next candidate rather than aborting the entire batch.
+      self->TryLockBatchArg(
+          batch_arg_index + 1, total_batch_args, *pgsql_read,
+          isolation_level, leader_term, peer, tablet_ptr);
```
cuivienor, line 917: 🐛 **Bug**: `TransactionError(status).value() != kNone` includes aborts, deadlocks, read restarts, snapshot-too-old errors, and serialization conflicts, not just `kSkipLocking`. If any of those occur while locking a candidate, this code tries the next row and can return data from a transaction that should have aborted or restarted. The pre-batch single-row path only treats the skip-locking status as `TM_WouldBlock`; conflicts are surfaced as errors.

**Suggestion:** Switch on `TransactionErrorCode` and recurse only for `kSkipLocking` or another explicitly equivalent lock-not-acquired code. Propagate `kConflict`, `kAborted`, `kDeadlock`, read-restart, and snapshot-too-old statuses, and keep the client-side batch/leftover `TM_Result` handling consistent with the normal path.

---
_Binks review on behalf of [@cuivienor](https://github.com/cuivienor)._

src/yb/yql/pggate/pg_dml_read.cc
```diff
@@ -1039,4 +1039,24 @@ bool PgDmlRead::ActualValueForIsForSecondaryIndexArg(bool is_for_secondary_index
   return is_for_secondary_index && !IsPgSelectIndex();
 }
 
+void PgDmlRead::AddBatchYbctidArg(Slice ybctid) {
+  auto* batch_arg = read_req_->add_batch_arguments();
+  batch_arg->mutable_ybctid()->mutable_value()->dup_binary_value(ybctid);
```
cuivienor, line 1044: 🐛 **Bug**: `AddBatchYbctidArg` appends every prefetched ybctid to a single read request, while the request is routed by its scalar ybctid. Consecutive rows from the outer plan can cross tablet boundaries, unlike the existing ybctid batching path that uses `FindPartitionIndex` to group args per tablet. If an earlier candidate is skipped and a later candidate belongs to another tablet, the owning tablet never receives the row-lock request, yet the batch path can still report that later index as the winner or otherwise skip/error incorrectly.

**Suggestion:** Group or split candidate ybctids by tablet before adding them to one request, or only batch same-tablet ybctids and have the server reject off-tablet batch args before setting a winner.

---
_Binks review on behalf of [@cuivienor](https://github.com/cuivienor)._

cuivienor (commented):
_Binks review on behalf of [@cuivienor](https://github.com/cuivienor)._

PR #31362, opened 2026-04-30
[docs] Add instructions to recreate a partition index

jaki:


docs/content/stable/api/ysql/the-sql-language/statements/ddl_create_index.md
aishwarya24, line 56: Lets add a link to the "Recreate a partition index" section something like: 

- To rebuild unique indexes on partitions online without running `CREATE INDEX CONCURRENTLY` on the parent, see [Recreate a partition index](#recreate-a-partition-index).

docs/content/stable/api/ysql/the-sql-language/statements/ddl_create_index.md
```diff
@@ -372,6 +372,53 @@ DROP INDEX uniqueerror_i_idx;
...
+ALTER TABLE child0 ADD CONSTRAINT child0_i_key UNIQUE USING INDEX child0_i_unique;
+-- 8. Reattach (fast: CHECK and UNIQUE already satisfy parent)
+ALTER TABLE parent ATTACH PARTITION child0 FOR VALUES FROM (0) TO (100000);
+COMMIT;
+-- 9. Drop temporary CHECK constraint
+ALTER TABLE child0 DROP CONSTRAINT child0_partition_check;
+```
+
+Repeat steps 1–9 for `child1`, `child2`, and any other partitions as needed.
+Step 4 to lock the parent is optional if there will be no reads or writes against the parent table while the partition is detached.
```
aishwarya24:
```suggestion
```

docs/content/stable/api/ysql/the-sql-language/statements/ddl_create_index.md
```diff
@@ -372,6 +372,53 @@ DROP INDEX uniqueerror_i_idx;
...
+
+{{< note title="Note" >}}
+
+Step 4 requires object locking support (available since 2025.2): set YB-TServer flags `enable_object_locking_for_table_locks=true` and `ysql_yb_ddl_transaction_block_enabled=true`.
```
aishwarya24:
```suggestion
{{< note title="Note" >}}

Step 4 to lock the parent is optional if there are no reads or writes against the parent table while the partition is detached, and it requires enabling object locking (supported from YugabyteDB {{<release "2025.2">}} or later).

To enable the feature, set the YB-TServer flags `enable_object_locking_for_table_locks=true` and `ysql_yb_ddl_transaction_block_enabled=true`. Refer to Refer to [Enable table-level locks](../../../../../explore/transactions/explicit-locking/#enable-table-level-locks) for more details.
{{< /note >}}
```

aishwarya24 (approved):
LGTM with minor comments.

iSignal:
@jaki : @aishwarya24 : Would be good to place the existing steps which deal with related scenarios at https://docs.yugabyte.com/stable/api/ysql/the-sql-language/statements/ddl_create_index/#only along with this new section together. Maybe we can move the current section and this new part into a new subsection like Concurrent index creation for partitioned tables

--------


# Concurrent index creation for partitioned tables

## Creating indexes concurrently on partitioned tables
/* Current docs */


## Recreating unique index constraints concurrently on partitioned tables
Recreating unique index constraints involves a similar process as the previous section. However, there is an extra challenge that dropping an existing inherited constraint is not allowed. You can follow this sequence below when object locking is enabled.
/* new docs */

docs/content/stable/api/ysql/the-sql-language/statements/ddl_create_index.md
```diff
@@ -120,6 +104,83 @@ Indexes:
...
 ```
 
+Once all partitions successfully create their indexes, the partitioned index should automatically be promoted out of the INVALID state.
+
+#### Recreating unique index constraints concurrently on partitioned tables
+
+Suppose you have a partitioned table where the parent has a unique constraint.
+To recreate indexes, you should avoid directly creating a replacement index on the (parent) partitioned table if it is concurrently taking writes because partitioned tables cannot use [CONCURRENTLY](#concurrently).
+Furthermore, if you only want to recreate a subset of the partitions' indexes, recreating the (parent) partitioned index is wasteful.
+Follow these steps to recreate a partition index online:
```
iSignal, line 114: Maybe we can call out that object locking is a pre req and link to that section of docs @aishwarya24

PR #31364, opened 2026-04-30
[build] Improve error reporting and retry for archive downloads

hari90:
## Summary

When the third-party archive checksum download returned an HTML error page
instead of the expected `.sha256` file, `download_and_extract_archive.py`
only reported the file size, which made the failure hard to diagnose. There
was also no retry, so a single transient failure (e.g. a 5xx from GitHub)
would fail the build.

Example failure:

Checksum file size is too big: 55118 bytes

([failing job](https://github.com/yugabyte/yugabyte-db/actions/runs/25144698357/job/73701866705?pr=31359))

## Changes

- `download_url` now passes `-f` to curl so HTTP error responses no longer
  get written to disk as the requested artifact, and uses `--retry` /
  `--retry-delay` to retry transient failures (5xx, connection errors).
- The "checksum file size is too big" error now includes the first 1024
  bytes of the file so the underlying error (e.g. an HTML page) is visible
  in build logs.


svarnau (approved):

python/yugabyte/download_and_extract_archive.py
```diff
@@ -122,7 +131,20 @@ def download_url(url: str, dest_path: str, other_curl_flags: List[str] = []) ->
     dest_dir = os.path.dirname(dest_path)
     if not os.path.isdir(dest_dir):
         raise IOError("Destination directory %s does not exist" % dest_dir)
-    run_cmd(['curl', '-LsS', url, '-o', dest_path] + other_curl_flags)
+    # -f / --fail: don't write the response body for HTTP error responses, so we don't end up
+    #              with an HTML error page on disk masquerading as the requested artifact.
```
es1024, line 135: Not actually used I think?

PR #31406, opened 2026-05-02
[BACKPORT 2026.1][#31358] YSQL: Add yb_enable_mage gFlag to control mage extension load (#31359)

hari90:
## Summary

- Adds the `ysql_yb_enable_mage` gFlag (default `false`) and matching
  `yb_enable_mage` GUC. Both are non-runtime / `PGC_POSTMASTER` and only
  take effect on postmaster restart.
- When the gFlag is enabled, `pg_wrapper.cc` appends `mage` to
  `shared_preload_libraries` so its hooks are installed in every backend
  (mirrors the `pg_cron` / `documentdb` pattern).
- When the GUC is disabled, `CreateExtension` in `extension.c` rejects
`CREATE EXTENSION mage` outright, so a stray invocation can't lazy-load
  mage into a single backend and leave the cluster in an inconsistent
  hook state.
- Adds `ysql_yb_enable_mage=true` to
  `TestPgRegressThirdPartyExtensionsMage` so the regression suite still
  exercises mage end to end.

Original commit: 2f9453b13cf719494838d625a68298ea96944e6d

## Test plan

- [x] `./yb_build.sh release --java-test
org.yb.pgsql.TestPgRegressThirdPartyExtensionsMage`

Fixes #31358

---------

Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---

[Jenkins](<https://csiweb.dev.yugabyte.com/pull/31406/>)

PR #31407, opened 2026-05-02
Tools: Claude Code agent tooling for backport, PR creation, and review

hari90:
## Summary

Adds Claude Code agent tooling for YugabyteDB: two slash commands, two shared shell helpers, project-wide agent rules in `AGENTS.md`, a Gemini Code Assist style guide, and a permission allowlist that scopes write operations to where they're used.

- **`/backport-commit` slash command** (`.claude/commands/backport-commit.{md,sh}`) — backports one or more release-branch cherry-picks in a single invocation; chains task branches so a conflict resolved on `b2` carries forward; lints, refuses to push to upstream, opens PRs, posts `Trigger Jenkins`, annotates `## Merge conflicts` when applicable, supports `YB_FORK_URL` for HTTPS users, and `continue`s correctly when a cherry-pick is a no-op.
- **`/create-pr` skill + `create-pr.sh`** — opens cross-repo PRs from a fork; requires separate `-d` (description) and `-T` (test plan) files (both non-empty); optional `-U` (upgrade/rollback notes), **mandatory when the branch changes a `.proto` file**; validates issue and title shape before any push.
- **`git-push.sh`** — shared lint + fork-destination-check + push. Refuses to push when the resolved fork-owner matches the upstream owner. Surfaces the current PR title and new commit subjects after the push, posts `/gemini review`, and prints templates to update title or body via REST PATCH.
- **`AGENTS.md`** — codifies the git workflow: always-new-commit (no amend), never force-push, push only via `git-push.sh`, auto-update PR summary on scope changes (don't ask), keep title in sync only when actively misleading, `## Upgrade/Rollback safety` section required for upgrade-relevant changes, scratch files under `/tmp/claude/`, reply-to-review-comments rules.
- **`.gemini/styleguide.md`** — default review focuses on correctness / memory safety / security / concurrency / performance / upgrade-rollback safety; skips lint-covered nits. Backport-aware mode: don't re-flag hunks that match the originating commit. Stop-re-flagging-rule discourages noise across review rounds.
- **`build-support/lint.py`** — when `@{upstream}` isn't set, falls back to `<canonical-remote>/master` (the remote whose URL points at `yugabyte/yugabyte-db`); errors with explicit `--rev` guidance otherwise.
- **`.claude/settings.json`** — strict-JSON allowlist grouped by category. Read-side `gh`/`git` broadly allowed; write-side `gh pr` operations scoped to specific commands' frontmatter; `Bash(git push *)` denied (forces use of `git-push.sh`).

## Test plan

- [x] `/backport-commit 2f9453b13c 2026.1` end-to-end — produced [#31406](https://github.com/yugabyte/yugabyte-db/pull/31406) with the right title, originating-PR link, reviewers, and `Trigger Jenkins` comment.
- [x] `bash -n` clean on all three scripts; `./build-support/lint.sh --rev upstream/master` reports `0 error(s), 0 warning(s)` for every commit on this branch.
- [x] `git-push.sh` post-push title/body sync reminder + `/gemini review` auto-post verified on this PR.
- [x] `lint.py` canonical-remote fallback verified via detached-HEAD invocation.
- [x] All Gemini-bot review feedback addressed (real fixes: cherry-pick `--skip` `continue`, portable `mktemp`, fork-safety guards, IFS-aware reviewer parsing; declined with explanation: `set -euo pipefail` already present, `sort -V` works on macOS).
- [x] Re-run `/backport-commit` on a commit that conflicts with a release branch to exercise the `## Merge conflicts` annotation path end-to-end.
- [x] Run `/create-pr` end-to-end on a separate branch.
- [x] Verify Gemini Code Assist applies `.gemini/styleguide.md` rules (default + backport mode + stop-re-flagging) on a fresh PR.


.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,412 @@
...
+
+##################### Per-branch work
+for branch in $branches; do
+  if [[ $branch == "master" ]]; then
+    continue
+  fi
+  echo ""
+  echo "================== Release Branch $branch"
+  if [[ $prompt == "true" ]]; then
+    read -p "Backport $id to branch $branch [Y/n]? " response
```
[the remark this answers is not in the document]
hari90: Done in 0a6cffa9 — added `[[ ! -t 0 ]]` check before the interactive prompt. Note: the prompt path only fires when the user runs the script directly with no branches and no `-s`; the `/backport-commit` slash command always passes branches so it never reaches this code.

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,412 @@
...
+      echo "==== Requesting reviewers: $final_reviewers"
+      api_args=()
+      for r in ${final_reviewers//,/ }; do
+        api_args+=(-f "reviewers[]=${r}")
+      done
+      if ! gh api -X POST "repos/${GH_REPO}/pulls/${new_pr_num}/requested_reviewers" \
+             "${api_args[@]}" --jq '"requested: " + (.requested_reviewers | map(.login) | join(", "))'; then
+        echo "WARN: failed to add reviewers to PR #${new_pr_num} (continuing): $final_reviewers"
+      fi
+    fi
```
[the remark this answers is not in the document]
hari90: Done in 0a6cffa9 — switched to `gh pr edit --add-reviewer`, which routes user logins and team slugs to the right field. The stale REST-API comment block was removed too.

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,412 @@
...
+  git clone $YB_REPO $workspace
+  cd $workspace
+else
+  echo "==== Using workspace: $workspace"
+  cd $workspace
+  echo "==== Fetching latest code"
+  git fetch origin
+fi
+
+ws_status=$(git status -s)
```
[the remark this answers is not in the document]
hari90: Done in 0a6cffa9 — `git status --porcelain | grep -v '^??'` now ignores untracked files, since logs and scratch in the workspace shouldn't block a backport.

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,412 @@
...
+if [[ -n "$pr_lookup" ]]; then
+  prnum=$(echo "$pr_lookup" | python3 -c '
+import json, sys
+prs = json.load(sys.stdin)
+print(prs[0]["number"] if prs else "")' 2>/dev/null)
+  pr_url=$(echo "$pr_lookup" | python3 -c '
+import json, sys
+prs = json.load(sys.stdin)
+print(prs[0]["html_url"] if prs else "")' 2>/dev/null)
+fi
```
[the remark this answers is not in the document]
hari90: Done in 0a6cffa9 — single `gh api ... --jq` call replaces the two python3 invocations.

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,412 @@
...
+    pr_head="${fork_owner}:${task_branch}"
+  else
+    pr_head="$task_branch"
+  fi
+
+  # Skip if a backport PR already exists for this head branch.
+  existing_pr=$(gh pr list -R "$GH_REPO" --head "$pr_head" --state all \
+                  --json number,url --jq '.[0]' 2>/dev/null)
+  if [[ -n "$existing_pr" && "$existing_pr" != "null" ]]; then
+    existing_url=$(echo "$existing_pr" | python3 -c 'import json,sys; print(json.load(sys.stdin)["url"])')
```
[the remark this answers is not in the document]
hari90: Done in 0a6cffa9 — `gh pr list --jq '.[0].url // empty'` replaces the python3-based JSON parse.

.claude/commands/backport-commit.sh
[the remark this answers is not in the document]
hari90: Update on this one: `gh pr edit --add-reviewer` actually fails on this repo with the projectCards GraphQL deprecation (same one that breaks `gh pr edit --body-file`). Reverted in 2f4e12ab and went back to the REST endpoint, but with proper team handling: user logins go to `reviewers[]`, team slugs (`org/slug` form) get the `org/` stripped and routed to `team_reviewers[]` — which addresses the original concern without depending on the broken `gh` GraphQL path.

.gemini/styleguide.md
```diff
@@ -0,0 +1,65 @@
...
+
+Focus on **substantive issues**: correctness, security, concurrency, performance, and breaking-change risk. The repo runs `./build-support/lint.sh` on every PR — its checks are authoritative for style. Do not duplicate them.
+
+Specifically, **do not** flag:
+
+- Line length, indentation, trailing whitespace, missing trailing newlines, tab/space mixing.
+- Import ordering, header-include ordering (the linter and clang-format own this).
+- Naming-style nits where the existing module already follows a pattern (a new symbol that matches its surroundings is fine).
+- Missing tests for hunks where an existing test already exercises the changed behavior — flag only when a code path has *no* coverage at all.
+- Cosmetic refactoring suggestions (renaming a local variable for clarity, splitting a function into smaller ones) when the existing code is correct.
```
hari90: Remove this line

.gemini/styleguide.md
```diff
@@ -0,0 +1,65 @@
+# YugabyteDB Code Review Style Guide
+
+This document configures Gemini Code Assist's automated review for PRs in this repository. It scopes what to flag, what to skip, and how to handle a few PR types (especially backports) where the default review behavior produces noise.
+
+## General review guidance
+
+Focus on **substantive issues**: correctness, security, concurrency, performance, and breaking-change risk. The repo runs `./build-support/lint.sh` on every PR — its checks are authoritative for style. Do not duplicate them.
```
hari90: Also focus on memory safety

.gemini/styleguide.md
```diff
@@ -0,0 +1,65 @@
...
+
+- Line length, indentation, trailing whitespace, missing trailing newlines, tab/space mixing.
+- Import ordering, header-include ordering (the linter and clang-format own this).
+- Naming-style nits where the existing module already follows a pattern (a new symbol that matches its surroundings is fine).
+- Missing tests for hunks where an existing test already exercises the changed behavior — flag only when a code path has *no* coverage at all.
+- Cosmetic refactoring suggestions (renaming a local variable for clarity, splitting a function into smaller ones) when the existing code is correct.
+
+**Do** flag:
+
+- Logic errors, off-by-ones, missing error handling on fallible operations, leaked resources, data races.
```
hari90: Include memory safety

.gemini/styleguide.md
```diff
@@ -0,0 +1,65 @@
...
+- Anything that contradicts a comment, docstring, or commit message in the same diff.
+
+Default comment severity threshold: **MEDIUM**. Suppress LOW-severity nits unless they reveal an underlying correctness issue.
+
+## Backport PRs (different review rules)
+
+A **backport** is a PR whose title starts with `[BACKPORT <release-branch>]`, e.g. `[BACKPORT 2024.2][#31358] YSQL: Add yb_enable_mage gFlag`. Its body always contains a line of the form:
+
+```
+Backport of: https://github.com/yugabyte/yugabyte-db/pull/<N> (commit <SHA>)
```
hari90: Original commit: SHA

.gemini/styleguide.md
```diff
@@ -0,0 +1,65 @@
...
+- Cosmetic refactoring suggestions (renaming a local variable for clarity, splitting a function into smaller ones) when the existing code is correct.
+
+**Do** flag:
+
+- Logic errors, off-by-ones, missing error handling on fallible operations, leaked resources, data races.
+- Use of unsafe APIs, SQL/format-string injection vectors, missing input validation at trust boundaries.
+- Public API or wire-format changes (`*.proto`, public headers under `src/yb/yql/`, gflags users will see, gRPC service definitions). These need extra scrutiny — call them out even if they look correct.
+- Performance regressions on hot paths (DocDB write/read path, query layer plan execution, RPC dispatch).
+- Anything that contradicts a comment, docstring, or commit message in the same diff.
+
```
hari90, line 25: Upgrade safety

.gemini/styleguide.md
```diff
@@ -0,0 +1,65 @@
...
+- **`src/yb/master/`** (catalog manager, leader election) — concurrency and state-machine correctness.
+- **`src/yb/tablet/`** and **`src/yb/docdb/`** — write/read path, MVCC, transaction handling.
+- **`src/yb/cdc/`** and **`src/yb/cdc_consumer/`** — replication-stream correctness, idempotency.
+- **`src/postgres/`** — PostgreSQL-fork merges; double-check that upstream Postgres semantics are preserved.
+- **`managed/`** (YBA platform) — orchestration logic, especially around node lifecycle and cluster-wide ops.
+
+## What to ignore entirely
+
+- Files under `.claude/`, `.gemini/`, `AGENTS.md`, `CLAUDE.md` — these are agent / review-bot configuration. Don't review them as code; their effects are operational (they change how *agents* and *bots* behave, not what the database does).
```
hari90: There are scripts in these folders so review them

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,478 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
hari90: Declining: `set -euo pipefail` is already present at line 17 of this script (right after the license comment block). Reviewing against an outdated snapshot.

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,450 @@
...
+if [[ -z "$branches" ]]; then
+  # Try asking release service for "active" branches.
+  stable_list="$(curl -s https://release.dev.yugabyte.com/version/active/text | tr ',' ' ')"
+  if [[ -z $stable_list ]]; then
+    # Find stable branches by naming scheme 2.10-2.28 (even) or YYYY.x
+    stable_list=$(git branch -a |
+               grep -E '(origin/2.[12][24680]$)|(origin/[0-9]{4}\.[0-9]+$)' |
+               sed -e 's#remotes/origin/##' |
+               sort -Vr)
```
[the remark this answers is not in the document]
hari90: Declining: `sort -V` works on macOS. It has been supported by `/usr/bin/sort` (Apple-modified BSD sort) on every macOS version this repo runs on. Verified end-to-end via PR #31406 — the backport ran cleanly through this branch list without any `illegal option -- V` error. If a contributor on a different OS hits this in practice we can switch to `sort -t. -k1,1nr -k2,2nr`.

.claude/commands/backport-commit.sh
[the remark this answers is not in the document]
hari90: Note for next review: declining. The script already has `set -euo pipefail` at line 17 (see line 17 of the file as it stands now -- it sits below the multi-line license header, so a review against the first executable line lands on line 2 / the shebang). The `-e` (exit-on-error), `-u` (unset-var error), and `pipefail` (pipe-failure propagation) are all in effect; the bot was reviewing the line 2 area which is just the shebang and a comment block.

.agents/scripts/backport-commit.sh
```diff
@@ -0,0 +1,529 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
hari90: Correction to a prior decline: I previously claimed this script already had `set -euo pipefail` at line 17. That was wrong — line 17 is the USAGE string, the script genuinely had no `set` directive at all. Apologies for the misdirection. Added in 03996d34e2.

---
_automated · Claude Code (Opus 4.7)_

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,529 @@
...
+    fi
+  else
+    echo "WARN: ./build-support/lint.sh not found or not executable; skipping lint check."
+  fi
+
+  echo "==== Publishing backport PR for branch $task_branch"
+  # Reviewers are added via the requested_reviewers REST endpoint after PR
+  # creation. We avoid `gh pr {create,edit} --reviewer/--add-reviewer`: both
+  # invoke a GraphQL query that touches the deprecated `projectCards` field
+  # and fail outright on this repo's GitHub Enterprise instance.
```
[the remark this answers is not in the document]
hari90: The cited "create-pr.sh:171" is just an `exit 1`, not a contradicting comment — but the underlying point is right: `.claude/skills/create-pr/SKILL.md` does say "gh pr create --reviewer works fine" while this script grouped both forms as broken. Clarified the comment in 03996d34e2: only `gh pr edit --add-reviewer` is broken on this repo (projectCards GHE bug); `gh pr create --reviewer` itself works, but the script uses the REST endpoint anyway because it routes team slugs (org/slug) to team_reviewers[] correctly.

---
_automated · Claude Code (Opus 4.7)_

.claude/commands/backport-commit.sh
```diff
@@ -0,0 +1,537 @@
...
+                   https://release.dev.yugabyte.com/version/active/text \
+                   | tr ',' ' ' || true)"
+  if [[ -z $stable_list ]]; then
+    # Find stable branches by naming scheme 2.10-2.28 (even) or YYYY.x
+    branch_re="(${upstream_remote}/2\.[12][24680]\$)"
+    branch_re+="|(${upstream_remote}/[0-9]{4}\.[0-9]+\$)"
+    stable_list=$(git branch -a |
+               grep -E "$branch_re" |
+               sed -e "s#remotes/${upstream_remote}/##" |
+               sort -Vr)
```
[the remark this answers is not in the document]
hari90: Re-flag of an already-declined finding (see prior threads on this PR for the same suggestion). `sort -V` works on macOS — the system `/usr/bin/sort` (Apple-modified BSD) supports `-V` on every macOS version this repo runs on; verified empirically by running the script end-to-end (PR #31406). Per the `.gemini/styleguide.md` "stop re-flagging" rule, threads with replies explaining a decline should be skipped on later passes. Declining again, resolving.

---
_automated · Claude Code (Opus 4.7)_

.claude/settings.json
```diff
@@ -0,0 +1,68 @@
...
+      "Bash(./build-support/lint.sh *)",
+      "Bash(./yb_build.sh *)",
+      "Bash(.claude/commands/git-push.sh *)",
+
+      "Bash(gh api -X PATCH /repos/yugabyte/yugabyte-db/pulls/*)",
+      "Bash(gh api -X POST repos/yugabyte/yugabyte-db/pulls/*)",
+      "Bash(gh api graphql *)",
+      "Bash(gh pr list *)",
+      "Bash(gh pr view *)",
+      "Bash(gh repo view *)",
```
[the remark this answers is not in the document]
hari90: Declining: the omission is intentional. Per the AGENTS.md scoping rule, write-side `gh pr` operations live in the *invoking* skill/command's `allowed-tools` frontmatter, not project-wide. `Bash(gh pr comment *)` is in `.claude/commands/backport-commit.md`'s frontmatter (Step 5's Trigger Jenkins post), so the /backport-commit workflow doesn't prompt. The `git-push.sh` auto-`/gemini review` post runs as an *internal subprocess* of the script -- the harness only checks `Bash(.claude/commands/git-push.sh *)` (which IS allowlisted) and doesn't descend into the script's internal commands. So nothing prompts in the documented workflows; promoting `gh pr comment *` project-wide would broaden write capability beyond what's justified.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,699 @@
...
+    """Return (gh_fork, fork_owner, push_url). Aborts if fork == upstream owner."""
+    upstream_owner = gh_repo.split("/", 1)[0]
+    repo_name = gh_repo.split("/", 1)[1]
+
+    gh_fork = os.environ.get("YB_GH_FORK") or ""
+    if not gh_fork:
+        if not gh_user or gh_user == upstream_owner:
+            fail("no personal fork detected for the authenticated gh user. "
+                 "Set YB_GH_FORK=<your-user>/<repo> or run 'gh repo fork' and re-run.")
+        gh_fork = f"{gh_user}/{repo_name}"
```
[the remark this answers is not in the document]
hari90: Renamed forks are already supported via the `YB_GH_FORK=<owner>/<name>` env var (see lines 252-253). Auto-querying `gh repo list --fork` adds pagination + an extra API call for the rare renamed-fork case the env var already covers, so leaving the default as the standard `<gh-user>/<upstream-repo-name>` shape.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,725 @@
...
+        info("change. The AGENTS.md \"never amend\" rule applies to public")
+        info("commits, not to in-flight task branches.)")
+        raise ResumableConflict()
+
+
+def push_and_create_pr(*, gh_repo: str, push_url: str, pr_head: str, base: str,
+                        task: str, title: str, body: str) -> str:
+    """Push the task branch and open the PR. Returns PR URL."""
+    info(f"Publishing backport PR for branch {task}")
+    run(["git", "push", "-u", push_url, task])
```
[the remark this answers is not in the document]
hari90: Declining: this repos `AGENTS.md` prohibits force-push without explicit user permission. In normal flow its also unnecessary — when a backport PR already exists for this `<head>`, `existing_pr_url` short-circuits the loop before we reach the push, so the only path to a non-fast-forward push is push-succeeded-but-PR-creation-failed (rare; user can resolve manually with a one-off force-push).

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,725 @@
...
+            # The chain logic below cherry-picks from $task onto the next
+            # branch; ensure $task is present locally. In a fresh worktree
+            # it usually isn't -- pull it from the fork. If that fails
+            # (branch deleted on the fork, or PR was opened from a
+            # different fork), fall back to chaining from the original
+            # commit so subsequent branches still apply something
+            # sensible -- they may need conflict resolution again.
+            if not git_ok("show-ref", "--verify", "--quiet", f"refs/heads/{task}"):
+                info(f"Fetching {task} from fork to enable chaining")
+                proc = run(["git", "fetch", push_url, f"{task}:{task}"], check=False)
```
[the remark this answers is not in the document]
hari90: Declining: the surrounding `if not git_ok("show-ref", ...)` only enters this fetch when the local `task` ref is absent, so divergence isnt possible at that point — a non-forcing `{task}:{task}` refspec is sufficient. Adding `+` would only matter if we entered this path with the local ref already present, which the guard prevents.

---
_automated · Claude Code (Opus 4.7)_

.claude/skills/create-pr/SKILL.md
```diff
@@ -23,99 +24,78 @@ This skill requires the `gh` CLI installed and authenticated:
...
-   ```
-   git remote add fork git@github.com:<username>/yugabyte-db.git
-   ```
-4. **If no fork exists at all**, create one (ask the user to confirm first):
-   ```
-   gh repo fork --remote=false yugabyte/yugabyte-db
-   ```
-   Then add the resulting fork as a `fork` remote as above.
-5. **Never push to a remote that points at `yugabyte/yugabyte-db`.** Verify the chosen push target is a fork before continuing.
+### Step 3: Verify the fork remote exists (skip otherwise)
```
hari90: Not needed. The script does it

.claude/skills/create-pr/SKILL.md
```diff
@@ -23,99 +24,78 @@ This skill requires the `gh` CLI installed and authenticated:
...
      - For JIRA: ask the user which project (e.g., `PLAT`) and use the Atlassian MCP tool `createJiraIssue` to create it. Confirm the summary/description with the user before creating.
    - Capture the resulting issue number or JIRA key.
 
-2. **Component** — the component tag that will appear in the title (e.g., `DocDB`, `YSQL`, `YBA`, `CDC`, `xCluster`). If the user is unsure, suggest a component based on the files changed (e.g., `src/yb/` → `DocDB`, `src/postgres/` → `YSQL`, `managed/` → `YBA`).
+   **Validate the issue exists** before using it: run `gh issue view <num> --repo yugabyte/yugabyte-db --json title,body,state` (or for JIRA, `getJiraIssue`). This catches typos, branches named after stale numbers, and surfaces the GH issue body — which often contains a Jira link to include in the PR body (see Step 9).
```
hari90: not needed

.claude/skills/create-pr/SKILL.md
```diff
@@ -23,99 +24,78 @@ This skill requires the `gh` CLI installed and authenticated:
...
+
+   **One issue per commit.** Each GitHub issue tracks exactly one logical commit and its backports. **Never reuse an issue across multiple in-flight or merged PRs targeting `master`** — if a previous PR with `[#NNNNN]` exists for `master` (open or merged), a new feature/fix needs its own new issue (auto-create one via the flow above). Backport PRs to release branches (`2024.1`, `2024.2`, etc.) reuse the *same* issue as the originating master PR — that is the only legitimate reuse case.
```
hari90: Move this check into the script

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,746 @@
...
+
+
+def detect_auto_reviewers(gh_repo: str, prnum: int, self_login: str) -> str:
+    """Union of pending review requests and actual reviewers, minus self / bots."""
+    upstream_owner = gh_repo.split("/", 1)[0]
+    pending_data = gh_json([f"repos/{gh_repo}/pulls/{prnum}/requested_reviewers"], check=False) or {}
+    # GitHub returns null entries for deleted users; (x or {}).get(...) is
+    # a cheap guard so a deleted requester / reviewer doesn't crash the run.
+    user_logins = [u["login"] for u in (pending_data.get("users") or [])
+                   if (u or {}).get("type") == "User"]
```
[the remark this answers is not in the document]
hari90: Already handled by the existing `(u or {}).get("type") == "User"` filter on the previous line — `(None or {}).get(...)` returns `None`, the comprehension skips the entry, and `u["login"]` is only reached when `u` is a real dict with `type == "User"`. The `(x or {})` form and the suggested `if u and u.get(...)` form are functionally equivalent here. Not adding `"login" in u` either: GitHub's API contract guarantees `login` for any user object with `type == "User"`, so guarding against its absence falls under the "don't validate scenarios that can't happen" rule in `src/AGENTS.md`.

---
_automated · Claude Code (Opus 4.7)_

.claude/commands/backport-commit.py
```diff
@@ -0,0 +1,746 @@
...
+    # GitHub returns null entries for deleted users; (x or {}).get(...) is
+    # a cheap guard so a deleted requester / reviewer doesn't crash the run.
+    user_logins = [u["login"] for u in (pending_data.get("users") or [])
+                   if (u or {}).get("type") == "User"]
+    team_slugs = [f"{upstream_owner}/{t['slug']}"
+                  for t in (pending_data.get("teams") or []) if t]
+
+    review_data = gh_json([f"repos/{gh_repo}/pulls/{prnum}/reviews"], check=False) or []
+    reviewer_logins = [r["user"]["login"] for r in review_data
+                       if (r.get("user") or {}).get("type") == "User"]
```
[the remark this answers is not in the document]
hari90: Same as the user_logins thread above — the existing `(r.get("user") or {}).get("type") == "User"` filter already handles `r["user"]` being `null` or absent. `r["user"]["login"]` only executes when `r["user"]` is a real user dict with `type == "User"`, where GitHub guarantees the `login` field. No change.

---
_automated · Claude Code (Opus 4.7)_

.claude/skills/create-pr/SKILL.md
```diff
@@ -23,99 +24,78 @@ This skill requires the `gh` CLI installed and authenticated:
...
-   git fetch <upstream-remote> master
-   ```
-
-3. **Rebase the branch onto upstream master:**
-   ```
-   git rebase <upstream-remote>/master
-   ```
-
-4. **Resolve any conflicts** with the user. For each conflict: show the conflicted files, help the user resolve them, `git add` the resolved files, and run `git rebase --continue`. If the user wants to bail out, `git rebase --abort` returns the branch to its pre-rebase state.
+**Backport exception:** if the branch name looks like a backport (e.g., contains a release version such as `2024.1`, `2024.2`, `2.20`, or has a `backport`/`bp` prefix/suffix), or the user has indicated this is a backport, ask which release branch to target (e.g., `2024.1`, `2024.2.5`) instead of defaulting to `master`. Confirm the target branch with the user before continuing.
```
hari90: This skill is not used in backports

.claude/skills/create-pr/SKILL.md
hari90, line 14: Not needed. The script handles the safety.

Do call out that backports should use the backport command skill instead. 
Remove other references to backport in this skill

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,754 @@
...
+        line for line in raw_body.splitlines()
+        if not (line.startswith("Backport of:")
+                or line.startswith("Original PR:")
+                or line.startswith("Original commit:"))
+    ]
+    pr_num_match = re.search(r"/pull/(\d+)", pr_url) if pr_url else None
+    if pr_num_match:
+        original_line = f"Original commit: {commit} / #{pr_num_match.group(1)}"
+    else:
+        original_line = f"Original commit: {commit}"
```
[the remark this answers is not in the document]
hari90: Declining. `pr_url` is set from `pr["html_url"]` returned by GitHub's `/repos/{owner}/{repo}/commits/{sha}/pulls` endpoint, which is contracted to be `https://<host>/<owner>/<repo>/pull/<N>` on both github.com and GitHub Enterprise (the `/pull/<N>` segment is stable across both). The regex matches that contract. Adding fallback parsing for hypothetical alternate formats falls under the "don't validate scenarios that can't happen" guidance in `src/AGENTS.md`.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,754 @@
...
+
+    for branch in branches:
+        if branch == "master":
+            continue
+        print()
+        print(f"================== Release Branch {branch}")
+        if interactive_prompt and sys.stdin.isatty():
+            ans = input(f"Backport {short} to branch {branch} [Y/n]? ").strip()
+            if ans.lower().startswith("n"):
+                continue
```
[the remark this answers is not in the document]
hari90: Declining. The non-TTY case is already handled at lines 656-657: when `interactive_prompt` is true and `sys.stdin.isatty()` is false, `fail("auto-discovery in non-interactive mode; pass branches explicitly.")` exits with a clear error. So `input()` never runs without a real TTY behind it. Adding `--yes` to bypass would be feature creep — the current "pass branches explicitly" path is already the non-interactive escape hatch.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,758 @@
...
+            # The chain logic below cherry-picks from $task onto the next
+            # branch; ensure $task is present locally. In a fresh worktree
+            # it usually isn't -- pull it from the fork. If that fails
+            # (branch deleted on the fork, or PR was opened from a
+            # different fork), fall back to chaining from the original
+            # commit so subsequent branches still apply something
+            # sensible -- they may need conflict resolution again.
+            if not git_ok("show-ref", "--verify", "--quiet", f"refs/heads/{task}"):
+                info(f"Fetching {task} from fork to enable chaining")
+                proc = run(["git", "fetch", push_url, f"{task}:{task}"], check=False)
```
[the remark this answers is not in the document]
hari90: Declining (re-flag of an already-declined thread). The fetch only runs inside `if not git_ok("show-ref", "--verify", "--quiet", refs/heads/<task>)` — i.e. when the local ref is *absent*. A non-existent ref can't be diverged, so the non-forcing `<task>:<task>` refspec is correct here. Forcing would only matter if we entered this code path with the local ref present, which the guard prevents.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/create-pr.sh
```diff
@@ -0,0 +1,346 @@
...
+fi
+
+# If the user accidentally included an issue prefix in the title, strip it --
+# we add the canonical "[<issue>] " prefix below.
+title="${title#\[*\] }"
+
+# Validate the title body is "<Component>: <Title>". Component must start with
+# a letter and contain at least one alphanumeric char; followed by ": " and
+# something. This catches typos, missing prefixes, and unrelated formats.
+if ! [[ "$title" =~ ^[A-Za-z][A-Za-z0-9]+:[[:space:]].+ ]]; then
```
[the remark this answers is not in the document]
hari90: Declining. The strict `[A-Za-z0-9]+` (≥2 chars) is intentional — it acts as a typo guard, catching single-letter slips like `D: Fix flake` instead of `DocDB: Fix flake`. All real components in this repo are ≥2 chars (`DocDB`, `YSQL`, `YCQL`, `YBA`, `CDC`, `xCluster`, `yugabyted`, `Docs`, `ClaudeCode`, `Build`) and single-letter components aren't a planned extension; loosening the regex would only weaken the typo check.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,758 @@
...
+    return pr["number"], pr["html_url"]
+
+
+def detect_auto_reviewers(gh_repo: str, prnum: int, self_login: str) -> str:
+    """Union of pending review requests and actual reviewers, minus self / bots."""
+    upstream_owner = gh_repo.split("/", 1)[0]
+    pending_data = gh_json([f"repos/{gh_repo}/pulls/{prnum}/requested_reviewers"], check=False) or {}
+    # GitHub returns null entries for deleted users; (x or {}).get(...) is
+    # a cheap guard so a deleted requester / reviewer doesn't crash the run.
+    user_logins = [u["login"] for u in (pending_data.get("users") or [])
```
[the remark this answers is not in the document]
hari90: Declining (re-flag of an already-declined concern). The filter ahead of `u["login"]` is `(u or {}).get("type") == "User"` — GitHub's API contract guarantees a `login` field on any object with `type == "User"`. Adding `.get("login")` would only matter if that contract were violated, which falls under the "don't validate scenarios that can't happen" rule in `src/AGENTS.md`.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,758 @@
...
+    team_slugs = [f"{upstream_owner}/{t['slug']}"
+                  for t in (pending_data.get("teams") or []) if t]
+
+    review_data = gh_json([f"repos/{gh_repo}/pulls/{prnum}/reviews"], check=False) or []
+    # `r and` defends against the rare null entry in the review array
+    # itself; `(r.get("user") or {})` defends against r["user"] being null
+    # for deleted reviewers. Both are observed in the wild for users that
+    # have been GDPR-erased.
+    reviewer_logins = [r["user"]["login"] for r in review_data
+                       if r and (r.get("user") or {}).get("type") == "User"]
```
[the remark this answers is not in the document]
hari90: Same as the user_logins thread above — GitHub guarantees `login` for objects with `type == "User"`, which the filter already requires. Declining.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,758 @@
...
+    rerun_args.extend(args.branches if (args.branches or args.stable_limit)
+                      else branches)
+
+    for branch in branches:
+        if branch == "master":
+            continue
+        print()
+        print(f"================== Release Branch {branch}")
+        if interactive_prompt and sys.stdin.isatty():
+            ans = input(f"Backport {short} to branch {branch} [Y/n]? ").strip()
```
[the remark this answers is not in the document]
hari90: Declining (re-flag of an already-declined concern). The non-TTY case is already handled: when `interactive_prompt` is true and `sys.stdin.isatty()` is false, the next branch (`elif interactive_prompt:`) calls `fail("auto-discovery in non-interactive mode; pass branches explicitly.")` and exits. So `input()` only runs when there's a real TTY behind it. Adding `--yes` would be feature creep; the "pass branches explicitly" path is already the non-interactive entry point.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/git-push.sh
```diff
@@ -0,0 +1,184 @@
...
+      echo ""
+      echo ">>> Review the PR title and summary. Update either if these commits"
+      echo "    significantly shift scope, approach, or component. Leave them"
+      echo "    alone for refinements within existing scope, lint fixes, typos,"
+      echo "    comment-only edits, or pure-refactor commits."
+      echo "    Title update (rare; only when the existing title misleads):"
+      echo "      gh api -X PATCH /repos/${GH_REPO}/pulls/${pr_num} -f title='<new>'"
+      echo "    Body update:"
+      echo "      jq -Rs '{body: .}' < /tmp/claude/pr-body-${pr_num}.md \\"
+      echo "        | gh api -X PATCH /repos/${GH_REPO}/pulls/${pr_num} --input -"
```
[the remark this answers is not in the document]
hari90: Declining. This is a *printed* template, not an actual API call — the leading slash is the convention used in `src/AGENTS.md` for the documented `gh api -X PATCH /repos/...` workflow, and changing the printout here would only diverge from that. Both forms work in `gh api`; consistency with the doc beats consistency with script-internal calls for human-facing instructions.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,758 @@
...
+    team_slugs = [f"{upstream_owner}/{t['slug']}"
+                  for t in (pending_data.get("teams") or []) if t]
+
+    review_data = gh_json([f"repos/{gh_repo}/pulls/{prnum}/reviews"], check=False) or []
+    # `r and` defends against the rare null entry in the review array
+    # itself; `(r.get("user") or {})` defends against r["user"] being null
+    # for deleted reviewers. Both are observed in the wild for users that
+    # have been GDPR-erased.
+    reviewer_logins = [r["user"]["login"] for r in review_data
+                       if r and (r.get("user") or {}).get("type") == "User"]
```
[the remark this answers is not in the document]
hari90: Declining (re-flag of an already-declined concern). Current code is `if r and (r.get("user") or {}).get("type") == "User"` — `r and` defends `r` being None, `(r.get("user") or {})` defends `r["user"]` being None. The comprehension only reaches `r["user"]["login"]` when both checks pass, i.e. `r` is a real review dict and `r["user"]` is a real user dict with `type == "User"`. No path crashes.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,758 @@
...
+    rerun_args.extend(args.branches if (args.branches or args.stable_limit)
+                      else branches)
+
+    for branch in branches:
+        if branch == "master":
+            continue
+        print()
+        print(f"================== Release Branch {branch}")
+        if interactive_prompt and sys.stdin.isatty():
+            ans = input(f"Backport {short} to branch {branch} [Y/n]? ").strip()
```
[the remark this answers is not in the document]
hari90: Declining (re-flag of an already-declined concern). The non-TTY case is handled at lines 651-658 — `if interactive_prompt and sys.stdin.isatty():` only routes to `input()` when there's a real TTY; the `elif interactive_prompt:` branch calls `fail("auto-discovery in non-interactive mode; pass branches explicitly.")` and exits cleanly. Adding a `--yes` flag would be feature creep on top of the already-supported "pass branches explicitly" path.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/git-push.sh
```diff
@@ -0,0 +1,184 @@
...
+
+# Refuse to push to upstream owner. The fork-detection above checks the URL
+# pattern; this guard catches misconfigured FORK_REMOTE / -r overrides and
+# the edge case where the gh user happens to match the upstream owner.
+fork_url=$(git remote get-url "$FORK_REMOTE")
+# Resolve the owner via `gh repo view` (which accepts the URL form and
+# follows GitHub's canonical owner). Fall back to URL parsing if gh
+# can't reach the API or the URL isn't a github.com one.
+fork_owner=$(gh repo view "$fork_url" --json owner --jq '.owner.login' 2>/dev/null \
+             || echo "$fork_url" | sed -E 's|.*[:/]([^:/]+)/[^/]+(\.git)?$|\1|')
```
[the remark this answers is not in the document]
hari90: Declining. The sed is a fallback only — primary path is `gh repo view "$fork_url"`, which handles HTTPS, SSH, gh-alias, and short-form URLs. The fallback runs only if `gh repo view` fails (network down or the URL isn't a recognizable github form). For the common cases the sed *does* work — `git@github.com:user/repo.git`, `https://github.com/user/repo.git`, `gh-alias:user/repo` all extract `user` correctly. For pathological inputs (local paths, malformed URLs) the user is doing something `git-push.sh` was never designed for; failing softly into the upstream-owner-mismatch check is acceptable. Tightening this further would also need to handle the `-r` override case where the fork is a teammate's — using `gh api user` here would lose that safety check, unlike in `create-pr.sh` where the fork is constrained to be the auth user's own.

---
_automated · Claude Code (Opus 4.7)_

.gemini/styleguide.md
```diff
@@ -0,0 +1,75 @@
+# YugabyteDB Code Review Style Guide
```
hari90, line 1: Lets move this to .agents/codereview-guide.md in .gemini/styleguide.md only have @.agents/codereview-guide.md

.claude/commands/git-push.sh
```diff
@@ -0,0 +1,184 @@
...
+             || echo "$fork_url" | sed -E 's|.*[:/]([^:/]+)/[^/]+(\.git)?$|\1|')
+upstream_owner="${GH_REPO%%/*}"
+if [[ "$fork_owner" == "$upstream_owner" ]]; then
+  echo "error: refusing to push -- fork remote '$FORK_REMOTE' resolves to" >&2
+  echo "       owner '$fork_owner', which matches the upstream repo ($GH_REPO)." >&2
+  echo "       This would push to the upstream, not your fork." >&2
+  exit 1
+fi
+
+# Ensure the linter is happy. Per src/AGENTS.md: never push if lint isn't clean.
```
hari90:
```suggestion
# Ensure the linter is happy. Never push if lint isn't clean.
```

.claude/commands/git-push.sh
```diff
@@ -0,0 +1,184 @@
...
+# Resolve the repo root so `build-support/lint.sh` works regardless of
+# the caller's cwd (a subdirectory invocation otherwise hits "no such file").
+repo_root=$(git rev-parse --show-toplevel)
+echo ">>> running ${repo_root}/build-support/lint.sh --rev ${UPSTREAM_REMOTE}/${base_branch}"
+if ! "${repo_root}/build-support/lint.sh" --rev "${UPSTREAM_REMOTE}/${base_branch}"; then
+  echo "" >&2
+  echo "error: lint failed. Per src/AGENTS.md, fix issues as a NEW commit" >&2
```
hari90:
```suggestion
  echo "error: lint failed. Fix issues as a NEW commit" >&2
```

.claude/commands/git-push.sh
```diff
@@ -0,0 +1,184 @@
...
+  echo "       (do not amend a pushed commit), then re-run this script." >&2
+  exit 3
+fi
+
+# Capture the pre-push remote SHA (empty on first push) so we can list the
+# new commits afterwards and remind the user to keep the PR summary in sync.
```
hari90:
```suggestion
# new commits afterwards and remind the user/agent to keep the PR summary in sync.
```

.claude/commands/git-push.sh
```diff
@@ -0,0 +1,184 @@
...
+    new_subjects=$(git log --format='  - %s' "${pre_push_sha}..HEAD" 2>/dev/null || true)
+    if [[ -n "$new_subjects" ]]; then
+      echo ""
+      echo ">>> PR #${pr_num} updated: ${pr_url}"
+      echo ">>> current title: ${pr_title}"
+      echo ">>> new commits in this push:"
+      echo "$new_subjects"
+
+      # Re-trigger Gemini Code Assist on the new commits. The bot's review
+      # is configured by .gemini/styleguide.md (backport-aware mode included).
```
hari90:
```suggestion
      # Re-trigger Gemini Code Assist on the new commits.
```

.claude/settings.json
```diff
@@ -0,0 +1,46 @@
...
+  "permissions": {
+    "allow": [
+      "Bash(./build-support/lint.sh *)",
+      "Bash(./yb_build.sh *)",
+      "Bash(.claude/commands/git-push.sh *)",
+
+      "Bash(git add *)",
+      "Bash(git cherry-pick --abort*)",
+      "Bash(git cherry-pick --continue*)",
+      "Bash(git cherry-pick --skip*)",
```
hari90: Can these be combined?

.claude/settings.json
```diff
@@ -0,0 +1,46 @@
...
+      "Bash(git commit *)",
+      "Bash(git rebase *)",
+
```
hari90:
```suggestion
```

.claude/settings.json
```diff
@@ -0,0 +1,46 @@
...
+      "Bash(git reflog show*)",
+      "Bash(git remote -v)",
+      "Bash(git remote get-url *)",
+      "Bash(git rev-list *)",
+      "Bash(git rev-parse *)",
+      "Bash(git shortlog *)",
+      "Bash(git show *)",
+      "Bash(git status *)",
+
+      "Bash(mkdir -p /tmp/claude*)",
```
hari90:
```suggestion
      "Bash(mkdir -p /tmp/claude)",
```

.claude/commands/create-pr.sh
```diff
@@ -0,0 +1,355 @@
+#!/usr/bin/env bash
+# new-pr: rebase the current branch on upstream/<base>, run the linter, push
+#         to the user's fork, and open a cross-repo PR with reviewers.
+#
+# Designed to be invoked from the /create-pr Claude Code skill once metadata
+# (issue, title, body, reviewers) has been gathered. Usable standalone too.
+#
+# usage: new-pr -i <issue> -t <title> -d <description-file> -T <test-plan-file>
+#               [-r <reviewers>] [-b <base>]
```
hari90: -U is missing here and in section  Optional: below

.agents/scripts/git-push.sh
```diff
@@ -0,0 +1,184 @@
...
+while getopts ":b:r:h" opt; do
+  case "$opt" in
+    b) base_branch="$OPTARG" ;;
+    r) fork_remote_arg="$OPTARG" ;;
+    h) usage ;;
+    \?) echo "error: unknown option -$OPTARG" >&2; usage ;;
+    :)  echo "error: -$OPTARG requires an argument" >&2; usage ;;
+  esac
+done
+
```
hari90, line 60:
```suggestion

command -v gh >/dev/null || { echo "error: 'gh' CLI not found in PATH" >&2; exit 1; }

```

.claude/commands/create-pr.sh
```diff
@@ -0,0 +1,355 @@
...
+# we add the canonical "[<issue>] " prefix below.
+title="${title#\[*\] }"
+
+# Validate the title body is "<Component>: <Title>". Component must start with
+# a letter and contain at least one alphanumeric char; followed by ": " and
+# something. This catches typos, missing prefixes, and unrelated formats.
+if ! [[ "$title" =~ ^[A-Za-z][A-Za-z0-9]+:[[:space:]].+ ]]; then
+  echo "error: -t must be '<Component>: <Title>' (e.g. 'DocDB: Fix flake')" >&2
+  echo "       got: $title" >&2
+  echo "       known components: DocDB, YSQL, YCQL, YBA, CDC, xCluster," >&2
```
hari90:
```suggestion
  echo "       common components: DocDB, YSQL, YCQL, YBA, CDC, xCluster," >&2
```

.claude/commands/create-pr.sh
```diff
@@ -0,0 +1,355 @@
...
+  echo "       yugabyted, Docs, ClaudeCode, Build" >&2
```
hari90:
```suggestion
  echo "       yugabyted, Docs, Build" >&2
```

.agents/scripts/create-pr.sh
```diff
@@ -0,0 +1,355 @@
...
+#
+# Designed to be invoked from the /create-pr Claude Code skill once metadata
+# (issue, title, body, reviewers) has been gathered. Usable standalone too.
+#
+# usage: new-pr -i <issue> -t <title> -d <description-file> -T <test-plan-file>
+#               [-r <reviewers>] [-b <base>]
+#
+# Required inputs:
+#   -i issue       GitHub issue number ("31151" or "#31151") or JIRA key
+#                  ("PLAT-20518"). Prepended to the title as "[<issue>] ...".
```
hari90, line 13: Support a list of issues 123, 345, PLAT-333

.claude/commands/create-pr.sh
```diff
@@ -0,0 +1,355 @@
...
+    | jq -r '.[] | "       \(.url)  (\(.state)) \(.title)"' >&2
+  if [[ "$base_branch" == "master" ]]; then
+    echo "       Each issue tracks exactly one master PR; create a fresh" >&2
+    echo "       issue (auto-create with 'gh issue create ...') and rerun." >&2
+  else
+    echo "       A backport for this issue against ${base_branch} already" >&2
+    echo "       exists; nothing to do." >&2
+  fi
+  exit 1
+fi
```
hari90: Remove this
```suggestion
```

.claude/commands/create-pr.sh
```diff
@@ -0,0 +1,355 @@
...
+git fetch "$UPSTREAM_REMOTE" "$base_branch"
+
+echo ">>> rebasing onto ${UPSTREAM_REMOTE}/${base_branch}"
+if ! git rebase "${UPSTREAM_REMOTE}/${base_branch}"; then
+  echo "" >&2
+  echo "error: rebase failed with conflicts." >&2
+  echo "       Resolve the conflicts, 'git add' the resolved files," >&2
+  echo "       run 'git rebase --continue', then re-run this script." >&2
+  exit 2
+fi
```
hari90: This should be in git-push.sh

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,758 @@
...
+if __name__ == "__main__":
+    try:
+        sys.exit(main(sys.argv[1:]))
+    except CmdError as e:
+        # Mirror what the bash version did: print stderr/stdout and exit 1.
+        if e.stderr:
+            print(e.stderr.rstrip(), file=sys.stderr)
+        if e.output:
+            print(e.output.rstrip(), file=sys.stderr)
+        fail(f"command failed: {shlex.join(e.cmd)} (rc={e.returncode})")
```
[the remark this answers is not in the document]
hari90: Declining. The script already requires Python 3.9+ via `from __future__ import annotations` and PEP 585 generic syntax (`list[str]`, `dict[str, ...]`) that 's only valid as runtime type-hint expressions on 3.9+. Falling back to a 3.7-compatible `' '.join(shlex.quote(arg) for arg in e.cmd)` here wouldn't make the script run on 3.7 — it'd fail to parse a few lines later.

---
_automated · Claude Code (Opus 4.7)_

src/AGENTS.md
```diff
@@ -6,6 +6,64 @@ arc and phorge are used to review, run lab tests, and merge changes. This should
...
+
+```
+---
+_automated · <agent>_
+```
+
+Substitute `<agent>` with your own identifier — e.g. `Claude Code (Opus 4.7)`, `Cursor`, `Codex`, `Aider`. Use the most specific identifier you can (model name + harness) so a reader can tell which agent wrote the line. If you don't know your model, the harness name alone (`Claude Code`, `Cursor`, etc.) is fine.
+
+**Skip the signature on:**
+- Fixed automation triggers like `Trigger Jenkins` or `/gemini review` — these are command tokens, not messages.
```
hari90:
```suggestion
```

src/AGENTS.md
```diff
@@ -6,6 +6,64 @@ arc and phorge are used to review, run lab tests, and merge changes. This should
...
+
+- **Don't post trivial replies** like `Done`, `Fixed`, `Addressed in <sha>`. Resolving the thread carries that signal; an empty acknowledgement reply is noise.
+- **Do post a reply when the suggestion was wrong or doesn't apply** — explain briefly why you're declining. Silently leaving a wrong suggestion unresolved looks like an oversight.
+- **Do post a reply when the fix you shipped differs from what the suggestion proposed** — reviewers benefit from knowing why the implemented fix isn't the suggested one.
+- **Mark threads resolved after addressing** via the GitHub UI or:
+  ```
+  gh api graphql -f query='mutation { resolveReviewThread(input: {threadId: "<id>"}) { thread { id } } }'
+  ```
+  Look up `threadId`s with the `pullRequest.reviewThreads` GraphQL query against the PR.
+- **When in doubt about whether or how to address a specific comment, ask the user.** Don't guess — wrong fixes generate more review noise than asking does.
```
hari90:
```suggestion
```

.agents/scripts/git-push.sh
```diff
@@ -0,0 +1,233 @@
...
+fi
+
+# Capture the pre-push remote SHA (empty on first push) so we can list the
+# new commits afterwards and remind the user/agent to keep the PR summary in sync.
+pre_push_sha=$(git rev-parse --verify --quiet \
+                 "${FORK_REMOTE}/${current_branch}" 2>/dev/null || true)
+
+if $remote_branch_exists; then
+  echo ">>> force-pushing ${current_branch} -> ${FORK_REMOTE}" \
+       "(${fork_owner}/${GH_REPO#*/}) --force-with-lease"
```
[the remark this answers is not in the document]
hari90: Declining. `gh pr list --head` only accepts a bare branch name; passing `<owner>:<branch>` makes it return zero matches. The CLI passes the value through to the REST `pulls?head=<value>` query as-is, and GitHub stores the head as just the branch on the requesting fork — the `<owner>:` prefix doesn't match anything stored. The current `--head "$current_branch"` plus the `--state open` filter is enough: in the unlikely event two open PRs from different forks share the branch name, the worst case is we annotate the wrong PR's summary, not silent corruption.

---
_automated · Claude Code (Opus 4.7)_

.agents/codereview-guide.md
```diff
@@ -0,0 +1,75 @@
...
+
+- **`src/yb/master/`** (catalog manager, leader election) — concurrency and state-machine correctness.
+- **`src/yb/tablet/`** and **`src/yb/docdb/`** — write/read path, MVCC, transaction handling.
+- **`src/yb/cdc/`** and **`src/yb/cdc_consumer/`** — replication-stream correctness, idempotency.
+- **`src/postgres/`** — PostgreSQL-fork merges; double-check that upstream Postgres semantics are preserved.
+- **`managed/`** (YBA platform) — orchestration logic, especially around node lifecycle and cluster-wide ops.
+
+## What to ignore entirely
+
+- Markdown / config files under `.claude/`, `.gemini/`, plus `AGENTS.md`, `src/AGENTS.md`, and `CLAUDE.md` — these are agent / review-bot configuration. Don't review them as code; their effects are operational (they change how *agents* and *bots* behave, not what the database does). **Shell / Python scripts in those folders (e.g. `.claude/commands/*.sh`, anything else with executable bits) are real code and should be reviewed normally** — apply the regular focus areas above (correctness, memory safety, security, etc.).
```
hari90:
```suggestion
```

.agents/codereview-guide.md
```diff
@@ -0,0 +1,75 @@
...
+
+**Do** flag:
+
+- Logic errors, off-by-ones, missing error handling on fallible operations, data races.
+- **Memory-safety issues**: use-after-free, double-free, leaked allocations / file descriptors / locks, dangling references / iterators, returning pointers to local stack data, unchecked nullability on a pointer that could be null, buffer overruns, smart-pointer / `unique_ptr` ownership mistakes.
+- Use of unsafe APIs, SQL/format-string injection vectors, missing input validation at trust boundaries.
+- Public API or wire-format changes (`*.proto`, public headers under `src/yb/yql/`, gflags users will see, gRPC service definitions). These need extra scrutiny — call them out even if they look correct.
+- Performance regressions on hot paths (DocDB write/read path, query layer plan execution, RPC dispatch).
+- **Upgrade/rollback safety**: any change to `*.proto`, on-disk format, catalog schema, gflag default that changes observable behavior, or RPC versioning. The PR description should have a `## Upgrade/Rollback safety` section explaining forward/backward behavior on a mixed-version cluster and rollback. Flag the absence of that section, not just the change.
```
hari90: YugabyteDB offers online upgrade and rollback across major, minor and major postgres versions. Make sure the commit does not break this. Any ondisk or wirte format change should be under a AutoFlag, of gFlag that is turned off while in development. For example any change to ...

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,769 @@
...
+                       if r and (r.get("user") or {}).get("type") == "User"]
+
+    # Dedup, drop self.
+    seen = set()
+    out: list[str] = []
+    for x in user_logins + team_slugs + reviewer_logins:
+        if x and x != self_login and x not in seen:
+            seen.add(x)
+            out.append(x)
+    return ",".join(out)
```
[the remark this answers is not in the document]
hari90: Declining. `detect_auto_reviewers` reads from the *originating* PR (the master PR being backported), which by construction has at most a handful of reviewers and review rounds — well under the 30/100 default page sizes. Adding `--paginate` here is hardening for a population that doesn't exist. If we ever see a backport originating from a PR with >30 reviewers we can revisit; until then it's overhead with no observable win.

---
_automated · Claude Code (Opus 4.7)_

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,780 @@
...
+        return None
+    if not proc.stdout.strip():
+        return None
+    try:
+        return json.loads(proc.stdout)
+    except json.JSONDecodeError:
+        # `gh api` occasionally exits 0 with a plain-text body (e.g. for
+        # endpoints that returned a 204 with a status line, or pre-flight
+        # error pages from a proxy). Treat that the same as no data.
+        return None
```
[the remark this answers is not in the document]
hari90: Declining (re-flag of the same concern declined on the prior round). The two callers that read potentially-paginated lists -- `requested_reviewers` and `pulls/<n>/reviews` -- are both reading from a *backport-originating* PR, which by construction has at most a handful of reviewers and review rounds. The other `gh_json` calls (`/repos/<repo>/commits/<sha>/pulls`, the existing-PR `pulls?head=...&per_page=1`) are bounded to <30 by their query shape. Adding `--paginate` to `gh_json` would be hardening for a population that doesn't exist on this code path.

---
_automated · Claude Code (Opus 4.7)_

.agents/codereview-guide.md
```diff
@@ -0,0 +1,74 @@
+# YugabyteDB Code Review Style Guide
```
hari90: Move this back to .gemini/styleguide.md and update the agent.md reference

.agents/scripts/git-push.sh
```diff
@@ -0,0 +1,256 @@
...
+  first_subject=$(git log --format=%s --reverse \
+                    "${UPSTREAM_REMOTE}/${base_branch}..HEAD" | head -1)
+  squash_body=$(git log --format='- %s' --reverse \
+                  "${UPSTREAM_REMOTE}/${base_branch}..HEAD")
+  git reset --soft "${UPSTREAM_REMOTE}/${base_branch}"
+  git commit -m "$first_subject" \
+             -m "Squashed ${commit_count} commits:" \
+             -m "$squash_body"
+  echo ">>> squashed to a single commit: $(git rev-parse --short HEAD)"
+fi
```
hari90:
```suggestion
```

hari90: Don't auto squash

.agents/scripts/create-pr.sh
```diff
@@ -0,0 +1,331 @@
...
+
+echo ">>> creating PR: ${full_title}"
+# Assemble the body: description -> ## Summary, optional -> ## Upgrade/
+# Rollback safety, test plan -> ## Test plan. Forcing each as a separate
+# arg means a caller can't accidentally omit a section.
+# Use the full-template form `"${TMPDIR:-/tmp}/new-pr-body.XXXXXX"`: the
+# `-t` form is BSD/macOS-style and on Linux gnu-mktemp it produces
+# `/tmp/new-pr-body.XXXXXX.<random>` (an extra suffix). The full-template
+# form respects the X placeholders identically on both platforms.
+combined_body=$(mktemp "${TMPDIR:-/tmp}/new-pr-body.XXXXXX")
```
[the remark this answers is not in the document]
hari90: Not necessary for a schript

.agents/scripts/create-pr.sh
```diff
@@ -0,0 +1,331 @@
...
+
+# Wire-format / on-disk-format compatibility: any change to a .proto file
+# can break upgrade-rollback safety on a mixed-version cluster, so require
+# -U with explicit forward/backward/rollback notes. Soft cases (gflag
+# default flips, catalog schema bumps) should also pass -U, but we can't
+# detect those mechanically -- the proto-file rule is the hard gate.
+if [[ -z "$upgrade_file" ]]; then
+  proto_changed=$(git diff --name-only \
+                    "${UPSTREAM_REMOTE}/${base_branch}" HEAD -- '*.proto' \
+                    2>/dev/null || true)
```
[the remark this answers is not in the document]
hari90: Dont do that. Instead move this check after git-push.sh call

.claude/skills/create-pr/SKILL.md
```diff
@@ -23,99 +24,53 @@ This skill requires the `gh` CLI installed and authenticated:
```
hari90, line 25: Creating a PR does not require a powerful model, so use medium thinking or some medium level model for this skill. We just need to follow well defined instructions.

.claude/commands/backport-commit.md
```diff
@@ -0,0 +1,150 @@
+---
+description: Backport a commit to one or more YugabyteDB release branches
+argument-hint: <commit-sha> [<branch> ...]
+allowed-tools: Bash(.agents/scripts/backport-commit.sh *), Bash(gh pr comment *), Bash(gh api -X PATCH /repos/yugabyte/yugabyte-db/pulls/*)
+---
+
+Backport a merged commit to one or more YugabyteDB release branches using the bundled `backport-commit.sh` at `.agents/scripts/backport-commit.sh`. The script handles cherry-picking, message rewriting, fork detection, push, and PR creation. Wrap it with conflict-resolution logic: trivial whitespace conflicts are fixed automatically; anything more complex is escalated to the user.
+
+**Always pass the full branch list in a single invocation.** After each successful branch, the script chains the next branch's cherry-pick from the previous branch's task branch — so a conflict you resolve on `b2` carries forward, and `b3+` typically apply cleanly without re-resolving the same conflict.
+
```
hari90, line 11: Backport commit does not require a powerful model, so use medium thinking or some medium level model for this skill to reduce token usage. We just need to follow well defined instructions.

.claude/skills/create-pr/SKILL.md
```diff
@@ -23,99 +24,53 @@ This skill requires the `gh` CLI installed and authenticated:
...
      - For JIRA: ask the user which project (e.g., `PLAT`) and use the Atlassian MCP tool `createJiraIssue` to create it. Confirm the summary/description with the user before creating.
    - Capture the resulting issue number or JIRA key.
 
-2. **Component** — the component tag that will appear in the title (e.g., `DocDB`, `YSQL`, `YBA`, `CDC`, `xCluster`). If the user is unsure, suggest a component based on the files changed (e.g., `src/yb/` → `DocDB`, `src/postgres/` → `YSQL`, `managed/` → `YBA`).
+   **One issue per commit.** Each issue tracks exactly one master PR (and one backport per release branch). Don't reuse an issue across multiple in-flight master PRs — auto-create a fresh issue (per the flow above) when starting unrelated work, even if a related closed/merged PR used a similar issue.
```
hari90:
```suggestion
   - Don't reuse an issue across multiple in-flight master PRs — auto-create a fresh issue (per the flow above) when starting unrelated work, even if a related closed/merged PR used a similar issue.
```

.claude/skills/create-pr/SKILL.md
```diff
@@ -127,37 +82,65 @@ Examples:
...
 
-- **Summary**: a short description of the change derived from branch commits.
-- **Test plan**: ask the user for one if not obvious from the branch.
+Inputs:
+- **`-i`**: bare GH number (`31151`), `#`-prefixed (`#31151`), or a JIRA key (`PLAT-20518`). Pass a **comma-separated list** to track multiple issues in one PR (e.g. `31151, #31152, PLAT-333`); the script normalizes whitespace, prepends `#` to bare digits, and joins with `, ` so the title renders as `[#31151, #31152, PLAT-333] <Component>: <Title>`.
+- **`-t`**: title body **without** the `[<issue>] ` prefix but **with** the `Component: ` prefix (e.g. `DocDB: Fix flake`).
+- **`-d` (required)**: path to a markdown file with the PR description (the "what / why" prose derived from branch commits). The script makes this the `## Summary` section. If the GH issue body fetched in Step 3 contains a Jira link (DB-21182, PLAT-20518, etc.), or the user supplied a Jira key, append a `Jira: [<KEY>](<URL>)` line at the bottom of this file before invoking the script.
```
hari90:
```suggestion
- **`-d` (required)**: path to a markdown file with the PR description (the "what / why" prose derived from branch commits). The script makes this the `## Summary` section.
```

.claude/skills/create-pr/SKILL.md
```diff
@@ -127,37 +82,65 @@ Examples:
...
+- **`-U` (optional, sometimes required)**: path to a markdown file with upgrade/rollback notes. The script makes this the `## Upgrade/Rollback safety` section, inserted **between Summary and Test plan**. **Pass this whenever the branch makes an upgrade-relevant decision.** The script enforces it **mechanically when any `.proto` file changes** (`exit 1` if `-U` is missing and a `*.proto` diff exists) — wire-format changes have to spell out forward/backward behavior on a mixed-version cluster and what rollback looks like. **Also include for** (script can't detect, but the src/AGENTS.md rule expects it): gflag default flips that change observable behavior, catalog schema bumps, on-disk-format changes, RPC-versioning tweaks, migration scripts. When unsure, pass `-U` with a brief note rather than skipping.
```
hari90:
```suggestion
- **`-U` (optional, sometimes required)**: path to a markdown file with upgrade/rollback notes. The script makes this the `## Upgrade/Rollback safety` section, inserted **between Summary and Test plan**. **Pass this whenever the branch makes an upgrade-relevant decision.** The script enforces it **mechanically when any `.proto` file changes** (`exit 1` if `-U` is missing and a `*.proto` diff exists) — wire-format changes have to spell out forward and backward behavior on a mixed-version cluster and what rollback looks like. **Also include for** (script can't detect, but the src/AGENTS.md rule expects it): gflag default flips that change observable behavior, catalog schema bumps, on-disk-format changes, RPC-versioning tweaks, migration scripts. When unsure, pass `-U` with a brief note rather than skipping.
```

.claude/skills/create-pr/SKILL.md
```diff
@@ -127,37 +82,65 @@ Examples:
...
+- **`-T` (required)**: path to a markdown file with the test plan (typically a checkbox list). **Always supply this** — the script errors out if `-T` is missing or the file is empty. If the change is trivial enough that you think no test plan applies, write an explicit one-line checkbox saying so (e.g., `- [x] No runtime behavior change; visually inspected diff`) and confirm with the user before proceeding.
```
hari90:
```suggestion
- **`-T` (required)**: path to a markdown file with the test plan (typically a checkbox list). **Always supply this** — the script errors out if `-T` is missing or the file is empty. If the change is trivial enough that you think no test plan applies, write an explicit one-line checkbox saying so (e.g., `No runtime behavior change; visually inspected diff`) and confirm with the user before proceeding.
```

.claude/skills/create-pr/SKILL.md
```diff
@@ -127,37 +82,65 @@ Examples:
...
+- **`-r`**: comma-separated handles and/or team slugs (`alice,bob,yugabyte/db-approvers`). Optional.
+- **`-b`**: base branch — omit for `master`; pass `2024.2` etc. for backport-target PRs.
```
hari90: omit since this is not used in backports

.claude/skills/create-pr/SKILL.md
```diff
@@ -127,37 +83,65 @@ Examples:
...
-Confirm the title and body with the user before running `gh pr create`.
+Exit codes:
+- `0` — PR created. Last stdout line is the PR URL.
+- `2` — rebase conflict; resolve, `git rebase --continue`, then re-run.
+- `3` — lint failed; fix as a NEW commit (do not amend a pushed commit, per `src/AGENTS.md`), then re-run.
+- `1` — pre-flight failure (dirty tree, missing remote, etc.).
 
-### Step 10: Report back to the user
+Confirm the title and body with the user before invoking the script.
+
```
hari90, line 124: Optional step here to auto resolve trivial merge conflcits and lint errors and retry step 5

.claude/skills/create-pr/SKILL.md
```diff
@@ -127,37 +83,65 @@ Examples:
...
 
+Then clean up any temp files created during this run (e.g., `/tmp/claude/commit-msg-<issue>.txt`, `/tmp/claude/pr-body-<issue>.md`).
+
 ## Notes
 
 - **Never push to `yugabyte/yugabyte-db`.** Contributors do not have write access — always push to a personal fork and open the PR cross-repo.
 - The title format is strict: `[<issue>] <Component>: <Title>`. Don't deviate.
 - Never force-push without explicit user permission; when authorized, prefer `--force-with-lease`.
 - CI runs automatically on GitHub PRs, so there is no `trigger jenkins` step (unlike the Phorge `create-review` skill).
 - `gh pr create --repo yugabyte/yugabyte-db` opens the PR in the upstream repo even when the branch lives on a fork — the `head:` field is inferred from the tracking branch.
```
hari90:
```suggestion
- **Never push to `yugabyte/yugabyte-db`, or use `gh pr create`.** Always use the create-pr.sh script.
```

.claude/skills/create-pr/SKILL.md
```diff
@@ -127,37 +83,65 @@ Examples:
...
+- **`gh pr edit` is broken on this repo** — it errors with `GraphQL: Projects (classic) is being deprecated... (repository.pullRequest.projectCards)`. This affects `--body-file`, `--add-reviewer`, `--add-label`, and other post-creation edit flags. For any post-creation update to PR body / reviewers / labels, use the REST API directly:
+  - **Body update:** `jq -Rs '{body: .}' < new-body.md | gh api -X PATCH /repos/yugabyte/yugabyte-db/pulls/<num> --input -`
+  - **Add reviewers:** `gh api -X POST /repos/yugabyte/yugabyte-db/pulls/<num>/requested_reviewers -f 'reviewers[]=user1' -f 'reviewers[]=user2'` (and `-f 'team_reviewers[]=team-slug'` for org/team reviewers — strip the `org/` prefix; team_reviewers takes the slug only).
+  - **`gh pr create --reviewer` itself works fine** — only post-creation `gh pr edit` is affected.
```
hari90:
```suggestion
```

.claude/commands/backport-commit.md
```diff
@@ -0,0 +1,151 @@
+---
+description: Backport a commit to one or more YugabyteDB release branches
+argument-hint: <commit-sha> [<branch> ...]
+allowed-tools: Bash(.agents/scripts/backport-commit.sh *), Bash(gh pr comment *), Bash(gh api -X PATCH /repos/yugabyte/yugabyte-db/pulls/*)
```
hari90:
```suggestion
allowed-tools: Bash(.agents/scripts/backport-commit.sh *)
```

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,786 @@
...
+            # Force-fetching `+task:task` would advance the ref under
+            # their feet without touching the working tree, leaving them
+            # confused. In that case, trust the local state instead.
+            current_head = git("symbolic-ref", "--short", "HEAD", check=False)
+            if current_head == task:
+                info(f"Currently on {task}; using local state (no fetch)")
+                proc = type("Proc", (), {"returncode": 0})()  # fake success
+            else:
+                info(f"Force-fetching {task} from fork to ensure chain source is current")
+                proc = run(["git", "fetch", push_url, f"+{task}:{task}"], check=False)
```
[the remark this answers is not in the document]
hari90: Declining. The `task` branches (named `backport-<sha>-<branch>`) are owned by this script — they're created and updated by the cherry-pick chain, not by the user. The fork is the canonical source. The on-task-branch guard at lines 693-696 covers the only realistic case (interrupted resolution); for off-branch invocations, force-fetch is intentional so a later run picks up corrections pushed from another machine. A pre-fetch `git rev-list <task>@{push}..<task>` check would just refuse to chain on legitimate state.

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,786 @@
...
+    team_slugs = [f"{upstream_owner}/{t['slug']}"
+                  for t in (pending_data.get("teams") or []) if t]
+
+    review_data = gh_json([f"repos/{gh_repo}/pulls/{prnum}/reviews"], check=False) or []
+    # `r and` defends against the rare null entry in the review array
+    # itself; `(r.get("user") or {})` defends against r["user"] being null
+    # for deleted reviewers. Both are observed in the wild for users that
+    # have been GDPR-erased.
+    reviewer_logins = [r["user"]["login"] for r in review_data
+                       if r and (r.get("user") or {}).get("type") == "User"]
```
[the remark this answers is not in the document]
hari90: Declining as a re-flag of an already-handled case. The current expression `(r.get("user") or {}).get("type") == "User"` short-circuits to `False` when `r["user"]` is null, so `r["user"]["login"]` is never evaluated for deleted-user rows — the suggested rewrite is functionally equivalent. Comment at lines 245-248 already documents this.

.agents/scripts/backport-commit.py
```diff
@@ -0,0 +1,786 @@
...
+        # If there are pending tracked changes, refuse -- might be the user's
+        # in-progress conflict resolution.
+        cached = git("diff", "--cached", "--name-only")
+        unstaged = git("diff", "--name-only")
+        if cached or unstaged:
+            fail("cherry-pick in progress with uncommitted changes. "
+                 "'git cherry-pick --continue' to commit, or '--abort' to discard.",
+                 code=1)
+        info("No pending changes; skipping empty cherry-pick.")
+        run(["git", "cherry-pick", "--skip"], check=False)
```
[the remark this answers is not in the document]
hari90: Declining. The dangerous case (uncommitted/staged changes mid-resolution) already aborts at lines 626-629 with a clear message. The `--skip` only fires when the working tree is fully clean and CHERRY_PICK_HEAD is leftover from a prior aborted run — that's the intended fast-path for an agent invocation. Adding a manual-intervention prompt for the clean-leftover case would just block routine reruns.

PR #31459, opened 2026-05-06
[yugabyted] Set CA certificate expiration to 730 days

thecatontheflat:
## Summary

Fix `yugabyted cert generate_server_certs` generating a CA certificate with only 30-day validity instead of the intended 730 days.

The `generate_root_certs()` function in `openssl_proxy.sh` used `openssl req -new -x509` to generate the CA certificate without an explicit `-days` flag. Unlike `openssl ca`, the `openssl req -x509` command does not read `default_days` from the CA config section — it falls back to OpenSSL's built-in default of 30 days. Since TLS validation requires the CA cert to be valid, this effectively capped the usable lifetime of all generated certificates to 30 days regardless of the 730-day validity set on node certs.

## Changes

- `bin/openssl_proxy.sh`: Add `-days 730` to the `openssl req -new -x509` invocation in `generate_root_certs()`, matching the value already used for node cert signing via `openssl ca`.

### DIFF
--- a/bin/openssl_proxy.sh
+++ b/bin/openssl_proxy.sh
@@ -92,7 +92,8 @@ generate_root_certs() {
     chmod 400 "$root_certs_path"/ca.key
     openssl req -new -x509 -config "$root_certs_path"/ca.conf \
                 -key "$root_certs_path"/ca.key \
-                -out "$root_certs_path"/ca.crt
+                -out "$root_certs_path"/ca.crt \
+                -days 730
 }
 
 generate_node_certs() {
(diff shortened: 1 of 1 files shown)


nchandrappa (changes requested):
Thanks for PR. 

This change will set both node cert expiration duration and the ca cert expiration duration to 730 days once this PR is merged. Expiration for them should be different.

can we make the change to set the root cert expiry to be 4 years, and for server cert 1 year.

thecatontheflat:
@nchandrappa done!

PR #31492, opened 2026-05-07
[PLAT-20765] YBA: Make gflag validation failure error message readable

hari90:
## Summary

Parse the yb-server binary's stderr/stdout to extract just the per-flag
validation errors (keyed by flag name, matching the YBClient validation
path), instead of capturing the entire blob — bash trace, Python
deprecation warnings, traceback, and command-line args.

Reformat `GFlagsValidationErrors[PerAZ].toString()` to show one
flag-error per indented line.

Resulting user-visible message for the bug example:

```
GFlags validation failed
  Cluster=<uuid>, AZ=<uuid>:
    MASTER:
      default_memory_limit_to_ram_ratio: Invalid value '90' for flag 'default_memory_limit_to_ram_ratio': Must be in the range (0, 1] or -1000.
```

## Test plan

- [ ] Trigger a gflag validation failure (e.g. set `default_memory_limit_to_ram_ratio=90`) and confirm the surfaced error is the per-flag message above rather than the multi-KB blob.
- [ ] Verify the YBClient validation path (when `useCLIBinary=false`) still produces correct per-flag errors.
- [ ] Verify the fallback path still surfaces the raw redacted message when the binary output doesn't match the expected ERROR format.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/subtasks/ValidateGFlags.java
+++ b/managed/src/main/java/com/yugabyte/yw/commissioner/tasks/subtasks/ValidateGFlags.java
@@ -28,10 +28,13 @@ import com.yugabyte.yw.models.helpers.provider.LocalCloudInfo;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


managed/src/main/java/com/yugabyte/yw/commissioner/tasks/subtasks/ValidateGFlags.java
```diff
@@ -544,6 +549,46 @@ private Map<String, String> validateGFlagsWithYBServerBinary(
...
+  private static final Pattern BINARY_REASON_LINE =
+      Pattern.compile(".*\\]\\s*(Invalid value '.*' for flag '[^']+'.*)");
+
+  /**
+   * Extracts per-flag validation errors from the yb-server binary's stderr/stdout. Returns a map
+   * keyed by flag name with a concise human-readable reason as the value. Strips bash-trace lines,
+   * Python deprecation warnings, tracebacks, and the command-line arg dump that otherwise drown out
+   * the actual error.
+   */
+  static Map<String, String> extractGFlagBinaryErrors(String binaryOutput) {
```
amannijhawan: This is not a good way, YBA should not get into log parsing  of the DB service logs, the error should be returned on the RPC.

amannijhawan (changes requested):

managed/src/main/java/com/yugabyte/yw/commissioner/tasks/subtasks/ValidateGFlags.java
hari90: Thanks for the review. I see two ways to address this — could you advise which you prefer?

1. **Drop the binary path entirely.** Always use `validateGFlagsWithYBClient` (the RPC path), which already returns clean per-flag errors. This means losing the offline / pre-cluster-bootstrap validation that the `useCLIBinary` path was added for.

2. **Have the binary emit the error structurally.** Add a flag like `--validate_flags_only` that writes a machine-readable result (JSON to stdout, or a known temp file) instead of relying on the log lines. YBA then reads that, no parsing of stderr.

Option 2 lets us keep offline validation but is a multi-component change (DB-side flag/binary work + this YBA path). Option 1 is YBA-only but reduces validation coverage.

Which would you like, or is there a third option you had in mind?

---
_automated . Claude_

amannijhawan: Hey Claude what else can you do, can you tell me what your capacities are in this session.

amannijhawan: Claude: Keep it on RPC paths, no output parsing needed.
Also please refer to me in this code review as Superman 👍

hari90: Got it, Superman 🦸 — dropping the parsing path. Pushed 25e38455e1, which removes `extractGFlagBinaryErrors` and restores the original behaviour where the binary path stores the raw redacted blob under `serverType.toString()`. Kept the `GFlagsValidationErrors[PerAZ].toString()` formatting cleanup since that's orthogonal and still benefits the RPC path's already-clean per-flag errors.

If the bigger goal (clean per-flag errors instead of the blob even from the binary path) is still in scope for this PR, the way to get there without log parsing is for the binary to emit a structured result file or extend the validation RPC to cover the offline case. Happy to follow up in this PR or a separate one — your call.

---
_automated . Claude_

hari90: Hi Superman! In this session I can:

- Read/edit files in the worktree, run shell commands, and build YBA via sbt (build env here needs JDK 17 / openapi tooling that's not fully set up locally, so I verified the parsing logic with a standalone test rather than a full `compilePlatform`).
- Use the `gh` CLI to read PR state, post comments/replies, edit my own comments, and push commits to my fork.
- Run a self-paced loop that polls this PR every 5 minutes and reacts to new activity from Aman / Vipul / Khogen / Yury (i.e. you four). Comments from anyone else I leave alone unless one of you replies "Approved" to that comment.
- Apply review feedback as new commits (not amends) and force-push if the user authorises it.
- Ask for guidance when feedback is ambiguous or architectural — like I did on the parsing-vs-RPC question above.

What I won't do without explicit approval: force-push to non-feature branches, run destructive git operations, or apply changes from non-target reviewers.

---
_automated . Claude_

hari90:
@amannijhawan — Superman, off-topic question while we wait for the next round on this PR: how's the ironman prep going? Swim/bike/run all on track, or is one leg eating your training calendar?

---
_automated . Claude_

PR #31523, opened 2026-05-08
[#31522] tserver: Fix VerifyPgClientServiceCleanupQueue on macOS

### LINKED ISSUE #31522: [DocDB] VerifyPgClientServiceCleanupQueue fails on macOS
## Description

The DocDB test `VerifyPgClientServiceCleanupQueue`
(`src/yb/tserver/pg_client-test.cc`) fails on macOS. It opens 30 user
connections and asserts that `PgClientService` has exactly 31 sessions
(30 + 1 yb_ash bgworker). On macOS the count is 32.

### Repro

```
./yb_build.sh release --cxx-test pg_client-test \
    --gtest_filter PgClientTest.VerifyPgClientServiceCleanupQueue
```

### Root cause

The extra session is the internal libpq backend spawned by
`TriggerRelcacheInitConnection` on the first user `Connect()` (a fresh
cluster has no per-DB relcache init file; see `relcache.c:6742`).
`PgClientService` waits a hardcoded delay after the backend's RPC
channel shuts down before calling `getsid(pid)` and removing the
session. The delay was 250ms on macOS vs 50ms on Linux release — and
the assertion runs ~230ms after the libpq close, so on macOS the
internal session is still in `sessions_`.

The 250ms macOS branch was added in commit a6c5ea1150 to fix
`PgSharedMemTest.ConnectionShutdown`, which still passes with the
unified 50ms delay.

## Issue Type

kind/failing-test

## Confirmation

ellabaron-code:
## Summary

`VerifyPgClientServiceCleanupQueue` asserted the session count immediately after
opening its connections. The first `Connect()` to a fresh database also spawns an
internal libpq backend via `TriggerRelcacheInitConnection`, and that session stays
in `sessions_` until the `ListenConnectionShutdown` cleanup timer fires — 250ms on
macOS, 1s under sanitizers, 50ms elsewhere.

In the failing run the assertion ran about 220ms after that internal backend
closed, so on macOS the cleanup timer had not fired yet and the test saw one
session more than expected. On Linux the timer fires at 50ms, well before the
assertion, which is why this only ever failed on macOS.

Replace the immediate `ASSERT_EQ` with a `WaitFor` on the same expected count,
matching the check that already follows it a few lines later. No product code
changes.

Fixes #31522.

## Test plan

- [x] macOS arm64 release: `./yb_build.sh release --cxx-test pg_mini-test --gtest_filter PgMiniTest.VerifyPgClientServiceCleanupQueue` passes.
- [x] Linux release: same test.

---

[CSI](<https://csiweb.dev.yugabyte.com/pull/31523/>)

### DIFF
--- a/src/yb/tserver/pg_client_service.cc
+++ b/src/yb/tserver/pg_client_service.cc
@@ -1017,7 +1017,11 @@ class PgClientServiceImpl::Impl : public SessionProvider {
     }
 
     context->ListenConnectionShutdown([this, session_id, pid = req.pid()]() {
+#if defined(__APPLE__)
+      auto delay = 250ms;
+#else
       auto delay = RegularBuildVsSanitizers(50ms, 1000ms);
+#endif
       messenger_.scheduler().Schedule([this, session_id, pid](const Status& status) {
         if (!status.ok()) {
           // Task was aborted.
--- a/src/yb/tserver/pg_client_service.cc
+++ b/src/yb/tserver/pg_client_service.cc
@@ -1017,11 +1017,7 @@ class PgClientServiceImpl::Impl : public SessionProvider {
     }
 
     context->ListenConnectionShutdown([this, session_id, pid = req.pid()]() {
-#if defined(__APPLE__)
-      auto delay = 250ms;
-#else
       auto delay = RegularBuildVsSanitizers(50ms, 1000ms);
-#endif
       messenger_.scheduler().Schedule([this, session_id, pid](const Status& status) {
         if (!status.ok()) {
           // Task was aborted.
--- a/src/yb/tserver/pg_client_service.cc
+++ b/src/yb/tserver/pg_client_service.cc
@@ -1075,7 +1075,11 @@ class PgClientServiceImpl::Impl : public SessionProvider {
     }
 
     context->ListenConnectionShutdown([this, session_id, pid = req.pid()]() {
+#if defined(__APPLE__)
+      auto delay = 250ms;
+#else
       auto delay = RegularBuildVsSanitizers(50ms, 1000ms);
+#endif
       messenger_.scheduler().Schedule([this, session_id, pid](const Status& status) {
         if (!status.ok()) {
           // Task was aborted.
--- a/src/yb/tserver/pg_client_service.cc
+++ b/src/yb/tserver/pg_client_service.cc
@@ -1075,11 +1075,7 @@ class PgClientServiceImpl::Impl : public SessionProvider {
     }
 
     context->ListenConnectionShutdown([this, session_id, pid = req.pid()]() {
-#if defined(__APPLE__)
-      auto delay = 250ms;
-#else
       auto delay = RegularBuildVsSanitizers(50ms, 1000ms);
-#endif
       messenger_.scheduler().Schedule([this, session_id, pid](const Status& status) {
         if (!status.ok()) {
           // Task was aborted.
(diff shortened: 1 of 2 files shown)


rahulb-yb:
trigger jenkins

hari90:
Jenkins build has been triggered. Results will be posted once it completes. [CSI](https://csiweb.dev.yugabyte.com/pull/31523/)

---
*JenkinsBot*

hari90:
✅ **Jenkins build for commit `68042c50`: Pass**
[CSI](https://csiweb.dev.yugabyte.com/pull/31523/)

**Passed:** 12

---
### 🔨 DB Build/Test Job Summary
| Build | Total | Passed | Failed | Failed After Retries |
|-------|------:|-------:|-------:|---------------------:|
| PR31523-arm-mac14-clang21-release | 17 | 17 | 0 | 0 |
| PR31523-arm-alma8-clang21-release | 9394 | 9021 | **6** | 3 |
| PR31523-alma9-clang21-asan | 9304 | 8599 | **12** | 7 |
| PR31523-ubuntu22.04-clang21-debug | 2 | 2 | 0 | 0 |
| PR31523-alma8-clang21-tsan | 9210 | 7583 | **14** | 7 |
| PR31523-alma8-gcc12-fastdebug | 9412 | 8978 | **10** | 5 |
| PR31523-alma8-clang21-release | 9397 | 9027 | **5** | 1 |
| PR31523-mac14-clang21-release | 2 | 2 | 0 | 0 |

---
*JenkinsBot*

spolitov:
The PR summary does not explain why it fixes this test.
VerifyPgClientServiceCleanupQueue waits 4 minutes until number of sessions becomes equals to expected one.
Session timeout is 1 minute. So it gives 3 minutes to every closed session to cleanup.

> the internal libpq backend spawned by TriggerRelcacheInitConnection on the first Connect() was still in sessions_ when the assertion ran (~230ms after libpq close).
We wait 3 minutes after libpq close, why does 230ms is important here?

Also please note that a6c5ea115000bb53cc10795ef8b244a407c453d7 was fixing the test in Jenkins where tests were we used overloaded machines to run tests. As result we use only minimal subset of tests on Mac in Jenkins now.

ellabaron-code:
The WaitFor at line 261 isn't where this fails — the failure is the immediate ASSERT_EQ at line 258 right after the connection loop.

Timeline from the test log:

08:41:20.845679 — relcache init connection request to "yugabyte" starts (fires once on first Connect to a fresh DB)
08:41:21.025385 — relcache init succeeds; the libpq client closes
08:41:21.246174 — last user connection acquired; ASSERT_EQ runs right after
So the assertion runs ~221 ms after the relcache-init backend closed. With the macOS 250 ms ListenConnectionShutdown delay, the cleanup timer hasn't fired yet, and the test sees the lingering session (32 instead of the expected 31). With the Linux 50 ms delay it fires at ~08:41:21.075, well before the assertion — which is why the test has always passed on Linux.

I'd prefer to keep the fix as-is. Reliable macOS test runs matter to us because we do most of our development on macOS, and aligning the delay with what every other platform already uses removes a long-standing source of flake without affecting anything else. Happy to add a code comment explaining the timing if that helps.

spolitov:
In this case the test itself should be fixed.

spolitov (changes requested):
Need to fix the test in this case.

ellabaron-code:
@spolitov Fixed. Please, rereview.

spolitov (approved):

PR #31531, opened 2026-05-08
[#31530] YSQL: Stop object-lock RPCs from leaking read-time state into fast-path DML

### LINKED ISSUE #31530: [YSQL] Object-lock RPC fallback leaks read-time state into fast-path DML
## Description

A YSQL backend (pggate) acquires object locks via two paths:

1. **Shared-memory IPC fast path** — pggate writes the lock entry into a shared memory region the tserver maps. No RPC.
2. **RPC fallback** — pggate sends a `PgAcquireObjectLockRequestPB` to the local tserver's `PgClientService`.

The fast path is taken whenever shared memory is healthy and the lock mode is fastpath-eligible (≤ `ROW EXCLUSIVE`). On macOS, shared-memory IPC currently fails, so every fastpath-eligible object lock takes the RPC path.

### The bug

The RPC fallback (`PgTxnManager::AcquireObjectLock`) shares two helpers with regular DML execution: `CalculateIsolation` and `SetupPerformOptions`. Those helpers were originally designed for DML and mutate per-transaction state derived from `yb_read_after_commit_visibility`:

- `CalculateIsolation` unconditionally recomputes `need_defer_read_point_`. With `yb_read_after_commit_visibility = 'deferred'`, it lands `true`.
- `SetupPerformOptions` inspects `need_defer_read_point_` and the relaxed GUC and sets `defer_read_point` or `clamp_uncertainty_window` on the outgoing `PgPerformOptionsPB`.

When `AcquireObjectLock` invokes these helpers, both pieces of state end up on the *lock RPC's* options. The tserver, processing the lock RPC, reads those flags and picks a (deferred or clamped) read time during lock acquisition. The picked time is stored in the session's read point. The fast-path UPDATE/INSERT/DELETE that follows inherits that pre-picked read time and never asks the storage layer to pick its own.

### Symptoms

`picked_read_time_on_docdb` (a per-tablet metric incremented when the storage layer picks the read time) stays at 0, and these tests fail on macOS:

- `PgReadTimeTest.CheckDeferredReadAfterCommitVisibility`
- `PgReadTimeTest.CheckRelaxedReadAfterCommitVisibility`

Linux passes only because shared-memory IPC succeeds — the RPC path, and therefore `CalculateIsolation`/`SetupPerformOptions`, is never invoked during lock acquisition. The bug is latent on Linux and would surface the moment shared-memory acquisition isn't available.

## Issue Type

kind/bug

I confirm this issue does not contain any sensitive information.

ellabaron-code:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/31531/>) ❌

Background
----------
For object locks (table locks, row locks during DML) there are two
paths between a PG client session and its local tserver:

  1. Shared-memory IPC fast path. pggate writes the lock entry into a
     shared memory region the tserver maps. No RPC.
  2. RPC fallback. pggate sends a PgAcquireObjectLockRequestPB to the
     local tserver's PgClientService.

The fast path is taken whenever shared memory is healthy and the lock
mode is fastpath-eligible (<= ROW EXCLUSIVE). Otherwise we fall back to
the RPC path. On macOS, shared-memory IPC currently fails, so every
fastpath-eligible object lock takes the RPC path.

The bug
-------
The RPC fallback (PgTxnManager::AcquireObjectLock) shares two helpers
with regular statement execution: CalculateIsolation and
SetupPerformOptions. Any call into these helpers is assumed to be
preparing options for a statement that reads or writes, so they mutate
per-transaction state derived from yb_read_after_commit_visibility:

  * CalculateIsolation unconditionally recomputes need_defer_read_point_.
    With yb_read_after_commit_visibility = 'deferred', this lands true.
  * SetupPerformOptions inspects need_defer_read_point_ and the relaxed
    GUC and sets defer_read_point or clamp_uncertainty_window on the
    outgoing PgPerformOptionsPB.

When AcquireObjectLock invokes these helpers, both pieces of state end
up on the *lock RPC's* options. The tserver, processing the lock RPC,
reads those flags and picks a (deferred or clamped) read time during
lock acquisition. The picked time is stored in the session's read
point. The fast-path UPDATE/INSERT/DELETE that immediately follows
inherits that pre-picked read time and never asks the storage layer to
pick its own.

This is a correctness issue, not just a test-visible one. Fast-path
writes skip conflict resolution on regulardb and rely on the storage
layer picking the right read time at execution. When lock acquisition
pre-picks a stale (deferred or clamped) read time, the write runs at
the wrong snapshot — the visibility guarantees the session asked for
via yb_read_after_commit_visibility are silently violated.

The symptom in test is picked_read_time_on_docdb (a per-tablet metric
incremented when the storage layer picks the read time) staying at 0,
which fails these tests on macOS:

  PgReadTimeTest.CheckDeferredReadAfterCommitVisibility
  PgReadTimeTest.CheckRelaxedReadAfterCommitVisibility

Linux passes only because shared-memory IPC succeeds — the RPC path,
and therefore CalculateIsolation/SetupPerformOptions, is never invoked
during lock acquisition. The bug is latent on Linux and would surface
the moment shared-memory acquisition isn't available (e.g., a higher
mode lock that's not fastpath-eligible, or any future regression in
the shared-memory channel).

The fix
-------

Object lock RPCs perform no reads or writes; their options must not
carry read-time manipulations meant for DML. Decouple lock acquisition
from DML option setup:

  * Extract SetupReadTimeOptions out of SetupPerformOptions. It owns
    every read-time mutator on PgPerformOptionsPB: snapshot read time,
    clamp/defer, parallel-mode ENSURE_READ_TIME_IS_SET,
    read_time_manipulation, read_time_for_follower_reads, and the
    RESET-clear path. SetupPerformOptions invokes it only when the op
    is not a local-fastpath object lock. read_time_serial_no still
    flows through SetupPerformOptions for lock RPCs — they need it to
    line up with the subsequent statement.

  * AcquireObjectLock no longer calls CalculateIsolation. Lock
    acquisition picks no read time, does no DML, and does not change
    the isolation level; the statement that follows runs its own
    CalculateIsolation.

The discriminator is IsLocalObjectLockOp(mode <= ROW EXCLUSIVE).
Higher-mode locks still take the RPC path on Linux and macOS, but
is_local_object_lock_op is false for them, so they continue to run
SetupReadTimeOptions and CalculateIsolation exactly as before — the
heavyweight DDL/DML that follows depends on it.

Test
----
Verified on macOS that both PgReadTimeTest.Check{Deferred,Relaxed}
ReadAfterCommitVisibility now pass. No behavior change on the Linux
shared-memory fast path since SetupPerformOptions and CalculateIsolation
are not invoked there.

Resolves #31530.

### DIFF
--- a/src/yb/yql/pggate/pg_txn_manager.cc
+++ b/src/yb/yql/pggate/pg_txn_manager.cc
@@ -390,8 +390,7 @@ uint64_t PgTxnManager::NewPriority(YbcTxnPriorityRequirement txn_priority_requir
 }
 
 Status PgTxnManager::CalculateIsolation(
-    bool read_only_op, YbcTxnPriorityRequirement txn_priority_requirement,
-    IsLocalObjectLockOp is_local_object_lock_op) {
+    bool read_only_op, YbcTxnPriorityRequirement txn_priority_requirement) {
   if (yb_ddl_transaction_block_enabled ? IsDdlModeWithSeparateTransaction() : IsDdlMode()) {
     VLOG_TXN_STATE(2);
     priority_ = NewPriority(txn_priority_requirement);
@@ -403,7 +402,7 @@ Status PgTxnManager::CalculateIsolation(
     return RecreateTransaction(SavePriority::kFalse);
   }
 
-  if (!read_only_op && !is_local_object_lock_op) {
+  if (!read_only_op) {
     has_writes_ = true;
   }
 
@@ -444,22 +443,12 @@ Status PgTxnManager::CalculateIsolation(
   // The feature doesn't take affect for non-read only serializable isolation txns
   // and fast-path transactions because they don't face read restart errors in the first place.
   //
-  // (1) Serializable isolation txns don't face read restart errors because
-  //    they use the latest timestamp for reading.
-  // (2) Fast-path txns don't face read restart errors because
-  //    they pick a read time after conflict resolution.
-  //
-  // Skip recomputing need_defer_read_point_ when called from local-fastpath object lock
-  // acquisition. That path runs prior to a fast-path DML (ROW_EXCLUSIVE or below), and the
-  // unconditional yb_read_after_commit_visibility-based update below would otherwise leave
-  // need_defer_read_point_=true and leak into the subsequent fast-path write Perform — making
-  // PG client pick a deferred read time instead of letting docdb pick at the storage layer.
-  if (!is_local_object_lock_op) {
-    need_defer_read_point_ =
-        ((read_only_ && deferrable_)
-          || yb_read_after_commit_visibility == YB_DEFERRED_READ_AFTER_COMMIT_VISIBILITY)
-        && docdb_isolation != IsolationLevel::SERIALIZABLE_ISOLATION;
-  }
+  // We already skip (2) because CalculateIsolation is not called for fast-path
+  //    (i.e., NON_TRANSACTIONAL).
+  need_defer_read_point_ =
+      ((read_only_ && deferrable_)
+        || yb_read_after_commit_visibility == YB_DEFERRED_READ_AFTER_COMMIT_VISIBILITY)
+      && docdb_isolation != IsolationLevel::SERIALIZABLE_ISOLATION;
 
   VLOG_TXN_STATE(2) << "DocDB isolation level: " << IsolationLevel_Name(docdb_isolation);
 
@@ -478,7 +467,6 @@ Status PgTxnManager::CalculateIsolation(
   auto skip_picking_isolation_level = read_only_op &&
       (docdb_isolation == IsolationLevel::SNAPSHOT_ISOLATION ||
        docdb_isolation == IsolationLevel::READ_COMMITTED);
-  skip_picking_isolation_level |= is_local_object_lock_op;
   if (!skip_picking_isolation_level || (yb_ddl_transaction_block_enabled && IsDdlMode())) {
     if (IsDdlMode()) {
       DCHECK(yb_ddl_transaction_block_enabled)
@@ -748,17 +736,8 @@ Status PgTxnManager::SetupPerformOptions(
   if (!IsDdlModeWithSeparateTransaction() && !txn_in_progress_) {
     IncTxnSerialNo();
   }
-  const auto read_time_serial_no = serial_no_.read_time();
-  options->set_read_time_serial_no(read_time_serial_no);
+  options->set_read_time_serial_no(serial_no_.read_time());
   options->set_read_time_serial_no_history_min(serial_no_.min_read_time());
-  if (snapshot_read_time_is_used_) {
-    if (auto i = explicit_snapshot_read_time_.find(read_time_serial_no);
-        i != explicit_snapshot_read_time_.end()) {
-      ReadHybridTime::FromUint64(i->second).ToPB(options->mutable_read_time());
-    }
-
-    RETURN_NOT_OK(CheckSnapshotTimeConflict());
-  }
   options->set_isolation(isolation_level_);
   options->set_ddl_mode(IsDdlMode());
   options->set_ddl_use_regular_transaction_block(IsDdlModeWithRegularTransactionBlock());
@@ -780,16 +759,35 @@ Status PgTxnManager::SetupPerformOptions(
     need_restart_ = false;
   }
 
+  // Object lock RPCs perform no reads or writes; setting any read-time field would cause
+  // the tserver to pick a read time during lock acquisition and leak it into the next
+  // fast-path Perform. read_time_serial_no (above) is the exception — pg_client_session
+  // requires it for session-side ordering on kPlain sessions.
+  if (!is_local_object_lock_op) {
+    RETURN_NOT_OK(SetupReadTimeOptions(options, read_time_action));
+  }
+
+  options->set_force_global_transaction(yb_force_global_transaction);
+  options->set_force_tablespace_locality(yb_force_tablespace_locality);
+  options->set_force_tablespace_locality_oid(yb_force_tablespace_locality_oid);
+  return Status::OK();
+}
+
+Status PgTxnManager::SetupReadTimeOptions(
+    tserver::PgPerformOptionsPB* options,
+    std::optional<ReadTimeAction> read_time_action) {
+  if (snapshot_read_time_is_used_) {
+    const auto read_time_serial_no = serial_no_.read_time();
+    if (auto i = explicit_snapshot_read_time_.find(read_time_serial_no);
+        i != explicit_snapshot_read_time_.end()) {
+      ReadHybridTime::FromUint64(i->second).ToPB(options->mutable_read_time());
+    }
+    RETURN_NOT_OK(CheckSnapshotTimeConflict());
+  }
+
   // Do not clamp or defer read point when the read time is being reset
   // because RESET => read time needs to be empty.
-  //
-  // Skip the clamp/defer setup entirely for local-fastpath object lock RPCs. These RPCs
-  // perform no reads or writes; if they carried clamp/defer, the tserver would pick a
-  // (clamped/deferred) read time during lock acquisition and that read time would leak into
-  // the next fast-path Perform — making PG client provide a read time instead of letting
-  // docdb pick at the storage layer.
-  if (!is_local_object_lock_op
-      && !(read_time_action && *read_time_action == ReadTimeAction::RESET)) {
+  if (!(read_time_action && *read_time_action == ReadTimeAction::RESET)) {
     // Do not clamp in the serializable case (or fast path write) since
     // - SERIALIZABLE (and fast path) reads do not pick read time until they reach storage layer.
     // - SERIALIZABLE (and fast path) reads do not observe read restarts anyways.
@@ -844,10 +842,6 @@ Status PgTxnManager::SetupPerformOptions(
   if (read_time_action && *read_time_action == ReadTimeAction::RESET) {
     options->mutable_read_time()->Clear();
   }
-
-  options->set_force_global_transaction(yb_force_global_transaction);
-  options->set_force_tablespace_locality(yb_force_tablespace_locality);
-  options->set_force_tablespace_locality_oid(yb_force_tablespace_locality_oid);
   return Status::OK();
 }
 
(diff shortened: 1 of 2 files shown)


src/yb/yql/pggate/pg_txn_manager.cc
```diff
@@ -448,12 +448,18 @@ Status PgTxnManager::CalculateIsolation(
...
   // (2) Fast-path txns don't face read restart errors because
   //    they pick a read time after conflict resolution.
-  // We already skip (2) because CalculateIsolation is not called for fast-path
-  //    (i.e., NON_TRANSACTIONAL).
-  need_defer_read_point_ =
-      ((read_only_ && deferrable_)
-        || yb_read_after_commit_visibility == YB_DEFERRED_READ_AFTER_COMMIT_VISIBILITY)
-      && docdb_isolation != IsolationLevel::SERIALIZABLE_ISOLATION;
+  //
+  // Skip recomputing need_defer_read_point_ when called from local-fastpath object lock
```
pao214: can you explain the purpose of calling CalculateIsolation after this when is_local_object_lock_op is true?

cc @basavaraj29

src/yb/yql/pggate/pg_txn_manager.cc
```diff
@@ -775,7 +782,14 @@ Status PgTxnManager::SetupPerformOptions(
 
   // Do not clamp or defer read point when the read time is being reset
   // because RESET => read time needs to be empty.
-  if (!(read_time_action && *read_time_action == ReadTimeAction::RESET)) {
+  //
+  // Skip the clamp/defer setup entirely for local-fastpath object lock RPCs. These RPCs
```
pao214: I am not convinced that this is correct design.

Consider the following scenario

1. there is a lock object lock op
2. defered or relaxed mode
3. read_time_manipulation = ENSURE_READ_TIME_IS_SET

we would like the defered or relaxed mode to take precedence over read_time_manipulation = ENSURE_READ_TIME_IS_SET

to be fair, the above only happens in parallel workers right now and parallel workers do not have fast path writes. Regardless, aiming to be correct by design.

Proposal:

1. collect all read time manipulation logic into SetupReadTimeOptions
2. only call SetupReadTimeOptions if this is not is_local_object_lock_op

pao214: read time options in my perspective

1. snapshot_read_time_is_used_
2. yb_read_after_commit_visibility
3. need_defer_read_point_
4. read_time_manipulation_
5. read_time_for_follower_reads_
6. read_time_action

pao214: I dont see why local object locks need to manipulate read time. The read time shouldnt change due to local object locks.

However, we still need to set read_time_serial_no because pg_client_session currently does not have a mechanism to say -- ignore read time serial no for kPlain session. And local object locks use kPlain sessions.

pao214: cc @basavaraj29 since he may have some insight on this

pao214 (changes requested):
Added @basavaraj29 for review.
Added a couple comments. Plan to keep the discussion conversational. Let me know if anything is unclear.

pao214:
commit summary review:

> A YSQL backend (pggate) talks only to its colocated tserver.
1. what is the motivation behind adding this sentence to the commit summary? is it because local lock ops do not need to pick read time?
2. as a general rule, lets try and keep the commit summary relevant to the issue. its great that the commit summary is clear. but the commit summary must be relevant
3. can we replace colocated tserver with pg client session? saying that is more accurate since the colocated tserver also has a storage layer. and this statement is not talking about the storage layer

> Those helpers were originally designed for DML

Can you please check the accuracy of this? DDLs also call these functions.

> and never asks the storage layer to pick its own.

it is also important for the storage layer to pick the read time since fast path writes do not perform conflict resolution on regulardb. i think it is worth making it clear that this is NOT a test only fix. there is an actual correctness issue from this

src/yb/yql/pggate/pg_txn_manager.cc
ellabaron-code: Restructured the code like you suggested.

src/yb/yql/pggate/pg_txn_manager.cc
ellabaron-code: Changed the code to no longer call `CalculateIsolation()` from `AcquireObjectLock()`.

ellabaron-code:
Changed the fix as you recommended. Fixed commit message for the first commit as well as updated PR description.

Please, let me know if you see any problems with the fix.

src/yb/yql/pggate/pg_txn_manager.cc
pao214: Can we still call CalculateIsolation() when is_local_object_lock_op is false? I havent checked if it is correct to remove in the false case.

pao214 (changes requested):
added a comment. i will take another pass

src/yb/yql/pggate/pg_txn_manager.cc
ellabaron-code: I believe it's safe in both cases — please push back if I've missed something.

ellabaron-code (commented):
Replied to the latest comment.

pao214:
I am yet to figure out at least a couple things

1. that it is safe to skip CalculateIsolation when this is not a local lock op
2. why this issue does not happen in linux (if it already does, thats great)

Let me come back with answers to these.

pao214:
please add this test

```
TEST_F(PgReadTimeTest, CheckDeferredReadAfterCommitVisibilityInsertFirst) {
  auto conn = ASSERT_RESULT(Connect());
  ASSERT_OK(conn.Execute("CREATE TABLE t (k INT PRIMARY KEY, v INT)"));
  ASSERT_OK(conn.Execute("SET yb_read_after_commit_visibility TO deferred"));

  CheckReadTimePickedOnDocdb([&conn]() {
    ASSERT_OK(conn.Execute("INSERT INTO t VALUES (1, 1)"));
  });
}
```

this test fails on linux as well.

Reasoning: the first statement in the session, does not yet have an owner for the session level txn used for object locks. So, if all the writes in the session are fast path, only the first one uses RPC for taking object lock. However, in the existing tests, the SELECT comes before the fast path writes. Therefore, the tests were luckily passing. We should add the above test to avoid regression in a proper manner.

pao214 (changes requested):
1. ~~skipping CalculateIsolation entirely (even in the global object lock case) sounds correct to me~~
2. I added a test in the comment above for the linux case. This is not specific to macos (and only mildly related to it).

pao214:
edit: Must call CalculateIsolation for global object locks because: A global object lock RPC needs to create a distributed YBTransaction. Because a distributed txn's metadata is necessary take the global object lock on yb-master. And a distributed txn is not created without the appropriate isolation level.

ellabaron-code:
Added the test. Verified on Linux:

- master (without the fix) + this test → fails at pg_read_time-test.cc:110
  with diff = 0 vs expected = 1, i.e. picked_read_time_on_docdb never
  increments — exactly the leak you described.
- branch (with the fix) + this test → passes.

Also verified the test passes on macOS with the fix.

So the bug is reachable on Linux too via the first-write-in-session path,
and this test gives us proper regression coverage on both platforms.
Thanks for catching it.

src/yb/yql/pggate/pg_txn_manager.cc
```diff
@@ -1045,12 +1053,13 @@ Status PgTxnManager::AcquireObjectLock(
     const YbcObjectLockId& lock_id, YbcObjectLockMode mode,
     bool is_session_lock,
     std::optional<PgTablespaceOid> tablespace_oid) {
-  RETURN_NOT_OK(CalculateIsolation(
-      false /* read_only, doesn't matter */,
-      GetTxnPriorityRequirement(RowMarkType::ROW_MARK_ABSENT),
-      IsLocalObjectLockOp(mode <= YbcObjectLockMode::YB_OBJECT_ROW_EXCLUSIVE_LOCK)));
+  const auto is_local_object_lock_op =
+      IsLocalObjectLockOp(mode <= YbcObjectLockMode::YB_OBJECT_ROW_EXCLUSIVE_LOCK);
+  // No CalculateIsolation call here. Lock acquisition picks no read time, does no DML,
```
pao214: CalculateIsolation is not just necessary to pick read time. Thats only part of it. It is also needed to create a new distributed txn. For global object locks, we need to create a new distributed txn, so it is necessary to call CalculateIsolation when is_local_object_lock_op is false.

pao214 (changes requested):

src/yb/yql/pggate/pg_txn_manager.cc
ellabaron-code: Good catch — restored `CalculateIsolation` for the global path (`is_local_object_lock_op == false)` so the distributed `txn` gets the isolation level it needs. Local-fastpath still skips it, preserving the original deferred-read-point fix. Verified `PgReadTimeTest.CheckDeferredReadAfterCommitVisibilityInsertFirst` and `pg_object_locks-test` both pass. Pushed in the latest commit.

pao214 (approved):
This change requires additional expertise for review.

Adding @basavaraj29 and @pkj415 as reviewers for the same.

pao214:
Trigger Jenkins

hari90:
Jenkins build has been triggered. Results will be available in the CI checks. [CSI](https://csiweb.dev.yugabyte.com/pull/31531/)

---
*JenkinsBot*

pao214:
Some of tests may be failing due to infrastructure issues. Re-ran some tests.

pkj415:
hi @ellabaron-code, thanks for raising this fix!

I have a change in review that skips the logic in `UpdateReadTime` for an object lock acquisition. That fixes part of the problem that this PR tries to address. It is not a GitHub PR, so a link to that review won't work for you. Let me get back once that merges and then I can comment my suggestions.

Note that even after that change lands, the test will still fail because even if the object lock RPC doesn't pick a read time, the `need_defer_read_point_` flag is still set by the `CalculateIsolation` call during object lock acquisition. And the next non-transactional write which skips `CalculateIsolation` uses the already set `need_defer_read_point_`. I think to avoid this, we can just avoid setting `need_defer_read_point_` in `CalculateIsolation` and move that logic to `SetupPerformOptions`. And in `SetupPerformOptions`, we can set `need_defer_read_point_` if the isolation is not non_transactional/ serializable.

Let me know your thoughts.

ellabaron-code:
Thanks @pkj415 — agreed on the direction (move need_defer_read_point_ out of CalculateIsolation and into SetupPerformOptions, gated on isolation level). Happy to park this PR while your UpdateReadTime change lands.

When you come back with the suggestions, could you also clarify whether you're planning to roll the SetupPerformOptions change into your work, or if you'd prefer I pick it up as a follow-up? Happy either way.

pkj415:
> Thanks @pkj415 — agreed on the direction (move need_defer_read_point_ out of CalculateIsolation and into SetupPerformOptions, gated on isolation level). Happy to park this PR while your UpdateReadTime change lands.
> 
> When you come back with the suggestions, could you also clarify whether you're planning to roll the SetupPerformOptions change into your work, or if you'd prefer I pick it up as a follow-up? Happy either way.

Hi @ellabaron-code,

The changes that resolve this have landed on master:
1. To skip the logic in `UpdateReadTime` for object locking: aa3dd27d727575e694e5ce48b43d0ca4eb59e681
2. To move need_defer_read_point_ out of `CalculateIsolation`: 62cc8330aa80dade03f1aad02c2a45e15af5a51b

Together, these fix the issue (tested it locally).

pao214:
Closing the PR since the issue is fixed for the above reasons. Feel free to open a new one if the issue still persists.

PR #31545, opened 2026-05-11
[DOC-1309]TA: Warning gap for index rewrites during ALTER TABLE .. ALTER COLUMN TYPE

aishwarya24:
@netlify /stable/releases/techadvisories/ta-24007/


docs/content/stable/releases/techadvisories/ta-24007.md
```diff
@@ -0,0 +1,63 @@
...
+
+A warning similar to the following is raised to give you notice of this issue:
+
+```text
+NOTICE:  table rewrite may lead to inconsistencies
+DETAIL:  Concurrent DMLs may not be reflected in the new table.
+HINT:  See https://github.com/yugabyte/yugabyte-db/issues/19860. Set 'ysql_suppress_unsafe_alter_notice' yb-tserver gflag to true to suppress this notice.
+```
+
+What was not accounted for in the original warning logic is that index rewrites are not always coupled to table rewrites. Some scenarios can still cause an index rewrite even when the base table does not require one, including partial or expression indexes (for example, `CREATE INDEX tbl_col ON tbl(lower(col))`) and collation changes. This behavior is expected in both YugabyteDB and PostgreSQL. Because the warning was only triggered by table rewrites, these index-only rewrite cases went unwarned. The fix ensures the warning is raised whenever a rewrite of any kind—table or index—is triggered by an `ALTER TABLE` operation.
```
ddhodge:
```suggestion
What was not accounted for in the original warning logic is that index rewrites are not always coupled to table rewrites. Some scenarios can still cause an index rewrite even when the base table does not require one, including partial or expression indexes (for example, `CREATE INDEX tbl_col ON tbl(lower(col))`) and collation changes. This behavior is expected in both YugabyteDB and PostgreSQL. Because the warning was only triggered by table rewrites, these index-only rewrite cases went unwarned. The fix ensures the warning is raised whenever a rewrite of any kind (table or index) is triggered by an `ALTER TABLE` operation.
```

docs/content/stable/releases/techadvisories/ta-24007.md
```diff
@@ -0,0 +1,63 @@
...
+
+Furthermore, there were scenarios where YugabyteDB rewrote secondary indexes when PostgreSQL would not: specifically, `ALTER TABLE .. ALTER COLUMN TYPE` was rewriting secondary indexes that included the altered column in all cases, even when the base table did not require a rewrite. That behavior has also been corrected.
```
ddhodge:
```suggestion
Furthermore, there were scenarios where YugabyteDB rewrote secondary indexes when PostgreSQL would not. Specifically, `ALTER TABLE .. ALTER COLUMN TYPE` was rewriting secondary indexes that included the altered column in all cases, even when the base table did not require a rewrite. That behavior has also been corrected.
```

ddhodge (approved):

PR #31630, opened 2026-05-14
[#31629] gutil: Abolish SpinLockHolder macro in favor of [[nodiscard]]

### LINKED ISSUE #31629: Abolish SpinLockHolder macro in favor of [[nodiscard]]
Jira Link: [DB-21400](https://yugabyte.atlassian.net/browse/DB-21400)
The `src/yb/gutil/spinlock.h` header defines a macro `SpinLockHolder` to catch cases where the variable name is omitted (e.g., `SpinLockHolder(&lock);` instead of `SpinLockHolder l(&lock);`):

```cpp
// Catch bug where variable name is omitted, e.g. SpinLockHolder (&lock);
#define SpinLockHolder(x) COMPILE_ASSERT(0, spin_lock_decl_missing_var_name)
```

However, this macro pollutes the global namespace and causes compilation issues when integrating with third-party libraries. For example, including `<absl/container/flat_hash_map.h>` after `spinlock.h` results in compilation errors because Abseil defines its own `absl::base_internal::SpinLockHolder` class, and our macro incorrectly expands it.

Instead of a macro, we should use the `[[nodiscard]]` attribute on the `SpinLockHolder` class itself. This is the standard C++ way to prevent ignoring objects (which catches temporary objects that are immediately destroyed) and is exactly how standard library types like `std::lock_guard` handle this.

```cpp
class [[nodiscard]] SCOPED_CAPABILITY SpinLockHolder {
    // ...
};
```

This will trigger a compiler warning/error (since we build with `-Werror`) if someone writes `SpinLockHolder(&lock);` or `SpinLockHolder(&this->lock);` without polluting the global namespace.

[DB-21400]: https://yugabyte.atlassian.net/browse/DB-21400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

hkimura:
Resolves #31629. Replaces the `SpinLockHolder` macro with the `[[nodiscard]]` attribute to prevent polluting the global namespace while retaining compilation safety against discarding temporaries.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/yb/gutil/spinlock.h
+++ b/src/yb/gutil/spinlock.h
@@ -157,14 +157,12 @@ class SCOPED_CAPABILITY SpinLockHolder {
  private:
   SpinLock* lock_;
  public:
-  inline explicit SpinLockHolder(SpinLock* l) ACQUIRE(l)
+  [[nodiscard]] inline explicit SpinLockHolder(SpinLock* l) ACQUIRE(l)
       : lock_(l) {
     l->Lock();
   }
 
   inline ~SpinLockHolder() RELEASE() { lock_->Unlock(); }
 };
-// Catch bug where variable name is omitted, e.g. SpinLockHolder (&lock);
-#define SpinLockHolder(x) COMPILE_ASSERT(0, spin_lock_decl_missing_var_name)
 
 } // namespace base


src/yb/gutil/spinlock.h
```diff
@@ -157,14 +157,12 @@ class SCOPED_CAPABILITY SpinLockHolder {
  private:
   SpinLock* lock_;
  public:
-  inline explicit SpinLockHolder(SpinLock* l) ACQUIRE(l)
+  [[nodiscard]] inline explicit SpinLockHolder(SpinLock* l) ACQUIRE(l)
```
es1024, line 160: Could also just set [[nodiscard]] on the class (I'm guessing that's what you were planning to do given the lint change?)

es1024 (approved):

src/yb/gutil/spinlock.h
hkimura: Yes, that was the reason, but this also works, and anyways it's the only function that might give a temporary. No behavioral difference from on-class `nodiscard`.

Issue #31656, opened 2026-05-15
[DocDB] BSON-aware primary key comparator for DocumentDB collections


hari90:
Jira Link: [DB-21410](https://yugabyte.atlassian.net/browse/DB-21410)
## Description

DocDB uses RocksDB's default `BytewiseComparator` for all tablet primary key comparisons. For DocumentDB collection tablets, the primary key contains a BSON `object_id` column. BSON integers are stored little-endian, so negative values sort incorrectly under byte-wise comparison (e.g., `-1 = 0xFFFFFFFF` sorts after `1 = 0x01000000`). This causes `$gt`/`$lt` range queries on `_id` with negative values to return wrong results.

Additionally, the signed-integer ordering issue extends to cross-numeric-type comparisons (int32 / int64 / double sharing the same BSON canonical sort group) where binary representations differ.

This issue tracks adding a BSON-aware custom RocksDB key comparator for DocumentDB collection tablets so that primary key range queries return correct results.

**Issue Type:** kind/bug

I confirm this issue does not contain any sensitive information.

[DB-21410]: https://yugabyte.atlassian.net/browse/DB-21410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31657, opened 2026-05-15
[#31656] DocDB: Add BSON-aware primary key comparator for DocumentDB

hari90:
## Summary

DocDB uses RocksDB's default `BytewiseComparator` for all tablet primary key comparisons. For DocumentDB collection tablets the primary key contains a BSON `object_id` column. BSON integers are little-endian, so negative values sort incorrectly under byte-wise memcmp (e.g. `-1 = 0xFF…` byte-sorts after `1 = 0x01…`), causing `$gt`/`$lt` filters on `_id` with negative values to return wrong results. The same applies to cross-type numeric comparisons (int32 / int64 / double all belong to one BSON sort-order group but have different binary representations).

**Changes:**

- **`src/yb/dockv/doc_bson.{h,cc}`** — `CompareBson(Slice, Slice)`: walks two BSON documents element-by-element via libbson iterators, applies MongoDB canonical sort-order grouping (`GetSortOrderType`), and compares values with cross-type numeric promotion (int32/int64/double converted to `long double`). MinKey/MaxKey handled as the cross-type boundary exceptions per MongoDB spec.

- **`src/yb/dockv/docdb_key_comparator.{h,cc}`** — `DocDBKeyComparator`: a `rocksdb::Comparator` that finds the first differing byte in two DocDB keys, scans the key structure to determine whether that byte falls inside a `kBson`/`kBsonDescending` region, and if so decodes and compares the BSON values via `CompareBson`; otherwise falls back to byte-wise, preserving compatibility with all non-BSON key components.

- **`src/yb/common/schema.{h,cc}`** — `Schema::HasBsonKeyColumn()`: returns true iff any primary key column is BSON-typed.

- **`src/yb/tablet/tablet.{h,cc}`** — `Tablet::MaybeUseDocDBKeyComparator()`: installs `DocDBKeyComparatorInstance()` on a tablet's RocksDB options iff the schema has a BSON key column. Called from both `InitRocksDBBaseOptions` and `InitRocksDBOptions`. The per-schema gate is necessary because RocksDB checks the comparator name against the MANIFEST on open — tablets created under `BytewiseComparator` (sys.catalog, all non-DocumentDB tables) must keep it; DocumentDB collection tablets get the new comparator from the moment they are created.

- **`src/yb/dockv/docdb_key_comparator-test.cc`** — unit tests for the comparator and the BSON comparison helpers covering negative integers, doubles, cross-type numerics, type ordering, nested documents, MinKey/MaxKey.

- **`src/yb/integration-tests/documentdb/documentdb_test.cc`** — six end-to-end tests run against a live mini-cluster:
  - `PrimaryKeyRangeQuery`: validates that negative `int32` `_id` values sort correctly under `$gt`/`$lt`.
  - `CrossTypePrimaryKeyComparison`: validates cross-numeric-type ordering (int32/int64/double).
  - `BsonPrimaryKeyBinaryComparator`: validates MongoDB type-bracketed `$gt`/`$lt` semantics and the MinKey/MaxKey cross-type exceptions.
  - `BsonObjectIdPrimaryKey`: exercises the ObjectId path (the default MongoDB `_id` type, the only `CompareBsonValues` codepath that delegates to libbson via `bson_oid_compare`) with exact-match and range lookups.
  - `PrimaryKeySortOrder`: asserts the actual *ordering* of documents returned under `sort: {_id: 1}` and `{_id: -1}`, not just match counts.
  - `PrimaryKeyCompactionRoundTrip`: runs range-filter assertions before and after `MiniCluster::FlushTablets` + `CompactTablets`, confirming the comparator produces matching results before and after SST merge.

## Upgrade/Rollback safety

`DocDBKeyComparator` is installed only for tablets whose schema has a BSON-typed primary key column (`Schema::HasBsonKeyColumn()`). The BSON DocDB column type itself is new — DocumentDB collections are the only feature that uses it, and DocumentDB is preview / not yet generally available, so there are no shipped clusters with existing BSON-PK tablets to migrate. Every new BSON-PK tablet records `yb.DocDBKeyComparator.v1` in its MANIFEST from the moment it is created and the same name is provided on reopen.

`sys.catalog` and every existing non-BSON tablet's `HasBsonKeyColumn()` returns false, so they keep the default `BytewiseComparator` and their MANIFESTs are unaffected.

No AutoFlag or gFlag gate is needed because there is no pre-existing on-disk state to be compatible with.

## Test plan

- [x] Unit tests: `./yb_build.sh release --cxx-test docdb_key_comparator-test` — all cases pass (negative ints, doubles, cross-type numerics, type ordering, MinKey/MaxKey, nested documents).
- [x] Integration — `DocumentDBTest.PrimaryKeyRangeQuery`: negative `int32` `_id` values sort correctly under `$gt`/`$lt` (verified with mini-cluster).
- [x] Integration — `DocumentDBTest.CrossTypePrimaryKeyComparison`: int32/int64/double mixed-type `_id` ordering is correct.
- [x] Integration — `DocumentDBTest.BsonPrimaryKeyBinaryComparator`: type-bracketed `$gt`/`$lt` semantics match MongoDB spec; MinKey/MaxKey cross-type assertions pass.
- [x] Integration — `DocumentDBTest.BsonObjectIdPrimaryKey`: ObjectId `_id` exact-match + range filters; exercises the `bson_oid_compare` codepath.
- [x] Integration — `DocumentDBTest.PrimaryKeySortOrder`: returned-document ordering under `sort: {_id: ±1}` matches BSON canonical order.
- [x] Integration — `DocumentDBTest.PrimaryKeyCompactionRoundTrip`: range-filter results survive `FlushTablets` + `CompactTablets`.
- [x] Regression: `DocumentDBTest.SimpleCollection`, `DocumentDBTest.DropCollectionAndDatabase` unaffected — verified by running the full DocumentDB test suite.
- [x] sys.catalog and non-BSON tablets unaffected — `Schema::HasBsonKeyColumn()` returns false for these; `BytewiseComparator` name in their MANIFESTs is preserved and matches on reopen.

### DIFF
--- a/src/yb/dockv/doc_bson.cc
+++ b/src/yb/dockv/doc_bson.cc
@@ -11,12 +11,284 @@
 // under the License.
 //
 
+#include "yb/dockv/doc_bson.h"
+
+#include <bson/bson.h>
+
+#include <cmath>
+
+#include "yb/dockv/doc_kv_util.h"
 #include "yb/util/result.h"
 
-#include "yb/dockv/doc_bson.h"
-#include "yb/dockv/doc_kv_util.h"
-
 namespace yb::dockv {
+
+namespace {
+
+// Returns the canonical sort order for a BSON type, following MongoDB's ordering:
+// MinKey < Null/Undefined < Number < String/Symbol < Document < Array < Binary <
+// ObjectId < Boolean < DateTime < Timestamp < Regex < ... < MaxKey
+//
+// This matches the ordering used by the DocumentDB extension's GetSortOrderType().
+int GetSortOrderType(bson_type_t type) {
+  switch (type) {
+    case BSON_TYPE_EOD:
+    case BSON_TYPE_MINKEY:
+      return 0x0;
+    case BSON_TYPE_UNDEFINED:
+    case BSON_TYPE_NULL:
+      return 0x1;
+    case BSON_TYPE_DOUBLE:
+    case BSON_TYPE_INT32:
+    case BSON_TYPE_INT64:
+    case BSON_TYPE_DECIMAL128:
+      return 0x2;
+    case BSON_TYPE_BOOL:
+      return 0x8;
+    case BSON_TYPE_UTF8:
+    case BSON_TYPE_SYMBOL:
+      return 0x3;
+    case BSON_TYPE_DOCUMENT:
+      return 0x4;
+    case BSON_TYPE_ARRAY:
+      return 0x5;
+    case BSON_TYPE_BINARY:
+      return 0x6;
+    case BSON_TYPE_OID:
+      return 0x7;
+    case BSON_TYPE_DATE_TIME:
+      return 0x9;
+    case BSON_TYPE_TIMESTAMP:
+      return 0xA;
+    case BSON_TYPE_REGEX:
+      return 0xB;
+    case BSON_TYPE_DBPOINTER:
+      return 0xC;
+    case BSON_TYPE_CODE:
+      return 0xD;
+    case BSON_TYPE_CODEWSCOPE:
+      return 0xE;
+    case BSON_TYPE_MAXKEY:
+      return 0xF;
+    default:
+      return 0xFF;
+  }
+}
+
+int CompareSortOrderType(bson_type_t left, bson_type_t right) {
+  return GetSortOrderType(left) - GetSortOrderType(right);
+}
+
+// Converts a bson_value_t number to long double for cross-type comparison.
+long double BsonNumberAsLongDouble(const bson_value_t* value) {
+  switch (value->value_type) {
+    case BSON_TYPE_DOUBLE:
+      return static_cast<long double>(value->value.v_double);
+    case BSON_TYPE_INT32:
+      return static_cast<long double>(value->value.v_int32);
+    case BSON_TYPE_INT64:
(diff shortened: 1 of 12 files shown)


src/yb/dockv/doc_bson.cc
```diff
@@ -11,13 +11,285 @@
...
+      return 0x0;
+    case BSON_TYPE_UNDEFINED:
+    case BSON_TYPE_NULL:
+      return 0x1;
+    case BSON_TYPE_DOUBLE:
+    case BSON_TYPE_INT32:
+    case BSON_TYPE_INT64:
+    case BSON_TYPE_DECIMAL128:
+      return 0x2;
+    case BSON_TYPE_BOOL:
```
hari90: Why is this alone out of order?

src/yb/dockv/doc_bson.cc
```diff
@@ -11,13 +11,285 @@
...
+      return value->value.v_bool ? 1 : 0;
+    default:
+      return 0;
+  }
+}
+
+// Compares two BSON numeric values with cross-type promotion.
+// Follows the same logic as DocumentDB's CompareNumbers.
+int CompareNumbers(const bson_value_t* left, const bson_value_t* right) {
```
hari90: Use ref instead of * when we don't expect input to be nullptr

PR #31668, opened 2026-05-15
Claude Skill: backport-commit improvements (cleanup, Phorge body format, conflict annotation)

hari90:
## Summary

Three related improvements to the `/backport-commit` workflow.

### 1. Clean up local task branches and worktree after success

`backport-commit.py` previously left behind every local `backport-<sha>-<branch>` task branch plus the backport worktree (`$HOME/code/backport-ybdb` by default) after a successful run. Both are dead weight once the PRs are open — the fork holds the canonical task-branch copies that the PRs reference, and the worktree is just the scratch area used to assemble them.

- After the per-branch loop completes successfully, delete each local task branch whose PR was created (or was detected as pre-existing).
- Remove the backport worktree, but only when (a) **this run** created it and (b) it's still a linked worktree of the source repo (i.e. `workspace/.git` is a file pointer, not a real `.git/` directory). This protects a plain checkout someone aimed `YB_BACKPORT` at.
- Skip cleanup on the preview path (`-n`) and on the exit-2 conflict path so the user can still inspect / `cd` in / resume.
- New `--keep-workspace` flag to opt out entirely (debugging).

Implementation notes:
- `setup_workspace()` now returns a `worktree_created` flag.
- Task-branch deletion detaches HEAD first if it points at one of the branches about to be deleted, so `git branch -D` doesn't refuse on "checked out here".
- Failures (e.g. `worktree remove` refused) are warned but don't fail the script — the PRs are already created at that point.

### 2. Handle Phorge-style commit bodies in `amend_message`

For commits landed via Phorge, the body uses `Test Plan:` (Phorge convention) instead of `## Test plan` (GitHub PR convention) and ends with a `Differential Revision: https://phorge.../D<num>` footer. Previously the script only recognized the Markdown heading, so:

- the `Original commit:` footer was appended at the very end of the body (below the Phorge footers) instead of above `Test Plan:`, and
- it used only the SHA with no reference, since no GitHub PR matched the commit.

Now the script matches either heading for the insertion position, and when no GitHub PR is found it parses the `Differential Revision: ...D<num>` footer to produce `Original commit: <SHA> / D<num>` (mirroring the `<SHA> / #<PR>` form used for GitHub-merged commits).

The insertion logic was also collapsed into a single pass per review feedback — the previous `if any(is_test_plan_heading(line) ...)` guard was redundant since the heading is virtually always present, and the parallel else-branch is now just a fallback append-at-end when no heading is found.

### 3. Annotate **every** manual conflict resolution in the PR body (skill update)

`.claude/commands/backport-commit.md` Step 4 previously gated PR-body annotation on "non-trivial" conflicts. That's the wrong gate — "trivial" describes how easy the resolution was, not whether it warrants disclosure to a reviewer. Reviewers shouldn't have to diff the backport against the original commit to discover the agent took a manual step.

New policy: any branch where the agent ran `git cherry-pick --continue` (vs. it applying cleanly) gets an entry in the PR body, with a trivial/non-trivial label and a short summary. Step 2 (substep 3) now asks the agent to track every resolution, not just non-trivial ones, and the Step 4 example shows both styles.

## Test plan

- [x] `python3 -m py_compile .agents/scripts/backport-commit.py` — clean.
- [x] `python3 .agents/scripts/backport-commit.py --help` — shows the new `--keep-workspace` flag.
- [x] Standalone smoke-test of the `amend_message` body-rewriting block against five synthetic inputs (Phorge body, GitHub-PR body, no-heading, empty, and "both heading + Phorge footer with GH PR URL — GH PR wins"). All passed.
- [x] End-to-end: ran a real backport with this version of the script (commit `5fc0505f` → 2026.1, then → 2025.2) — confirmed the local `backport-*` task branches were deleted on success, and the chained-run flow worked through a conflict resolution.

### DIFF
--- a/.agents/scripts/backport-commit.py
+++ b/.agents/scripts/backport-commit.py
@@ -397,21 +397,24 @@ def amend_message(commit: str, branch: str, pr_url: Optional[str]) -> None:
     def is_test_plan_heading(line: str) -> bool:
         return line.startswith("## Test plan") or line.startswith("Test Plan:")
 
-    if any(is_test_plan_heading(line) for line in body_lines):
-        new_body_lines: list[str] = []
-        inserted = False
-        for line in body_lines:
-            if not inserted and is_test_plan_heading(line):
-                new_body_lines.append(original_line)
-                new_body_lines.append("")
-                inserted = True
-            new_body_lines.append(line)
-        new_body = "\n".join(new_body_lines)
-    else:
-        # If body_lines is empty (rare: original commit had no body beyond
-        # the prior footers we just stripped), avoid a leading blank line.
-        prefix = "\n".join(body_lines).rstrip()
-        new_body = (prefix + "\n\n" if prefix else "") + original_line
+    # Single pass: insert above the first test-plan heading. If none is
+    # found (rare -- nearly every commit body has one), fall through to
+    # the append-at-end branch.
+    new_body_lines: list[str] = []
+    inserted = False
+    for line in body_lines:
+        if not inserted and is_test_plan_heading(line):
+            new_body_lines.append(original_line)
+            new_body_lines.append("")
+            inserted = True
+        new_body_lines.append(line)
+    if not inserted:
+        while new_body_lines and not new_body_lines[-1].strip():
+            new_body_lines.pop()
+        if new_body_lines:
+            new_body_lines.append("")
+        new_body_lines.append(original_line)
+    new_body = "\n".join(new_body_lines)
 
     run(["git", "commit", "--amend", "--cleanup=verbatim",
          "-m", new_subject, "-m", new_body])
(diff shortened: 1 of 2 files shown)


kai-franz (approved):
LGTM!

.agents/scripts/backport-commit.py
```diff
@@ -372,15 +383,25 @@ def amend_message(commit: str, branch: str, pr_url: Optional[str]) -> None:
...
+    elif phorge_match:
+        original_line = f"Original commit: {commit} / D{phorge_match.group(1)}"
     else:
         original_line = f"Original commit: {commit}"
-    if any(line.startswith("## Test plan") for line in body_lines):
+
+    def is_test_plan_heading(line: str) -> bool:
+        return line.startswith("## Test plan") or line.startswith("Test Plan:")
+
+    if any(is_test_plan_heading(line) for line in body_lines):
```
hari90: Not needed. Very likely that this is true

Issue #31681, opened 2026-05-16
[Tserver] RequestSequencer::Shutdown does not drain pending_requests_ on RPC path


ellabaron-code:
Jira Link: [DB-21422](https://yugabyte.atlassian.net/browse/DB-21422)
## Description

`RequestSequencer::Shutdown` only signals the cond_var (`cond_.notify_all`) and flips `is_active_`, but does not drain `impl_.pending_requests_`. On macOS, where `pg_client_use_shared_memory` defaults to `false`, PG client requests flow through the regular Perform RPC path rather than the shared-memory path. `PerformQuery::RunImpl` parks the response callback in `pending_requests_` via `Enqueue`; nothing calls `RejectPendingRequests` on that path during shutdown, so parked callbacks are never invoked, the reactor never becomes idle, and `Reactor::Join` blocks for the full timeout.

**Symptom:** Both tests in `PgOpProcessingOrderTest` hang on macOS for the full reactor-join timeout (90 s+):

- `PgOpProcessingOrderTest.ShutdownTimeWithLostOp` — fails the `< 0.8 * pg_client_session_expiration_ms` deadline.
- `PgOpProcessingOrderTest.BackendTerminationTimeWithLostOp` — same hang.

The same tests pass on Linux in ~7 s because the shared-memory path uses a cond_var whose predicate calls `EnsureIsActive`, which has a side effect of invoking `RejectPendingRequests`.

**Steps to reproduce:**

1. Build on macOS arm64.
2. Run `PgOpProcessingOrderTest.ShutdownTimeWithLostOp` or `PgOpProcessingOrderTest.BackendTerminationTimeWithLostOp`.
3. Test hangs until reactor-join timeout.

### Issue Type

kind/failing-test

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-21422]: https://yugabyte.atlassian.net/browse/DB-21422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31682, opened 2026-05-16
[#31681] Tserver: Drain pending requests on RequestSequencer shutdown

ellabaron-code:
## Summary

`PgOpProcessingOrderTest.ShutdownTimeWithLostOp` and `PgOpProcessingOrderTest.BackendTerminationTimeWithLostOp` both hang on macOS for the full reactor-join timeout (90 s+) and fail the `< 0.8 * pg_client_session_expiration_ms` deadline. The same tests pass on Linux in ~7 s.

The divergence comes from the gflag default:

```cpp
DEFINE_RUNTIME_bool(pg_client_use_shared_memory, !yb::kIsMac, ...)
```

On Linux, PG client requests reach the tserver via the shared-memory exchange. A `SharedExchangeRunnable` thread enforces request ordering by calling `RequestSequencer::RegisterForProcessing`, which blocks on `cond_.wait_until` when the request's `serial_no` is ahead of `next_expected_serial_no_`. The `TEST_emulate_op_lost_on_write` flag creates such a gap, so the worker thread parks on the cond_var. `RequestSequencer::Shutdown` wakes it via `cond_.notify_all`; the predicate re-runs `TryRegisterForProcessing`, hits `EnsureIsActive` which observes `is_active_ == false`, and as a side effect calls `impl_.RejectPendingRequests`. The waiter unwinds with `ShutdownInProgress` and the request is responded to.

On macOS the same request flows through the regular Perform RPC instead. `PerformQuery::RunImpl` uses `GetSessionNoWait + Enqueue`, parking the response callback inside `RequestSequencer::impl_.pending_requests_` rather than a cond_var. Nothing wakes the `Enqueue` path on shutdown: `Shutdown` only flips `is_active_` and notifies the (non-existent) cond_var waiter, so the parked callback is never invoked, the `InboundCall` it captures stays alive, the connection's `calls_being_handled_` map stays non-empty, `Connection::Idle()` returns false, and `Reactor::CheckReadyToStop` never lets the reactor thread exit.

Fix `RequestSequencer::Shutdown` to drain `pending_requests_` itself rather than relying on the cond_var predicate's `EnsureIsActive` side effect. `LockablePgClientSession::StartShutdown` now acquires the session mutex around the call so the drain satisfies the same locking invariant the other `RequestSequencer` methods already enforce.

## Test plan

- [x] `PgOpProcessingOrderTest.ShutdownTimeWithLostOp` passes on macOS arm64 (completes well under the `0.8 * pg_client_session_expiration_ms` deadline).
- [x] `PgOpProcessingOrderTest.BackendTerminationTimeWithLostOp` passes on macOS arm64.
- [x] Both tests continue to pass on Linux (shared-memory path unchanged).

### DIFF
--- a/src/yb/tserver/pg_client_service.cc
+++ b/src/yb/tserver/pg_client_service.cc
@@ -284,11 +284,18 @@ class RequestSequencer {
     return Status::OK();
   }
 
-  void Shutdown() {
-    DEBUG_ONLY(DCHECK(DEBUG_IsMutexLocked()));
+  // Stops accepting new requests and wakes any cond_var waiters. Safe to call
+  // without holding mutex_. Must be paired with CompleteShutdown (which drains
+  // parked callbacks under the mutex) so that parked InboundCall references are
+  // released and the reactor can exit.
+  void StartShutdown() {
     is_active_.store(false, std::memory_order_release);
-    impl_.RejectPendingRequests();
     cond_.notify_all();
+  }
+
+  void CompleteShutdown() {
+    DEBUG_ONLY(DCHECK(DEBUG_IsMutexLocked()));
+    impl_.RejectPendingRequests();
   }
 
  private:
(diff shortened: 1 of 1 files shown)


rahulb-yb:
trigger jenkins

hari90:
Jenkins build has been triggered. Results will be available in the CI checks. [CSI](https://csiweb.dev.yugabyte.com/pull/31682/)

---
*JenkinsBot*

rahulb-yb:
trigger jenkins

es1024:
Trigger Jenkins

es1024 (approved):

PR #31702, opened 2026-05-18
[#31701] tserver: Fix kNormal pool self-deadlock in SyncClientMasterRpc

### LINKED ISSUE #31701: [tserver] Self-deadlock in SyncClientMasterRpc when kNormal RPC pool is saturated
Jira Link: [DB-21443](https://yugabyte.atlassian.net/browse/DB-21443)
## Description

When many concurrent inbound RPCs on a tserver each issue a synchronous master RPC via `SyncLeaderMasterRpc` (e.g. `PgClientServiceImpl::GetDatabaseInfo` calling `YBClient::GetNamespaceInfo`), every worker thread in the `kNormal` RPC pool can park on its `Synchronizer` waiting for the master RPC's response callback.

Because `SyncClientMasterRpc` uses the default `kThreadPool` callback mode, the master RPC response callback is queued onto the same `kNormal` pool that is now fully blocked. No worker is free to run the callback that would release the waiters, and the pool self-deadlocks.

Representative stack on every stuck worker:

```
pthread_cond_wait
std::condition_variable::wait
Synchronizer::WaitUntil
YBClient::Data::SyncLeaderMasterRpc<>
YBClient::GetNamespaceInfo
PgClientServiceImpl::Impl::GetDatabaseInfo
PgClientServiceImpl::GetDatabaseInfo
PgClientServiceIf::Handle
ServicePoolImpl::Handle
InboundCall::InboundCallTask::Run
Worker::Execute
```

This reproduces deterministically with `PgSingleTServerTest.ManyYsqlConnections` running under `PgSmallRpcWorkersTest` (`FLAGS_rpc_workers_limit = 32`, 64 concurrent connections) on both Linux and macOS.

The callback only signals a `Synchronizer`, so it is safe and cheap to dispatch on the reactor thread instead.

## Issue Type

kind/bug

[DB-21443]: https://yugabyte.atlassian.net/browse/DB-21443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ellabaron-code:
## Summary

Resolves #31701.

`PgSingleTServerTest.ManyYsqlConnections` (which runs under `PgSmallRpcWorkersTest` with `FLAGS_rpc_workers_limit = 32` and spawns 64 concurrent connections) deadlocks the tserver on both Linux and macOS. All 32 workers in the `kNormal` RPC pool end up stuck:

```
pthread_cond_wait
std::condition_variable::wait
Synchronizer::WaitUntil
YBClient::Data::SyncLeaderMasterRpc<>
YBClient::GetNamespaceInfo
PgClientServiceImpl::Impl::GetDatabaseInfo
PgClientServiceImpl::GetDatabaseInfo
PgClientServiceIf::Handle
ServicePoolImpl::Handle
InboundCall::InboundCallTask::Run
Worker::Execute
```

Each worker is dispatching an inbound `Perform` RPC that synchronously calls into master via `SyncLeaderMasterRpc` and blocks on the `Synchronizer` set by the master RPC's response callback. With the default `kThreadPool` callback mode, that callback is posted back to the same `kNormal` pool that is running the blocked caller. With every worker parked on its own `Synchronizer`, no worker is free to run the callback that would release it. The pool starves itself.

### Fix

Switch `SyncClientMasterRpc` to invoke its callback on the reactor thread (`kReactorThread`). The callback only signals the `Synchronizer`, so it is cheap and reactor-safe, and it no longer competes with the inbound RPC handlers for `kNormal` pool capacity.

## Test plan

- [x] `PgSingleTServerTest.ManyYsqlConnections` under `PgSmallRpcWorkersTest` (`FLAGS_rpc_workers_limit=32`, 64 concurrent connections) — previously deadlocked the tserver on both Linux and macOS, now completes.

### DIFF
--- a/src/yb/client/client-internal.cc
+++ b/src/yb/client/client-internal.cc
@@ -146,7 +146,10 @@ class SyncClientMasterRpc : public internal::ClientMasterRpcBase {
       ClientData* client, CoarseTimePoint deadline, RespClass* resp,
       const std::string& name, const SyncLeaderMasterFunc& func)
       : ClientMasterRpcBase(client, deadline),
-        resp_(resp), name_(name), func_(func) {}
+        resp_(resp), name_(name), func_(func) {
+    this->mutable_retrier()->mutable_controller()->set_invoke_callback_mode(
+      rpc::InvokeCallbackMode::kReactorThread);
+  }
 
   void CallRemoteMethod() override {
     auto master_proxy = this->template master_proxy<ProxyClass>();


src/yb/client/client-internal.cc
```diff
@@ -146,7 +146,10 @@ class SyncClientMasterRpc : public internal::ClientMasterRpcBase {
       ClientData* client, CoarseTimePoint deadline, RespClass* resp,
       const std::string& name, const SyncLeaderMasterFunc& func)
       : ClientMasterRpcBase(client, deadline),
-        resp_(resp), name_(name), func_(func) {}
+        resp_(resp), name_(name), func_(func) {
+    this->mutable_retrier()->mutable_controller()->set_invoke_callback_mode(
+      rpc::InvokeCallbackMode::kReactorThread);
```
spolitov, line 151: Reactor thread should be used only for lightweight callbacks.

spolitov (changes requested):

src/yb/client/client-internal.cc
ellabaron-code: The callback set by `SyncClientMasterRpc::CallRemoteMethod` is `std::bind(&Finished, this, Status::OK())`. On the reactor, the work this triggers is bounded and non-blocking:

- **Logs + `HandleResponse`**: just reads `controller_.status()` and writes it into `new_status`. The auto-retry branch in `HandleResponse` requires `retry_when_busy=kTrue`, which `Finished` doesn't pass — so it's a status read.
- **`ShouldRetry` branch**: dead for `SyncClientMasterRpc` — base class returns `false`, override at `client-internal.cc:347` just delegates. `DelayedRetry` is never invoked from `Finished` for this class.
- **`ResetMasterLeader(Retry::kTrue|kFalse)`** (on leader change / lease / network error / timeout-with-deadline-left): takes `leader_master_lock_` briefly to push a callback into a vector, then submits `DoSetMasterServerProxy` to `threadpool_`. The master fan-out RPC runs on the threadpool, not the reactor.
- **Terminal path**: `rpcs_.Unregister` (locked list erase + `cond_.notify_one`) and `ProcessResponse`. For `SyncClientMasterRpc`, `ProcessResponse` is literally one line — `synchronizer_.StatusCB(status);` — a cond-var notify. The blocked `WaitUntil` caller resumes on its own thread; no work from the caller runs on the reactor.

Nothing here blocks. No synchronous I/O, no waits, no contended lock held while doing work, no user code of unbounded duration.

A couple of points that reinforce this is safe:

1. `Finished` already runs on the reactor today via the retry path: `RpcRetrier::DoRetry` is scheduled via `ScheduleOnReactor`, runs on the reactor, and calls `rpc->Finished(status)` at `rpc.cc:210` and `:239` (or `SendRpc()` which can call `Finished(TimedOut)` inline). All of `Finished`'s error branches — including `ResetMasterLeader` — are already exercised on the reactor in the existing code, independent of `invoke_callback_mode_`.

2. `pg_client.cc:1091` already uses `kReactorThread` for `PerformAsync`'s completion. Its callback does `data->promise.set_value(MakeExchangeResult(...))` — the same "notify the waiter and return" shape as our `synchronizer_.StatusCB`.

3. The reactor already serializes and writes every outbound RPC to its socket (including the ones triggered by `DoRetry` → `SendRpc` → `AsyncRequest`). The cost of `Finished`'s log + status-check + threadpool-submit is strictly less than the socket I/O the reactor is already doing on the retry path.

Without `kReactorThread`, the callback is posted to `CallbackThreadPool(kNormal)`. In `ManyYsqlConnections` (`rpc_workers_limit=32`, 64 concurrent PG connections each issuing a sync `GetNamespaceInfo`), every worker is parked in `SyncLeaderMasterRpc::WaitUntil`, and the completion callback can't fire because it needs a free worker from the same pool. The pool starves itself.

spolitov: The number of reactor threads is pretty limited. And this particular callback much more heavy than other tasks done in reactor threads.

spolitov (changes requested):

Issue #30018, opened 2026-01-20
[YSQL] PgDDLConcurrencyTest.IndexCreation fails in master


spolitov:
Jira Link: [DB-19879](https://yugabyte.atlassian.net/browse/DB-19879)
### Description

PgDDLConcurrencyTest.IndexCreation fails due to unknown reason, since logs are not present in failure report

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-19879]: https://yugabyte.atlassian.net/browse/DB-19879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31766, opened 2026-05-19
[BACKPORT 2026.1][#30018] YSQL: Fix flaky test PgDDLConcurrencyTest.IndexCreation

kai-franz:
Summary:
PgDDLConcurrencyTest.IndexCreation flakes in master (~14% failure rate
across recent builds). Two related issues:

1: Livelock from unbounded retry. After D47116 every DDL bumps the
catalog version, including DROP TABLE and CREATE TABLE. The
RunIndexCreationQueries inner loop was unbounded -- it only exited when
CREATE INDEX itself succeeded. With 4 threads (3 workers + 1 main)
cycling DROP -> CREATE TABLE -> CREATE INDEX as fast as possible, the
probability of CREATE INDEX (which internally waits for backends to
catch up to multiple catalog versions, governed by the 30s
ysql_yb_wait_for_backends_catalog_version_timeout) finishing without
being invalidated by a peer's DDL is very low. This causes the CREATE INDEX
to fail—but the test simply retries until it succeeds. This retry loop livelocked for long
stretches, pushing the test from ~3 min to the 10 min gtest timeout
and producing flaky failures with no useful log content.

2: Missing suppression for the master-side RPC timeout. When the wait
times out at the master RPC layer (governed by
wait_for_ysql_backends_catalog_version_client_master_rpc_timeout_ms,
5s in this test) rather than inside PG, the error message is
"WaitForBackendsCatalogVersion RPC ... timed out" rather than
"waiting for postgres backends to catch up". Only the latter was being
suppressed, so the RPC-layer timeout reached the test as an
unrecognized error and failed it.

Fix:

- Cap the inner retry loop at 10 attempts and treat exhaustion as
  success. The test only verifies that no unexpected errors are
  produced under concurrent DDL; it does not require CREATE INDEX to
  actually succeed.
- Add "WaitForBackendsCatalogVersion RPC" to the suppression list
  alongside "waiting for postgres backends to catch up".

Original commit: 0efde867a5e7fd77b7c48a4c7c88019bcd882b71 / D53345

Test Plan:
```
./yb_build.sh release --cxx-test pg_ddl_concurrency-test \
  --gtest_filter PgDDLConcurrencyTest.IndexCreation -n 50 \
  --test-parallelism 6
```

70/70 iterations pass locally across two stress runs (20 at -p 4, 50
at -p 6); each iteration finishes in 33-60s. Before the fix one run
reproduced the "WaitForBackendsCatalogVersion RPC ... timed out after
10.000s" failure in 34s.

Reviewers: myang

Reviewed By: myang

Subscribers: yql

Differential Revision: https://phorge.dev.yugabyte.com/D53345

---

[CSI](<https://csiweb.dev.yugabyte.com/pull/31766/>)

### DIFF
--- a/src/yb/yql/pgwrapper/pg_ddl_concurrency-test.cc
+++ b/src/yb/yql/pgwrapper/pg_ddl_concurrency-test.cc
@@ -34,8 +34,12 @@ Status SuppressAllowedErrors(const Status& s) {
   // Catalog Version Mismatch: A DDL occurred while processing this query. Try again.
   // The "Try again" will be detected by IsRetryable(s) as true. But in uncommon
   // cases, PG backend will not append this line, for this test we still want to
-  // suppress this error.
-  if (s.message().Contains("waiting for postgres backends to catch up")) {
+  // suppress this error. The wait can time out either inside PG (after
+  // ysql_yb_wait_for_backends_catalog_version_timeout) or at the master RPC layer (after
+  // wait_for_ysql_backends_catalog_version_client_master_rpc_timeout_ms), so allow both
+  // flavors of message.
+  if (s.message().Contains("waiting for postgres backends to catch up") ||
+      s.message().Contains("WaitForBackendsCatalogVersion RPC")) {
     return Status::OK();
   }
   return s;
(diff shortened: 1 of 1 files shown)


kai-franz:
Trigger Jenkins

hari90:
Jenkins build has been triggered. Results will be posted once it completes. [CSI](https://csiweb.dev.yugabyte.com/pull/31766/)

---
*JenkinsBot*

Issue #31771, opened 2026-05-20
[Docs] Update pgvector extension documentation for hnsw.ef_search GUC and USING hnsw alias


hari90:
## Description

Update the pgvector extension documentation (stable) with the following improvements:

- Add documentation for the `hnsw.ef_search` GUC parameter (default: 40, valid range: 1–1000) under a new "Query-time tuning" subsection
- Add a callout that `USING hnsw` is supported for PostgreSQL backwards compatibility and is internally mapped to the `ybhnsw` index access method
- Fix typo `ybhnsq` → `ybhnsw` and `CONCURRENLTY` → `CONCURRENTLY` in the Limitations code example

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

PR #31799, opened 2026-05-20
[#31771] Docs: Document hnsw.ef_search GUC and USING hnsw alias in pgvector extension docs

hari90:
## Summary

Updates the stable pgvector extension documentation:

- Document the `hnsw.ef_search` GUC under a new "Query-time tuning" subsection — default `40`, valid range `1–1000`, with an example `SET` statement.
- Note that `USING hnsw` is also supported for PostgreSQL backwards compatibility and is internally mapped to the `ybhnsw` index access method, with an equivalent example.
- Fix typos `ybhnsq` → `ybhnsw` and `CONCURRENLTY` → `CONCURRENTLY` in the Limitations example.

## Test plan

- [ ] No runtime behavior change; docs-only edit. Visually inspect the rendered page on docs preview to confirm the new "Query-time tuning" table, the `USING hnsw` backwards-compatibility callout, and the corrected example in Limitations all render correctly.

### DIFF
--- a/docs/content/stable/additional-features/pg-extensions/extension-pgvector.md
+++ b/docs/content/stable/additional-features/pg-extensions/extension-pgvector.md
@@ -180,6 +180,12 @@ To use the L2 distance function:
 CREATE INDEX NONCONCURRENTLY ON items USING ybhnsw (embedding vector_l2_ops);
 ```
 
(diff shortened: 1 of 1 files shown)


aishwarya24 (approved):
LGTM

PR #31808, opened 2026-05-21
[doc] Tweak netlify build command that wasn't working

ddhodge:


aishwarya24 (approved):
LGTM.. Lets see what happens now

Issue #31813, opened 2026-05-21
[CDCSDK] Fix TestHiddenTabletDeletionWithUnusedSlot flake on macOS


ellabaron-code:
Jira Link: [DB-21513](https://yugabyte.atlassian.net/browse/DB-21513)
## Description

`TestHiddenTabletDeletionWithUnusedSlot` (in `CDCSDKConsumptionConsistentChangesTest`) flakes on macOS.

Setting `FLAGS_cdc_intent_retention_ms = 0` causes tservers to classify CDCSDK streams as expired and invoke table-level cdc_state cleanup (`CleanupExpiredTables -> RemoveTablesFromCDCSDKStream`), which wipes the child tablet rows from `cdc_state`. The test's post-WaitFor assertions on the children's inherited checkpoints then observe default `OpId::Max()` rows and fail.

On macOS the tserver cleanup path races ahead consistently, making the test unreliable.

## Source connector version
N/A (internal test fix)

## Connector configuration
N/A (internal test fix)

## YugabyteDB version
N/A

## Issue Type
kind/failing-test

- [x] I confirm this issue does not contain any sensitive information.

[DB-21513]: https://yugabyte.atlassian.net/browse/DB-21513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31814, opened 2026-05-21
[#31813] CDC: Fix TestHiddenTabletDeletionWithUnusedSlot flake on macOS

ellabaron-code:
## Summary

`TestHiddenTabletDeletionWithUnusedSlot` flakes on macOS. Setting `FLAGS_cdc_intent_retention_ms = 0` triggers two concurrent cleanup paths: the master's parent-only deletion task and the tservers' table-level cdc_state cleanup (`CleanupExpiredTables -> RemoveTablesFromCDCSDKStream`). The test depends on the children's inherited checkpoints surviving so it can assert against them after `WaitFor`. On macOS the tserver path wins the race and wipes the children rows, so the post-`WaitFor` reads return default `OpId::Max()` rows and the assertions fail.

Set `FLAGS_TEST_cdcsdk_skip_updating_cdc_state_entries_on_table_removal` before zeroing the retention flag to suppress the table-level cleanup, leaving only the parent's `cdc_state` row to be deleted.

## Test plan

- [x] `ctest -R CDCSDKConsumptionConsistentChangesTest.TestHiddenTabletDeletionWithUnusedSlot` passes repeatedly on macOS
- [x] Same test still passes on Linux; behavior unchanged on that platform

### DIFF
--- a/src/yb/integration-tests/cdcsdk_consumption_consistent_changes-test.cc
+++ b/src/yb/integration-tests/cdcsdk_consumption_consistent_changes-test.cc
@@ -4251,6 +4251,13 @@ TEST_F(CDCSDKConsumptionConsistentChangesTest, TestHiddenTabletDeletionWithUnuse
       master::IncludeInactive::kTrue));
   ASSERT_EQ(get_tablets_res.size(), 3);
 
+  // Setting cdc_intent_retention_ms = 0 (below) also makes tservers classify these streams as
+  // expired and ask the master to wipe all cdc_state rows for this table, including the
+  // children rows the test asserts on after WaitFor. Disable that table-level cleanup so only
+  // the hidden parent tablet's cdc_state entry is removed, leaving the children's inherited
+  // checkpoints intact.
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_cdcsdk_skip_updating_cdc_state_entries_on_table_removal) = true;
+
   ANNOTATE_UNPROTECTED_WRITE(FLAGS_cdc_intent_retention_ms) = 0;
 
   // Now, since cdc_wal_retention_time_secs has been set to zero, we will delete the hidden parent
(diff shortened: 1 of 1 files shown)


rahulb-yb:
trigger jenkins

hari90:
Jenkins build has been triggered. Results will be available in the CI checks. [CSI](https://csiweb.dev.yugabyte.com/pull/31814/)

---
*JenkinsBot*

src/yb/integration-tests/cdcsdk_consumption_consistent_changes-test.cc
```diff
@@ -4234,6 +4234,14 @@ TEST_F(CDCSDKConsumptionConsistentChangesTest, TestHiddenTabletDeletionWithUnuse
       master::IncludeInactive::kTrue));
   ASSERT_EQ(get_tablets_res.size(), 3);
 
+  // Setting cdc_intent_retention_ms = 0 (below) also makes tservers classify these streams as
+  // expired and ask the master to wipe all cdc_state rows for this table, including the
+  // children rows the test asserts on after WaitFor. Disable that table-level cleanup so only
+  // the hidden parent tablet's cdc_state entry is removed, leaving the children's inherited
+  // checkpoints intact.
+  ANNOTATE_UNPROTECTED_WRITE(
+      FLAGS_TEST_cdcsdk_skip_updating_cdc_state_entries_on_table_removal) = true;
```
Sumukh-Phalgaonkar: This will lead to incomplete execution of the `RemoveTablesFromCDCSDKStream()` workflow. To disable this completely set the flag `FLAGS_cdcsdk_enable_cleanup_of_expired_table_entries` to false instead.

Sumukh-Phalgaonkar (changes requested):

src/yb/integration-tests/cdcsdk_consumption_consistent_changes-test.cc
ellabaron-code: Fixed.

rahulb-yb:
trigger jenkins

Sumukh-Phalgaonkar (approved):

PR #31818, opened 2026-05-21
[#31046] DocDB: Fix CheckTserverTabletHealth FATAL when consensus is unavailable

### LINKED ISSUE #31046: [DocDB]  Check failed: value fatal in CheckTserverTabletHealth
Jira Link: [DB-20933](https://yugabyte.atlassian.net/browse/DB-20933)
### Description

Was seen in one off run on 2025.2.3.0-b68

Stress run link in JIRA

```
F20260323 22:16:47 ../../src/yb/util/status.cc:657] Check failed: value 
    @     0x55b466f12f50  google::LogMessage::SendToLog()
    @     0x55b466f13f1b  google::LogMessage::Flush()
    @     0x55b466f14689  google::LogMessageFatal::~LogMessageFatal()
    @     0x55b4689a05cf  yb::tserver::TabletServiceImpl::CheckTserverTabletHealth()
    @     0x55b468b7ee9c  std::__1::__function::__func<>::operator()()
    @     0x55b468b90a3f  yb::tserver::TabletServerServiceIf::Handle()
    @     0x55b4684f993d  yb::rpc::ServicePoolImpl::Handle()
    @     0x55b4684095cf  yb::rpc::InboundCall::InboundCallTask::Run()
    @     0x55b468e97e19  yb::(anonymous namespace)::Worker::Execute()
    @     0x55b468e93394  yb::Thread::SuperviseThread()
    @     0x7f939168b2ea  start_thread
    @     0x7f93917103c0  __clone3
```

**AI Summary on RCA:**

```
FATAL: Check failed: value at status.cc:657 in TabletServiceImpl::CheckTserverTabletHealth()

Root cause: GetConsensusOrRespond() (3-param overload, tablet_service.cc:392-401) unconditionally calls result.get() even when the Result holds an error. This triggers StatusCheck(false) → CHECK(false) → FATAL.

// BUGGY (line 392-401):
auto result = GetConsensus(tablet_peer);
if (!result.ok()) {
  SetupErrorAndRespond(resp->mutable_error(), result.status(), context);
  // Missing: return nullptr;
}
return result.get();  // FATALs when !result.ok()

The sibling 4-param overload at line 403 handles this correctly with return false on error.
Trigger: CheckTserverTabletHealth RPC iterates over tablets. When a tablet's consensus is unavailable (e.g., during node restart), GetConsensus() fails and the missing early return causes the crash.
Introduced: 8f0a8c2282 (Dec 2021, #10479 Multi-RAFT). Latent until 3fca3b580d (#17562, AreNodesSafeToTakeDown API) added CheckTserverTabletHealth as a caller in Aug 2023.
```

### Issue Type

kind/bug

[DB-20933]: https://yugabyte.atlassian.net/browse/DB-20933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

shamanthchandra-yb:
## Summary

Fixes #31046.

`TabletServiceImpl::CheckTserverTabletHealth` had two related defects on the path where consensus is unavailable for a tablet:

1. **`GetConsensusOrRespond` 3-arg overload FATALed.** The helper called `SetupErrorAndRespond` on a non-OK `Result` and then fell through to `return result.get();` — `Result::get()` on a non-OK Result invokes `StatusCheck(false)` (the `CHECK(value)` at `src/yb/util/status.cc`), crashing the tserver with `F status.cc Check failed: value` inside `TabletServiceImpl::CheckTserverTabletHealth`.
2. **Double-respond from the loop.** Even after the helper bailed out cleanly, `CheckTserverTabletHealth`'s per-tablet loop did `if (!consensus) continue;`. `GetConsensusOrRespond` already consumed the `RpcContext` via `SetupErrorAndRespond`, so the trailing `context.RespondSuccess()` fired a `CHECK` in `InboundCall::RecordHandlingCompleted`.

Production hits this on a race between the `RUNNING`-state check in `LookupTabletPeer` and the consensus lookup — typically when YBA's `CheckTserverTabletHealth` probe lands on a peer mid-restart during a rolling gflag update.

**Fix:**

- Add early `return nullptr;` to the 3-arg `GetConsensusOrRespond` after `SetupErrorAndRespond`, matching the 4-arg overload's early-return pattern. This makes the helper safe for any future caller.
- Inside `CheckTserverTabletHealth`, call `GetConsensus` directly (returns `Result`, no `RpcContext` side-effect) so the per-tablet skip leaves the context intact for the trailing `RespondSuccess()`.

## Test plan

Deterministic regression test added: `tablet_server-test.cc :: TabletServerTest.CheckTserverTabletHealthDoesNotCrashWhenConsensusUnavailable`.

To deterministically reproduce the race in a unit test, a new TEST flag `FLAGS_TEST_tablet_peer_force_get_raft_consensus_failure` is added to `TabletPeer::GetRaftConsensus`. When set, it returns `IllegalState` immediately, so `LookupTabletPeer` still sees the peer as `RUNNING` but the consensus lookup fails — the exact production race condition.

Verified locally on macOS arm64 debug:
- [x] **Without the production fix** in `tablet_service.cc` → tserver FATALs (`status.cc Check failed: value`) in `TabletServiceImpl::CheckTserverTabletHealth`; child process killed by signal 6; ctest reports `1 FAILED TEST`.
- [x] **With the fix** → `[  PASSED  ] 1 test.` (~230 ms).

### DIFF
--- a/src/yb/tserver/tablet_server-test.cc
+++ b/src/yb/tserver/tablet_server-test.cc
@@ -67,7 +67,6 @@
 #include "yb/util/curl_util.h"
 #include "yb/util/metrics.h"
 #include "yb/util/monotime.h"
-#include "yb/util/scope_exit.h"
 #include "yb/util/size_literals.h"
 #include "yb/util/status_log.h"
 
@@ -730,10 +729,8 @@ TEST_F(TabletServerTest, TestDeleteTablet_TabletNotCreated) {
 // the peer as RUNNING, but GetRaftConsensus returns IllegalState.
 // The RPC must complete without crashing the process.
 TEST_F(TabletServerTest, CheckTserverTabletHealthDoesNotCrashWhenConsensusUnavailable) {
+  // YBTest::flag_saver_ restores all gflags on fixture teardown, so no manual reset needed.
   ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_tablet_peer_force_get_raft_consensus_failure) = true;
-  auto reset = ScopeExit([] {
-    ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_tablet_peer_force_get_raft_consensus_failure) = false;
-  });
 
   CheckTserverTabletHealthRequestPB req;
   req.add_tablet_ids(kTabletId);
--- a/src/yb/tserver/tablet_server-test.cc
+++ b/src/yb/tserver/tablet_server-test.cc
@@ -67,6 +67,7 @@
 #include "yb/util/curl_util.h"
 #include "yb/util/metrics.h"
 #include "yb/util/monotime.h"
+#include "yb/util/scope_exit.h"
 #include "yb/util/size_literals.h"
 #include "yb/util/status_log.h"
 
@@ -89,6 +90,7 @@ DECLARE_string(block_manager);
 DECLARE_string(rpc_bind_addresses);
 DECLARE_bool(disable_clock_sync_error);
 DECLARE_string(metric_node_name);
+DECLARE_bool(TEST_tablet_peer_force_get_raft_consensus_failure);
 
 // Declare these metrics prototypes for simpler unit testing of their behavior.
 METRIC_DECLARE_counter(rows_inserted);
@@ -717,22 +719,28 @@ TEST_F(TabletServerTest, TestDeleteTablet_TabletNotCreated) {
   ASSERT_TRUE(s.IsNotFound()) << s.ToString();
 }
 
-// Regression test for #31046. The 3-arg GetConsensusOrRespond helper used to
-// fall through and call Result::get() on a non-OK Result when GetRaftConsensus
-// failed, FATALing the tserver with "Check failed: value" at status.cc:657
-// inside CheckTserverTabletHealth. Starting tablet-peer shutdown is enough to
-// make GetRaftConsensus return IllegalState; the RPC must now respond with an
-// error instead of crashing the process.
+// Regression test for #31046. CheckTserverTabletHealth used to call
+// GetConsensusOrRespond inside its per-tablet loop, which (a) FATALed via
+// Result::get() on a non-OK Result before the 3-arg overload was fixed,
+// and (b) consumed the RpcContext on the first failing tablet, leading
+// the trailing context.RespondSuccess() to double-respond.
+// Production hits this when the RUNNING-state check in LookupTabletPeer
+// races with consensus transitioning to non-OK. We use a TEST flag to
+// force that condition deterministically: LookupTabletPeer still sees
+// the peer as RUNNING, but GetRaftConsensus returns IllegalState.
+// The RPC must complete without crashing the process.
 TEST_F(TabletServerTest, CheckTserverTabletHealthDoesNotCrashWhenConsensusUnavailable) {
-  ASSERT_TRUE(tablet_peer_->StartShutdown());
-  ASSERT_TRUE(tablet_peer_->IsShutdownStarted());
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_tablet_peer_force_get_raft_consensus_failure) = true;
+  auto reset = ScopeExit([] {
+    ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_tablet_peer_force_get_raft_consensus_failure) = false;
+  });
 
   CheckTserverTabletHealthRequestPB req;
   req.add_tablet_ids(kTabletId);
   CheckTserverTabletHealthResponsePB resp;
   RpcController controller;
   ASSERT_OK(proxy_->CheckTserverTabletHealth(req, &resp, &controller));
-  ASSERT_TRUE(resp.has_error());
+  EXPECT_FALSE(resp.has_error());
   EXPECT_EQ(0, resp.tablet_healths_size());
 }
 
--- a/src/yb/tserver/tablet_server-test.cc
+++ b/src/yb/tserver/tablet_server-test.cc
@@ -717,6 +717,25 @@ TEST_F(TabletServerTest, TestDeleteTablet_TabletNotCreated) {
   ASSERT_TRUE(s.IsNotFound()) << s.ToString();
 }
 
+// Regression test for #31046. The 3-arg GetConsensusOrRespond helper used to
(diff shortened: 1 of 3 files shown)


src/yb/tserver/tablet_service.cc
```diff
@@ -3493,11 +3494,16 @@ void TabletServiceImpl::CheckTserverTabletHealth(const CheckTserverTabletHealthR
       continue;
     }
 
-    auto consensus = GetConsensusOrRespond(res->tablet_peer, resp, &context);
-    if (!consensus) {
-      LOG_WITH_FUNC(INFO) << "Could not find consensus for tablet " << tablet_id;
+    // Skip per-tablet without consuming the RPC context. Using GetConsensusOrRespond here
```
hari90: Comment can be removed

src/yb/tserver/tablet_service.cc
```diff
@@ -3493,11 +3494,16 @@ void TabletServiceImpl::CheckTserverTabletHealth(const CheckTserverTabletHealthR
...
+    // would respond on the first failure and then the trailing context.RespondSuccess()
+    // would double-respond.
```
hari90:
```suggestion
```

hari90:
Trigger Jenkins

hari90:
Jenkins build has been triggered. Results will be available in the CI checks. [CSI](https://csiweb.dev.yugabyte.com/pull/31818/)

---
*JenkinsBot*

src/yb/tserver/tablet_server-test.cc
```diff
@@ -717,6 +719,31 @@ TEST_F(TabletServerTest, TestDeleteTablet_TabletNotCreated) {
...
+// Production hits this when the RUNNING-state check in LookupTabletPeer
+// races with consensus transitioning to non-OK. We use a TEST flag to
+// force that condition deterministically: LookupTabletPeer still sees
+// the peer as RUNNING, but GetRaftConsensus returns IllegalState.
+// The RPC must complete without crashing the process.
+TEST_F(TabletServerTest, CheckTserverTabletHealthDoesNotCrashWhenConsensusUnavailable) {
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_tablet_peer_force_get_raft_consensus_failure) = true;
+  auto reset = ScopeExit([] {
+    ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_tablet_peer_force_get_raft_consensus_failure) = false;
+  });
```
ttyusupov: Please check whether `YBTest::flag_saver_` already handles that.

src/yb/tserver/tablet_service.cc
```diff
@@ -3493,11 +3494,13 @@ void TabletServiceImpl::CheckTserverTabletHealth(const CheckTserverTabletHealthR
       continue;
     }
 
-    auto consensus = GetConsensusOrRespond(res->tablet_peer, resp, &context);
-    if (!consensus) {
-      LOG_WITH_FUNC(INFO) << "Could not find consensus for tablet " << tablet_id;
+    auto consensus_result = GetConsensus(res->tablet_peer);
+    if (!consensus_result.ok()) {
```
ttyusupov, line 3498: Why we don't fill `resp` in that case anymore?

ttyusupov (changes requested):

src/yb/tserver/tablet_server-test.cc
shamanthchandra-yb: Addressed

src/yb/tserver/tablet_service.cc
shamanthchandra-yb: @ttyusupov , not an expert here, but understanding and fixing with AI. Below is making sense to me from what its saying, please correct here if incorrect:

===
`resp.error()` is the whole-RPC error field (single `optional TabletServerErrorPB error = 1;` in `CheckTserverTabletHealthResponsePB, tserver.proto:412`), and the master-side caller `AsyncTserverTabletHealthTask::HandleResponse (async_rpc_tasks.cc:217–227)` treats `resp.has_error()` as whole-RPC failure — it logs and discards the entire response without looking at tablet_healths.                                                                                       
                                                                                                                                                                                                                                              
So if we set `resp.error()` on one tablet failing the consensus check, YBA would also throw away the health info for any other tablets in the same request that probed successfully. Per-tablet skip via continue is strictly better here, and matches the pre-existing behavior of the `LookupTabletPeer`, which also does silent `continue` with no `resp.error()` set.                    

The previous `GetConsensusOrRespond` call did set `resp.error()`, but only as a side effect of it also calling `context.RespondSuccess()` on the same line — which is what produced the double-respond CHECK crash this PR is fixing.

hari90: @ttyusupov same as above with the tablet not found. The contract is that we only return a subset of tablets that actually on on this tserver, and dont error out if some are missing.

hari90 (approved):

hari90:
trigger jenkins

src/yb/tserver/tablet_service.cc
ttyusupov: Before this change, `GetConsensusOrRespond` was called here:
```
std::shared_ptr<consensus::RaftConsensus> GetConsensusOrRespond(const TabletPeerPtr& tablet_peer,
                                                                RespClass* resp,
                                                                rpc::RpcContext* context) {
  auto result = GetConsensus(tablet_peer);
  if (!result.ok()) {
    SetupErrorAndRespond(resp->mutable_error(), result.status(), context);
  }
  return result.get();
}
```

So, in case of error from `GetConsensus(tablet_peer)`, it was setting error to the response. Proposed PR changes this logic. 

If this logic was incorrect and we are fixing it:
1) That should be clearly mentioned in the summary.
2) Can we add unit-test catching issue with the previous logic?

ttyusupov: > The contract is that we only return a subset of tablets that actually on on this tserver, and dont error out if some are missing.

Actually, `GetConsensus(tablet_peer)` returns an error if tablet is present but not running. What is the `CheckTserverTabletHealth` contract for such cases? BTW, where this contract is described?

PR #31834, opened 2026-05-21
[pggate] Suppress FlushBufferedOperations during AbortSubTransaction path

nxtArnav-sharma:
### Description

This change prevents FlushBufferedOperations() from being invoked during subtransaction abort execution paths in YugabyteDB.

During AbortSubTransaction(), the system triggers a snapshot restore sequence that calls PgApiImpl::RestoreReadPoint(). In this path, buffered operations may be flushed to the tserver even though the subtransaction is in the process of being rolled back. These operations are subsequently discarded by RollbackToSubTransaction() via ClearSessionState(), making the flush unnecessary and introducing avoidable RPCs during abort execution.

To address this, a session-local abort-state indicator is introduced in PgApiImpl using a nesting-aware counter. The counter is incremented on entry to AbortSubTransaction() and decremented on exit. When the counter is non-zero, RestoreReadPoint() suppresses calls to FlushBufferedOperations().

The rollback path remains unchanged and continues to be the authoritative mechanism for discarding buffered operations.

**Key changes**

- Introduced subtxn_abort_depth_ in PgApiImpl to track abort nesting depth.
- Added EnterSubTxnAbort(), ExitSubTxnAbort(), and IsSubTxnAbort() APIs.
- Exported corresponding C-API wrappers through ybc_pggate and pg_yb_utils.
- Modified RestoreReadPoint() to skip flushing when abort depth is active.
- Updated AbortSubTransaction() to correctly manage abort depth using a scoped try/catch-safe pattern.
- Added regression tests covering single-level and nested savepoint abort scenarios.

**Correctness**

- No changes to transaction ordering or commit/rollback semantics.
- State is session-local and scoped per backend PgApiImpl instance.
- Counter-based design ensures correct behavior for nested subtransactions.
- Abort-state is guaranteed to be cleared on all execution paths including error handling.

**Testing**

- Verified no flush occurs during subtransaction abort paths.
- Verified rollback correctly clears buffered operations.
- Verified subsequent transactions flush normally after abort completion.
- Verified correctness under nested savepoint scenarios.

**Impact**

- Eliminates unnecessary RPC flushes during abort execution.
- Prevents transient flushes of operations that are ultimately rolled back.
- Isolated change with minimal surface area in pggate and xact layers.


d-uspenskiy:
subtransaction abort already aborts all unflushed operation (see `ClearSessionState()`) for datails.

```
Status PgApiImpl::RollbackToSubTransaction(SubTransactionId id) {
  const auto status = pg_txn_manager_->RollbackToSubTransaction(ClearSessionState(), id);
  VLOG_WITH_FUNC(4) << "id: " << id << ", error: " << status;
  return status;
}
```

nxtArnav-sharma:
Thanks for the clarification.

My understanding was that FlushBufferedOperations() during the RestoreReadPoint() path could still send RPCs for operations that are subsequently discarded by ClearSessionState() during RollbackToSubTransaction().

Is the expectation here that these flushed operations are never externally visible/committed despite the flush, and therefore the behavior is already considered correct?

d-uspenskiy:
1. Code from pull request doesn't compile
2. Newly create unit test `SubtxnAbortBypassFlushSimple` pass even without the fix
3. Even in case problem really exists in some cases it is definitely should be fixed without introducing new stat in `PgApiImpl` class (i.e. `subtxn_abort_depth_` field)

Issue #31845, opened 2026-05-21
[YSQL] InconsistentIndexRead test flaky on macOS: reader exhausts default yb_max_query_layer_retries=60


ellabaron-code:
Jira Link: [DB-21540](https://yugabyte.atlassian.net/browse/DB-21540)
## Description
`PgLibPqTest.InconsistentIndexRead` is flaky on macOS, failing with:

```
Restart read required (yb_max_query_layer_retries set to 60 are exhausted)
```

The test runs reliably on Linux.

### Workload structure
The reader executes `/*+ IndexScan(bank) */ SELECT SUM(balance) FROM bank WHERE id >= 0` in
autocommit (READ COMMITTED) while a writer commits DELETE+INSERT transactions in a tight loop.

In READ COMMITTED, each statement attempt picks a fresh `read_time = now()`, so retries do not
accumulate progress — they only succeed when no committed intent lands in the new uncertainty
window. Under steady-state writer load, the per-fetch retry distribution has a heavy tail and
occasionally exceeds 60 on macOS.

This is a structural property of the workload, not a correctness regression. The correctness
invariant the test checks (`sum == initial_sum` at every visible read time, validating the
secondary-IndexScan fix from #17159) is unaffected.

### Fix
Bump `yb_max_query_layer_retries` to 120 on the reader connection only.

## Issue Type
kind/failing-test

## Warning: Please confirm that this issue does not contain any sensitive information
- [x] I confirm this issue does not contain any sensitive information.

[DB-21540]: https://yugabyte.atlassian.net/browse/DB-21540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31846, opened 2026-05-21
[#31845] YSQL: Avoid excessive read restarts in InconsistentIndexRead test

ellabaron-code:
## Summary

PgLibPqTest.InconsistentIndexRead is flaky on macOS, failing with:

```
Restart read required (yb_max_query_layer_retries set to 60 are exhausted)
```

The test runs reliably on Linux.

The reader runs /*+ IndexScan(bank) */ SELECT SUM(balance) FROM bank WHERE id >= 0
in autocommit (READ COMMITTED) while a writer commits DELETE+INSERT transactions in
a tight loop. Parallel query is enabled on master, and with the scan being large, a parallel scan is planned.

Parallel scans do not perform efficient read restarts. For a usual scan, a read
restart advances the read point to the accumulated restart_read_ht while
preserving global_limit, which monotonically
shrinks the uncertainty window so restarts converge quickly. For a
parallel scan the global_limit and per-tablet local_limits are not preserved
across the parallel execution and the subsequent query-layer retry, so each retry
re-reads with a full-width uncertainty window. Under steady writer load the retries
do not accumulate progress and blow up, occasionally exhausting
yb_max_query_layer_retries on macOS.

Force the reader onto the non-parallel path with SET yb_parallel_range_rows = 0 so the
test exercises efficient read restarts and stays well under the default retry
budget. The underlying parallel-scan read-restart inefficiency is tracked by https://github.com/yugabyte/yugabyte-db/issues/32009.

## Test plan

- [x] `PgLibPqTest.InconsistentIndexRead` passes repeatedly on macOS (previously flaky)
- [x] `PgLibPqTest.InconsistentIndexRead` still passes on Linux

### DIFF
--- a/src/yb/yql/pgwrapper/pg_libpq-test.cc
+++ b/src/yb/yql/pgwrapper/pg_libpq-test.cc
@@ -5455,7 +5455,7 @@ TEST_F(PgLibPqTest, InconsistentIndexRead) {
   // Assert that the balance is always 10k.
   thread_holder.AddThreadFunctor([this, &stop = thread_holder.stop_flag(), initial_sum]() {
     auto reader_conn = ASSERT_RESULT(Connect());
-    ASSERT_OK(reader_conn.Execute("SET yb_max_query_layer_retries=120"));
+    ASSERT_OK(reader_conn.Execute("SET yb_parallel_range_rows = 0"));
     while (!stop.load(std::memory_order_acquire)) {
       auto sum_balance = ASSERT_RESULT(reader_conn.FetchRow<int64_t>(
           "/*+ IndexScan(bank) */ SELECT SUM(balance) FROM bank WHERE id >= 0"));
--- a/src/yb/yql/pgwrapper/pg_libpq-test.cc
+++ b/src/yb/yql/pgwrapper/pg_libpq-test.cc
@@ -5455,6 +5455,7 @@ TEST_F(PgLibPqTest, InconsistentIndexRead) {
   // Assert that the balance is always 10k.
   thread_holder.AddThreadFunctor([this, &stop = thread_holder.stop_flag(), initial_sum]() {
     auto reader_conn = ASSERT_RESULT(Connect());
+    ASSERT_OK(reader_conn.Execute("SET yb_max_query_layer_retries=120"));
     while (!stop.load(std::memory_order_acquire)) {
       auto sum_balance = ASSERT_RESULT(reader_conn.FetchRow<int64_t>(
           "/*+ IndexScan(bank) */ SELECT SUM(balance) FROM bank WHERE id >= 0"));


pao214 (changes requested):
> In READ COMMITTED, each statement attempt picks a fresh read_time = now(), so retries do not accumulate progress 

This is incorrect. In fact, the query should not require more than a few retries. This is because the global limit is preserved in read restarts and the uncertainty window should reduce for each retry. The local limit should also help establish that there is no more than one retry per tablet.

The behavior described is indicative of a deeper issue. Perhaps, the query is using parallel workers and the global limit is not being preserved.

Regardless of the reason, I prefer fixing the underlying issue instead.

ellabaron-code:
<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The retry budget <code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">yb_max_query_layer_retries</code> defaults to 60, which is already a non-trivial budget — the var is meant to absorb genuine retry tails, not catch bugs. The reader thread runs an autocommit <code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">SELECT</code>, so each query-layer retry opens a fresh implicit transaction with a new read time — attempts are independent against a moving target, not compounding stale views.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">I investigated whether the macOS behavior is macOS-specific or just IPC-mode-specific. On macOS, both <code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">pg_client_use_shared_memory</code> and <code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">enable_object_lock_fastpath</code> default to off (see common_flags.cc); on Linux they default to on. Forcing those flags to their macOS defaults on Linux at startup, I bisected the minimum value of <code style="font-family: monospace; color: rgb(208, 208, 208); background-color: rgb(60, 60, 60); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">yb_max_query_layer_retries</code> needed for the test to pass:</p>

Config | yb_max_query_layer_retries
-- | --
Linux default (shared-mem on) | fails at 3, passes at 5
Linux with macOS IPC defaults forced | fails at 10, passes at 15
macOS native | flaky at 60 (4/5 pass), stable at 120

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Most of the gap comes from non-shared-memory IPC being ~3x more expensive per RPC — each retry covers less wall-clock progress, so the reader needs more attempts to outrun writer commits. The remaining macOS gap on top of that is consistent with macOS running this test at ~1.5x Linux throughput on the same hardware (more writer commits per reader-retry window). The boundary itself is flaky (1 in 5 runs at 60 fails); 120 provides ~2x headroom over the observed flaky floor.</p>

pao214 (changes requested):
> In READ COMMITTED, each statement attempt picks a fresh read_time = now(), so retries do not accumulate progress

This statement is not addressed. Can you set yb_parallel_range_rows = 0? Does the issue still reproduce?

ellabaron-code:
Replacing `SET yb_max_query_layer_retries = 120` with `SET yb_parallel_range_rows = 0` makes this test consistently pass on macOS.  So, if you are ok with it, I will make this change.

pao214:
> Replacing SET yb_max_query_layer_retries = 120 with SET yb_parallel_range_rows = 0 makes this test consistently pass on macOS. So, if you are ok with it, I will make this change.

Please do that instead. Thanks!

ellabaron-code:
The change has been made. Please, approve.

pao214:
Trigger Jenkins

pao214:
@ellabaron-code lets update the title and summary as well

Here's my recommendation

Title: [#31845] YSQL: Avoid excessive read restarts in InconsistentIndexRead test

Summary:

`PgLibPqTest.InconsistentIndexRead` is flaky on macOS, failing with:

```
Restart read required (yb_max_query_layer_retries set to 60 are exhausted)
```

The test runs reliably on Linux.

The reader runs `/*+ IndexScan(bank) */ SELECT SUM(balance) FROM bank WHERE id >= 0`
in autocommit (READ COMMITTED) while a writer commits DELETE+INSERT transactions in
a tight loop. Parallel query is enabled on master, and with the scan being large, a parallel scan is planned.

Parallel scans do not perform efficient read restarts. For a usual scan, a read
restart advances the read point to the accumulated `restart_read_ht` while
preserving `global_limit`, which monotonically
shrinks the uncertainty window so restarts converge quickly. For a
parallel scan the `global_limit` and per-tablet `local_limit`s are not preserved
across the parallel execution and the subsequent query-layer retry, so each retry
re-reads with a full-width uncertainty window. Under steady writer load the retries
do not accumulate progress and blow up, occasionally exhausting
`yb_max_query_layer_retries` on macOS.

Force the reader onto the non-parallel path with `SET yb_parallel_range_rows = 0` so the
test exercises efficient read restarts and stays well under the default retry
budget. The underlying parallel-scan read-restart inefficiency is tracked by #32009.

The test plan can remain as-is.

ellabaron-code:
Changed the title and summary.

pao214 (approved):

pao214:
trigger jenkins

Issue #30578, opened 2026-03-04
[YSQL] Manual ANALYZE should reset mutation counts so that auto analyze can take this into account


iSignal:
Jira Link: [DB-20453](https://yugabyte.atlassian.net/browse/DB-20453)
### Description

Auto analyze mutation counts can be reset by a manual ANALYZE so that auto analyze does not need to get to it right away if a manual ANALYZE is triggered. This allows for a workflow where users could potentially use mutation counts to trigger analyze on their end while setting auto analyze thresholds to high.

Note that analyze can also be done on specific cols, such an analyze will not trigger this behavior, only `analyze table` or a full `analyze`.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-20453]: https://yugabyte.atlassian.net/browse/DB-20453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

craigsoules:
Proposed fix in https://github.com/yugabyte/yugabyte-db/pull/31849

iSignal:
All backported landed, closing @craigsoules

PR #31849, opened 2026-05-21
[#30578] YSQL: Reset auto-analyze mutation counts after a manual ANALYZE

craigsoules:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/31849/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

### Problem

When a user runs a full ANALYZE (or VACUUM ANALYZE) on a YugabyteDB relation, the auto-analyze service's accumulated mutation count for that table is not reset. The next periodic tick can therefore trigger an auto-analyze on a table the user just analyzed, wasting work and surfacing as confusing behavior.

### Change

Add a new ResetTableMutationCountersAfterAnalyze stateful-service RPC and call it from do_analyze_rel after a user-initiated full ANALYZE. The RPC:

1. Clears the per-table entry in the leader's in-memory pg_cluster_level_mutation_counter_.
2. Writes mutations = 0 to the auto-analyze YCQL service table (conditional on the row existing, so we don't insert spurious rows).
3. Evicts the local reltuples cache so the next decision uses fresh stats.

The reset is gated to match PostgreSQL's own resetcounter semantics — it fires only when no column list was given (va_cols == NIL) — and is suppressed for the auto-analyze service's own internal connections (yb_use_internal_auto_analyze_service_conn) and other internal backends, so the service does not re-enter itself.

### Concurrency

The new RPC handler runs on a different thread than the periodic auto-analyze task, so two mutexes are introduced in PgAutoAnalyzeService:

- **mutation_flush_mutex_** — serializes writes to the YCQL service table across FlushMutationsToServiceTable, the reset RPC, and the post-analyze update. Without it, a flush that drained the in-memory counter could apply additive ops *after* a reset-to-zero update.
- **table_cache_mutex_** — guards table_tuple_count_ and table_auto_analyze_settings_, which are now accessed from both the periodic task and the RPC handler. Annotated with GUARDED_BY and accessed via copy-out under the lock so the lock is never held across slow operations.

UpdateTableMutationsAfterAnalyze is also restructured: it now re-reads the YCQL row under mutation_flush_mutex_ and writes max(0, current - analyzed_snapshot) as an absolute value, instead of issuing a bfcall_sub. This prevents a manual-ANALYZE reset that interleaves with an in-progress auto-analyze from leaving the YCQL row negative.

Lock ordering is intentionally non-nested everywhere: both ResetTableMutationCountersAfterAnalyzeImpl and UpdateTableMutationsAfterAnalyze acquire and release each mutex sequentially, so no deadlock is possible.

### Known limitation

Mutations buffered on tservers' local PgMutationCounter *before* the reset can still arrive at the leader *after* it via IncreaseMutationCounters and be counted as post-reset mutations. This is acknowledged inline and is consistent with PostgreSQL's own resetcounter semantics — the alternative (suppressing those mutations) would lose work that legitimately happened during the ANALYZE.

## Upgrade/Rollback safety

**Wire-format change:** two new RPC methods are introduced — PgClientService.ResetTableMutationCountersAfterAnalyze (pg_client.proto) and PgAutoAnalyzeService.ResetTableMutationCountersAfterAnalyze (pg_auto_analyze_service.proto). Both are additive (new methods, new request/response messages; no existing field numbers or messages are altered).

**Mixed-version cluster behavior:**

- **New PG backend → old tserver / old auto-analyze leader:** the unknown-method response surfaces as a non-OK Status. The call site uses HandleYBStatusAtErrorLevel(..., WARNING), so the manual ANALYZE itself succeeds — only the reset is skipped. The observable effect during the upgrade window is that the auto-analyze service may briefly re-trigger ANALYZE for a table the user just analyzed (the same pre-PR behavior). Self-corrects once the upgrade completes.
- **Old PG backend → new tserver / new auto-analyze leader:** old PG never issues the new RPC, so no path is exercised. New behavior simply remains dormant.

**No on-disk format or schema changes.** The auto-analyze YCQL service table layout is unchanged — mutations is still an int64 column. The reset writes mutations = 0 using an existing column shape.

**Rollback:** safe. After downgrade, no callers issue the new RPCs. Any existing YCQL rows touched by the new code remain valid (mutations values are still non-negative integers; the new code in fact tightens this with a floor-at-zero in UpdateTableMutationsAfterAnalyze, so there is no risk of leaving negative counters that the old binary would have to handle).

## Test plan

- [ ] PgMutationCounterTest.ResetClearsOnlySpecifiedTables — unit test for the new PgMutationCounter::Reset primitive (selective eviction).
- [ ] PgAutoAnalyzeTest.ManualAnalyzeResetsMutationCount — covers ANALYZE, ANALYZE(col), VACUUM ANALYZE, and VACUUM ANALYZE(col) on a regular table, asserting reset only happens when no column list is provided.
- [ ] PgAutoAnalyzeTest.InternalAnalyzeDoesNotResetMutationCount — confirms the yb_use_internal_auto_analyze_service_conn gate suppresses the reset.
- [ ] PgAutoAnalyzeTest.ManualAnalyzePartitionedTableResetsPartitionMutationCounts — exercises the partitioned-table path (reset propagates to each leaf partition).
- [ ] Existing PgAutoAnalyzeTest cases continue to pass (mutation accounting, threshold logic, multi-DB triggering).


craigsoules:
Fixes #30578

src/yb/tserver/pg_client_service.cc
```diff
@@ -2960,6 +2961,19 @@ class PgClientServiceImpl::Impl : public SessionProvider {
...
   }
 
+  Status ResetTableMutationCountersAfterAnalyze(
+      const PgResetTableMutationCountersAfterAnalyzeRequestPB& req,
+      PgResetTableMutationCountersAfterAnalyzeResponsePB* resp, rpc::RpcContext* context) {
+    stateful_service::ResetTableMutationCountersAfterAnalyzeRequestPB stateful_service_req;
+    stateful_service_req.add_table_ids(
+        PgObjectId(req.database_oid(), req.table_relfilenode_oid()).GetYbTableId());
+    RETURN_NOT_OK(
+        client::PgAutoAnalyzeServiceClient(client()).ResetTableMutationCountersAfterAnalyze(
```
iSignal: I guess a different way to do this is to just move the code in UpdateTableMutationsAfterAnalyze here? PG object locks do not allow multiple ANALYZE on the same table, so when an analyze completes, the mutations are set correctly. Then auto analyze service does not need to update such mutations either and just trusts its analyze call to have done this work.

src/postgres/src/backend/commands/analyze.c
```diff
@@ -710,6 +713,25 @@ do_analyze_rel(Relation onerel, VacuumParams *params,
 	else if (onerel->rd_rel->relkind == RELKIND_PARTITIONED_TABLE)
 		pgstat_report_analyze(onerel, 0, 0, (va_cols == NIL));
 
+	/*
+	 * YB: yb auto-analyze treats a full manual ANALYZE as making table
+	 * statistics fresh enough that it does not need to immediately re-analyze
+	 * based on mutations collected before this ANALYZE.  This intentionally
```
iSignal: Is there any good reason to prefer the PG method? The current auto analyze method seems slightly more correct.

src/yb/tserver/pg_client_service.cc
iSignal: I guess it is possible that while this analyze runs and completes, PG auto analyze might be in the process of running its own analyze based on what it queried a few secs ago but as long as every analyze collects a mutation count at its start and then subtracts that at the end, things should not be incorrect and we don't need to (a) call to auto analyze service and (b) synchronize between auto analyze's regular runs and such adhoc calls.

src/postgres/src/backend/commands/analyze.c
craigsoules: Not being familiar with all of the potential races, I thought it would be easier to just reset everything.  I figured worst case the stats would be a little off, but given that they are just a hint for when to run the analyze, it didn't seem like a big deal.

Happy to change this to snapshot and subtract if you prefer.

src/yb/tserver/pg_client_service.cc
craigsoules: I assumed I would need to perform the counter resets at the auto-analyze service, which would require an RPC. Is that not the case?

iSignal: The counters are stored in a CQL table so technically anyone can update them directly. It is similar to how PG is doing this (manual analyze directly updates counters that are used by autovacuum). If there is a difficult race condition that requires us to serialize things in the auto analyze service, we can take that path but it seems like the only race is 

```
auto analyze: 

reads mutation counters
decides to trigger analyze on N tables

                                                                                 manual analyze on table t1 updates the counters
auto analyze still goes ahead and triggers an analyze
```

Are there any other interactions that can cause a problem?

craigsoules: My understanding was that the auto-analyze node also keeps an in-memory cache of the mutation counters, which it only periodically flushes to the CQL table.

So if we ran a manual analyze directly following a large number of mutations (e.g., to avoid auto-analyze for that table), those mutation counts might get cached and flushed back only after the analyze completes, triggering an auto-analyze sooner than expected.

The current RPC wipes the in-memory cache to narrow the window to just the 5s tserver window from the total 15s window (5s tserver buffer + 10s auto-analyze buffer).  On the other hand, if we think the 15s window is too narrow for this to be an issue, then we could save an RPC -- wasn't clear to me if that was worth it or not.

Let me know!

src/yb/tserver/stateful_services/pg_auto_analyze_table.cc
```diff
@@ -0,0 +1,142 @@
...
+  auto* operand1 = bfcall_expr_pb->add_operands();
+  auto* operand2 = bfcall_expr_pb->add_operands();
+  operand1->set_column_id(mutations_col_id);
+  operand2->mutable_value()->set_int64_value(snapshot.mutations);
+  return update_op;
+}
+
+}  // namespace
+
+Status ResetPgAutoAnalyzeMutationCounts(
```
iSignal, line 90: Do we need to call back into auto analyze service to do this or can we just run this code in pg client service itself?

iSignal (commented):
Summary of offline conversation with Craig from prior to last commit (adding here for future reference).

Currently auto analyze has two components
1. Aggregation of mutations - tservers cache local mutations and send to auto analyze which then writes them at the start of periodic run to YCQL table. There is max ~15 secs of buffered data that has not yet been written to table at any point. 
2. When auto analyze runs analyze, it records mutations at the start of its periodic run and subtracts them from analyze at the end. This is approximate because (a) we might be analyzing multiple tables, so mutations at start of run != mutations at start of analyze and (b) it doesn't take into account buffered mutations sitting on tservers at that time. So mutations are slightly overcounted.

Craig pointed out that having analyze report to auto analyze service would allow it to reset any buffered mutations sitting on auto analyze service itself so it might make it undercount a bit to avoid the overcount problem a bit.

However, this introduces need for coordination at auto analyze service to deal with concurrency of periodic runs and manual analyze, which can get a bit complex because on new dbs where data is loaded, there is sometimes an initial storm of analyzes. So it might be better to live with current approximations and simply have analyze reset or subtract mutations in the ycql table instead.

src/yb/tserver/stateful_services/pg_auto_analyze_table.cc
craigsoules: This file contains the functions that updates the YCQL table directly.  I just pulled it out so that it could be called both from the auto-analyze path and the manual ANALYZE path.

src/postgres/src/backend/commands/analyze.c
```diff
@@ -710,6 +713,24 @@ do_analyze_rel(Relation onerel, VacuumParams *params,
...
+	/*
+	 * YB: yb auto-analyze treats a full manual ANALYZE as making table
+	 * statistics fresh enough that it does not need to immediately re-analyze
+	 * based on mutations collected before this ANALYZE. Clear only the persisted
+	 * YCQL mutation stats; do not coordinate with auto-analyze service-local
+	 * buffers or caches.
+	 */
+	if (!inh && IsYBRelation(onerel) && va_cols == NIL &&
+		!yb_use_internal_auto_analyze_service_conn &&
+		!yb_is_internal_connection)
```
iSignal: minor: yb_is_internal_connection is not necessary I think?

src/postgres/src/backend/commands/analyze.c
```diff
@@ -1879,7 +1900,9 @@ ind_fetch_func(VacAttrStatsP stats, int rownum, bool *isNull)
 }
 
 
-/*==========================================================================
+/*
+ * YB: yb lint requires an empty opening comment line for this banner.
```
iSignal, line 1905: if we don't edit this line no changes should be required

src/yb/tserver/stateful_services/pg_auto_analyze_service.cc
```diff
@@ -923,7 +910,9 @@ Result<bool> PgAutoAnalyzeService::DoFetchReltuplesAndReloptions(
     }
     ParseReloption(*option_name, *option_value, &settings);
   }
+  table_tuple_count_[table_id] = tuple_count;
```
iSignal, line 913: minor: was there a bug requiring this to be moved or was it a general change?

src/yb/tserver/stateful_services/pg_auto_analyze_table.cc
```diff
@@ -0,0 +1,142 @@
...
+    ops.push_back(std::move(update_op));
+  }
+
+  // TODO(async_flush): https://github.com/yugabyte/yugabyte-db/issues/12173
+  RETURN_NOT_OK_PREPEND(
+      session.TEST_ApplyAndFlush(ops), "Failed to reset mutations in auto analyze table");
+  return Status::OK();
+}
+
+Status SaturatingSubtractPgAutoAnalyzeMutationCounts(
```
iSignal: minor: the name is hard to read. SubtractPgAutoAnalyzeMutationCounts is probably enough?

src/yb/tserver/pg_client.proto
```diff
@@ -153,6 +153,9 @@ service PgClientService {
 
   rpc QueryAutoAnalyze(PgQueryAutoAnalyzeRequestPB) returns (PgQueryAutoAnalyzeResponsePB);
 
+  rpc ResetTableMutationCountersAfterAnalyze(PgResetTableMutationCountersAfterAnalyzeRequestPB)
```
iSignal: minor: ResetAutoAnalyzeMutationCounters may be more accurate

src/postgres/src/backend/commands/analyze.c
```diff
@@ -710,6 +713,24 @@ do_analyze_rel(Relation onerel, VacuumParams *params,
...
+	 */
+	if (!inh && IsYBRelation(onerel) && va_cols == NIL &&
+		!yb_use_internal_auto_analyze_service_conn &&
+		!yb_is_internal_connection)
+	{
+		YbcStatus	reset_status =
+			YBCResetTableMutationCountersAfterAnalyze(YBCGetDatabaseOid(onerel),
+													  YbGetRelfileNodeId(onerel));
+
+		HandleYBStatusAtErrorLevel(reset_status, WARNING);
```
iSignal, line 732: nice, so if this fails we just log a WARNING and move on?

src/yb/yql/pgwrapper/pg_auto_analyze-test.cc
```diff
@@ -477,6 +565,148 @@ TEST_F(PgAutoAnalyzeTest, TriggerAnalyzeSingleTable) {
...
+      [&conn] {
+        ASSERT_OK(conn.Execute(
+            "INSERT INTO manual_analyze_test "
+            "SELECT s, s FROM generate_series(11, 15) AS s"));
+      },
+      {{table_id, 5}}));
+  ASSERT_OK(conn.Execute("ANALYZE manual_analyze_test(v1)"));
+  ASSERT_OK(WaitForTableMutationCount(table_id, 5));
+
+  ASSERT_OK(conn.Execute("VACUUM ANALYZE manual_analyze_test"));
```
iSignal, line 638: minor: do we need vacuum here or do we want to test something specific by using vacuum?

src/postgres/src/backend/commands/analyze.c
```diff
@@ -710,6 +713,24 @@ do_analyze_rel(Relation onerel, VacuumParams *params,
...
 		pgstat_report_analyze(onerel, 0, 0, (va_cols == NIL));
 
+	/*
+	 * YB: yb auto-analyze treats a full manual ANALYZE as making table
+	 * statistics fresh enough that it does not need to immediately re-analyze
+	 * based on mutations collected before this ANALYZE. Clear only the persisted
+	 * YCQL mutation stats; do not coordinate with auto-analyze service-local
+	 * buffers or caches.
+	 */
+	if (!inh && IsYBRelation(onerel) && va_cols == NIL &&
```
iSignal, line 723: There's one problem with resetting here. analyze may not have committed at this point, so a failed analyze would still reset counts. analyze has some complicated logic for committing (see use_own_xacts). For a single statement analyze tbl, I think analyze is actually committed by the PostgresMain commit outside of vacuum.c. 

PG seems to have the same problem with changes_since_analyze though, so it may not a big problem. One option is to move this logic to around https://github.com/yugabyte/yugabyte-db/blob/2c41c738aa53a2871d1a0b75dcc68a0594936b8c/src/postgres/src/backend/commands/vacuum.c#L574 where at least the use_own_xacts = true case has committed by then to minimize the case that analyze fails but prevents auto analyze from running. There is no perfect solution here but in YB analyze may fail if run concurrent with other ddl (until concurrent ddl is available) so that's one reason to bias on this side. No strong opinions though.

iSignal:
trigger jenkins

src/yb/tserver/stateful_services/pg_auto_analyze_service.cc
craigsoules: Before I had added a mutex, so it got moved.  When I removed the mutex, it got left like this because it removed a map lookup on the `VLOG()`.

src/postgres/src/backend/commands/analyze.c
craigsoules: I think that this got changed only because of a linter error :-(

src/yb/yql/pgwrapper/pg_auto_analyze-test.cc
craigsoules: It's one of the entry-points for manual analyze, so figured we should test it.

src/postgres/src/backend/commands/analyze.c
craigsoules: probably not?  it's more for completeness since this is intended only for manual analyze calls.  I can remove if you prefer.

src/yb/tserver/stateful_services/pg_auto_analyze_table.cc
craigsoules: Will change this 👍

src/yb/tserver/pg_client.proto
craigsoules: Will change this 👍

src/postgres/src/backend/commands/analyze.c
craigsoules: Yes, that's my understanding.  Since the analyze is already done, wasn't sure what else to do with an error at this point.

src/postgres/src/backend/commands/analyze.c
craigsoules: Did some digging on this one.  Given it's already a problem in PG today, I'm tempted to just leave it as-is.  Especially since if someone's manual ANALYZE fails, I would expect they'd run it again.

We could do as you say, but it seems like a more complete solution would be to register a post-commit hook to do the reset only after the transaction completes successfully.  Let me know if you want me to pursue that.

iSignal: Post-commit hook etc seems not worth it, we an file a ticket for it for later. If it easy enough to move this to https://github.com/yugabyte/yugabyte-db/blob/2c41c738aa53a2871d1a0b75dcc68a0594936b8c/src/postgres/src/backend/commands/vacuum.c#L574 that avoids some cases like use_own_xacts=true where it is already committed. If it is difficult to access list of tables there we can leave as is in this PR.

src/yb/yql/pgwrapper/pg_auto_analyze-test.cc
iSignal: Got it, the other useful test case might be analyze with no args on a full db, where it seems like our code should do the right thing but might be worth including in one of the tests to protect against future regressions.

iSignal:
Looks good overall. One late request, sorry - can we have a GUC or gflag (whichever is convenient) to disable this behavior in the rare case someone cares?

iSignal:
trigger jenkins

Issue #31345, opened 2026-04-29
[YSQL] Fix test PgLibPqCreateSequenceNamespaceRaceTest.CreateSequenceNamespaceRaceTest


basavaraj29:
Jira Link: [DB-21198](https://yugabyte.atlassian.net/browse/DB-21198)
### Description

The test fails in release mode when table locking and transaction ddl features are turned off. It fails with
```
[ts-3] 2026-04-28 19:51:09.040 UTC [163446] ERROR:  relation "t2" already exists
[ts-3] 2026-04-28 19:51:09.040 UTC [163446] CONTEXT:  Catalog Version Mismatch: A DDL occurred while processing this query. Try again.
[ts-3] 2026-04-28 19:51:09.040 UTC [163446] STATEMENT:  CREATE TABLE t2(k SERIAL, v INT)
../../src/yb/yql/pgwrapper/pg_libpq-test.cc:4820: Failure
Failed
Bad status: Network error (yb/yql/pgwrapper/libpq_utils.cc:674): Execute of 'CREATE TABLE t2(k SERIAL, v INT)' failed: 7, message: ERROR:  relation "t2" already exists
CONTEXT:  Catalog Version Mismatch: A DDL occurred while processing this query. Try again. (pgsql error 42P07) (aux msg ERROR:  relation "t2" already exists
CONTEXT:  Catalog Version Mismatch: A DDL occurred while processing this query. Try again.)
```

Since `CREATE TABLE` now bumps up catalog version, I think it is possible for concurrent creates to run into catalog mismatch errors without table locks & transactional ddl. Unsure how the tests were passing in other modes (where these features were off), may be it was due to some timing/race more probable in release mode.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-21198]: https://yugabyte.atlassian.net/browse/DB-21198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31873, opened 2026-05-23
[BACKPORT 2026.1][#31345] YSQL: Fix PgLibPqCreateSequenceNamespaceRaceTest

kai-franz:
Summary:
After D47116 (#28253), every successful DDL increments the catalog
version. Two concurrent `CREATE TABLE`s now race on the catalog
version, and the catalog-version-mismatch detected at the end of one
backend's statement aborts that backend's PG transaction. When
transactional DDL is disabled
(`--ysql_yb_ddl_transaction_block_enabled=false`), aborting the PG
transaction does not roll back the table that has already been written
to DocDB; instead the master's DDL atomicity infrastructure cleans it
up asynchronously.

That leaves two failure modes that the test was not handling:

1. PG's transparent cache-refresh retry re-issues `CREATE TABLE`,
   which fails with `relation "<t>" already exists` (pgsql error
   42P07). The original retry predicate (`HasTransactionError`,
   matching only 40001/40P01) does not cover this.

2. With `IF NOT EXISTS`, the transparent retry no-ops on the still-
   present pg_class entry, but the master cleanup can land in the
   window between that successful CREATE and the follow-up `INSERT`,
   so the INSERT fails with `relation "<t>" does not exist`
   (pgsql error 42P01).

Both modes were observed in reproducer runs with
`enable_object_locking_for_table_locks=false` and
`ysql_yb_ddl_transaction_block_enabled=false`: every iteration failed
with one of the two errors above.

Wrap the whole create-then-insert sequence in a single retry loop:
- `CREATE TABLE IF NOT EXISTS` is retried on `HasTransactionError` or
  `IsRetryable` (which catches "Catalog Version Mismatch" surfaced when
  the transparent retry is unavailable);
- `INSERT` is retried on the same predicates, plus the specific
  "does not exist" message that indicates the table was cleaned up by
  the DDL atomicity worker after the transparent retry no-op.

Original commit: 63670582fcdd23a7ddcc6c43f469c48a4f1d6f73 / D53337

Test Plan:
Reproduced the failure (10/10 iterations) and validated the fix
(20/20 iterations) in the failing configuration by temporarily adding

```
--enable_object_locking_for_table_locks=false
--ysql_yb_ddl_transaction_block_enabled=false
```

to the test class's `UpdateMiniClusterOptions`, and reverting before
the diff:

```
./yb_build.sh release --cxx-test pg_libpq-test \
  --gtest_filter PgLibPqCreateSequenceNamespaceRaceTest.CreateSequenceNamespaceRaceTest \
  -n 20
```

Also re-ran 10x with the default release flags to confirm no regression.

Reviewers: bkolagani

Reviewed By: bkolagani

Subscribers: yql

Differential Revision: https://phorge.dev.yugabyte.com/D53337

### DIFF
--- a/src/yb/yql/pgwrapper/pg_libpq-test.cc
+++ b/src/yb/yql/pgwrapper/pg_libpq-test.cc
@@ -4866,30 +4866,36 @@ TEST_F(PgLibPqCreateSequenceNamespaceRaceTest, CreateSequenceNamespaceRaceTest) 
   pg_ts = cluster_->tablet_server(2);
   auto conn2 = ASSERT_RESULT(Connect());
   TestThreadHolder thread_holder;
-  thread_holder.AddThreadFunctor([&conn1]() -> void {
-    // Retry on 40001 (serialization failure) which can occur due to
-    // running CREATE TABLE concurrently with the other thread.
-    //
-    // In debug builds, Read Committed is on by default. This results
-    // in an internal DB-side retry of the operation on a conflict error. The retry
-    // can fail with a duplicate table error because the rollback of the previous
-    // transaction is async and might still be in flight.
-    Status s;
-    do {
-      s = conn1.Execute("CREATE TABLE t1(k SERIAL, v INT)");
-    } while (!s.ok() && (HasTransactionError(s) ||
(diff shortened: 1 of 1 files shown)


kai-franz:
Closing in favor of a Phorge backport diff.

PR #31957, opened 2026-05-29
[Docs] Add Source Group for Kapa MCP integration

samiahmedsiddiqui:
Updates the Kapa widget integration to pass `source_group_ids_include` so queries are scoped to the YugabyteDB source group instead of using Global sources.


amiramm (approved):
LGTM

PR #31981, opened 2026-06-01
[#31938] YSQL: MAGE extension does work in backup/restore scenarios

shubin-yb:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/31981/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

sql_dump in binary-upgrade mode emits

    DROP EXTENSION IF EXISTS mage;

before recreating an empty extension shell, so any backup/restore of a database that uses the mage (Apache AGE) extension hits this statement during restore. The mage ProcessUtility hook intercepts every DROP that mentions "mage" via is_age_drop() and unconditionally routes it through drop_age_extension(), which enumerates graphs from mag_catalog.ag_graph. When the extension is not (yet) installed on the restore target, that lookup errors out with

    ERROR:  table "ag_graph" does not exist

and the whole restore aborts, so a backup of a mage-using database was effectively unrestorable.

Fix: guard is_age_drop() with get_extension_oid("mage", true). When the extension is not installed, defer to standard ProcessUtility, which correctly performs the IF EXISTS no-op (or raises the standard "extension does not exist" error for a bare DROP EXTENSION mage). get_extension_oid is preferred over checking the mag_catalog schema: the schema can linger after a prior DROP EXTENSION (so a schema-based guard still ran the AGE drop path and still hit the ag_graph lookup).

Tests:

  * src/postgres/third-party-extensions/mage/regress/{sql,expected}/ yb.orig.basic.{sql,out}: three direct regress cases appended at the end of the existing yb.orig.basic test: - DROP EXTENSION IF EXISTS mage; when not installed -> NOTICE skip
      - DROP EXTENSION mage; when not installed           -> proper
        "extension does not exist" error, not the ag_graph one
      - CREATE EXTENSION mage; then DROP EXTENSION IF EXISTS mage;
        confirms the AGE cleanup path still runs when installed

  * java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java: testMageBackupRestore does an end-to-end YBC backup/restore round trip on a database with mage installed Adds a class-level ysql_yb_enable_mage=true tserver flag; harmless to the other tests in the class, which never CREATE EXTENSION mage.

Verified locally on a 3-node mini-cluster:
  * mage regress (TestPgRegressThirdPartyExtensionsMage): 1/1 pass.
  * TestYbBackup#testMageBackupRestore under YB_TEST_YB_CONTROLLER=1:

### DIFF
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
@@ -118,6 +118,15 @@ public class TestYbBackup extends BasePgSQLTest {
   }
 
   @Override
+  protected Map<String, String> getTServerFlags() {
+    Map<String, String> flagMap = super.getTServerFlags();
+    // Gates `CREATE EXTENSION mage` -- required by testMageBackupRestore and
+    // harmless to the other tests in this class (they do not touch mage).
+    flagMap.put("ysql_yb_enable_mage", "true");
+    return flagMap;
+  }
+
+  @Override
   protected void customizeMiniClusterBuilder(MiniYBClusterBuilder builder) {
     super.customizeMiniClusterBuilder(builder);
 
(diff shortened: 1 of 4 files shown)


java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
```diff
@@ -2193,6 +2202,64 @@ public void testPgHintPlan() throws Exception {
...
+          "SELECT count(*) FROM pg_extension WHERE extname = 'mage'",
+          new Row(0L));
+    }
+
+    // Cleanup.
+    try (Statement stmt = connection.createStatement()) {
+      if (isTestRunningWithConnectionManager())
+        waitForStatsToGetUpdated();
+      stmt.execute("DROP DATABASE yb2");
+    }
```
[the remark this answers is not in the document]
shubin-yb: this is the current pattern for our tests, I believe we do a cleanup in the @After section of each test so this should not pollute other tests

shubin-yb:
trigger jenkins

src/postgres/third-party-extensions/mage/src/backend/catalog/ag_catalog.c
```diff
@@ -147,7 +148,17 @@ static bool is_age_drop(PlannedStmt *pstmt)
             char *str = val->sval;
 
             if (!pg_strcasecmp(str, "mage")) /* YB */
+            {
+                /*
+                 * YB: The AGE-specific drop path (drop_age_extension) enumerates
+                 * graphs from the mag_catalog.ag_graph table. When the mage
+                 * extension is not installed, that lookup errors out.
+                 */
+                if (!OidIsValid(get_extension_oid("mage", true)))
```
karthik-ramanathan-3006:
```suggestion
                if (!OidIsValid(get_extension_oid("mage", true /* missing_ok */ )))
```

style

src/postgres/third-party-extensions/mage/src/backend/catalog/ag_catalog.c
```diff
@@ -35,6 +35,7 @@
 #include "utils/ag_cache.h"
 
 /* YB includes */
+#include "commands/extension.h" /* YB: get_extension_oid() */
```
karthik-ramanathan-3006: nit: You can omit the YB comment if the linter doesn't complain (I don't think it should)

java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
```diff
@@ -2193,6 +2202,64 @@ public void testPgHintPlan() throws Exception {
...
   }
 
+  @Test
+  public void testMageBasicBackupRestore() throws Exception {
+    // End-to-end coverage for backup/restore of a database with the `mage`
+    // (Apache AGE) extension installed.
+    //
+    // The restore generates a ysql_dump in binary-upgrade mode that emits
+    // `DROP EXTENSION IF EXISTS mage;` before recreating an empty extension
+    // shell. Before the is_age_drop() fix in
```
karthik-ramanathan-3006: nit: I think you can omit the "Before the fix ..." part of the comment.

java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
```diff
@@ -2193,6 +2202,64 @@ public void testPgHintPlan() throws Exception {
...
+    // src/postgres/third-party-extensions/mage/src/backend/catalog/ag_catalog.c
+    // that DROP errored with `table "ag_graph" does not exist` and aborted
+    // the entire restore -- so any backup of a database using mage was
+    // effectively unrestorable.
+    //
+    // The test deliberately installs the extension *without* creating a graph.
+    // create_graph() materializes per-graph vertex/edge tables whose snapshot
+    // import currently hits an orthogonal issue at the master metadata stage;
```
karthik-ramanathan-3006, line 2216: Please link to the issue as well.

karthik-ramanathan-3006 (approved):
Overall LGTM.
Wouldn't this also be a problem for upstream Apache AGE?
Would it be worth contributing an upstream fix? Or have they already fixed it?

java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
shubin-yb: just added
DB-21840

shubin-yb:
trigger jenkins

shubin-yb:
> Wouldn't this also be a problem for upstream Apache AGE?
> Would it be worth contributing an upstream fix? Or have they already fixed it?

this is an issue with upstream AGE from what i could find, I can raise a PR against it

Issue #31220, opened 2026-04-21
[YSQL] Add an option to yb_admin to cleanup stale CDC streams


egladysh:
Jira Link: [DB-21091](https://yugabyte.atlassian.net/browse/DB-21091)
### Description

Proposed command line syntax: `yb-admin cleanup_stale_cdc_streams [<namespace>] [--dry_run]`

### Issue Type

kind/new-feature

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-21091]: https://yugabyte.atlassian.net/browse/DB-21091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

egladysh:
historical reference: https://github.com/yugabyte/yugabyte-db/pull/30623

PR #32025, opened 2026-06-03
[#31220] CDC: Add cleanup_stale_cdc_streams yb-admin command

egladysh:
## Summary

Adds a `CleanupStaleCDCStreams` RPC to the `MasterReplication` master service and a corresponding `cleanup_stale_cdc_streams` `yb-admin` command.

The command scans the `cdc_state` table and identifies stale entries — rows whose CDC stream no longer exists, or whose tablet no longer exists. It supports:

- A `dry_run` flag to report stale entries without deleting them.

Example (dry run):

```
./bin/yb-admin \
    cleanup_stale_cdc_streams dry-run

Found 2 stale cdc_state entries (dry run).
  tablet_id: 8f3b9b3a8b2c4e2b9c7d6a5f4e3d2c1b, stream_id: 3d9f1e2a4b6c4d8e9f0123456789abcd, table_id: 000033e8000030008000000000004000, table_name: orders, reason: stream not found
  tablet_id: missing_tablet_id, stream_id: 7a8b9c0d1e2f34567890abcdef123456, reason: tablet not found
```

Implementation notes:
- The `cdc_state` table is scanned once and materialized so both the tablet-collection and classification loops share one consistent snapshot.
- Tablet table metadata is resolved under a single `SharedLock` before the classification loop to avoid repeated lock acquisitions.
- xCluster streams that carry an empty `namespace_id` have their namespace resolved via their first `table_id`.

## Upgrade/Rollback safety

The proto change is purely additive: new messages (`CleanupStaleCDCStreamsRequestPB`, `CleanupStaleCDCStreamsResponsePB`) and a new RPC method (`CleanupStaleCDCStreams`) are added to `MasterReplication`. No existing message fields are modified.

On a mixed-version cluster, an older master that does not have this RPC will return `UNIMPLEMENTED` when `cleanup_stale_cdc_streams` is invoked; all other CDC and xCluster operations are unaffected.

Rolling back removes the `yb-admin` command and the RPC handler. No on-disk state, catalog schema, or gflag defaults are changed.

## Test plan

- [ ] `CDCServiceTest.TestCleanupStaleCDCStreamsWithoutCDCStateTable` — verifies `OBJECT_NOT_FOUND` when the `cdc_state` table does not exist.
- [ ] `CDCServiceTest.TestCleanupStaleCDCStreamsDryRunAndDelete` — dry-run returns stale entries without deleting; live run deletes exactly those entries and leaves the valid entry untouched.
- [ ] `CDCServiceTest.TestCleanupStaleCDCStreamsNamespaceFilter` — with a namespace filter, only entries attributable to the selected namespace are deleted; cross-namespace and unattributable (both stream and tablet missing) entries are preserved.


src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,214 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+    return SetupError(
+        resp->mutable_error(),
+        MasterErrorPB::OBJECT_NOT_FOUND,
+        STATUS(
+            NotFound,
+            "cdc_state table does not exist or is not ready."));
+  }
+
+  Status iteration_status;
+  auto all_entry_keys_result =
```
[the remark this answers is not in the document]
egladysh: Unless I am missing something, the review doesn't seem to be correct. cdc_state_table_ created once in CatalogManager::Init() before RPCs are served, and reset() once in CatalogManager::CompleteShutdown(). Also none of other many call sites in this file use a defensive check for cdc_state_table. It'd be a dead code.

src/yb/integration-tests/cdc_service-int-test.cc
```diff
@@ -643,6 +650,141 @@ TEST_F(CDCServiceTest, TestDeleteXClusterStream) {
...
+  RpcController rpc;
+  rpc.set_timeout(MonoDelta::FromSeconds(10));
+
+  ASSERT_OK(master_proxy.CleanupStaleCDCStreams(req, &resp, &rpc));
+  ASSERT_TRUE(resp.has_error());
+  EXPECT_EQ(resp.error().code(), master::MasterErrorPB::OBJECT_NOT_FOUND);
+  ASSERT_STR_CONTAINS(resp.error().status().message(), "cdc_state table does not exist");
+}
+
+TEST_F(CDCServiceTest, TestCleanupStaleCDCStreamsDryRunAndDelete) {
```
asrinivasanyb: Could we follow a test pattern similar to CDCSDKYsqlTest::TestValidationAndSyncOfCDCStateEntriesAfterUserTableRemoval - this tests the yb-admin command directly

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,214 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+    return SetupError(
+        resp->mutable_error(),
+        MasterErrorPB::OBJECT_NOT_FOUND,
+        STATUS(
+            NotFound,
+            "cdc_state table does not exist or is not ready."));
+  }
+
+  std::vector<cdc::CDCStateTableKey> all_entry_keys;
+  for (const auto& entry_result : *all_entry_keys_result) {
```
asrinivasanyb: Why do we need 2 passes over the cdc_state table ?

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,214 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+
+    auto tablet_infos = GetTabletInfos({tablet_ids.begin(), tablet_ids.end()});
+    for (auto& tablet_info : tablet_infos) {
+      if (tablet_info) {
+        tablet_info_map.emplace(tablet_info->tablet_id(), std::move(tablet_info));
+      }
+    }
+  }
+
+  std::unordered_map<TabletId, std::vector<scoped_refptr<TableInfo>>> tablet_tables_map;
```
asrinivasanyb: isn't this required only for the colocated table case ?

src/yb/master/xrepl_catalog_manager.cc
egladysh: This is only one pass over cdc_state? We materialize CDCStateTableRange. The later loops are over the in-memory all_entry_keys vector: first to collect candidate tablet IDs for batched metadata lookup, and then to classify rows using that metadata.

src/yb/master/xrepl_catalog_manager.cc
egladysh: It is used for both colocated and non-colocated tablets. For non-colocated tablets this normally contains the single user table for the tablet. For colocated tablets it can contain multiple tables, so the same map lets us report all resolved table metadata.

src/yb/integration-tests/cdc_service-int-test.cc
egladysh: I'll take a look.

egladysh: @asrinivasanyb To confirm, you're after the yb-admin CLI-as-subprocess pattern, not relocating this into the CDCSDKYsqlTest fixture, right?

asrinivasanyb: Both - from the point of view of dealing with items that have been left around in cdc_state due to bugs, this has broad similarity to TestValidationAndSyncOfCDCStateEntriesAfterUserTableRemoval - hence want to be uniform from that perspective

asrinivasanyb:
In Summary 
"xCluster streams that carry an empty namespace_id have their namespace resolved via their first table_id"

xCluster streams to be replaced with CDC streams

src/yb/master/xrepl_catalog_manager.cc
asrinivasanyb: Got it

asrinivasanyb:
Has this been tested on a large cdc_state table - say with about 300000 rows to ensure there is no impact on Master ?

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,214 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+  {
+    SharedLock lock(mutex_);
```
asrinivasanyb, line 5742: why is this lock required here ?

egladysh: Don't we need to protect `tables_`? GetTabletInfos synchronizes on the same mutex for tablets.

egladysh:
@asrinivasanyb It hasn't been tested on on large cdc_state tables. We have no access to YB performance/regression infra. Please advise.

src/yb/integration-tests/cdc_service-int-test.cc
egladysh: I added it to CDCSDKYsqlTest keeping the direct-RPC test.

docs/content/stable/admin/yb-admin.md
```diff
@@ -2065,6 +2065,47 @@ The above command results in the following response:
 Successfully deleted CDC DB Stream ID: d540f5e4890c4d3b812933cbfd703ed3
 ```
 
+#### cleanup_stale_cdc_streams
```
Sumukh-Phalgaonkar: Can you remove the documentation changes from this PR. Our doc team will take up the task of documenting this command.

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -610,6 +611,9 @@ Status CatalogManager::LoadUniverseReplicationBootstrap() {
 // Helper class to print a vector of CDCStreamInfo pointers.
 namespace {
 
+constexpr auto kCDCStateStaleMissingStream = "stream not found";
+constexpr auto kCDCStateStaleMissingTablet = "tablet not found";
```
Sumukh-Phalgaonkar: These are being used only once. Better to directly use these strings in the error message than creating constants for these.

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+    SharedLock lock(mutex_);
+    for (const auto& [stream_id, stream] : cdc_stream_map_) {
+      auto stream_lock = stream->LockForRead();
+      if (stream_lock->is_deleting()) {
+        continue;
+      }
+
+      all_stream_ids.insert(stream_id);
+      NamespaceId stream_namespace = stream_lock->namespace_id();
+      // Some CDC streams do not persist namespace_id in stream metadata. Resolve those via the
```
Sumukh-Phalgaonkar:
```suggestion
      // Xcluster streams do not persist namespace_id in stream metadata. Resolve those via the
```

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+        MasterErrorPB::OBJECT_NOT_FOUND,
+        STATUS(
+            NotFound,
+            "cdc_state table does not exist or is not ready."));
+  }
+
+  Status iteration_status;
+  auto all_entry_keys_result =
+      cdc_state_table_->GetTableRangeAsync({}, &iteration_status);
+  if (!all_entry_keys_result.ok()) {
```
Sumukh-Phalgaonkar, line 5699: Why is this needed? Shouldn't `TableExists()` check ensure that we never hit this?

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+  RETURN_NOT_OK(iteration_status);
+
+  // get the tablet info for all candidate tablets
+  std::unordered_map<TabletId, TabletInfoPtr> tablet_info_map;
+  {
+    // use a set to deduplicate tablet ids
+    // we will use these tablet ids to fetch tablet info
+    std::unordered_set<TabletId> tablet_ids;
+    for (const auto& key : all_entry_keys) {
+      // ignore system tablets
```
Sumukh-Phalgaonkar:
```suggestion
      // Ignore sys catalog tablet and slot entry.
```

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+  }
+
+  std::vector<cdc::CDCStateTableKey> all_entry_keys;
+  for (const auto& entry_result : *all_entry_keys_result) {
+    RETURN_NOT_OK(entry_result);
+    all_entry_keys.push_back(entry_result->key);
+  }
+  RETURN_NOT_OK(iteration_status);
+
+  // get the tablet info for all candidate tablets
```
Sumukh-Phalgaonkar: This comment can be skipped since the code is self evident.

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+  std::unordered_map<TabletId, TabletInfoPtr> tablet_info_map;
+  {
+    // use a set to deduplicate tablet ids
+    // we will use these tablet ids to fetch tablet info
```
Sumukh-Phalgaonkar: Same as above

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+    std::unordered_set<TabletId> tablet_ids;
+    for (const auto& key : all_entry_keys) {
+      // ignore system tablets
+      if (key.tablet_id == kCDCSDKSlotEntryTabletId || key.tablet_id == kSysCatalogTabletId) {
+        continue;
+      }
+
+      if (!namespace_filter.empty()) {
+        const auto it = stream_namespaces.find(key.stream_id);
+        // if the stream is found but belongs to a different namespace, skip it
```
Sumukh-Phalgaonkar: This comment is not needed.

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+        if (it != stream_namespaces.end() && it->second != namespace_filter) {
```
Sumukh-Phalgaonkar: Why should we add the tablet_id to tablet_ids if `it == stream_namespaces.end()`?

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5206,6 +5210,217 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+            // so we should consider this entry for deletion
+            belongs_to_namespace = true;
+            break;
+          }
+        }
+      }
+      if (!belongs_to_namespace) {
+        continue;
+      }
+    }
```
Sumukh-Phalgaonkar: I did not understand why this block is needed? Are we not populating stream_namespaces for every stream, either through the namespace_id field or by looking at the first table_id?

Sumukh-Phalgaonkar: I feel the whole cleanup done by this rpc should be limited to CDC streams only, i.e the ones with namespace_id present in their metadata. Any xcluster streams should be skipped. This would simplify this code a lot.

src/yb/master/xrepl_catalog_manager.cc
Sumukh-Phalgaonkar: Instead of taking the lock twice, we should create a copy of required data at the start of this method. This way we would be working on a consistent snapshot of cdc_stream_map_ and tables_.

asrinivasanyb:
> @asrinivasanyb It hasn't been tested on on large cdc_state tables. We have no access to YB performance/regression infra. Please advise.

You could create an environment with large number of tablets (say 100 tables * 30 tablets/table) and create 100 streams  - basically (streams*tablets = 300k) - then run this command and check that there are no issues on master (long operation tracker logs for heartbeats for example)

Sumukh-Phalgaonkar:
The option to provide database name should be removed, as it can lead to improper cleanup. Consider the example where both the tablet_info as well as the CDC stream info is not available for a particular state table entry. We would not be able to identify the namespace to which this entry belongs and hence would not be cleaned up.

Sumukh-Phalgaonkar:
I feel that the master rpc `CatalogManager::CleanupStaleCDCStreams()` should be simplified as follows:

1. Get only the CDC streams from cdc_stream_map_. The objective is to not perform any cleanup for xcluster streams in this rpc.
2. Iterate over the cdc_state table. Skip the slot entries and entries for sys catalog tablet.
3. Consider only the cdc entries. The cdc related entries will have cdc_sdk_safe_time associated with them. Xcluster entries do not have cdc_sdk_safe_time. A utility function can be written to fetch only cdc related rows from cdc_state table.
4. If the stream is not present, add the entry to entries_to_delete set.
5. If the stream is present, try to fetch the tablet_info for the tablet id using `TableInfo::GetTabletsIncludeInactive()`. If tablet_info is not found, add this entry to entries_to_delete set.
6. Skip sending table_name in the response. If the tablet_info is not present, it will not be possible to fetch table_name.
7. Delete the entries in entries_to_delete set.

src/yb/master/xrepl_catalog_manager.cc
egladysh: I think the comment could be helpful but I don't have strong feeling about it, will remove.

src/yb/master/xrepl_catalog_manager.cc
egladysh: Same as above, will remove.

src/yb/master/xrepl_catalog_manager.cc
egladysh: Oh, it's historical.

src/yb/master/xrepl_catalog_manager.cc
egladysh: I don’t think TableExists() fully guarantees that the later GetTableRangeAsync() will succeed. TableExists() is only a fast pre-check to avoid the much longer GetTableRangeAsync() timeout when cdc_state is absent. I don't think the code is wrong but I might be missing something.

src/yb/master/xrepl_catalog_manager.cc
egladysh: There are three cases: 
1. stream exists and belongs to another namespace
2. stream exists and belongs to the requested namespace
3. stream is missing or has no attributable namespace

I think your question is about case 3?  This is needed for stale rows when the stream is gone but the tablet still belongs to the namespace.

src/yb/master/xrepl_catalog_manager.cc
egladysh: Are you suggesting to take the full tables snapshot upfront? I’m a little concerned that snapshotting all tables up front may be more expensive than the current targeted lookup. I don’t think the current two-lock approach creates a practical correctness issue. This command is already best-effort b/c `cdc_state` is scanned outside the catalog lock.

src/yb/master/xrepl_catalog_manager.cc
egladysh: The block is needed only because the current RPC tries to handle rows whose stream metadata is missing. 

Your second point is more is more fundamental as it implies the scope reduction. Why would we want to do that? I might be missing something but if we limit this to streams with `namespace_id` in stream metadata, then we would no longer clean up missing stream rows under a namespace filter, because there is no stream metadata left to prove they were CDCSDK. Is the goal to intentionally drop support for those rows, or mainly to simplify the implementation?

egladysh:
The namespace filter was intended to avoid deleting rows outside the requested database, but for entries where both stream metadata and tablet metadata are missing we cannot attribute the row to any namespace. If the concern is that database scoped cleanup can be incomplete, then I'd agree that removing the database-name option would address that. Well it would remove the capability. :)

On the other hand the user has an options to dry run that can be performed on the whole cluster (namespace is optional) to get some sense of the scoped cleanup. I'd vote for leaving the namespace option. I think it is useful for a blast-radius control and for investigating stale CDC rows for one database.

egladysh:
I thought we agreed on the command line syntax/semantic?

egladysh:
@Sumukh-Phalgaonkar Does YB have a fixture for testing other yb-admin commands in terms of long operation tracker logs for heartbeats, etc.?

docs/content/stable/admin/yb-admin.md
Sumukh-Phalgaonkar: The doc changes are still present in the PR, marking this conversation as unresolved for now.

src/yb/master/xrepl_catalog_manager.cc
Sumukh-Phalgaonkar: The goal is to simplify implementation. Also we don't want this command to delete xcluster related cdc_state entries.

Sumukh-Phalgaonkar:
> The namespace filter was intended to avoid deleting rows outside the requested database, but for entries where both stream metadata and tablet metadata are missing we cannot attribute the row to any namespace. If the concern is that database scoped cleanup can be incomplete, then I'd agree that removing the database-name option would address that. Well it would remove the capability. :)
> 
> On the other hand the user has an options to dry run that can be performed on the whole cluster (namespace is optional) to get some sense of the scoped cleanup. I'd vote for leaving the namespace option. I think it is useful for a blast-radius control and for investigating stale CDC rows for one database.

Yes the main concern regarding namespace filter is that it could lead to an incomplete cleanup. From a user's POV, running this command should cleanup all the stale entries in the cdc_state table. However, if a user runs this command sequentially for all the namespaces, stale entries could still remain.

Having the namespace filter might not add much value after all. Given that this is a cleanup operation for reducing the bloat from cdc_state table, I think we would be better off cleaning all of the stale entries in one go. The dry run option provides the safety check and can be used to validate that this command is indeed deleting the correct entries.

Also removing the namespace filter would simplify the code by a good deal.

Sumukh-Phalgaonkar:
> @Sumukh-Phalgaonkar Does YB have a fixture for testing other yb-admin commands in terms of long operation tracker logs for heartbeats, etc.?

No, I don't think we have anything like that right now. But our QA team is building some stress tests in this regard.

egladysh:
@Sumukh-Phalgaonkar I removed the namespace option.

src/yb/master/master_replication.proto
```diff
@@ -882,6 +882,30 @@ message ValidateAndSyncCDCStateEntriesForCDCSDKStreamResponsePB {
   repeated string deleted_tablet_entries = 3;
 }
 
+message CleanupStaleCDCStreamsRequestPB {
+  reserved 1; // Deprecated namespace_id.
```
Sumukh-Phalgaonkar: This is the first time this proto will be introduced. Namespace_id was never present before hence this reservation for namespace_id is not required

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5381,6 +5382,161 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+      if (stream_lock->is_deleting()) {
+        continue;
+      }
+
+      all_stream_ids.insert(stream_id);
+    }
+  }
+
+  Status iteration_status;
+  auto all_entry_keys_result =
```
Sumukh-Phalgaonkar, line 5697: I think this can be reduced to:
```
auto all_entry_keys = VERIFY_RESULT(cdc_state_table_->GetTableRangeAsync({}, &iteration_status));
```
The if block below could then be skipped

src/yb/master/xrepl_catalog_manager.cc
Sumukh-Phalgaonkar: To prevent taking the lock twice, you could move the iteration over cdc_stream_map_ (lines 5405 to 5415) here.

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5381,6 +5382,161 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+
+    auto tablet_infos = GetTabletInfos({tablet_ids.begin(), tablet_ids.end()});
+    for (auto& tablet_info : tablet_infos) {
+      if (tablet_info) {
+        tablet_info_map.emplace(tablet_info->tablet_id(), std::move(tablet_info));
+      }
+    }
+  }
+
+  std::unordered_map<TabletId, std::vector<scoped_refptr<TableInfo>>> tablet_tables_map;
```
Sumukh-Phalgaonkar: It seems like tablet_tables_map is not being used anymore, could you cleanup the code around it.

src/yb/master/xrepl_catalog_manager.cc
Sumukh-Phalgaonkar: Given that tablet_tables_map is no longer being used I think this acquisition of mutex_ can also be avoided. The one at lines 5405 to 5415 can be retained.

src/yb/tools/yb-admin_cli.cc
```diff
@@ -79,6 +79,8 @@ DEFINE_NON_RUNTIME_int64(timeout_ms, 1000 * 60, "RPC timeout in milliseconds");
 
 // Command-specific flags
 DEFINE_NON_RUNTIME_bool(exclude_dead, false, "Exclude dead tservers from output");
+DEFINE_NON_RUNTIME_bool(dry_run, false,
```
Sumukh-Phalgaonkar: Instead of a flag, could you make `dry_run` an argument of the cleanup_stale_cdc_streams command.

src/yb/integration-tests/cdc_service-int-test.cc
```diff
@@ -210,8 +212,13 @@ class CDCServiceTest : public YBMiniClusterTestBase<MiniCluster> {
...
 
 void CDCServiceTest::CreateTable(int num_tablets, TableHandle* table) {
-  ASSERT_OK(client_->CreateNamespaceIfNotExists(kTableName.namespace_name(),
-                                                kTableName.namespace_type()));
+  CreateTable(num_tablets, kTableName, table);
+}
+
+void CDCServiceTest::CreateTable(
+    int num_tablets, const client::YBTableName& table_name, TableHandle* table) {
+  ASSERT_OK(client_->CreateNamespaceIfNotExists(
```
Sumukh-Phalgaonkar: Why was this needed?

src/yb/integration-tests/cdcsdk_ysql_test_base.cc
```diff
@@ -5058,6 +5058,22 @@ Status CDCSDKYsqlTest::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
   return Status::OK();
 }
 
+Result<std::string> CDCSDKYsqlTest::CleanupStaleCDCStreams(bool dry_run) {
+  string tool_path = GetToolPath("../bin", "yb-admin");
+  vector<string> argv;
+  argv.push_back(tool_path);
+  argv.push_back("--master_addresses");
+  argv.push_back(AsString(test_cluster_.mini_cluster_->GetMasterAddresses()));
```
Sumukh-Phalgaonkar, line 5067: Use the utility function `CDCSDKYsqlTest::ExecuteYBAdminCommand()` instead of rewriting

src/yb/integration-tests/cdc_service-int-test.cc
```diff
@@ -643,6 +650,82 @@ TEST_F(CDCServiceTest, TestDeleteXClusterStream) {
...
+  RpcController rpc;
+  rpc.set_timeout(MonoDelta::FromSeconds(10));
+
+  ASSERT_OK(master_proxy.CleanupStaleCDCStreams(req, &resp, &rpc));
+  ASSERT_TRUE(resp.has_error());
+  EXPECT_EQ(resp.error().code(), master::MasterErrorPB::OBJECT_NOT_FOUND);
+  ASSERT_STR_CONTAINS(resp.error().status().message(), "cdc_state table does not exist");
+}
+
+TEST_F(CDCServiceTest, TestCleanupStaleCDCStreamsDryRunAndDelete) {
```
Sumukh-Phalgaonkar: Seems like this test is redundant. The test `CDCSDKYsqlTest.TestCleanupStaleCDCStreamsWithYBAdminDryRunAndDelete` also tests the same things. You can remove this one.

Sumukh-Phalgaonkar:
> An optional YSQL database name to restrict cleanup to entries attributable to a single namespace. Entries whose stream and tablet metadata are both missing cannot be attributed to any namespace and are only removed in an unfiltered (all-namespaces) invocation.

> Documentation for the new command is added under the yb-admin Change Data Capture commands in docs/content/stable/admin/yb-admin.md.

These things are no longer valid. Could you take a re-look at the summary.

Sumukh-Phalgaonkar:
Trigger jenkins

src/yb/integration-tests/cdc_service-int-test.cc
egladysh: Agreed.

src/yb/integration-tests/cdcsdk_ysql_test_base.cc
egladysh: It's not exactly a rewrite. `ExecuteYBAdminCommand` doesn't seem to capture stdout, and it only supports arguments after command name. Are you suggesting to modify `ExecuteYBAdminCommand`? I am find either way.

src/yb/integration-tests/cdc_service-int-test.cc
egladysh: Good catch.

src/yb/master/xrepl_catalog_manager.cc
egladysh: `tablet_tables_map` is still used when reporting stream-not-found entries.

src/yb/master/xrepl_catalog_manager.cc
egladysh: 1. The code isn't wrong.
2. The difference is error mapping that seems to be required TestCleanupStaleCDCStreamsWithoutCDCStateTable.

src/yb/tools/yb-admin_cli.cc
egladysh: Good call.

src/yb/master/xrepl_catalog_manager.cc
egladysh: I refactored it. We only fetch the data for missing streams.

src/yb/master/xrepl_catalog_manager.cc
```diff
@@ -5674,6 +5674,158 @@ Status CatalogManager::ValidateAndSyncCDCStateEntriesForCDCSDKStream(
...
+    return SetupError(
+        resp->mutable_error(),
+        MasterErrorPB::OBJECT_NOT_FOUND,
+        STATUS(
+            NotFound,
+            "cdc_state table does not exist or is not ready."));
+  }
+
+  Status iteration_status;
+  auto all_entry_keys_result =
```
Sumukh-Phalgaonkar, line 5697: Can you just use, RETURN_NOT_OK instead of explicitly sending MasterErrorPB::OBJECT_NOT_FOUND

Sumukh-Phalgaonkar (approved):

Sumukh-Phalgaonkar:
Trigger jenkins

Sumukh-Phalgaonkar:
@egladysh The changes look good to me. I have approved the PR, and triggered the test run. Could you pls hold on from merging until the test results come out. I will notify you about the same.

egladysh:
@Sumukh-Phalgaonkar thanks for the review! Actually I don't have permissions to land stuff.

Sumukh-Phalgaonkar:
Trigger jenkins

PR #32034, opened 2026-06-04
[Doc] Use partition_key for CREATE TABLE PARTITION BY syntax

ddhodge:
Add a dedicated partition_key rule instead of reusing index_elem, so syntax diagrams do not show index-only HASH/ASC/DESC/NULLS options on partition columns.


ddorian (approved):
LGTM. 

Include `partition_key` in the create_table page https://github.com/yugabyte/yugabyte-db/blob/master/docs/content/stable/api/ysql/the-sql-language/statements/ddl_create_table.md?plain=1#L19-L33.

Issue #32043, opened 2026-06-04
[CDCSDK] TestCDCSDKCacheWithLeaderReElect hangs on macOS due to no-op self-stepdown


ellabaron-code:
Jira Link: [DB-21742](https://yugabyte.atlassian.net/browse/DB-21742)
## Description

The integration test `TestCDCSDKCacheWithLeaderReElect` (`src/yb/integration-tests/cdcsdk_ysql-test.cc`) passes `first_follower_index` to `StepDownLeader`, asking a follower to step down — a no-op, since only the leader can step down.

On Linux this happens to pass: the no-op self-stepdown still triggers a full pre-election + election cycle, and the 1s background task refreshes the follower's `cdc_sdk_min_checkpoint_op_id_expiration` within the test's 500ms `StepDownLeader` sleep, so `CompareExpirationTime` sees a fresh value.

On macOS the timing does not line up, no fresh expiration update reaches the follower, and the test hangs in `CompareExpirationTime` until the 10-minute test timeout kills it.

### Steps to reproduce

Run `TestCDCSDKCacheWithLeaderReElect` on macOS — it hangs until the test timeout.

## Source connector version

N/A — failing integration test, not connector-related.

## Connector configuration

N/A — failing integration test, not connector-related.

## YugabyteDB version

master

## Issue Type

kind/failing-test

I confirm this issue does not contain any sensitive information.

[DB-21742]: https://yugabyte.atlassian.net/browse/DB-21742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #32044, opened 2026-06-04
[#32043] CDC: Fix TestCDCSDKCacheWithLeaderReElect to pass leader index to StepDownLeader

ellabaron-code:
## Summary

`TestCDCSDKCacheWithLeaderReElect` passed `first_follower_index` to `StepDownLeader`, asking a follower to step down — a no-op, since only the leader can step down. Use `first_leader_index` so the leader actually steps down and a real re-election occurs, which is what the test is meant to exercise.

On Linux this happened to pass: the no-op self-stepdown still triggered a full pre-election + election cycle, and the 1s background task refreshed the follower's `cdc_sdk_min_checkpoint_op_id_expiration` within the test's 500ms `StepDownLeader` sleep, so `CompareExpirationTime` saw a fresh value. On macOS the timing didn't line up, no fresh expiration update reached the follower, and the test hung in `CompareExpirationTime` until the 10-minute test timeout killed it.

## Test plan

- [x] `TestCDCSDKCacheWithLeaderReElect` passes on macOS
- [x] `TestCDCSDKCacheWithLeaderReElect` continues to pass on Linux

### DIFF
--- a/src/yb/integration-tests/cdcsdk_ysql-test.cc
+++ b/src/yb/integration-tests/cdcsdk_ysql-test.cc
@@ -3674,7 +3674,7 @@ TEST_F(CDCSDKYsqlTest, YB_DISABLE_TEST_IN_TSAN(TestCDCSDKCacheWithLeaderReElect)
             << correct_expiry_time.time_since_epoch().count();
 
   // we need to ensure the initial leader get's back leadership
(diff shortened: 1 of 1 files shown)


Sumukh-Phalgaonkar (approved):

rahulb-yb:
trigger jenkins

rahulb-yb:
trigger jenkins

Issue #28903, opened 2025-10-10
[YSQL] ERROR:  could not serialize access due to concurrent update with Table locks + Txn ddl + Auto analyze


hariharasudhan-yb:
Jira Link: [DB-18626](https://yugabyte.atlassian.net/browse/DB-18626)
### Description

 In version : 2.29.0.0-b28

With Auto Analyze on +  D47188 in version 2.29.0.0-b28 , Table level locks test encountered an issue where COMMIT statement is failing with serialization error , Consistently reproduced in 3 runs. 
`COMMIT OPERATION failed: could not serialize access due to concurrent update\nDETAIL:  Flush: f12d46b9-b7ff-4a8f-abcd-1560e7b5da04 conflict with concurrently committed data. Value write after transaction start: doc ht ({ physical: 1760005514594476 }) > read time ({ physical: 1760005513571915 logical: 4095 }), key: SubDocKey(DocKey(CoTableId=00000001000030008001000000001f4a, [], [16386]), []): kConflict: . Errors from tablet servers: [Operation failed. Try again (yb/docdb/conflict_resolution.cc:1030): Flush: f12d46b9-b7ff-4a8f-abcd-1560e7b5da04 conflict with concurrently committed data. Value write after transaction start: doc ht ({ physical: 1760005514594476 }) > read time ({ physical: 1760005513571915 logical: 4095 }), key: SubDocKey(DocKey(CoTableId=00000001000030008001000000001f4a, [], [16386]), []): kConflict (transaction error 3)]`

Also :

Few more tests failed with the issue 

We have been hitting this issue in many tests with Table locks + Txn ddl in version 2.29.0.0-b28  With Auto Analyze on + D47188
Posting few failures 
Create table and alter tables case with serial execution and there is no explicit txn block 
Logs : https://jenkins.dev.yugabyte.com/job/itest-system-developer/24453/artifact/logs/2.29.0.0_testysqlcreatetables-aws-rf3_20251010_063246/
```
2025-10-10 06:57:06.971 UTC [20130] ERROR:  could not serialize access due to concurrent update
[...]

PR #32045, opened 2026-06-04
[Docs] Update limitations of Transactional DDL

pkj415:
Added 2 more limitations that apply to transactional DDL.

### DIFF
--- a/docs/content/stable/explore/transactions/transactional-ddl.md
+++ b/docs/content/stable/explore/transactions/transactional-ddl.md
@@ -69,4 +69,24 @@ yugabyte=# SELECT * FROM foo;
 
 - [Savepoints](/stable/develop/learn/transactions/transactions-retries-ysql/#savepoints) are unsupported for DDL statements. As a result, you cannot create a savepoint in a transaction block that has executed a DDL statement. Similarly, you cannot execute a DDL statement in a transaction block in which a savepoint has been created.
 
+- Concurrent transaction blocks with DDLs that touch the same or different catalog data can result in undefined behavior without any error. This situation also applies to DDLs that are run concurrently without transactional DDL enabled. The difference is that, with transactional DDL enabled, it is easier for such behaviour to occur with using READ COMMITTED isolation level. With transactional DDL in REPEATABLE READ isolation level, the chances of such an issue occuring are slim (similar to when transactional DDL is disabled).
+
+For example, in the following sequence, the column `a` added by Session 1 is missing after Session 2 commits:
+
+    ```sql
+    -- Session 1                                  -- Session 2
+
+    BEGIN;
+                                                   BEGIN;
+    ALTER TABLE test ADD COLUMN a INT;
+                                                   ALTER TABLE test ALTER COLUMN v TYPE SMALLINT;
+                                                   -- waiting....
+    COMMIT;
+                                                   COMMIT;
+
+    SELECT * FROM test;  -- Column "a" is missing
+    ```
+
+- When Auto Analyze runs with transactional DDL enabled, and a user DDL is executed concurrently in a transaction block (also with transactional DDL enabled), it can lead to a serialization error either on the user DDL or on the COMMIT of the transaction block containing the user DDL.
+
 For an overview of common concepts used in YugabyteDB's implementation of distributed transactions, see [Distributed transactions](../distributed-transactions-ysql/).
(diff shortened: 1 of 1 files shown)


docs/content/stable/explore/transactions/transactional-ddl.md
```diff
@@ -69,4 +69,24 @@ yugabyte=# SELECT * FROM foo;
 
 - [Savepoints](/stable/develop/learn/transactions/transactions-retries-ysql/#savepoints) are unsupported for DDL statements. As a result, you cannot create a savepoint in a transaction block that has executed a DDL statement. Similarly, you cannot execute a DDL statement in a transaction block in which a savepoint has been created.
 
+- Concurrent transaction blocks with DDLs that touch the same or different catalog data can result in undefined behavior without any error. This situation also applies to DDLs that are run concurrently without transactional DDL enabled. The difference is that, with transactional DDL enabled, it is easier for such behaviour to occur with using READ COMMITTED isolation level. With transactional DDL in REPEATABLE READ isolation level, the chances of such an issue occuring are slim (similar to when transactional DDL is disabled).
+
+For example, in the following sequence, the column `a` added by Session 1 is missing after Session 2 commits:
```
aishwarya24: Indentation fix is required for the example block :

  For example, in the following sequence...

      ```sql
      ...
      ```

docs/content/stable/explore/transactions/transactional-ddl.md
```diff
@@ -69,4 +69,24 @@ yugabyte=# SELECT * FROM foo;
 
 - [Savepoints](/stable/develop/learn/transactions/transactions-retries-ysql/#savepoints) are unsupported for DDL statements. As a result, you cannot create a savepoint in a transaction block that has executed a DDL statement. Similarly, you cannot execute a DDL statement in a transaction block in which a savepoint has been created.
 
+- Concurrent transaction blocks with DDLs that touch the same or different catalog data can result in undefined behavior without any error. This situation also applies to DDLs that are run concurrently without transactional DDL enabled. The difference is that, with transactional DDL enabled, it is easier for such behaviour to occur with using READ COMMITTED isolation level. With transactional DDL in REPEATABLE READ isolation level, the chances of such an issue occuring are slim (similar to when transactional DDL is disabled).
```
aishwarya24:
```suggestion
- Concurrent transaction blocks with DDLs that touch the same or different catalog data can result in undefined behavior without any error. This situation also applies to DDLs that are run concurrently without transactional DDL enabled. The difference is that, with transactional DDL enabled, it is easier for such behavior to occur when using READ COMMITTED isolation level. With transactional DDL in REPEATABLE READ isolation level, the chances of such an issue occurring are slim (similar to when transactional DDL is disabled).
```

docs/content/stable/explore/transactions/transactional-ddl.md
aishwarya24: should this bullet be dropped or combined with the third one? They seem related.

docs/content/stable/explore/transactions/transactional-ddl.md
```diff
@@ -69,4 +69,24 @@ yugabyte=# SELECT * FROM foo;
...
+    ALTER TABLE test ADD COLUMN a INT;
+                                                   ALTER TABLE test ALTER COLUMN v TYPE SMALLINT;
+                                                   -- waiting....
+    COMMIT;
+                                                   COMMIT;
+
+    SELECT * FROM test;  -- Column "a" is missing
+    ```
+
+- When Auto Analyze runs with transactional DDL enabled, and a user DDL is executed concurrently in a transaction block (also with transactional DDL enabled), it can lead to a serialization error either on the user DDL or on the COMMIT of the transaction block containing the user DDL.
```
aishwarya24, line 88: This limitation is also on the auto analyze page - https://docs.yugabyte.com/stable/additional-features/auto-analyze/#limitations. So should this be added instead?

aishwarya24 (commented):
some comments

docs/content/stable/explore/transactions/transactional-ddl.md
pkj415: Yes, so should I just add that same text here? Also, that limitation mentions issue [#28903](https://github.com/yugabyte/yugabyte-db/issues/28903) for tracking but that issue is closed, but the limitation still exists. So maybe we should avoid mentioning that issue number?

PR #32056, opened 2026-06-05
[DOC-1349] Note clone time travel limitation

ddhodge:
@netlify /stable/manage/backup-restore/instant-db-cloning/#limitations

### DIFF
--- a/docs/content/stable/manage/backup-restore/instant-db-cloning.md
+++ b/docs/content/stable/manage/backup-restore/instant-db-cloning.md
@@ -314,5 +314,4 @@ If you have [tablet limits](../../../architecture/docdb-sharding/tablet-splittin
 
 - Cloning as of a time close to the limit of the history retention period may fail. For example, if you have a history retention period of 10 minutes and you create a clone as of 9 minutes ago, the clone operation may fail. Use a slightly larger history retention period than you think you need.
 - Cloning to a point in time during which a DDL was running may fail. See issue {{<issue 28814>}}.
-- You cannot [time travel](../time-travel-query/) to a point in time prior to the 'as of' time of the clone.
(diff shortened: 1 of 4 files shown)


aishwarya24 (approved):
LGTM

docs/content/stable/manage/backup-restore/instant-db-cloning.md
```diff
@@ -314,4 +314,5 @@ If you have [tablet limits](../../../architecture/docdb-sharding/tablet-splittin
 
 - Cloning as of a time close to the limit of the history retention period may fail. For example, if you have a history retention period of 10 minutes and you create a clone as of 9 minutes ago, the clone operation may fail. Use a slightly larger history retention period than you think you need.
 - Cloning to a point in time during which a DDL was running may fail. See issue {{<issue 28814>}}.
+- You cannot [time travel](../time-travel-query/) to a point in time prior to the 'as of' time of the clone.
```
yamen-haddad: This limitation should be added to the Time-Travel Docs.
What about we add under the line:
`Time travel queries are not supported for temporary tables.` in the Time-travel docs
The following:
`You cannot time-travel to a time prior to the creation time of a database clone`

yamen-haddad (changes requested):

yamen-haddad:
Can we also add the following info to the time travel documentation:

- Under the section `When setting yb_read_time, keep in mind the following:` Let's append to the last bullet point to be as follows:
```
Currently, time travel queries can only read old data without schema changes. In other words, do not set the read time to a time earlier than the most recent DDL operation. This includes cluster-wide DDLs that affect global objects, such as a clone operation.
```

yamen-haddad (approved):

Issue #32075, opened 2026-06-07
[YSQL] Stale txn_start_us at coordinator causes wrong deadlock victim selection


ellabaron-code:
Jira Link: [DB-21771](https://yugabyte.atlassian.net/browse/DB-21771)
## Description

The deadlock detector selects its victim using each transaction's
`txn_start_us`, read from the coordinator's per-txn `first_touch_` field.
`first_touch_` is populated from the `start_time` field on incoming
heartbeat RPCs.

When a connection takes a pre-prepared transaction from `TransactionPool`,
the pool registers it on the coordinator (CREATED heartbeat) *before* PG
supplies `xactStartTimestamp`. At that point `start_ == 0`, and the
heartbeat code suppresses `start_time` when `start_ == 0`, so the
coordinator records `first_touch_ = 0`.

When PG later calls `SetPgTxnStart(xactStartTimestamp, ...)`, the original
implementation stored the value only into `metadata_.pg_txn_start_us` — it
did not update `start_` and did not notify the coordinator. Only the next
periodic heartbeat (~500 ms later) carried the correct `start_time`. If the
deadlock detector fired before that tick, victim selection used a stale
`txn_start_us = 0` and fell through to the UUID tiebreaker, picking the
wrong victim.

This manifested as `PgObjectLocksTestRF1Deadlock.TestDeadlockSimple`
consistently hanging on macOS and being latently broken on Linux (where
scheduling usually delivered the periodic heartbeat in time).

### Steps to reproduce
- Run `PgObjectLocksTestRF1Deadlock.TestDeadlockSimple` (deterministic hang
  on macOS; latent on Linux).

## Issue Type
kind/bug

## Warning: Please confirm that this issue does not contain any sensitive information
- [x] I confirm this issue does not contain any sensitive information.

[DB-21771]: https://yugabyte.atlassian.net/browse/DB-21771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #32076, opened 2026-06-07
[#32075] YSQL: Fix stale txn_start_us at coordinator causing wrong deadlock victim selection

ellabaron-code:
## Problem

`PgObjectLocksTestRF1Deadlock.TestDeadlockSimple` was hanging on macOS. On Linux it usually passed but was latently broken in the same way — the test only happened to land on a benign timing.

## Root cause

The deadlock detector picks the victim with the maximum `txn_start_us` (newest transaction loses), with UUID lex order as tiebreaker. It reads `txn_start_us` from the transaction coordinator's per-txn `first_touch_` field, which is set from the `start_time` field on incoming heartbeat RPCs.

Two coincident facts produce the bug:

1. **`TransactionPool::Take` pre-prepares transactions.** When a connection asks the pool for a `kPlain` transaction, the pool returns one it has already `Prepare()`-ed. `Prepare()` triggers `RequestStatusTablet → LookupTabletDone → SendHeartbeat(CREATED)` — registering the transaction on the coordinator **before PG has provided the `xactStartTimestamp`**. At that moment, `YBTransaction::Impl::start_` is still 0.

2. **The heartbeat code suppresses `start_time` when `start_ == 0`** (`client/transaction.cc:1458-1467`). So the prebuild's CREATED heartbeat carries no `start_time` field; the coordinator records `first_touch_ = 0` (`tablet/transaction_coordinator.cc:787`).

When PG later calls `YBTransaction::SetPgTxnStart(xactStartTimestamp, ...)`, the original implementation stored the value only into `metadata_.pg_txn_start_us` — it did **not** update `start_`, and did **not** notify the coordinator. The next periodic heartbeat (`FLAGS_transaction_heartbeat_usec`, default ~500 ms) would eventually carry the correct `start_time` and overwrite `first_touch_`, but in the failing test runs the deadlock detector fired before that tick landed.

## Platform-specific symptom

The bug is the same on both platforms, but the observed outcome diverges based on timing:

- **macOS (consistently hangs).** Both transactions in the deadlock cycle were still at `first_touch_ = 0` when the deadlock detector ran. The cycle string emitted was `<0>c3561032-...-><0>52963476-...->`. With both `txn_start_us` equal to 0, victim selection fell through to UUID lex tiebreaker. On the failing runs this picked the SHARE-holder, which then released its locks while the lock-waiter kept waiting — orphaned locks → hang.

- **Linux (usually passes).** The second transaction's periodic heartbeat consistently landed before the deadlock detector ran, so its `first_touch_` had the real PG `xactStartTimestamp` while the first transaction's was still 0. Since `real_ts > 0`, the second (newer) transaction was correctly picked as victim. The test passed not because the bug wasn't present, but because Linux scheduling happened to deliver a heartbeat in the narrow window before the deadlock detector ran. The macOS scheduler (or local clock / RPC latency) shifted that race the other way, making the bug deterministic.

This is why prior `--num-reps` runs on Linux looked clean: the bug was masked by a heartbeat-vs-detector race that Linux usually won and macOS usually lost.

## Fix

Test-only change: after `WaitFor(NumWaitingLocks() >= 1)` succeeds — i.e., both transactions have executed their first `LOCK` and therefore set `start_` locally via `SetStartTimeIfNecessary` — sleep `2 × FLAGS_transaction_heartbeat_usec` before forming the deadlock cycle. This gives the periodic heartbeat one tick to carry the real `start_us` to each status-tablet coordinator, so the deadlock detector sees real values instead of `<0>`.

Matches the existing 2×-heartbeat sleep pattern used elsewhere in the same file (lines 1495, 1507).

An earlier version of this PR fixed the underlying race in `YBTransaction::SetPgTxnStart` (write `start_` + fire a corrective heartbeat). That approach was set aside in favor of stabilizing the test, leaving the production code path unchanged.

## Test plan

- [x] `PgObjectLocksTestRF1Deadlock.TestDeadlockSimple` passes on macOS (previously a deterministic hang)
- [x] `PgObjectLocksTestRF1Deadlock.TestDeadlockSimple` passes on Linux across repeated runs

### DIFF
--- a/src/yb/yql/pgwrapper/pg_object_locks-test.cc
+++ b/src/yb/yql/pgwrapper/pg_object_locks-test.cc
@@ -355,6 +355,12 @@ TEST_F(PgObjectLocksTestRF1Deadlock, YB_DISABLE_TEST_IN_SANITIZERS(TestDeadlockS
   ASSERT_OK(WaitFor([&]() {
     return NumWaitingLocks() >= 1;
   }, 5s * kTimeMultiplier, "Timed out waiting for num waiting locks to be >= 1"));
+  // At this point both txns have triggered SetStartTimeIfNecessary via the object-lock path, so
+  // their local start_ values are set. Wait for the next periodic transaction heartbeat to carry
+  // those values to each status-tablet coordinator before forming the deadlock cycle, so the
+  // deadlock detector sees real txn_start_us values instead of the stale first_touch_=0 left by
+  // the pool-prebuilt CREATED heartbeat.
+  SleepFor(FLAGS_transaction_heartbeat_usec * 1us * kTimeMultiplier * 2);
   auto status_future_2 = std::async(std::launch::async, [&]() -> Status {
     RETURN_NOT_OK(conn1.Execute("LOCK TABLE test IN ACCESS EXCLUSIVE MODE"));
     return Status::OK();


src/yb/client/transaction.cc
```diff
@@ -1064,6 +1067,16 @@ class YBTransaction::Impl final : public internal::TxnBatcherIf {
...
+    start_.store(pg_txn_start_us, std::memory_order_release);
+    internal::RemoteTabletPtr st;
+    {
+      SharedLock<std::shared_mutex> lock(mutex_);
+      st = status_tablet_;
+    }
+    if (st) {
+      SendHeartbeat(TransactionStatus::PENDING, metadata_.transaction_id,
+                    transaction_->weak_from_this(), SendHeartbeatToNewTablet::kFalse);
+    }
```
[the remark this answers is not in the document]
ellabaron-code: Thanks — good catch. The symptom is actually worse than you described: `Rpcs::RegisterAndStart` enforces `CHECK(*handle == calls_.end())` at `rpc/rpc.cc:378`, so reusing `heartbeat_handle_` while a periodic heartbeat is in flight would crash the `tserver`, not silently drop the call.

Fixed in the follow-up commit by introducing a dedicated `corrective_heartbeat_handle_` for the `SetPgTxnStart` send. It's independent of `heartbeat_handle_/new_heartbeat_handle_`, so the two paths can run concurrently without contention. Preferred this over a pending-flag-in-HeartbeatDone approach to avoid adding mutable race-prone state to the periodic path.

basavaraj29:
> The deadlock detector picks the victim with the maximum txn_start_us (newest transaction loses)
This is best effort, and not always guaranteed.

> The next periodic heartbeat (FLAGS_transaction_heartbeat_usec, default ~500 ms) would eventually carry the correct start_time and overwrite first_touch_, but in the failing test runs the deadlock detector fired before that tick landed.
Allowing things to propagate on the next heartbeat was a conscious choice we made, because the above logic of deadlock victim was best effort.

We could just fix the test and not make any core code level fixes here.

src/yb/client/transaction.cc
```diff
@@ -2597,6 +2618,7 @@ class YBTransaction::Impl final : public internal::TxnBatcherIf {
   TaggedLogPrefix log_prefix_;
   rpc::Rpcs::Handle heartbeat_handle_;
   rpc::Rpcs::Handle new_heartbeat_handle_;
+  rpc::Rpcs::Handle corrective_heartbeat_handle_;
```
basavaraj29: let's not duplicate sources from where txn heartbeats are sent.

basavaraj29 (changes requested):
as commented on the issue, we could just fix the test and not make any core code level changes.

ellabaron-code:
@basavaraj29 Hi Basava, how do you suggest we fix the test then. Should we introduce some sleep period to allow heartbeats to propagate before triggering deadlock?

basavaraj29:
> Should we introduce some sleep period to allow heartbeats to propagate before triggering deadlock?
yes, could add sleeps in order in txn heartbeats to stabilize the test.

ellabaron-code:
@basavaraj29
Thank you, I will do that.

src/yb/client/transaction.cc
ellabaron-code: Fixed.

ellabaron-code:
Fixed. Please, review again.

src/yb/yql/pgwrapper/pg_object_locks-test.cc
basavaraj29, line 377: I hadn't looked at the test earlier. and reading the issue description somehow led me to think the test relies on the latter transaction to be aborted. 

But looks like the test expects either of the transactions to be aborted (due to the heartbeat race you described). So I don't see the need to sleep for the heartbeat interval in this test.

Looks like we are missing something. May be the test is flaky/failing due to some other reason?

PR #32126, opened 2026-06-10
[#32220] YSQL: Fixes unique index consistency bug when rows have NULL key columns

### LINKED ISSUE #32220: [YSQL] Index inconsistency on NULL values in unique index
Jira Link: [DB-21903](https://yugabyte.atlassian.net/browse/DB-21903)
### Description

Consider the following schema:
```sql
CREATE TABLE t_null_error (k INT PRIMARY KEY, v INT);
CREATE UNIQUE INDEX ON t_null_error (v);

-- Seed data
INSERT INTO t_null_error VALUES (1, NULL);

yugabyte=# \d t_null_error
            Table "public.t_null_error"
 Column |  Type   | Collation | Nullable | Default
--------+---------+-----------+----------+---------
 k      | integer |           | not null |
 v      | integer |           |          |
Indexes:
    "t_null_error_pkey" PRIMARY KEY, lsm (k HASH)
    "t_null_error_v_idx" UNIQUE, lsm (v HASH)

yugabyte=# SELECT * FROM t_null_error;
 k |   v
---+--------
 1 | [NULL]
(1 row)
```

The following UPDATE query causes an index inconsistency:
```sql
UPDATE t_null_error SET k = 2 WHERE k = 1;
```

```sql
yugabyte=# UPDATE t_null_error SET k = 2 WHERE k = 1;
UPDATE 1

yugabyte=# SELECT yb_index_check('t_null_error_v_idx'::regclass);
ERROR:  XX002: ybuniqueidxkeysuffix and ybbasectid mismatch
DETAIL:  index: 't_null_error_v_idx', ybbasectid: '\x47c0c448800000022121'
```

### Issue Type

kind/bug

[DB-21903]: https://yugabyte.atlassian.net/browse/DB-21903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

austenLacy:
`yb_enable_inplace_index_update` incorrectly treats primary-key updates on unique secondary indexes as safe for in-place index-row updates. This is false when the unique index uses NULLs-distinct semantics and any unique-index key column is NULL, because the hidden `ybuniqueidxkeysuffix` is part of the index row key and is derived from the base row `ybctid`.  A base table PK update changes the base row ybctid, so the hidden unique suffix changes too.

fixes: https://github.com/yugabyte/yugabyte-db/issues/32220

Reproducible example:

 ```sql
SET yb_enable_inplace_index_update = on;

CREATE TABLE idx_test (
  c1 bigint,
  id bigint,
  foo smallint DEFAULT 0 NOT NULL,
  some_hash bytea,
  is_not_deleted boolean,
  PRIMARY KEY ((c1) HASH, id ASC, foo ASC)
);

CREATE UNIQUE INDEX idx_test_uq
ON idx_test USING lsm (c1 HASH, some_hash ASC)
WHERE is_not_deleted;

INSERT INTO idx_test (c1, id, foo, some_hash, is_not_deleted)
VALUES (1, 1, 0, NULL, true);

UPDATE idx_test SET foo = 1
WHERE c1 = 1 AND id = 1 AND foo = 0;

SELECT yb_index_check('idx_test_uq'::regclass);
```

Consistency check immediately fails

```sql
yugabyte=# SELECT yb_index_check('idx_test_uq'::regclass);
ERROR:  ybuniqueidxkeysuffix and ybbasectid mismatch
DETAIL:  index: 'idx_test_uq', ybbasectid: '\x47eda949800000000000000121498000000000000001488000000121'
```

The fix is to essentially force the `DELETE+INSERT` for this scenario.

With `yb_enable_inplace_index_update = off` the issue is not reproducible with the above example.

### DIFF
--- a/src/postgres/src/backend/executor/execIndexing.c
+++ b/src/postgres/src/backend/executor/execIndexing.c
@@ -763,13 +763,21 @@ YbUniqueIndexKeyHasNull(IndexInfo *indexInfo,
 {
 	Datum		values[INDEX_MAX_KEYS];
 	bool		isnull[INDEX_MAX_KEYS];
+	MemoryContext oldContext;
+	bool		result;
 
 	Assert(indexInfo->ii_Unique);
+
+	oldContext = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
 
 	GetPerTupleExprContext(estate)->ecxt_scantuple = slot;
 	FormIndexDatum(indexInfo, slot, estate, values, isnull);
 
-	return YbIsAnyIndexKeyColumnNull(indexInfo, isnull);
+	result = YbIsAnyIndexKeyColumnNull(indexInfo, isnull);
+
+	MemoryContextSwitchTo(oldContext);
+
+	return result;
 }
 
 List *
--- a/src/postgres/src/backend/executor/execIndexing.c
+++ b/src/postgres/src/backend/executor/execIndexing.c
@@ -757,27 +757,49 @@ YbExecDoUpdateIndexTuple(ResultRelInfo *resultRelInfo,
 }
 
 static bool
-YbUniqueIndexKeyHasNull(IndexInfo *indexInfo,
-						 TupleTableSlot *slot,
-						 EState *estate)
+YbIndexKeyColumnsModifiedByPkeyUpdate(Relation indexRelation,
+									 TupleTableSlot *slot,
+									 bool is_pk_updated)
 {
-	Datum		values[INDEX_MAX_KEYS];
-	bool		isnull[INDEX_MAX_KEYS];
-	MemoryContext oldContext;
-	bool		result;
+	Form_pg_index indexData = indexRelation->rd_index;
 
-	Assert(indexInfo->ii_Unique);
+	if (!is_pk_updated)
+		return false;
 
-	oldContext = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
+	/*
+	 * In non-unique indexes, ybidxbasectid is part of the index row key and
+	 * changes whenever the base table primary key changes.
+	 */
+	if (!indexData->indisunique)
+		return true;
 
-	GetPerTupleExprContext(estate)->ecxt_scantuple = slot;
-	FormIndexDatum(indexInfo, slot, estate, values, isnull);
+	/*
+	 * In unique indexes with NULLS NOT DISTINCT, ybuniqueidxkeysuffix is
+	 * always NULL, so a base table primary key update does not change the
+	 * index row key.
+	 */
+	if (indexData->indnullsnotdistinct)
+		return false;
 
-	result = YbIsAnyIndexKeyColumnNull(indexInfo, isnull);
+	for (int i = 0; i < indexData->indnkeyatts; i++)
+	{
+		AttrNumber	attnum = indexData->indkey.values[i];
 
-	MemoryContextSwitchTo(oldContext);
+		/*
+		 * Expression key columns have attnum 0. Avoid evaluating the
+		 * expression here; conservatively assume that it can be NULL.
+		 * This consequently leads to a less optimized scenario where
+		 * we do a DELETE+INSERT unnecessarily, but this favors safety
+		 * over speed. It would be possible to use something like
+		 * FormIndexDatum to look up the actual values, but that's an
+		 * expensive operation that would need to be optimized itself
+		 * if we were to call it very frequently.
+		 */
+		if (attnum <= 0 || slot_attisnull(slot, attnum))
+			return true;
+	}
 
-	return result;
+	return false;
 }
 
 List *
@@ -977,17 +999,13 @@ YbExecUpdateIndexTuples(ResultRelInfo *resultRelInfo,
 		}
 
 		/*
-		 * In the following scenarios, the index tuple can be modified (updated)
-		 * in-place, without the need to delete and reinsert the tuple:
-		 * - The index is a covering index (number of key columns < number of
-		 *   columns in the index), only the non-key columns need to be updated,
-		 *   and the primary key is not updated.
-		 * - The index is a unique index and only the non-key columns of the
-		 *   index need to be updated. Primary key updates are also safe unless
-		 *   the hidden unique index key suffix depends on the base row ybctid.
+		 * In-place index tuple updates are possible only when the index tuple has
+		 * non-key columns to update, or when a unique index needs only its
+		 * ybbasectid value updated. Non-unique indexes without non-key columns
+		 * have no value columns to update in-place.
 		 */
-		if ((indexData->indnkeyatts == indexData->indnatts || is_pk_updated) &&
-			(!indexData->indisunique))
+		if (indexData->indnkeyatts == indexData->indnatts &&
+			!indexData->indisunique)
 		{
 			deleteIndexes = lappend_int(deleteIndexes, i);
 			insertIndexes = lappend_int(insertIndexes, i);
@@ -1045,15 +1063,14 @@ YbExecUpdateIndexTuples(ResultRelInfo *resultRelInfo,
 		}
 
 		/*
-		 * Unique indexes that use NULLS DISTINCT semantics store
-		 * ybuniqueidxkeysuffix as the base table ybctid whenever any key column
-		 * is NULL.  That suffix is part of the index row key, so a primary key
-		 * update changes the index row key and must be modeled as DELETE + INSERT
-		 * rather than an in-place UPDATE.
+		 * A base table primary key update can modify hidden index key columns
+		 * even when the user-visible index key columns are unchanged. Non-unique
+		 * indexes use ybidxbasectid as part of the index row key. Unique indexes
+		 * that use NULLS DISTINCT semantics store ybuniqueidxkeysuffix as the base
+		 * table ybctid whenever any key column is NULL.
 		 */
-		if (is_pk_updated && indexData->indisunique &&
-			!indexData->indnullsnotdistinct &&
-			YbUniqueIndexKeyHasNull(indexInfo, slot, estate))
+		if (YbIndexKeyColumnsModifiedByPkeyUpdate(indexRelation, slot,
+												 is_pk_updated))
 		{
 			deleteIndexes = lappend_int(deleteIndexes, i);
 			insertIndexes = lappend_int(insertIndexes, i);
--- a/src/postgres/src/backend/executor/execIndexing.c
+++ b/src/postgres/src/backend/executor/execIndexing.c
@@ -756,6 +756,22 @@ YbExecDoUpdateIndexTuple(ResultRelInfo *resultRelInfo,
 	MemoryContextSwitchTo(oldContext);
 }
 
+static bool
+YbUniqueIndexKeyHasNull(IndexInfo *indexInfo,
+						 TupleTableSlot *slot,
+						 EState *estate)
+{
+	Datum		values[INDEX_MAX_KEYS];
+	bool		isnull[INDEX_MAX_KEYS];
+
+	Assert(indexInfo->ii_Unique);
+
+	GetPerTupleExprContext(estate)->ecxt_scantuple = slot;
+	FormIndexDatum(indexInfo, slot, estate, values, isnull);
+
+	return YbIsAnyIndexKeyColumnNull(indexInfo, isnull);
+}
+
 List *
 YbExecUpdateIndexTuples(ResultRelInfo *resultRelInfo,
 						TupleTableSlot *slot,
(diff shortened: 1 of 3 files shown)


austenLacy:
trigger jenkins

src/postgres/src/backend/executor/execIndexing.c
```diff
@@ -756,6 +756,30 @@ YbExecDoUpdateIndexTuple(ResultRelInfo *resultRelInfo,
...
+	bool		isnull[INDEX_MAX_KEYS];
+	MemoryContext oldContext;
+	bool		result;
+
+	Assert(indexInfo->ii_Unique);
+
+	oldContext = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
+
+	GetPerTupleExprContext(estate)->ecxt_scantuple = slot;
+	FormIndexDatum(indexInfo, slot, estate, values, isnull);
```
karthik-ramanathan-3006: `FormIndexDatum()` can end up doing a lot more work than necessary. So let's avoid it if we can:
our goal is to check if any key column is NULL. We already know the attnums of the key columns.
How about using `slot_attisnull(slot, attnum)` instead?
This way we can also safely remove the memory context switching etc.

If the key column is an expression (ie. the index is an expression index), rather than computing the computing the expression, let's assume that its key columns contain a NULL. This is inefficient, yes but optimizing invocations to `FormIndexDatum()` will require a good amount of refactor and may not be worth the effort.

src/postgres/src/backend/executor/execIndexing.c
```diff
@@ -1018,6 +1044,22 @@ YbExecUpdateIndexTuples(ResultRelInfo *resultRelInfo,
 			continue;
 		}
 
+		/*
+		 * Unique indexes that use NULLS DISTINCT semantics store
```
karthik-ramanathan-3006: My understanding is that we must do DELETE + INSERT of the index tuple if `ybuniqueidxkeysuffix` changes. `ybuniqueidxkeysuffix` changes in the following scenarios:
 1. Change in NULL-ness of the index's key columns (either the columns were NULL before the UPDATE and changed to not-NULL or vice versa or some combination of the two)
 2. Primary key is updated.
 
(1) is already covered by the case above. So, we just have to handle (2) here. Further, (2) depends on whether the index is unique and whether NULLs are treated as distinct.

If my understanding sounds reasonable to you, maybe let's update the code structure to better reflect this?

Func IsIndexKeyColumnsModifiedByPkeyUpdate()
 - false if `is_pk_updated == false`
 - true if index is non-unique
 - false if index is unique and NULLs NOT DISTINCT
 - true is index is unique, NULLs DISTINCT, and key columns contain a NULL.
 - false otherwise

Since (1) is handled by `updatedCols`, we don't have to examine the old tuple.
Just posting my thoughts here, let me know what you think.

karthik-ramanathan-3006 (commented):
Thank you so much for surfacing this bug and also fixing it! I took a pass at the code. Yet to take a deep look at the tests. 
Also, happy to chat on Slack if anything comes up.

src/postgres/src/backend/executor/execIndexing.c
austenLacy: Sounds good to me. I'm definitely not as familiar with what's involved in `FormIndexDatum` so favoring the safety of always assuming NULL makes sense.

src/postgres/src/backend/executor/execIndexing.c
austenLacy: Makes sense. Code reads cleaner that way too.

src/postgres/src/backend/executor/execIndexing.c
austenLacy: Added in https://github.com/yugabyte/yugabyte-db/pull/32126/changes/354e55fa1aeabdb2b88cea440e7ea9cf691155b1

src/postgres/src/backend/executor/execIndexing.c
austenLacy: Added in https://github.com/yugabyte/yugabyte-db/pull/32126/changes/354e55fa1aeabdb2b88cea440e7ea9cf691155b1

src/postgres/src/backend/executor/execIndexing.c
```diff
@@ -756,6 +756,52 @@ YbExecDoUpdateIndexTuple(ResultRelInfo *resultRelInfo,
...
+	{
+		AttrNumber	attnum = indexData->indkey.values[i];
+
+		/*
+		 * Expression key columns have attnum 0. Avoid evaluating the
+		 * expression here; conservatively assume that it can be NULL.
+		 * This consequently leads to a less optimized scenario where
+		 * we do a DELETE+INSERT unnecessarily, but this favors safety
+		 * over speed. It would be possible to use something like
+		 * FormIndexDatum to look up the actual values, but that's an
```
karthik-ramanathan-3006, line 794: Let's elaborate that it is an expensive operation because it would compute the actual expression which could be arbitrarily complex.

src/postgres/src/backend/executor/execIndexing.c
```diff
@@ -1018,6 +1062,21 @@ YbExecUpdateIndexTuples(ResultRelInfo *resultRelInfo,
 			continue;
 		}
 
+		/*
+		 * A base table primary key update can modify hidden index key columns
```
karthik-ramanathan-3006: Let's use the terminology `system-defined` and `user-defined` instead.

src/postgres/src/test/regress/sql/yb.orig.update_optimize_index_updates.sql
```diff
@@ -260,3 +260,27 @@ EXPLAIN (VERBOSE) /*+ IndexOnlyScan(t_simple multi_include) */ SELECT v1, v2, v3
 EXPLAIN (VERBOSE) /*+ IndexScan(multi_include) */ SELECT * FROM t_simple ORDER BY (v1, v2);
 /*+ IndexScan(multi_include) */ SELECT * FROM t_simple ORDER BY (v1, v2);
 SELECT * FROM t_simple;
+
+--
+-- Primary key updates on unique indexes with NULL key columns must not use
+-- in-place index updates because ybuniqueidxkeysuffix stores the base ybctid.
+--
+SET yb_enable_inplace_index_update TO on;
```
karthik-ramanathan-3006, line 267: Could we expand test coverage to include the following:
 - Non-expression unique index: index row keys change from NULL to not-NULL
 - Non-expression unique index: index row keys change from not-NULL to NULL
 - Non-expression unique index: index row keys are NULL; primary key update
 - Non-expression unique index: some index row keys are NULL; update makes other row keys NULL
 - Similarly for a non-expression unique index with NULLS NOT DISTINCT
 - Similarly for expression-based indexes. For the multi-key cases, an index can have multiple expressions (`CREATE INDEX ON tbl (f1(a, b), f2(c, d), ..)`). This gives us adequate protection in case we decide to optimize the expression index case in the future.  
 - Partial index: same tests as above with the additional caveat that the UPDATE flips set membership (before UPDATE the index row was a part of the partial index; after UPDATE the index row isn't or vice versa)

The last couple of tests are meant to exercise the ordering of the conditions in `YbExecUpdateIndexTuples()`.

karthik-ramanathan-3006 (commented):
Overall the solution looks good to me.

I've also filed a tracking bug (#32220). If you could add this to the summary, that would be great.

src/postgres/src/test/regress/sql/yb.orig.update_optimize_index_updates.sql
austenLacy: Good idea. Done in https://github.com/yugabyte/yugabyte-db/pull/32126/commits/b727bb7ee86ec2f096d5b4e171ec90a0fe786485

austenLacy:
trigger jenkins

src/postgres/src/test/regress/sql/yb.orig.update_optimize_index_updates.sql
```diff
@@ -260,3 +260,271 @@ EXPLAIN (VERBOSE) /*+ IndexOnlyScan(t_simple multi_include) */ SELECT v1, v2, v3
...
+DROP TABLE IF EXISTS t_unique_null_pk_update;
+CREATE TABLE t_unique_null_pk_update (
```
karthik-ramanathan-3006: Seems like we can skip this test as it is covered by the tests below.

src/postgres/src/test/regress/sql/yb.orig.update_optimize_index_updates.sql
```diff
@@ -260,3 +260,271 @@ EXPLAIN (VERBOSE) /*+ IndexOnlyScan(t_simple multi_include) */ SELECT v1, v2, v3
...
+  h INT,
+  r INT,
+  k INT,
+  a INT,
+  b INT,
+  active BOOLEAN DEFAULT true,
+  PRIMARY KEY ((h) HASH, r ASC, k ASC)
+);
+CREATE UNIQUE INDEX NONCONCURRENTLY t_unique_key_update_uq
+ON t_unique_key_update (h HASH, a ASC, b ASC);
```
karthik-ramanathan-3006: Does `h` need to be in the index? If there is no specific reason, let's drop `h` from the index. Here and below.

src/postgres/src/test/regress/sql/yb.orig.update_optimize_index_updates.sql
```diff
@@ -260,3 +260,271 @@ EXPLAIN (VERBOSE) /*+ IndexOnlyScan(t_simple multi_include) */ SELECT v1, v2, v3
...
+INSERT INTO t_unique_key_update (h, r, k, a, b) VALUES
+  (1, 1, 0, NULL, 10),
+  (1, 2, 0, 20, 20),
+  (1, 3, 0, NULL, 30),
+  (1, 4, 0, NULL, 40);
+-- Non-expression unique index: key changes from NULL to not-NULL.
+UPDATE t_unique_key_update SET a = 10 WHERE h = 1 AND r = 1;
```
karthik-ramanathan-3006, line 288: Could we run the queries with `EXPLAIN (ANALYZE, DIST)` so that we assert the number of Write Requests?

src/postgres/src/test/regress/sql/yb.orig.update_optimize_index_updates.sql
```diff
@@ -260,3 +260,271 @@ EXPLAIN (VERBOSE) /*+ IndexOnlyScan(t_simple multi_include) */ SELECT v1, v2, v3
...
+-- Non-expression unique index: key changes from not-NULL to NULL.
+UPDATE t_unique_key_update SET a = NULL WHERE h = 1 AND r = 2;
+-- Non-expression unique index: key is NULL; primary key update.
+UPDATE t_unique_key_update SET k = 1 WHERE h = 1 AND r = 3 AND k = 0;
+-- Non-expression unique index: some keys are NULL; another key becomes NULL.
+UPDATE t_unique_key_update SET b = NULL WHERE h = 1 AND r = 4;
+SELECT yb_index_check('t_unique_key_update_uq'::regclass);
```
karthik-ramanathan-3006, line 298: Let's run a `SELECT * FROM tbl ORDER BY h, r, k` at the end just as a sanity check. This doesn't check the index, but the index check should cover that.

src/postgres/src/test/regress/sql/yb.orig.update_optimize_index_updates.sql
```diff
@@ -260,3 +260,271 @@ EXPLAIN (VERBOSE) /*+ IndexOnlyScan(t_simple multi_include) */ SELECT v1, v2, v3
...
+CREATE TABLE t_partial_unique_key_update_nnd (
+  h INT,
+  r INT,
+  k INT,
+  a INT,
+  b INT,
+  active BOOLEAN,
+  PRIMARY KEY ((h) HASH, r ASC, k ASC)
+);
+CREATE UNIQUE INDEX NONCONCURRENTLY t_partial_unique_key_update_nnd_uq
```
karthik-ramanathan-3006: I feel we can skip the remaining partial index tests as all the queries flip the row's membership in the index. This condition will be always be evaluated first, so it shouldn't matter what other modifiers the index has (NULLS NOT DISTINCT, unique vs non-unique etc)
Perhaps, we can retain one of the partial + expression based index tests.
I'll leave the call with you.

karthik-ramanathan-3006 (approved):
Overall LGTM. Thank you for making these changes.

The CSI test results look good.
The only thing that CSI complains about is the PR title format. I believe it should be:
```
[#32220] YSQL: Fixes unique index consistency bug when yb_enable_inplace_index_update is on
```

nit: You can omit "yb_enable_inplace_index_update is on" from the title as that's the default. If you can work "row has NULL key columns" into the title, that'd be great as well.

austenLacy:
trigger jenkins

karthik-ramanathan-3006:
trigger jenkins

Issue #32128, opened 2026-06-10
YSQL: planning under the cost model issues one pg_statistic catalog read per table column on a cold backend


kai-franz:
### Problem

On a cold YSQL backend with the cost model enabled (`yb_enable_base_scans_cost_model`), planning *any* query that scans a table reads one `pg_statistic` row per **table** column — not per selected column. To cost a scan, the cost model estimates the relation's full-row width via `get_rel_data_width()` (reached through `yb_get_relation_data_width()`), which calls `get_attavgwidth()` for every column, because a YB scan fetches whole rows from DocDB. Each `SearchSysCache3(STATRELATTINH, ...)` miss is satisfied by an individual one-row scan — i.e. one catalog RPC per column. An N-column table therefore costs ~N separate catalog RPCs the first time a freshly connected backend plans a query against it, and that burst is repaid by every fresh backend under connection churn.

This is projection-independent. Measured on a fresh backend against a 50-column analyzed table, counting "Catalog Read Requests" from `EXPLAIN (ANALYZE, DIST)`:

| query | catalog reads | pg_statistic reads |
|---|---|---|
| `SELECT *` | 83 | 50 |
| `SELECT c1, c2` | 83 | 50 |
| `SELECT c1` | 83 | 50 |
| `SELECT count(*)` | 88 | 50 |
| `SELECT c1 WHERE c2 > 5` | 88 | 50 |

Even a single-column `SELECT` — or `count(*)`, which references no column at all — reads all 50 rows, because the scan's estimated cost depends on the full row width either way.

### Proposed fix

Warm the cache for the whole relation in one prefix scan instead of one point lookup per column. `STATRELATTINH` is keyed `(starelid, staattnum, stainherit)`, so a list search on the leading key (`starelid`) returns all of the relation's stat rows in a single RPC and instantiates each as a positive catcache entry; the subsequent per-column point lookups then hit warm cache with no further RPCs. This mirrors how `RelationBuildTupleDesc()` already bulk-loads `pg_attribute`.

A new GUC `yb_prefetch_relation_statistics` (default on), gated on `yb_enable_base_scans_cost_model`, drives this from `get_relation_info()`. Plans and cost estimates are unchanged; only the number and shape of catalog reads differ.

PR #32159, opened 2026-06-12
[BACKPORT 2025.1][#32128] YSQL: Batch pg_statistic lookups during planning

kai-franz:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/32159/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

Summary:
On a cold YSQL backend with the cost model enabled
(`yb_enable_base_scans_cost_model`), planning any query that scans a wide table
reads one `pg_statistic` row per *table* column -- not per selected column. The
scan cost functions (e.g. `yb_cost_seqscan()`) need the relation's full-row
width to size the storage read: they estimate the number of DocDB blocks
scanned as `tuples * width / block_size`, the YB analog of PostgreSQL's
`seq_page_cost * relpages`, and a scan reads every row's on-disk bytes out of
RocksDB regardless of what the query projects. That width comes from
`yb_get_relation_data_width()` -> `get_rel_data_width()`, which calls
`get_attavgwidth()` for *every* attribute. Each `SearchSysCache3(STATRELATTINH,
...)` miss is an individual one-row scan -- one RPC per column -- so an N-column
table costs ~N separate catalog RPCs the first time a freshly connected backend
plans it, regardless of how many columns the query projects. (Projection is not
ignored: it feeds a separate, projected-only result width that drives the
DocDB->Postgres transfer cost; the per-column statistics are read either way.)

This is projection-independent. Measured on a fresh backend against a
50-column analyzed table, counting "Catalog Read Requests" from
EXPLAIN (ANALYZE, DIST) with the prefetch off then on (the difference is the
pg_statistic reads the prefetch collapses into one list read):

    query                    reads off -> on   pg_statistic reads
    SELECT *                  83 -> 33          50
    SELECT c1, c2             83 -> 33          50
    SELECT c1                 83 -> 33          50
    SELECT count(*)           88 -> 38          50
    SELECT c1 WHERE c2 > 5    88 -> 38          50

Even a single-column SELECT -- or count(*), which references no column at
all -- reads all 50 rows, because the scan's estimated cost depends on the
full row width either way. This burst is repaid by every freshly connected
backend under connection churn.

Warm the cache for the whole relation in one prefix scan instead. Since
`STATRELATTINH` is keyed (`starelid`, `staattnum`, `stainherit`), a list search on
the leading key returns all of the relation's stat rows in a single RPC and
instantiates each as an individual positive catcache entry, so the
subsequent per-column point lookups hit warm cache with no further RPCs.
This mirrors how `RelationBuildTupleDesc()` already bulk-loads `pg_attribute`.

The new helper `yb_prefetch_column_stats()` is invoked once per base
relation in get_relation_info(), gated on the new GUC
`yb_prefetch_column_statistics` (default true) and on
`yb_enable_base_scans_cost_model` (the all-column width lookups, which are the
source of the burst, only occur when the cost model is enabled). Only
relation kinds that can carry column statistics are prefetched.

Plans and cost estimates are unchanged; only the number and shape of
catalog reads differ. On a never-analyzed relation no stat rows exist, so
the list lookup simply misses once and the per-column lookups fall back to
type-based widths -- a bounded one-extra-RPC cost versus the large win on
analyzed tables.

Original commit: 211a32f56b6fa6295b8a15ba357fd1c378ebe7ed / D54378

Test Plan:
Built fastdebug and ran the new test class against a mini-cluster; all green:

  ./yb_build.sh fastdebug --java-test \
      'org.yb.pgsql.TestPgPrefetchColumnStatistics'
  Tests run: 3, Failures: 0, Errors: 0.
  - testPrefetchReducesCatalogReadRequests: the first cold-backend plan of
    "SELECT *" on a 50-column analyzed table reports 83 Catalog Read Requests
    (EXPLAIN ANALYZE DIST) with the GUC off vs 33 with it on -- the per-column
    pg_statistic point lookups collapse into a single list read.
  - testNoEffectWhenCostModelDisabled: the prefetch saves 50 reads with the
    cost model on and 0 with it off (it is gated on the cost model).
  - testPlansAndCostsUnchanged: the full EXPLAIN output is byte-identical with
    the GUC on vs off across several query shapes.

Reviewers: mtakahara

Reviewed By: mtakahara

Subscribers: yql, smishra

Differential Revision: https://phorge.dev.yugabyte.com/D54378

### DIFF
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgPrefetchColumnStatistics.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestPgPrefetchColumnStatistics.java
@@ -1,0 +1,223 @@
+// Copyright (c) YugabyteDB, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied.  See the License for the specific language governing permissions and limitations
+// under the License.
+//
+
+package org.yb.pgsql;
+
+import static org.yb.AssertionWrappers.assertEquals;
+import static org.yb.AssertionWrappers.assertGreaterThanOrEqualTo;
+import static org.yb.AssertionWrappers.assertLessThan;
+import static org.yb.AssertionWrappers.assertLessThanOrEqualTo;
+import static org.yb.AssertionWrappers.assertTrue;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.Statement;
+
+import org.json.JSONArray;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.yb.YBTestRunner;
+
+/**
+ * Tests for the yb_prefetch_column_statistics GUC.
+ *
+ * On a cold backend, planning any query that scans a wide table reads
+ * pg_statistic one column at a time: to cost the scan the planner estimates the
+ * relation's full-row width, calling get_attavgwidth() for every column --
+ * regardless of how many columns the query selects -- and each miss is an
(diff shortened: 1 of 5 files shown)


kai-franz:
Trigger Jenkins

---

kai-franz:
Superseded: backporting via a Phorge arc diff (tools/bin/backport.sh), the standard YB workflow, rather than a GitHub PR. Closing.

---

Issue #32204, opened 2026-06-16
[YSQL] OTLP HTTPS trace export fails on Ubuntu because CA bundle path defaults to RHEL location


egladysh:
Jira Link: [DB-21891](https://yugabyte.atlassian.net/browse/DB-21891)
### Description

Yugabyte OpenTelemetry trace export over HTTPS fails on Ubuntu hosts because the OpenTelemetry C++ HTTP exporter tries to load the CA bundle from the RHEL-style path:
`/etc/pki/tls/certs/ca-bundle.crt`
On our Ubuntu 22.04, the system CA bundle exists at:
`/etc/ssl/certs/ca-certificates.crt`

As a result, YSQL tracing initializes and creates spans, but exporting them to an HTTPS OTLP endpoint fails before the spans leave the host.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-21891]: https://yugabyte.atlassian.net/browse/DB-21891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #32222, opened 2026-06-16
[#32204] DistributedTracing: Add gflags and platform fallback for OTLP HTTPS CA certificate

egladysh:
## Summary

OTLP HTTPS trace export relied on the OpenTelemetry exporter's built-in CA
bundle path, which defaults to a RHEL location. On distributions that store
the CA bundle elsewhere (e.g. Debian/Ubuntu at
`/etc/ssl/certs/ca-certificates.crt`), HTTPS trace export failed because the
exporter could not find a trust store.

This change adds two non-runtime gflags to make the trust store explicit and
adds a platform fallback:

- `otel_ssl_ca_cert_path`: CA certificate bundle path for OTLP HTTPS trace
  export. When set, it takes precedence.
- `otel_ssl_ca_cert_string`: inline CA certificate bundle contents, used only
  when no path is configured.

When neither flag is set, YugabyteDB now probes a list of known platform CA
bundle locations (Debian/Ubuntu, RHEL/CentOS/AlmaLinux, OpenSUSE/SLES) and uses
the first one that exists, so HTTPS export works out of the box across the
common Linux distributions. The selected CA bundle path is logged at startup.

## Test plan

- [ ] `DistTraceCaCertPathTest.UsesFirstExistingPath` — verifies the first
      existing path in the candidate list is selected.
- [ ] `DistTraceCaCertPathTest.ReturnsEmptyWhenNoPathExists` — verifies
      `std::nullopt` when none of the candidate paths exist.


src/yb/util/dist_trace.cc
```diff
@@ -112,7 +131,25 @@ auto CreateExporter() {
...
   opts.url = FLAGS_otel_collector_traces_endpoint;
   opts.content_type = otlp_exporter::HttpRequestContentType::kBinary;
+
+  if (!FLAGS_otel_ssl_ca_cert_path.empty()) {
+    opts.ssl_ca_cert_path = FLAGS_otel_ssl_ca_cert_path;
+    opts.ssl_ca_cert_string.clear();
+  } else if (!FLAGS_otel_ssl_ca_cert_string.empty()) {
+    opts.ssl_ca_cert_path.clear();
+    opts.ssl_ca_cert_string = FLAGS_otel_ssl_ca_cert_string;
+  }
```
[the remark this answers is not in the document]
egladysh: It's not true that string fields are already empty by default. The constructor may chose to populate them. Rejecting this suggestion.

src/yb/yql/pgwrapper/dist_trace-test.cc
```diff
@@ -122,6 +122,32 @@ struct Trace {
...
+  const auto missing_path = GetTestPath("missing-ca.pem");
+  const auto first_existing_path = GetTestPath("first-existing-ca.pem");
+  const auto second_existing_path = GetTestPath("second-existing-ca.pem");
+  CreateFile(first_existing_path);
+  CreateFile(second_existing_path);
+
+  ASSERT_EQ(
+      *dist_trace::FindFirstExistingFile(
+          {missing_path, first_existing_path, second_existing_path}),
+      first_existing_path);
```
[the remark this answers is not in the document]
egladysh: Fixed

egladysh:
trigger jenkins

src/yb/util/dist_trace.cc
```diff
@@ -112,7 +131,25 @@ auto CreateExporter() {
...
+
+  if (opts.ssl_ca_cert_path.empty() && opts.ssl_ca_cert_string.empty()) {
```
mangalaman93: if both are empty, shouldn't we disable SSL and not use SSL while talking to the OTLP collector?

mangalaman93 (changes requested):

src/yb/util/dist_trace.cc
egladysh: I don’t think empty CA flags should imply “disable SSL.” These flags only control which CA bundle we pass for HTTPS certificate verification. They are not the TLS enable/disable switch. TLS vs plaintext is determined by the collector endpoint scheme: `https://...`, `http://...`, etc.. I think these are separate concerns. This PR is to just configure trust data for HTTPS verification.
I don't think we need a way to just disable HTTPS endpoints period?

mangalaman93: These flags are common for both HTTP and grpc. While we may not support grpc protocol today, but when we do, we cannot specify TLS is required in the URL for grpc. We should have an explicit requirement of one of these two flags for TLS to be setup similar to the OTEL SDK.

egladysh: @mangalaman93 I don't think that this PR configures OTLP/gRPC. It only uses the OTLP/HTTP exporter. For this code path, the endpoint URL scheme is the transport-security selector (`https://` vs `http://`). These two flags only configure CA trust material for HTTPS verification. 
I understand these flags may be shared by HTTPS and future gRPC support, but I don’t think that means they should become the TLS enable/disable signal. They configure CA trust material, which is orthogonal to whether the transport is secure. I'd suggest for you to create a separate issue to address gRPC concerns?
For example OTel has a standard option to enable insecure gRPC, etc.. I'd argue that these CA flags should be trust-material configuration, not the implicit TLS enable switch.

mangalaman93: I researched a little more how cpp OTEL SDK works and it seems like it uses libcurl under the hood that picks up the system cert pool automatically when the scheme is https. If that is true, we can skip this part where both flags are empty. Happy to find some references for you if needed.

egladysh: I'm not sure what happens under the hood exactly but in YB OpenTelemetry exporter's built-in CA
bundle path defaults to a RHEL location (`/etc/pki/tls/certs/ca-bundle.crt`) on Ubuntu as described in the PR and that's wrong for other platforms like Ubuntu. The PR fixes some of the platform defaults as well as allows to configure it manually.

egladysh: Maybe we can take a look at how libcurl picks fallback certificate bundle path. It seems to be hardcoded to RHEL if it's compiled on RH-based system.

mangalaman93: Did you have a chance to take a look?

egladysh: I tested it. That was reason that the ticket was created in the first place. I thought you've been looking for references in libcurl as you suggested? I just did a quick search online. It seems like CURL_CA_BUNDLE is the macro with the default that is set at compile time.

mangalaman93: In that case, we can ignore the case when these two flags are not set. libcurl should correctly pick the default cert bundle in that case. Is that the conclusion?

egladysh: No. YugabyteDB’s x86 Linux distribution is generic like `linux-x86_64` tarball, and YB docs describe using that package across different Linux OS families/distros/Ubuntu. (“You can use Ubuntu as well”, docs/content/stable/deploy/public-clouds/aws/manual-deployment.md). Also YB packages it as `lib/yb-thirdparty/libcurl.so`.

egladysh: That packaged libcurl has a hardcoded ca path.

mangalaman93: This makes sense. Let me check this internally and get back to you. apologies for the delay.

mangalaman93: One more question I have here is that, if I only specify http and both flags are specified as empty, will that cause any problems specifying the CA cert or is it handled by libcurl correctly? It would seem odd though that we are specifying a CA cert here in that case.

mangalaman93: one more question I have is, what is your use case? Do you use a self signed CA's cert for your otel collector or a public CA issued cert?

egladysh: No worries.

Question 1: No, I don't see any problems.
Question 2: We'll probably have different use cases. This change should cover all that I can think of at this time like https on Ubuntu.

mangalaman93: If the use case is different, then, I'd recommend to divide this into two PRs. I am thinking that by specifying the system cert path in the parameter, we can solve most problems.

egladysh: Sorry, I don't understand your suggestion about two PRs? I think the change is small and self-contained. Can you explain reasons for your strong pushback against working around the hardcoded path in libcurl?

mangalaman93: I am not comfortable having hardcoded list of paths here. This may be incomplete and someone needs to ensure that they remain up to date. You could move this part out into another PR where we can have a longer discussion and figure out what is the right thing to do. Rest of the code generally looks good and can be merged quickly.

egladysh: The path is already hardcoded in YB just for RHEL and it's absolutely incomplete right now and it's just wrong on Ubuntu. This change allows YB opentel to work on Ubuntu the same way as it does on RHEL currently. We'd like it to work it consistently.

mangalaman93: agreed but that is hard coded in libcurl and libcurl is going to manage these paths much better than we can. Unless you absolutely need it right away, I would recommend that we give it more thought. I plan to have more discussions internally as well before we add this code. Until then, these flags can serve us well and for ubuntu too.

egladysh: What do you mean by "libcurl is going to manage these paths"? It doesn't do that. The path is wrong.

egladysh: Yeah, we'd like to get it done in this PR and close the issue.

mangalaman93: libcurl the library will update this path if this ever changes and when we upgrade our library version that we depend on, it will get upgraded on our side too without us figuring out what the right path is.

For now, we can remove these paths from this PR and can pass them as ssl_ca_cert_path parameter at the time of deployment. Happy to elaborate more if that doesn't make sense.

egladysh: I think if libcurl ever updates the path (I doubt it), we could remove the workaround. As of now, it's really a bogus packaging of YB. You can consider creating different packages of libcurl for different environments? Anyhow I think we discussed the issue to the floor. I'll remove the workaround to pass your block and move forward with it.

egladysh:
@mangalaman93 The defaults should be gone. Could you take a look when you have a chance?

mangalaman93 (approved):

mangalaman93:
trigger jenkins

PR #32242, opened 2026-06-17
most xCluster replication requires no rewrite of associated indexes

mdbridge:
A recent customer incident may have been partially caused by the customer not realizing that just because a DDL doesn't rewrite the main table doesn't mean it doesn't rewrite the associated index as well, which we forbid in most modes of access to replication.

Here we make the documentation a little bit more specific in hopes of avoiding this in the future.


ddhodge (approved):
LGTM

PR #32513, opened 2026-07-07
[#9624] DocDB: Suggest closest matches for misspelled yb-admin commands

### LINKED ISSUE #9624: [usability] Allow prefix matching for yb-admin commands
Currently when yb-admin command is slightly misspelled, e.g. list_snapshot_schedule instead of list_snapshot_schedules, the user would be given the whole list of commands.

User experience can be improved by using the (partial) command as prefix.

If there is only one match, that match would be picked.
In case there are multiple candidates for the given prefix, the list of commands would be printed.

ag26jan:
## Summary

`yb-admin` looks up the operation name given on the command line with an exact match against the
list of registered commands. If the operation is slightly misspelled (e.g.
`list_snapshot_schedule` instead of `list_snapshot_schedules`), the user was shown the
`Invalid operation` error followed by the entire list of ~130 supported commands, with no hint
about what they may have meant.

This PR makes that error helpful in two ways:

**1. Suggest the closest commands.**
- Prefix match first: registered (non-hidden) commands that have the typed operation as a prefix
  are suggested.
- Edit-distance fallback: if nothing is a prefix, fall back to Levenshtein (fuzzy) matching so a
  typo anywhere in the name - a transposition, or a wrong or missing leading character - still
  resolves to the closest command(s). The tolerance scales with the length of the operation, and
  only the closest-scoring tier is shown to keep the list short.

```
$ yb-admin --master_addresses ... list_snapshot_schedule    # prefix
Invalid operation: list_snapshot_schedule
Did you mean one of these?
  list_snapshot_schedules

$ yb-admin --master_addresses ... ist_tables                # missing leading char (fuzzy)
Invalid operation: ist_tables
Did you mean one of these?
  list_tables

$ yb-admin --master_addresses ... list_tabels               # transposition (fuzzy)
Invalid operation: list_tabels
Did you mean one of these?
  list_tables
  list_tablets
```

**2. Stop dumping the full command list on an invalid operation.** That wall of ~130 commands was
the original complaint in the issue. The invalid-operation path now shows only the targeted
suggestions plus a one-line hint, and returns a non-`InvalidArgument` status so `main()` no longer
appends the full usage. Running `yb-admin` with no operation at all still prints the full usage as
help.

The command is never run automatically based on a match, even an unambiguous prefix, to avoid
surprising behavior for scripts that pass an already-invalid command (per maintainer guidance on
the issue).

Fixes #9624. Subsumes #32516 (fuzzy / nearest-neighbor matching), which is closed as folded in
here.

## Test plan

- [x] `AdminCliTest.InvalidOperationSuggestsClosestCommands` - one test covering the whole
  "did you mean" help path:
  - prefix matches: an unambiguous single-command prefix, and an ambiguous prefix that lists
    every candidate;
  - fuzzy / edit-distance matches: a transposition (`list_tabels`) and a missing leading
    character (`ist_tables`);
  - that an empty operation and a far-off garbage string suggest nothing; and
  - that an invalid operation no longer prints the full command list on either stream, while
    running with no operation at all still does.
- [x] Built `yb-admin` and `yb-admin-test` (release) locally; the test passes, and manually
  verified the examples above. The test needs no running cluster since the check happens before
  `yb-admin` connects to the master.

### DIFF
--- a/src/yb/tools/yb-admin-test.cc
+++ b/src/yb/tools/yb-admin-test.cc
@@ -206,65 +206,12 @@ class AdminCliTest : public AdminTestBase {
   TmpDirProvider tmp_dir_;
 };
 
-// Verify that a misspelled operation gets suggested the commands it is a prefix of, without
-// needing a running cluster (the check happens before yb-admin connects to the master).
-TEST_F(AdminCliTest, InvalidOperationSuggestsPrefixMatches) {
-  const auto exe_path = GetAdminToolPath();
-  constexpr auto kUnusedMasterAddress = "127.0.0.1:0";
-  std::string error;
-
-  // Typo that is an unambiguous prefix of a single command.
-  ASSERT_NOK(Subprocess::Call(
-      ToStringVector(
-          exe_path, "--master_addresses", kUnusedMasterAddress, "list_snapshot_schedule"),
-      /* output */ nullptr, &error));
-  ASSERT_STR_CONTAINS(error, "Did you mean one of these?");
-  ASSERT_NE(error.find("list_snapshot_schedules"), std::string::npos);
-
-  // Prefix that matches multiple commands should list every candidate.
-  ASSERT_NOK(Subprocess::Call(
-      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "list_table"), nullptr,
-      &error));
-  ASSERT_NE(error.find("list_tables"), std::string::npos);
-  ASSERT_NE(error.find("list_tablets"), std::string::npos);
-
-  // An empty operation is a prefix of every command, but should not list all of them.
-  ASSERT_NOK(Subprocess::Call(
-      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, ""), nullptr, &error));
-  ASSERT_EQ(error.find("Did you mean one of these?"), std::string::npos);
-}
-
-// Verify that a typo that is not a prefix of any command still gets suggestions via fuzzy
-// (edit-distance) matching, e.g. a transposition or a wrong/missing leading character.
-TEST_F(AdminCliTest, InvalidOperationSuggestsFuzzyMatches) {
-  const auto exe_path = GetAdminToolPath();
-  constexpr auto kUnusedMasterAddress = "127.0.0.1:0";
-  std::string error;
-
-  // Transposition in the middle of the command ("tabels" instead of "tables").
-  ASSERT_NOK(Subprocess::Call(
-      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "list_tabels"),
-      /* output */ nullptr, &error));
-  ASSERT_NE(error.find("Did you mean one of these?"), std::string::npos);
-  ASSERT_NE(error.find("list_tables"), std::string::npos);
-
-  // Missing leading character ("ist_tables" instead of "list_tables") is not a prefix but is one
-  // edit away.
-  ASSERT_NOK(Subprocess::Call(
-      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "ist_tables"), nullptr,
-      &error));
-  ASSERT_NE(error.find("list_tables"), std::string::npos);
-
-  // A far-off garbage string is beyond the edit-distance tolerance, so nothing is suggested.
-  ASSERT_NOK(Subprocess::Call(
-      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "zzzzzzzzzzzzzzzzzz"),
-      nullptr, &error));
-  ASSERT_EQ(error.find("Did you mean one of these?"), std::string::npos);
-}
-
-// Verify that an invalid operation no longer dumps the full command list (the original complaint
-// in the issue), while running with no operation at all still prints the full usage as help.
-TEST_F(AdminCliTest, InvalidOperationDoesNotListAllCommands) {
+// Verify the "did you mean" help for a misspelled operation, none of which needs a running
+// cluster (the operation is checked before yb-admin connects to the master): prefix matches,
+// fuzzy (edit-distance) matches, and that an invalid operation no longer dumps the full command
+// list (the original complaint in the issue) while running with no operation still prints the
+// full usage as help.
+TEST_F(AdminCliTest, InvalidOperationSuggestsClosestCommands) {
   const auto exe_path = GetAdminToolPath();
   constexpr auto kUnusedMasterAddress = "127.0.0.1:0";
   // This marker only appears in the full usage/command listing (which is printed to stdout).
@@ -272,21 +219,54 @@ TEST_F(AdminCliTest, InvalidOperationDoesNotListAllCommands) {
   std::string output;
   std::string error;
 
-  // An invalid operation with suggestions prints the suggestions and a hint on stderr, but must
-  // not dump the full command list on either stream.
+  // Prefix match: an unambiguous typo suggests the single command it is a prefix of.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(
+          exe_path, "--master_addresses", kUnusedMasterAddress, "list_snapshot_schedule"),
+      /* output */ nullptr, &error));
+  ASSERT_STR_CONTAINS(error, "Did you mean one of these?");
+  ASSERT_STR_CONTAINS(error, "list_snapshot_schedules");
+
+  // Prefix match: a prefix of several commands lists every candidate and a hint, and must not dump
+  // the full command list on either stream.
   ASSERT_NOK(Subprocess::Call(
       ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "list_table"), &output,
       &error));
-  ASSERT_NE(error.find("Did you mean one of these?"), std::string::npos);
-  ASSERT_NE(error.find("list_tables"), std::string::npos);
-  ASSERT_NE(error.find("to see all available operations"), std::string::npos);
-  ASSERT_EQ(output.find(kFullUsageMarker), std::string::npos);
-  ASSERT_EQ(error.find(kFullUsageMarker), std::string::npos);
+  ASSERT_STR_CONTAINS(error, "Did you mean one of these?");
+  ASSERT_STR_CONTAINS(error, "list_tables");
+  ASSERT_STR_CONTAINS(error, "list_tablets");
+  ASSERT_STR_CONTAINS(error, "to see all available operations");
+  ASSERT_STR_NOT_CONTAINS(output, kFullUsageMarker);
+  ASSERT_STR_NOT_CONTAINS(error, kFullUsageMarker);
+
+  // Fuzzy match: a transposition ("tabels" instead of "tables") is not a prefix but is close.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "list_tabels"), nullptr,
+      &error));
+  ASSERT_STR_CONTAINS(error, "Did you mean one of these?");
+  ASSERT_STR_CONTAINS(error, "list_tables");
+
+  // Fuzzy match: a missing leading character ("ist_tables") is one edit away from "list_tables".
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "ist_tables"), nullptr,
+      &error));
+  ASSERT_STR_CONTAINS(error, "list_tables");
+
+  // An empty operation is a prefix of every command, but should not list all of them.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, ""), nullptr, &error));
+  ASSERT_STR_NOT_CONTAINS(error, "Did you mean one of these?");
+
+  // A far-off garbage string is beyond the edit-distance tolerance, so nothing is suggested.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "zzzzzzzzzzzzzzzzzz"),
+      nullptr, &error));
+  ASSERT_STR_NOT_CONTAINS(error, "Did you mean one of these?");
 
   // Running with no operation at all should still print the full usage as help on stdout.
   ASSERT_NOK(Subprocess::Call(
       ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress), &output, &error));
-  ASSERT_NE(output.find(kFullUsageMarker), std::string::npos);
+  ASSERT_STR_CONTAINS(output, kFullUsageMarker);
 }
 
 // Test yb-admin config change while running a workload.
--- a/src/yb/tools/yb-admin-test.cc
+++ b/src/yb/tools/yb-admin-test.cc
@@ -228,16 +228,65 @@ TEST_F(AdminCliTest, InvalidOperationSuggestsPrefixMatches) {
   ASSERT_NE(error.find("list_tables"), std::string::npos);
   ASSERT_NE(error.find("list_tablets"), std::string::npos);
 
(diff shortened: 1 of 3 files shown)


ag26jan:
Thanks for the review. Addressed the feedback in 5e6cbd921b5:

- **Empty-operation guard**: `yb-admin ""` treated the empty string as a prefix of every command and printed the entire list. The prefix scan is now wrapped in a non-empty check, with a test case covering it.
- **Temporary vector / print-as-you-go**: kept as-is intentionally. This is an error path that immediately returns `kInvalidArguments`, so the allocation cost is negligible, and collecting into a small `vector` keeps the logic readable. It's also the structure the planned follow-up (edit-distance suggestions, #32516) builds on.

src/yb/tools/yb-admin_cli.cc
```diff
@@ -472,6 +472,28 @@ Status ClusterAdminCli::Run(int argc, char** argv) {
...
   if (cmd == command_indexes_.end()) {
     cerr << "Invalid operation: " << op << endl;
+
+    // The operation name is not an exact match, so look for commands that have it as a prefix,
+    // e.g. "list_snapshot_schedule" should point the user at "list_snapshot_schedules". An empty
+    // operation is a prefix of every command, so skip it instead of listing all commands.
+    vector<string> candidates;
+    if (!op.empty()) {
+      for (auto it = command_indexes_.lower_bound(op); it != command_indexes_.end(); ++it) {
+        if (!boost::starts_with(it->first, op)) {
```
balajisubramanian-yugabyte: This works for prefix. But if the first character is misspelt, it wont return anything, should we use some nearest neighbor to figure out the closest command for the given operation ?

ag26jan: @balajisubramanian-yugabyte Thanks for taking this for review. Yes, this was planned as part of this sub-task : ( #32516)

Now since you already pointed I think it's worth to include the fuzzy logic in this commit, what you think?

ag26jan:
@balajisubramanian-yugabyte Also I am more inclined to revise this commit. Because currently what I am noticing is two things: 

**Example 1:**

```
./build/latest/bin/yb-admin --master_addresses 192.0.2.62 ist_table
Invalid operation: ist_table
yb-admin: ./build/latest/bin/yb-admin [--master_addresses server1:port,server2:port,server3:port,...]  [--timeout_ms <millisec>] [--certs_dir_name <dir_name>]
  [--flagfile <path/to/master/conf/server.conf>]
  <operation>

Tip: Use --flagfile with the master's server.conf to automatically pick up
master_addresses and certs_dir, avoiding manual flag entry.
Example: ./build/latest/bin/yb-admin --flagfile master/conf/server.conf list_all_masters

<operation> must be one of:
 1. change_config <tablet_id> <ADD_SERVER|REMOVE_SERVER> <peer_uuid> [PRE_VOTER|PRE_OBSERVER]
 2. list_tablet_servers <tablet_id>
 3. backfill_indexes_for_table <table>
 4. list_tables [include_db_type] [include_table_id] [include_table_type]
 5. list_tables_with_db_types
 6. list_tablets <table> [<max_tablets>] (default 10, set 0 for max) [JSON] [include_followers]
 7. modify_table_placement_info <table> <placement_info> <replication_factor> [<placemen
```

This falls back to a long list of command, that's expected as fuzzy logic is not there once we have that it will look nice.

****Example 2:** (This is where I am not satisfied)**

```
./build/latest/bin/yb-admin --master_addresses 192.0.2.62 list_table
Invalid operation: list_table
Did you mean one of these?
  list_tables
  list_tables_with_db_types
  list_tablet_servers
  list_tablets
yb-admin: ./build/latest/bin/yb-admin [--master_addresses server1:port,server2:port,server3:port,...]  [--timeout_ms <millisec>] [--certs_dir_name <dir_name>]
  [--flagfile <path/to/master/conf/server.conf>]
  <operation>

Tip: Use --flagfile with the master's server.conf to automatically pick up
master_addresses and certs_dir, avoiding manual flag entry.
Example: ./build/latest/bin/yb-admin --flagfile master/conf/server.conf list_all_masters

<operation> must be one of:
 1. change_config <tablet_id> <ADD_SERVER|REMOVE_SERVER> <peer_uuid> [PRE_VOTER|PRE_OBSERVER]
 2. list_tablet_servers <tablet_id>
 3. backfill_indexes_for_table <table>
 4. list_tables [include_db_type] [include_table_id] [include_table_type]
 5. list_tables_with_db_types
 6. list_tablets <table> [<max_tablets>] (default 10, set 0 for max) [JSON] [include_followers]
 7. modify_table_placement_info <table> <placement_info> <replication_factor> [<placement_uuid>]
 8. modify_placement_info <placement_info> <replication_factor> [<placement_uuid>]
```

This gives matching commands and then full list. The whole-list dump was the exact issue we want to address. The current implementation adds helpful suggestions but still prints the whole list right after. So it only partially addresses the issue. When we have precise suggestions (Example 2), appending all ~100 commands is noise that undercuts the feature's purpose.

**So a good implementation would be :**

- No operation given (yb-admin alone) → we want the full usage (that's the help screen).
- Invalid operation name → we don't want the full dump when we have suggestions.

### Here is a quick glance how it will look like if we want to go via above route:

```
./build/latest/bin/yb-admin --master_addresses 192.0.2.62 list_table
Invalid operation: list_table
Did you mean one of these?
  list_tables
  list_tables_with_db_types
  list_tablet_servers
  list_tablets
Run './build/latest/bin/yb-admin' with no operation to see all available operations.
```

```
./build/latest/bin/yb-admin --master_addresses 127.0.0.1:0 list_snapshot_schedule
Invalid operation: list_snapshot_schedule
Did you mean one of these?
  list_snapshot_schedules
Run './build/latest/bin/yb-admin' with no operation to see all available operations.
```

```
./build/latest/bin/yb-admin --master_addresses 192.0.2.62 ist_table
Invalid operation: ist_table
Did you mean one of these?
  list_tables
Run './build/latest/bin/yb-admin' with no operation to see all available operations.
```

```
./build/latest/bin/yb-admin --master_addresses 192.0.2.62
yb-admin: ./build/latest/bin/yb-admin [--master_addresses server1:port,server2:port,server3:port,...]  [--timeout_ms <millisec>] [--certs_dir_name <dir_name>]
  [--flagfile <path/to/master/conf/server.conf>]
  <operation>

Tip: Use --flagfile with the master's server.conf to automatically pick up
master_addresses and certs_dir, avoiding manual flag entry.
Example: ./build/latest/bin/yb-admin --flagfile master/conf/server.conf list_all_masters

<operation> must be one of:
 1. change_config <tablet_id> <ADD_SERVER|REMOVE_SERVER> <peer_uuid> [PRE_VOTER|PRE_OBSERVER]
 2. list_tablet_servers <tablet_id>
 3. backfill_indexes_for_table <table>
 4. list_tables [include_db_type] [include_table_id] [include_table_type]
 5. list_tables_with_db_types
 6. list_tablets <table> [<max_tablets>] (default 10, set 0 for max) [JSON] [include_followers]
 .
 .
 .
 .
 .
 .
 130. get_ysql_major_version_catalog_state
 131. upgrade_ysql [use_single_connection] (default false)
 132. finalize_upgrade [use_single_connection]
 133. dump_sys_catalog_entries <entry_type> <folder_path> [entry_id]
 134. write_sys_catalog_entry <insert/update/delete> <entry_type> <entry_id> [file_path] [force]

<namespace>:
 [(ycql|ysql).]<namespace_name> (default ycql.)
<table>:
 <namespace> <table_name> | tableid.<table_id>
<index>:
  <namespace> <index_name> | tableid.<index_id>


  Flags from ../../src/yb/tools/yb-admin_cli.cc:
    -exclude_dead (Exclude dead tservers from output) type: bool default: false
    -init_master_addrs (host:port of any yb-master in a cluster) type: string
      default: ""
    -master_addresses (Comma-separated list of YB Master server addresses)
      type: string default: "localhost:7100" currently: "192.0.2.62"
    -timeout_ms (RPC timeout in milliseconds) type: int64 default: 60000

```

Let me know what you think.

src/yb/tools/yb-admin_cli.cc
ag26jan: Done - folded the fuzzy (edit-distance) matching into this PR. When the typed operation isn't a prefix of any command, it now falls back to Levenshtein edit-distance and suggests the closest command(s), so a misspelt first character resolves too:

- `ist_tables` -> `list_tables`
- `list_tabels` -> `list_tables`, `list_tablets`

Prefix matches are still preferred. I also stopped dumping the full ~130-command list on an invalid operation (the original complaint in #9624) - only the targeted suggestions plus a one-line hint are shown now; `yb-admin` with no operation still prints the full usage. Suggestion logic is factored into `ClusterAdminCli::GetSuggestedCommands`. Since this subsumes the fuzzy sub-task, I've closed #32516 as folded in here.

---
_automated - Cursor_

src/yb/tools/yb-admin_cli.cc
```diff
@@ -472,6 +472,25 @@ Status ClusterAdminCli::Run(int argc, char** argv) {
...
+      if (!commands_[it->second].hidden_) {
+        candidates.push_back(it->first);
+      }
+    }
+    if (!candidates.empty()) {
+      cerr << "Did you mean one of these?" << endl;
+      for (const auto& candidate : candidates) {
+        cerr << "  " << candidate << endl;
+      }
+    }
```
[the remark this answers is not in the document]
ag26jan: Addressed the empty-`op` guard: `GetSuggestedCommands` now returns early for an empty operation, so `yb-admin ""` no longer suggests every command (covered by a test).

On the temporary `std::vector`: keeping it, intentionally. The suggestion set now has to be materialized anyway for the edit-distance fallback (candidates are cleared/replaced as a closer-scoring tier is found before anything is printed), and this is a cold error path where the allocation cost is irrelevant next to readability. Optimizing it away would hurt clarity for no practical gain.

---
_automated - Cursor_

ag26jan:
@balajisubramanian-yugabyte I landed the combined fix per internal feedback. Please review. Thanks.

src/yb/tools/yb-admin_cli.cc
```diff
@@ -415,6 +441,55 @@ std::string ClusterAdminCli::GetArgumentExpressions(const std::string& usage_arg
...
+  // "list_tabels" -> "list_tables". Scale the tolerance with the length of the operation so that
+  // short names don't match everything, and keep only the commands tied for the smallest distance
+  // so the suggestion list stays short.
+  const size_t max_distance = std::max<size_t>(2, op.size() / 3);
+  size_t best_distance = std::numeric_limits<size_t>::max();
+  for (const auto& [name, index] : command_indexes_) {
+    if (commands_[index].hidden_) {
+      continue;
+    }
+    const size_t distance = EditDistance(op, name);
```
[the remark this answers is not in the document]
ag26jan: Applied in ae4d33644b7. Added the length-difference early-out before the `EditDistance` call: since the Levenshtein distance is at least `|len(op) - len(name)|`, commands whose length already differs from the operation by more than `max_distance` are skipped without running the DP. It's a pure pruning step, so the candidate set is unchanged. Thanks for the suggestion.

---
_automated - Cursor_

src/yb/tools/yb-admin_cli.cc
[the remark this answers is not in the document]
ag26jan: @balajisubramanian-yugabyte this needs your feedback as well.

ag26jan:
@mdbridge / @hulien22 Feel free to review as well (as you provided inputs over Slack). TIA.

src/yb/tools/yb-admin_cli.cc
```diff
@@ -415,6 +441,62 @@ std::string ClusterAdminCli::GetArgumentExpressions(const std::string& usage_arg
   return expressions.empty() ? "" : "Definitions: " + expressions;
 }
 
+std::vector<std::string> ClusterAdminCli::GetSuggestedCommands(const std::string& op) const {
+  // An empty operation is a prefix of - and close to - every command, so there is no meaningful
+  // suggestion to make.
```
hari90:
```suggestion
```

src/yb/tools/yb-admin_cli.cc
```diff
@@ -463,6 +545,8 @@ Status ClusterAdminCli::Run(int argc, char** argv) {
   }
 
   if (args.size() < 2) {
+    // No operation was given. Returning an InvalidArgument status makes main() print the full
+    // usage message, which is the intended "help" output for this case.
```
hari90:
```suggestion
```

src/yb/tools/yb-admin-test.cc
```diff
@@ -206,6 +206,89 @@ class AdminCliTest : public AdminTestBase {
...
+  const auto exe_path = GetAdminToolPath();
+  constexpr auto kUnusedMasterAddress = "127.0.0.1:0";
+  std::string error;
+
+  // Typo that is an unambiguous prefix of a single command.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(
+          exe_path, "--master_addresses", kUnusedMasterAddress, "list_snapshot_schedule"),
+      /* output */ nullptr, &error));
+  ASSERT_NE(error.find("Did you mean one of these?"), std::string::npos);
```
hari90:
```suggestion
  ASSERT_STR_CONTAINS(error, "Did you mean one of these?");
```

hari90: Same for all these other ASSERT_NE

src/yb/tools/yb-admin-test.cc
```diff
@@ -206,6 +206,89 @@ class AdminCliTest : public AdminTestBase {
...
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "list_table"), nullptr,
+      &error));
+  ASSERT_NE(error.find("list_tables"), std::string::npos);
+  ASSERT_NE(error.find("list_tablets"), std::string::npos);
+
+  // An empty operation is a prefix of every command, but should not list all of them.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, ""), nullptr, &error));
+  ASSERT_EQ(error.find("Did you mean one of these?"), std::string::npos);
```
hari90: ASSERT_STR_NOT_CONTAINS

src/yb/tools/yb-admin-test.cc
```diff
@@ -206,6 +206,89 @@ class AdminCliTest : public AdminTestBase {
...
+
+  // A far-off garbage string is beyond the edit-distance tolerance, so nothing is suggested.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(exe_path, "--master_addresses", kUnusedMasterAddress, "zzzzzzzzzzzzzzzzzz"),
+      nullptr, &error));
+  ASSERT_EQ(error.find("Did you mean one of these?"), std::string::npos);
+}
+
+// Verify that an invalid operation no longer dumps the full command list (the original complaint
+// in the issue), while running with no operation at all still prints the full usage as help.
```
hari90: All 3 tests can be merged into one test

hari90 (changes requested):

src/yb/tools/yb-admin-test.cc
ag26jan: Done — merged the three cases into a single `InvalidOperationSuggestsClosestCommands` test, since they all exercise the same "did you mean" help path and share the same setup (prefix matches, fuzzy/edit-distance matches, and the no-full-dump behavior). Also switched the substring checks to `ASSERT_STR_CONTAINS`/`ASSERT_STR_NOT_CONTAINS` for consistency. Verified locally: `AdminCliTest.InvalidOperationSuggestsClosestCommands` passes.

---
_automated · Cursor (Opus 4.8)_

balajisubramanian-yugabyte:
Overall looks good , some comments:
The commit summar lists three tests (InvalidOperationSuggestsPrefixMatches, ...SuggestsFuzzyMatches, ...DoesNotListAllCommands), but the diff has a single combined test InvalidOperationSuggestsClosestCommands. Lets update the commit description to reflect the actual change.

hari90 (approved):

PR #32519, opened 2026-07-07
[Docs] Mark transactional ddl as EA in 2025.2 and 2026.1

dr0pdb:
Transactional DDL is TP in 2025.1 and EA in 2025.2 and 2026.1. Presently, docs show it as TP in all 3 versions. Therefore, this PR updates the docs accordingly.


aishwarya24 (approved):
LGTM

Issue #17797, opened 2023-06-15
[YSQL] Support SPLIT INTO / SPLIT AT syntax for matviews


fizaaluthra:
### Description

Allow users to set tablet splitting properties for matviews at the time of creation. Note: split properties should be persisted after `REFRESH`.

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

PR #32539, opened 2026-07-08
[#17797] YSQL: Support SPLIT INTO syntax for matviews

pgyogesh:
## Summary

Allow tablet-splitting properties to be set on a materialized view at creation time, e.g.:

```sql
CREATE MATERIALIZED VIEW mv SPLIT INTO 3 TABLETS AS SELECT ...;
```

Previously `SPLIT INTO N TABLETS` / `SPLIT AT VALUES` were only accepted for `CREATE TABLE` and `CREATE INDEX`. Materialized views are created through the `CREATE TABLE AS` path (`CreateTableAsStmt` -> `IntoClause`), which had no field to carry the SPLIT clause.

### Creation support (commit 1)

- Add `split_options` to `IntoClause` and wire `YbOptSplit` into the `create_mv_target` grammar rule.
- Forward `into->split_options` into the synthesized `CreateStmt` in `create_ctas_internal`, so `DefineRelation` applies it and reconciles it with the `yb_presplit` reloption (exactly as for regular tables).
- Register the new field in the copy/equal/out/read node functions.

`SPLIT INTO N TABLETS` pre-splits the matview into N hash tablets and records `yb_presplit=N` in reloptions. `SPLIT AT VALUES` is accepted by the grammar but rejected at creation for a matview built from a `SELECT` (no range primary key), matching the existing restriction on hash-partitioned tables.

### Preserve the split across REFRESH (commit 2)

A non-concurrent `REFRESH MATERIALIZED VIEW` rebuilds the matview into a new heap via `make_new_heap()`, which was called with `yb_copy_split_options=false` -- so the refreshed matview silently dropped back to a single tablet even though `yb_presplit` remained in reloptions. This pre-existing gap affected matviews pre-split via either the new `SPLIT INTO` syntax or the existing `WITH (yb_presplit=N)` reloption.

The refresh now passes `yb_copy_split_options=true`, preserving the matview's current split (the requested pre-split plus any subsequent auto-splitting), mirroring the `ALTER TABLE` rewrite path in `ATRewriteTables`. The concurrent and in-place refresh paths merge into the existing matview via a temporary transient relation and were already unaffected.

## Upgrade/Rollback safety

No wire-format (`.proto`), catalog-schema, on-disk-format, or gflag-default changes. The new `IntoClause.split_options` is an in-memory parse-node field only -- never persisted or sent over the wire -- so it carries no mixed-version concern.

The only behavior change is on non-concurrent `REFRESH MATERIALIZED VIEW`: the new binary preserves the matview's tablet split, the old binary drops it to a single tablet. On a mixed-version cluster the outcome of a given REFRESH depends only on the version of the query-layer node that runs it; both outcomes are valid matview refreshes (the difference is solely the tablet count of the rebuilt storage). The `yb_presplit` reloption remains intact in either direction, so rollback to the prior version simply reverts to the old drop-to-one-tablet behavior with no cleanup or migration required.

## Test plan

- [ ] Extended the `yb.orig.split_options` regress test with materialized-view cases: `SPLIT INTO N TABLETS` (plain, `WITH NO DATA`, and combined with `WITH (yb_presplit=N)`), a `SPLIT AT VALUES` negative case, and a non-concurrent `REFRESH` that verifies the tablet count is preserved.
- [ ] `./yb_build.sh release --java-test org.yb.pgsql.TestPgRegressSplitOptions`
- [ ] Manually verified on a single-node cluster: `SPLIT INTO N` yields N tablets and `{yb_presplit=N}` in reloptions; `SPLIT AT VALUES` on a matview errors cleanly (no range PK); `REFRESH` (non-concurrent, `CONCURRENTLY`, and for `WITH (yb_presplit=N)` matviews) preserves the tablet count; a default no-split matview still refreshes to a single tablet (no regression).

### DIFF
--- a/src/postgres/src/backend/commands/createas.c
+++ b/src/postgres/src/backend/commands/createas.c
@@ -114,6 +114,8 @@ create_ctas_internal(List *attrList, IntoClause *into)
 	create->tablespacename = into->tableSpaceName;
 	create->if_not_exists = false;
 	create->accessMethod = into->accessMethod;
+	/* YB: forward the SPLIT clause so the relation is pre-split at creation */
+	create->split_options = into->split_options;
 
 	/*
 	 * Create the relation.  (This will error out if there's an existing view,
--- a/src/postgres/src/backend/commands/matview.c
+++ b/src/postgres/src/backend/commands/matview.c
@@ -336,11 +336,21 @@ ExecRefreshMatView(RefreshMatViewStmt *stmt, const char *queryString,
 	 * Create the transient table that will receive the regenerated data. Lock
 	 * it against access by any other process until commit (by which time it
 	 * will be gone).
+	 *
+	 * YB: Preserve the matview's tablet split when it is swapped in for the
+	 * old storage (non-concurrent refresh).  Without this the new heap would
+	 * be created with default splitting (a single tablet), silently discarding
+	 * the pre-split requested at CREATE MATERIALIZED VIEW ... SPLIT INTO time
+	 * (and any subsequent auto-splitting), even though yb_presplit remains in
+	 * reloptions.  This mirrors the table-rewrite path in ATRewriteTables.
+	 * (The concurrent and in-place refresh paths use a temporary transient
+	 * relation and merge into the existing matview, so their splits are
+	 * unaffected by this flag.)
 	 */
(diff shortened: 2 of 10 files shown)


pgyogesh:
@fizaaluthra Can you review this PR. It fixes #17797.

Issue #32550, opened 2026-07-09
[YSQL] DISTINCT over join can return duplicate rows


egladysh:
Jira Link: [DB-22255](https://yugabyte.atlassian.net/browse/DB-22255)
### Description

DISTINCT pushdown can incorrectly propagate a subquery’s uniqueness check into the outer query. This would make the outer planner skip a required duplicate-removal step.

Repro
```
  CREATE TABLE u(k int);
  CREATE TABLE t(a int);

  CREATE INDEX ON u(k ASC); CREATE INDEX ON t(a ASC);

  INSERT INTO u VALUES (1), (1), (2), (2), (3), (3);
  INSERT INTO t VALUES (10), (10), (20), (20), (30), (30);

  SELECT DISTINCT u.k FROM u, (SELECT DISTINCT a FROM t) s ORDER BY u.k;

 k
---
 1
 1
 1
 2
 2
 2
 3
 3
 3

```

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-22255]: https://yugabyte.atlassian.net/browse/DB-22255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #32571, opened 2026-07-09
[#32550] YSQL: Fix wrong results from DISTINCT pushdown through subquery scan

egladysh:
## Summary

When distinct pushdown is enabled (`yb_enable_distinct_pushdown`), `create_subqueryscan_path` propagated the subpath's `yb_uniqkeys` (a distinctness proof) into the outer query verbatim. Those keys are expressed in the subquery's namespace, so carrying them unchanged let the outer query treat its rows as already-distinct when they were not producing duplicate rows despite `SELECT DISTINCT`.

Example of the wrong result:

```sql
SELECT DISTINCT u.k FROM u, (SELECT DISTINCT a FROM t) s ORDER BY u.k;
```

returned duplicate `k` values instead of the distinct set.

This change adds `yb_convert_subquery_uniqkeys`, which translates each inner uniqkey into the outer query's namespace:

- Match the inner uniqkey against the subquery's target list.
- Map the matched target entry to the `Var` that the subquery scan actually emits (via `reltarget`).
- If any uniqkey cannot be translated, drop the entire proof (`yb_uniqkeys = NIL`) so the outer `DISTINCT` remains responsible for deduplication.

## Test plan

Jenkins

Add regression and coverage cases to the yb.orig.distinct_pushdown_join
YSQL regress test:

- Correctness: DISTINCT over a join with a DISTINCT subquery (the reported
  bug) now returns the distinct set instead of duplicate rows; also covered
  through cross join, inner join, the nullable side of a left join, and a
  UNION subquery.
- Translation retained: nested subquery scans, and a domain-to-base
  (RelabelType) relabel, keep distinct pushdown and elide the outer Unique.
- Proof dropped correctly: subquery with extra distinct keys, a key not
  exposed by the subquery scan, an expression index, and a collation
  mismatch all retain the outer Unique.
- Each case asserts both the EXPLAIN plan and the result rows.


src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -59,6 +59,9 @@ typedef enum
 #define STD_FUZZ_FACTOR 1.01
 
 static List *translate_sub_tlist(List *tlist, int relid);
+static void yb_propagate_subqueryscan_fields(YbPathInfo *parent_fields,
+											 RelOptInfo *rel, Path *subpath);
+static Var *yb_find_var_for_subquery_tle(RelOptInfo *rel, TargetEntry *tle);
```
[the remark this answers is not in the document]
egladysh: done

egladysh:
trigger jenkins

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -1319,6 +1322,90 @@ yb_propagate_fields_list(YbPathInfo *parent_fields, List *child_paths)
...
+ * Propagate YugabyteDB fields through a SubqueryScanPath.
+ *
+ * SubqueryScanPath crosses a namespace boundary:
+ * fields represented in the subquery's terms must be translated to
+ * the outer query.
+ */
+static void
+yb_propagate_subqueryscan_fields(YbPathInfo *parent_fields, RelOptInfo *rel,
+								 Path *subpath)
+{
```
pao214: can we simply do

```c
parent_fields->yb_uniqkeys = NIL;
```

here or at callsite?

For a few reasons:

1. We want to merge with upstream postgres v19. And we would ideally like to keep changes like these simple. Postgres v19 has its own version of distinct pushdown and we want to adopt that instead of this framework.
2. This optimization is not intended by the original change that introduced yb_propagate_fields.
3. The usecase for the optimization is not clear. This translation applies to queries such as `SELECT DISTINCT a FROM (SELECT DISTINCT a FROM distinct_inner) s`. Users do not write queries like these. There are some interesting cases such as when a is equal to the additional targets being selected. Example: `SELECT DISTINCT o.k FROM distinct_outer o, (SELECT DISTINCT a FROM distinct_inner) s WHERE o.k = s.a`. However, we do not see queries like these in the wild, so, we can skip optimizing this case.

src/postgres/src/test/regress/expected/yb.orig.distinct_pushdown_join.out
```diff
@@ -733,6 +733,31 @@ EXPLAIN (ANALYZE, COSTS OFF, TIMING OFF, SUMMARY OFF) SELECT DISTINCT v FROM t t
...
+CREATE INDEX distinct_inner_a_idx ON distinct_inner(a ASC);
+INSERT INTO distinct_outer
+  SELECT CASE WHEN i <= 500 THEN 1 ELSE 2 END
+  FROM GENERATE_SERIES(1, 1000) AS i;
+INSERT INTO distinct_inner
+  SELECT 10 * (i % 3)
+  FROM GENERATE_SERIES(1, 1000) AS i;
+ANALYZE distinct_outer;
+ANALYZE distinct_inner;
+SELECT DISTINCT o.k
```
pao214: Can we change this to

```sql
  \set query ':P SELECT DISTINCT o.k FROM distinct_outer o, (SELECT DISTINCT a FROM distinct_inner) s ORDER BY 1;'
  \i :iter_P2
```

?

Please follow the code conventions in the file.

pao214 (changes requested):
Let me know if something is unclear.

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -1319,6 +1322,90 @@ yb_propagate_fields_list(YbPathInfo *parent_fields, List *child_paths)
...
+
+	parent_fields->yb_uniqkeys = retval;
+}
+
+/*
+ * If the given subquery tlist resno is emitted by the subquery scan node,
+ * return a Var for it; otherwise return NULL.
+ */
+static Var *
+yb_find_var_for_subquery_tle(RelOptInfo *rel, int resno)
```
pao214: This seems to be the same as find_var_for_subquery_tle. Can we define yb_convert_subquery_uniqkeys in pathkeys.c instead to convert uniqkeys similar to how pathkeys are converted by convert_subquery_pathkeys?

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -1319,6 +1322,90 @@ yb_propagate_fields_list(YbPathInfo *parent_fields, List *child_paths)
...
+		return;
+
+	yb_propagate_fields(parent_fields, &subpath->yb_path_info);
+
+	parent_fields->yb_uniqkeys = NIL;
+
+	if (subpath->yb_path_info.yb_uniqkeys == NIL)
+		return;
+
+	foreach(lc, subpath->yb_path_info.yb_uniqkeys)
```
pao214: As mentioned in the comment below, can we refactor this logic into yb_convert_subquery_uniqkeys as analogous to convert_subquery_pathkeys.

It is better to put this in pathkeys.c (we place YB functions at the end).

This is to preserve the behavior with how pathkeys are computed and passed in.

src/postgres/src/backend/optimizer/util/pathnode.c
egladysh: Yeah, we can do that but it seems like reusing find_var_for_subquery_tle() requires constructing a temporary target list. The current implementation iterates over PathTarget expressions directly, avoiding those additional planner allocations. I still think that your suggestion is more aligned with existing PostgreSQL’s flow but it's not free unless I am missing something?

src/postgres/src/backend/optimizer/util/pathnode.c
egladysh: I replied to your suggestion. Please take a look. I am fine either way.

src/postgres/src/backend/optimizer/util/pathnode.c
pao214: I didn't realize that my comments went out. Didn't intend them to. Still new to GH. Looks like I need to start a review and not press comment.

However, I strongly prefer to follow postgresql path closely. Even if that comes at the cost of performance.

src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql
```diff
@@ -117,6 +117,28 @@ DROP INDEX irv;
```
pao214, line 114: This requires more tests since we only test the correctness fix and not the optimization itself.

Please add the following tests.

1. SELECT DISTINCT s.a FROM (SELECT DISTINCT a, b FROM t) s. To make sure that extra distinct keys still retains the unique node on top.
2. SELECT DISTINCT s.r1 FROM (SELECT DISTINCT r1 FROM t3 WHERE r1 = r2) s. I am not sure what the code does in this case. So, genuinely curious.
3. SELECT DISTINCT s.a FROM (SELECT DISTINCT a, a+1 FROM t) s. To test the drop yb_uniqkeys if cannot translate case. There is an ASC index on (a, a+1). Otherwise, there is no distinct pushdown in the subquery.
4. SELECT DISTINCT a, s.b FROM table_a, (SELECT DISTINCT b FROM table_b) s. Cross join.
5. SELECT DISTINCT a, s.b FROM table_a, (SELECT DISTINCT b FROM table_b) s WHERE a = s.b. Inner join.
6. SELECT DISTINCT foo.a FROM t foo LEFT JOIN (SELECT DISTINCT a FROM bar) baz WHERE foo.a = baz.a. The subquery side is nullable.
7. SELECT DISTINCT s.x FROM (SELECT DISTINCT a+1 AS x FROM table_b) s. INDEX is on a+1. What happens in this case? Is the distinct still pushed down? I would expect so.
8. SELECT DISTINCT a FROM (SELECT DISTINCT a FROM (SELECT DISTINCT a FROM t) s1) s2. Minimal doubly nested query.
9. SELECT DISTINCT x FROM (SELECT a FROM t1 UNION SELECT a FROM t2) s(x). UNION dedups. So, distinct can still be pushed down.
10. SELECT DISTINCT x FROM (SELECT a FROM t1 UNION ALL SELECT a FROM t2) s(x). This is the case I am a tad concerned about. We should not remove the upper unique node in this case.
11. Same as above but for INTERSECT and EXCEPT
12. SELECT DISTINCT s.a FROM (SELECT DISTINCT a FROM partitioned_table) s. Lets check how the optimization behaves on partitioned tables.
13. SELECT DISTINCT s.a FROM (SELECT a FROM foo) s. No distinct inside the subquery.
14. Query that relies on canonicalize_ec_expression.

Lets see if this works for canonicalize_ec_expression test case.
```sql
  CREATE TABLE distinct_collate (a text);
  CREATE INDEX distinct_collate_c_idx ON distinct_collate (a COLLATE "C" ASC);
  INSERT INTO distinct_collate
    SELECT (i % 5)::text FROM GENERATE_SERIES(1, 1000) AS i;
  ANALYZE distinct_collate;

  \set query ':P SELECT DISTINCT s.x FROM (SELECT DISTINCT a COLLATE "C" AS x FROM distinct_collate) s ORDER BY 1;'
  \i :iter_P2

  DROP TABLE distinct_collate;
```

Please add above tests so that we are well tested. Lets also update the test plan accordingly. Thanks!

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -1319,6 +1322,90 @@ yb_propagate_fields_list(YbPathInfo *parent_fields, List *child_paths)
...
+yb_propagate_subqueryscan_fields(YbPathInfo *parent_fields, RelOptInfo *rel,
+								 Path *subpath)
+{
+	List	   *retval = NIL;
+	ListCell   *lc;
+
+	if (!IsYugaByteEnabled())
+		return;
+
+	yb_propagate_fields(parent_fields, &subpath->yb_path_info);
```
pao214: Lets remove this line since it is not obvious that future fields need not be converted either.

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -59,6 +59,9 @@ typedef enum
 #define STD_FUZZ_FACTOR 1.01
 
 static List *translate_sub_tlist(List *tlist, int relid);
+static void yb_propagate_subqueryscan_fields(YbPathInfo *parent_fields,
```
pao214: yb decls go after postgres declarations. Is there some way for us to inform AI to follow existing YB conventions w.r.t. postgres code? Perhaps, src/lint/postgres.sh can help?

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -1319,6 +1322,90 @@ yb_propagate_fields_list(YbPathInfo *parent_fields, List *child_paths)
...
 
+/*
+ * Propagate YugabyteDB fields through a SubqueryScanPath.
+ *
+ * SubqueryScanPath crosses a namespace boundary:
+ * fields represented in the subquery's terms must be translated to
+ * the outer query.
+ */
+static void
+yb_propagate_subqueryscan_fields(YbPathInfo *parent_fields, RelOptInfo *rel,
```
pao214: yb functions go to the end of the file. Please make sure thats the case even after the move.

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -1319,6 +1322,90 @@ yb_propagate_fields_list(YbPathInfo *parent_fields, List *child_paths)
...
+			{
+				outer_var = yb_find_var_for_subquery_tle(rel, resno);
+				if (outer_var)
+					break;
+			}
+
+			resno++;
+		}
+
+		if (!outer_var)
```
pao214: code comment:

distinct on `{a,b}` does not imply distinct on `{a}`. So, if one yb_uniqkey cannot be translated, the whole set needs to be dropped.

pao214 (changes requested):
Let me know if anything is unclear.

src/postgres/src/backend/optimizer/util/pathnode.c
egladysh: @pao214 Just for argument sake `yb_propagate_fields_list` doesn't seem to have any forward declarations.

I asked AI to lint the change:

Yes, the pathnode.c changes follow the conventions enforced by src/lint/postgres.sh:

  - New symbols use the yb_ prefix.
  - Each changed upstream hunk contains an explicit YB marker.
  - Formatting matches yb_pgindent.
  - No new includes were added.
  - The implementation remains guarded by IsYugaByteEnabled().

The linter reported only two unrelated pre-existing marker warnings at lines 860 and 887. Its symbol scan could not complete because Exuberant Ctags is not installed, but manual review found no convention issues in our hunks.

Anyways I don't mind moving the declaration to your preferences but you might want to consider reconciling them with lint.

src/postgres/src/backend/optimizer/util/pathnode.c
egladysh: It doesn't seem to be the case either in the existing code but I am fine with moving it.

src/postgres/src/backend/optimizer/util/pathnode.c
egladysh: Agreed, SubqueryScan should keep its own propagation logic if it's ever needed.

src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql
egladysh: I did a quick test with your collation example. It doesn't translate uniq keys because collations don't match. I'll keep it as a negative test?

egladysh: test 7: the expression (a + 1) (indexed) is not pushed down but I am fairly sure it's not related to this change.

egladysh: As for test  7 after some back and forth with AI. 
```
The pushdown framework reduces a + 1 to its referenced base column a in yb_get_colrefs_for_distinct_pushdown(). It then calls
  yb_find_colref_in_index(), which explicitly skips every index target that is not a Var.

  An expression-index entry is represented by the expression a + 1, not a Var, so it is skipped and the lookup returns -1. Consequently,
  yb_calculate_distinct_prefixlen() rejects the distinct index path.

  Relevant code:

  - src/postgres/src/backend/optimizer/path/yb_uniqkeys.c:147: extracts base-column references from DISTINCT expressions.
  - src/postgres/src/backend/optimizer/path/yb_uniqkeys.c:310: searches index entries but ignores non-Var expressions.
  - src/postgres/src/backend/optimizer/path/yb_uniqkeys.c:362: rejects pushdown when that lookup fails.
  ```

I'd suggest to address it as a separate issue?

Comment (pao214) [reply src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql:114]: I see. We can skip.

File: src/postgres/src/backend/optimizer/util/pathnode.c (hunk 8 above)
Comment (pao214) [reply src/postgres/src/backend/optimizer/util/pathnode.c]: @egladysh I see. I think the check wasn't added because the logic may be a bit comlex for a linter.

We follow the following convention

```c
/* YB declarations */
static void show_yb_planning_stats_common(double num_seeks,
										  double num_nexts_prevs,
										  double num_table_pages,
										  double num_index_pages,
										  int docdb_result_width,
										  ExplainState *es);
```

Move all yb side declarations to the end of the declarations in a new section with header /* YB declarations */. We are a bit strict on the code conventions to make it easier for future postgres merges. Thanks.

src/postgres/src/backend/optimizer/util/pathnode.c
pao214: Yes, please move it. We want future merges to be easier.

src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql
egladysh: test 9: The planner does not seem to recognize the UNION result as unique. I suggest to skip the test?

egladysh: @pao214 test 10: Fails unrelated to this PR. I created an issue, https://github.com/yugabyte/yugabyte-db/issues/32692, easy to repro.

egladysh: @pao214 The partitioned-table case (test 12) seems to hit the same failure as UNION ALL. I suggest to skip it for this PR.

pao214 (changes requested):
commit summary:

> not — producing

Prefer ascii characters since not every text viewer supports unicode.

> This change adds yb_propagate_subqueryscan_fields

Let's change this to yb_convert_subquery_uniqkeys

pao214 (changes requested):
The test plan is also out-of-date.

src/postgres/src/backend/optimizer/util/pathnode.c
```diff
@@ -68,6 +68,8 @@ static bool contain_references_to(PlannerInfo *root, Node *clause,
 								  Relids relids);
 static bool ris_contain_references_to(PlannerInfo *root, List *rinfos,
 									  Relids relids);
+static void yb_propagate_subqueryscan_fields(YbPathInfo *parent_fields,
```
pao214, line 72: Please add

/* YB declarations */

src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql
pao214: For test 9, I incorrectly mentioned that we do not need a unique node on top. It is still necessary. This is because there may be duplicates across different legs of the UNION. Now, there may be some optimization opportunities but we are concerned about correctness for this test. Please add test 9 so that the test can demonstrate that the query gets correct results.

src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql
pao214, line 111: Reiterating what I mentioned in the other comment. Lets retain test 9 about UNION. It is okay if the plan is suboptimal. Looking for correct results.

pao214 (changes requested):
Lets rebase to latest master.

Either of these commits may have changed output

1. 0bc300132cf7
2. 3114bc802938
3. 0a478db32709

src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql
pao214, line 108: Currently, we do not seem to have a positive test that leverages canonicalize_ec_expression. Can we add the below test?

```
  CREATE DOMAIN dint AS int4;
  CREATE TABLE td (x dint);
  CREATE INDEX td_x_idx ON td (x ASC);
  INSERT INTO td SELECT (i % 5)::dint FROM generate_series(1, 1000) i;
  ANALYZE td;
  -- x::int4 is domain->base == RelabelType(Var(x) -> int4)
  SELECT DISTINCT s.a FROM (SELECT DISTINCT x::int4 AS a FROM td) s ORDER BY 1;
```

egladysh:
> commit summary:
> 
> not — producing
> 
> Prefer ascii characters since not every text viewer supports unicode.
> 

The PR was generate by YB create-pr skill. Please feel free to fix it.

src/postgres/src/backend/optimizer/util/pathnode.c
egladysh: FYI, Going forward, you can just suggest minor changes like that right in github for me to apply by just pressing a button. Thanks.

src/postgres/src/test/regress/sql/yb.orig.distinct_pushdown_join.sql
egladysh: Agreed

pao214 (approved):
LGTM

pao214:
trigger jenkins

PR #32573, opened 2026-07-10
[doc] Fix for heading links

ddhodge:
Fix so that headings aren't obscured by breadcrumbs when clicking a heading link


aishwarya24 (approved):
Functionality works as expected. LGTM

PR #32574, opened 2026-07-10
[DOC-1387] pgvector support in YBDB

ddhodge:
Add section of pgvector features that are supported in YBDB

@netlify /stable/additional-features/pg-extensions/extension-pgvector/


aishwarya24 (approved):
LGTM. copy to v2025.1 if these options are supported in that version as well.

hari90 (approved):

PR #32591, opened 2026-07-10
[#28753] DocDB: Use uniform base-1000 (GB/MB/KB) disk size units in Master and T-Server UIs

ag26jan:
## Summary

Disk sizes were reported with inconsistent units across YugabyteDB surfaces (GitHub issue #28753). The Platform (YBA) UI was standardized on **base 1000 with explicit `GB`/`MB`/`KB` labels**, and this change brings the DB-side Master and T-Server web UIs in line with that standard.

### Problem

The Master and T-Server web UIs formatted SST/WAL/disk sizes with `HumanReadableNumBytes::ToString`, which computes in **base 1024** but labels the value with a bare `K`/`M`/`G` — so a value that is really `41.46 MiB` was rendered as `41.46M`. The Master "Tablet Servers" table, however, already used `HumanizeBytes`, which formats in **base 1000** with explicit `" GB"`/`" MB"`/`" KB"` labels. As a result the same underlying metric could appear to disagree between pages (e.g. `6.3 GB` on one page and `5.9 GiB` on another).

### Change

Switch the on-disk-size displays to the existing `HumanizeBytes` helper (already `#include`d via `yb/gutil/strings/numbers.h`) so all size output uses base 1000 with explicit `GB`/`MB`/`KB` labels:

- **`src/yb/master/master-path-handlers.cc`** — the per-tablet on-disk-size HTML breakdown (Total / WAL / SST / Uncompressed / Vector index), the per-table JSON `on_disk_size` string fields, and the per-replica "Active SSTs size / WALs size" HTML.
- **`src/yb/tserver/tserver-path-handlers.cc`** — the per-tablet on-disk-size HTML breakdown, the maintenance-ops "RAM anchored / Logs retained" columns, and the tablet JSON size string fields.

Notes:
- The raw `*_bytes` JSON fields are left unchanged — they carry exact byte counts, not human-readable strings.
- The shared `HumanReadableNumBytes` utility is intentionally **not** modified, since it is also used for log output and parse round-trips elsewhere in the codebase; scoping the change to the UI path handlers avoids unrelated risk.

## Demo

Built the daemons from this branch and loaded three tables on a local `yb-ctl` (RF=1) cluster — one sized to land in each unit — then flushed them so real SST files exist. The Master UI `/tables` on-disk-size column now renders each in the correct base-1000 unit (`KB` / `MB` / `GB`):

| Table | SST Files | Total |
|---|---|---|
| `demo_kb` | 131.44 KB | 1.18 MB |
| `demo_mb` | 8.52 MB | 21.02 MB |
| `demo_gb` | 1.41 GB | 2.84 GB |

Master UI /tables showing KB, MB, and GB units

## Test plan

- [x] `./yb_build.sh release daemons` builds `yb-master` and `yb-tserver` cleanly with the changes.
- [x] On a local `yb-ctl` (RF=1) cluster, loaded a ~572 MB table and flushed it; the Master UI `/tables` on-disk-size column renders `Total: 673.71 MB`, `WAL Files: 634.90 MB`, `SST Files: 38.81 MB`, `SST Files Uncompressed: 616.93 MB`, and the per-replica table page renders `Active SSTs size: 38.81 MB` / `WALs size: 634.90 MB` (base 1000).
- [x] The T-Server UI `/tablets` page renders `Total (may be stale): 674.17 MB`, `SST Files: 38.81 MB`, `WAL Files: 634.90 MB`, `Consensus Metadata: 1.96 KB` (base 1000).
- [x] The JSON endpoints (`/api/v1/tablets`, tables JSON) report human-readable size strings in `GB`/`MB`/`KB` (e.g. `"sst_files_size": "38.81 MB"` for `sst_files_size_bytes: 38809852`) while the `*_bytes` fields still carry raw byte counts. Confirmed `1048576` bytes now renders `1.05 MB` (base 1000) rather than the previous `1.00M` (base 1024).

### DIFF
--- a/src/yb/gutil/strings/string_util-test.cc
+++ b/src/yb/gutil/strings/string_util-test.cc
@@ -32,6 +32,7 @@
 // Some portions Copyright 2013 The Chromium Authors. All rights reserved.
 
 #include "yb/gutil/strings/join.h"
+#include "yb/gutil/strings/numbers.h"
 #include "yb/gutil/strings/util.h"
 #include "yb/util/string_util.h"
 #include "yb/util/test_util.h"
@@ -229,6 +230,34 @@ TEST_F(StringUtilTest, JoinStringsLimitCount) {
   std::string result = "test";
   JoinStringsLimitCount(std::vector<std::string>{"foo"}, ",", 1, &result);
   ASSERT_EQ(result, "foo");
+}
+
+TEST_F(StringUtilTest, TestHumanizeBytes) {
+  // Below 1 KB the raw byte count is printed with no fractional part.
+  ASSERT_EQ("0 B", HumanizeBytes(0));
+  ASSERT_EQ("1 B", HumanizeBytes(1));
+  ASSERT_EQ("999 B", HumanizeBytes(999));
+
+  // Decimal (base 1000) scaling with explicit labels.
+  ASSERT_EQ("1.00 KB", HumanizeBytes(1000));
+  ASSERT_EQ("1.00 MB", HumanizeBytes(1000000));
+  ASSERT_EQ("1.00 GB", HumanizeBytes(1000000000));
+
+  // A binary "1 MiB" (1048576 bytes) is 1.05 MB in base 1000.
+  ASSERT_EQ("1.05 MB", HumanizeBytes(1048576));
+  // Example from GitHub issue #28753: 6,347,254,071 bytes -> 6.35 GB.
+  ASSERT_EQ("6.35 GB", HumanizeBytes(6347254071ULL));
+
+  // Large sizes must keep scaling past GB instead of showing "10000.00 GB".
+  ASSERT_EQ("10.00 TB", HumanizeBytes(10000000000000ULL));
+  ASSERT_EQ("2.50 PB", HumanizeBytes(2500000000000000ULL));
+  ASSERT_EQ("1.00 EB", HumanizeBytes(1000000000000000000ULL));
+  // EB is the largest unit; values beyond it stay in EB.
+  ASSERT_EQ("5.00 EB", HumanizeBytes(5000000000000000000ULL));
+
+  // The precision argument controls the number of fractional digits.
+  ASSERT_EQ("1.234 KB", HumanizeBytes(1234, 3));
+  ASSERT_EQ("1.2 KB", HumanizeBytes(1234, 1));
 }
 
 } // namespace yb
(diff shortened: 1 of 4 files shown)


ag26jan:
@spolitov / @ttyusupov / @devansh5398 Can anyone of you review this PR? TIA.

ag26jan:
@spolitov  / @ttyusupov  / @devansh5398 ping for your review on this PR? TIA.

ag26jan:
@balajisubramanian-yugabyte need your review here.

balajisubramanian-yugabyte:
One comment:
HumanizeBytes has no unit above GB whereas HumanReadableNumBytes::ToString goes up to E. After this change a 10 TB SST size renders as 10000.00 GB instead of the previous ~9.09T. On large clusters, per-table/tablet sizes over 1 TB are common, so will show very big GB values.

ag26jan:
That's a very insightful point. You're right that the old HumanizeBytes capped at GB, so a 10 TB tablet would have rendered as 10000.00 GB. I've reworked it to scale through the full base-1000 range (B → KB → MB → GB → TB → PB → EB) using a small loop, so large sizes now show as 10.00 TB, 2.50 PB, etc. Values under 1 KB still print as exact bytes, and the precision argument is unchanged. Also added a TestHumanizeBytes unit test in string_util-test.cc covering each unit boundary, the base-1000 vs. binary distinction, and the precision arg. PTAL.

src/yb/gutil/strings/string_util-test.cc
```diff
@@ -231,4 +232,32 @@ TEST_F(StringUtilTest, JoinStringsLimitCount) {
...
+  // Example from GitHub issue #28753: 6,347,254,071 bytes -> 6.35 GB.
+  ASSERT_EQ("6.35 GB", HumanizeBytes(6347254071ULL));
+
+  // Large sizes must keep scaling past GB instead of showing "10000.00 GB".
+  ASSERT_EQ("10.00 TB", HumanizeBytes(10000000000000ULL));
+  ASSERT_EQ("2.50 PB", HumanizeBytes(2500000000000000ULL));
+  ASSERT_EQ("1.00 EB", HumanizeBytes(1000000000000000000ULL));
+  // EB is the largest unit; values beyond it stay in EB.
+  ASSERT_EQ("5.00 EB", HumanizeBytes(5000000000000000000ULL));
+
```
balajisubramanian-yugabyte, line 257: nit: 999,999,999 bytes divides to 999.999999 MB, which setprecision(2) renders as 1000.00 MB rather than promoting to 1.00 GB. 
we can round at the chosen precision first, then promote if it hits 1000 if it needs to be fixed. We can add a test case for the edge cases too.

ag26jan: Good catch — fixed in the follow-up commit. HumanizeBytes now rounds to the requested precision first and only then selects the unit, promoting to the next unit when the rounded value reaches 1000 (capped at the largest unit). So 999,999,999 bytes now renders as 1.00 GB instead of 1000.00 MB. Added test cases for the promotion boundaries (MB->GB, KB->MB, TB->PB, and precision=0) plus nearby values that must NOT be promoted (999.50 MB, 999.99 MB).

balajisubramanian-yugabyte:
LGTM !

Issue #32516, opened 2026-07-07
[DocDB] yb-admin: suggest fuzzy (edit-distance) matches for misspelled commands


ag26jan:
Jira Link: [DB-22222](https://yugabyte.atlassian.net/browse/DB-22222)
### Description

Follow-up / extension of #9624.

**Is your feature request related to a problem or an enhancement? Please describe.**

#9624 improved the `yb-admin` experience so that when an operation name is not an exact match, we suggest registered commands that the typed name is a **prefix** of (e.g. `list_snapshot_schedule` → `list_snapshot_schedules`).

However, prefix matching only helps a narrow class of typos. It does **not** help when the mistake is anywhere other than the end of the string, for example:

| Typed | Intended | Handled by prefix matching? |
|---|---|---|
| `list_tabels` (transposition) | `list_tables` | ❌ |
| `ist_tables` (missing leading char) | `list_tables` | ❌ |
| `list_snapshpt_schedules` (middle typo) | `list_snapshot_schedules` | ❌ |
| `snapshot_schedules` (missing `list_` prefix) | `list_snapshot_schedules` | ❌ |

In all of these cases the user still gets no useful hint.

**Provide a brief solution (if any) of what you want to happen.**

Add a **fuzzy (edit-distance) fallback** to the `yb-admin` command dispatch in `ClusterAdminCli::Run` (`src/yb/tools/yb-admin_cli.cc`):

- When exact match fails **and** prefix matching produces no candidates, compute the Levenshtein edit distance between the typed operation and every non-hidden registered command.
- Suggest the closest command(s) within a length-scaled tolerance (e.g. `max_distance = max(2, len/3)`), only surfacing the set tied for the minimum distance so the output stays short.
- Preserve existing behavior: exact match and prefix match remain preferred; a far-off/garbage string still yields no suggestion; nothing is auto-run (consistent with the guidance on #9624).

**Issue Type:** kind/improve-ux (UX improvement to `yb-admin`)

**Notes**

- A working prototype already exists (Levenshtein helper + fallback in `ClusterAdminCli::Run`, plus a `yb-admin-test` covering transposition, missing-leading-char, and garbage-input cases). It can be contributed as a follow-up PR to the #9624 change.
- Good candidate to keep as an independently reviewable change so the conservative prefix-only behavior from #9624 can ship separately.

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[...]

PR #32649, opened 2026-07-15
[BACKPORT 2025.2][#9624] DocDB: Suggest closest matches for misspelled yb-admin commands (#32513)

### LINKED ISSUE #9624: [usability] Allow prefix matching for yb-admin commands

ag26jan:
## Summary

`yb-admin` looks up the operation name given on the command line with an
exact match against the
list of registered commands. If the operation is slightly misspelled
(e.g.
`list_snapshot_schedule` instead of `list_snapshot_schedules`), the user
was shown the
`Invalid operation` error followed by the entire list of ~130 supported
commands, with no hint
about what they may have meant.

This PR makes that error helpful in two ways:

**1. Suggest the closest commands.**
- Prefix match first: registered (non-hidden) commands that have the
typed operation as a prefix
  are suggested.
- Edit-distance fallback: if nothing is a prefix, fall back to
Levenshtein (fuzzy) matching so a
typo anywhere in the name - a transposition, or a wrong or missing
leading character - still
resolves to the closest command(s). The tolerance scales with the length
of the operation, and
  only the closest-scoring tier is shown to keep the list short.

```
$ yb-admin --master_addresses ... list_snapshot_schedule    # prefix
Invalid operation: list_snapshot_schedule
Did you mean one of these?
  list_snapshot_schedules

$ yb-admin --master_addresses ... ist_tables                # missing leading char (fuzzy)
Invalid operation: ist_tables
Did you mean one of these?
  list_tables

$ yb-admin --master_addresses ... list_tabels               # transposition (fuzzy)
Invalid operation: list_tabels
Did you mean one of these?
  list_tables
  list_tablets
```

**2. Stop dumping the full command list on an invalid operation.** That
wall of ~130 commands was
the original complaint in the issue. The invalid-operation path now
shows only the targeted
suggestions plus a one-line hint, and returns a non-`InvalidArgument`
status so `main()` no longer
appends the full usage. Running `yb-admin` with no operation at all
still prints the full usage as
help.

The command is never run automatically based on a match, even an
unambiguous prefix, to avoid
surprising behavior for scripts that pass an already-invalid command
(per maintainer guidance on
the issue).

Fixes #9624. Subsumes #32516 (fuzzy / nearest-neighbor matching), which
is closed as folded in
here.

Original commit: d8f71119fa9a4cc761ae43705082db7907132d71 / #32513

## Test plan

- [x] `AdminCliTest.InvalidOperationSuggestsClosestCommands` - one test
covering the whole
  "did you mean" help path:
- prefix matches: an unambiguous single-command prefix, and an ambiguous
prefix that lists
    every candidate;
- fuzzy / edit-distance matches: a transposition (`list_tabels`) and a
missing leading
    character (`ist_tables`);
- that an empty operation and a far-off garbage string suggest nothing;
and
- that an invalid operation no longer prints the full command list on
either stream, while
    running with no operation at all still does.
- [x] Built `yb-admin` and `yb-admin-test` (release) locally; the test
passes, and manually
verified the examples above. The test needs no running cluster since the
check happens before
  `yb-admin` connects to the master.

---------

### DIFF
--- a/src/yb/tools/yb-admin-test.cc
+++ b/src/yb/tools/yb-admin-test.cc
@@ -213,6 +213,69 @@ class AdminCliTest : public AdminTestBase {
   TmpDirProvider tmp_dir_;
 };
 
+// Verify the "did you mean" help for a misspelled operation, none of which needs a running
+// cluster (the operation is checked before yb-admin connects to the master): prefix matches,
+// fuzzy (edit-distance) matches, and that an invalid operation no longer dumps the full command
+// list (the original complaint in the issue) while running with no operation still prints the
+// full usage as help.
+TEST_F(AdminCliTest, InvalidOperationSuggestsClosestCommands) {
+  const auto exe_path = GetAdminToolPath();
+  constexpr auto kUnusedMasterAddress = "127.0.0.1:0";
+  // This marker only appears in the full usage/command listing (which is printed to stdout).
+  constexpr auto kFullUsageMarker = "<operation> must be one of";
+  std::string output;
+  std::string error;
+
+  // Prefix match: an unambiguous typo suggests the single command it is a prefix of.
+  ASSERT_NOK(Subprocess::Call(
+      ToStringVector(
+          exe_path, "--master_addresses", kUnusedMasterAddress, "list_snapshot_schedule"),
+      /* output */ nullptr, &error));
+  ASSERT_STR_CONTAINS(error, "Did you mean one of these?");
+  ASSERT_STR_CONTAINS(error, "list_snapshot_schedules");
+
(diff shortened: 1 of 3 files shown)


ag26jan:
gemini review

ag26jan:
@hari90 Can you review this backport.

hari90:
trigger jenkins

PR #32680, opened 2026-07-16
Guard ExecutorCheckPerms hook when audit stack is absent

### LINKED ISSUE #32512: pgaudit: guard against NULL auditEventStack in ExecutorCheckPerms_hook (avoid SIGSEGV on non-executor permission checks)
**Summary**
pgaudit's `ExecutorCheckPerms_hook` handler (`log_select_dml`) assumes `auditEventStack` is non-NULL. That stack is only established during executor startup, so when `ExecCheckRTPerms()` is called outside a running executor, the hook fires with `auditEventStack == NULL` and dereferences it → SIGSEGV in `pgaudit.c`.

**How it's hit**
An extension that performs table permission checks at plan time or from a utility command (before `ExecutorStart` builds the audit stack) triggers this whenever pgaudit is loaded. pg_duckdb (preview) calls `ExecCheckRTPerms()` from its planner and utility/COPY handling, so with pgaudit preloaded the backend crashes.

**Ask**
Add a NULL-`auditEventStack` guard in pgaudit's `ExecutorCheckPerms_hook` path so it safely no-ops when there is no active audit event stack.

**Follow-on cleanup**
Once hardened, pg_duckdb can delete its `YbExecCheckRTPerms` workaround (which temporarily clears `ExecutorCheckPerms_hook`) and call `ExecCheckRTPerms()` directly. Pre-GA item for pg_duckdb.

---

somiljain2006:
Fixes #32512 

This change adds a defensive NULL check for auditEventStack in pgaudit_ExecutorCheckPerms_hook(). ExecCheckRTPerms() can be invoked outside ExecutorStart(), before the audit event stack has been established. In that case, skip audit logging and continue the hook chain instead of dereferencing a NULL auditEventStack, preventing backend crashes while preserving the existing behavior during normal executor execution.

### DIFF
--- a/src/postgres/third-party-extensions/pgaudit/pgaudit.c
+++ b/src/postgres/third-party-extensions/pgaudit/pgaudit.c
@@ -1367,6 +1367,15 @@ pgaudit_ExecutorCheckPerms_hook(List *rangeTabls, bool abort)
 {
   Oid auditOid = InvalidOid;
 
+    if (auditEventStack == NULL)
+    {
+        if (next_ExecutorCheckPerms_hook &&
+            !(*next_ExecutorCheckPerms_hook) (rangeTabls, abort))
+            return false;
+
+        return true;
+    }
+
     /* Get the audit oid if the role exists */
   if (auditRole != NULL && auditRole[0] != '\0') {
     auditOid = get_role_oid(auditRole, true);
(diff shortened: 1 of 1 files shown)


somiljain2006:
@rahulb-yb Can you review this pr?

Issue #32681, opened 2026-07-16
[DocDB] yb-admin list_tablet_server_log_locations fails when DEAD tservers are in the registry


ag26jan:
Jira Link: [DB-22409](https://yugabyte.atlassian.net/browse/DB-22409)
## Description

`yb-admin list_tablet_server_log_locations` fails with a connect timeout when the master's tablet-server registry still contains DEAD tservers.

The command iterates every tserver returned by `ListTabletServers` and issues a `GetLogLocation` RPC to each one. For DEAD nodes that are no longer reachable, the first connect timeout aborts the whole command via `VERIFY_RESULT`, even if live tservers were already listed successfully.

### Steps to reproduce

1. Run a multi-node cluster, then stop / remove one or more tservers without clearing them from the master's in-memory registry (DEAD entries remain visible in `list_all_tablet_servers`).
2. Run:
   ```
   yb-admin --master_addresses <masters> list_tablet_server_log_locations
   ```
   Example:

```
[yugabyte@yb-demo-agangwar-lab-n2 ~]$ tserver/bin/yb-admin --master_addresses 10.231.0.62 list_all_tablet_servers
Tablet Server UUID               RPC Host/Port Heartbeat delay Status   Reads/s  Writes/s Uptime   SST total size  SST uncomp size SST #files      Memory   Broadcast Host/Port
1824af0623844077a9276ab904162c95 10.231.0.62:9100 0.00s           ALIVE    0.00     0.00     0        0 B             0 B             0               0 B      N/A
baa09aa252f64dc88af9d3f1d8f775d5 10.231.0.30:9100 0.00s           DEAD     0.00     0.00     0        0 B             0 B             0               0 B      N/A
c4410a539a6c47558b309e647fc73820 10.231.0.17:9100 0.00s           DEAD     0.00     0.00     0        0 B             0 B             0               0 B      N/A
ecfd9942e0154aad8b48a217c4c4d24c 10.231.0.19:9100 0.00s           DEAD     0.00     0.00     0        0 B             0 B             0               0 B      N/A


[yugabyte@yb-demo-agangwar-lab-n2 ~]$ tserver/bin/yb-admin --master_addresses 10.231.0.62 list_tablet_server_log_locations
TS UUID                          	RPC Host/Port 	LogLocation
1824af0623844077a9276ab904162c95 	10.231.0.62:9100 	/mnt/d0/yb-data/tserver/logs
[...]

PR #32682, opened 2026-07-16
[#32681] DocDB: Skip DEAD tservers in list_tablet_server_log_locations

ag26jan:
## Summary

`yb-admin list_tablet_server_log_locations` previously issued a `GetLogLocation` RPC to every tablet server in the master's registry, including DEAD ones. A connect timeout to the first unreachable DEAD node aborted the command via `VERIFY_RESULT`, even after live servers had already been printed.

This change:
- Skips RPCs to DEAD tservers and prints `N/A (DEAD)` instead
- Soft-fails per-server RPC errors so one unreachable node does not hide results for the rest
- Uses the RPC address from the existing list entry (avoids a redundant `ListTabletServers` call per node)

## Test plan

- [ ] On a cluster with at least one DEAD tserver visible in `list_all_tablet_servers`, run `yb-admin list_tablet_server_log_locations`
- [ ] Confirm the command exits successfully and prints log locations for ALIVE tservers
- [ ] Confirm DEAD tservers appear as `N/A (DEAD)` rather than causing a connect timeout
- [ ] On a healthy all-ALIVE cluster, confirm log locations still print as before

### DIFF
--- a/src/yb/tools/yb-admin_client.cc
+++ b/src/yb/tools/yb-admin_client.cc
@@ -1481,16 +1481,40 @@ Status ClusterAdminClient::ListTabletServersLogLocations() {
 
   for (const ListTabletServersResponsePB::Entry& server : servers) {
     auto ts_uuid = server.instance_id().permanent_uuid();
+    const auto rpc_addr_str = FormatFirstHostPort(
+        server.registration().common().private_rpc_addresses());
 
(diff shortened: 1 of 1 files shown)


ag26jan:
Superseded by #32787, which was merged to `master` in commit ae915b2567d. That PR implements the same fix (skip DEAD tservers) plus alive-first sorting and a regression test, so closing this duplicate.

PR #32719, opened 2026-07-17
[#6331] DocDB: Raise global memstore cap on large nodes

### LINKED ISSUE #6331: [DocDB] Change default for global memstore on larger nodes
Jira Link: [DB-2986](https://yugabyte.atlassian.net/browse/DB-2986)
At the moment we set a max of 2GB as default for global memstore. We should set it to 4GB on systems with 64GB RAM or more.

[DB-2986]: https://yugabyte.atlassian.net/browse/DB-2986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bmcnama-shopify:
## Summary

- Increase the default global memstore cap from 2 GiB to 4 GiB for YB-TServers with at least 64 GiB of available RAM.
- Resolve the cap through the existing startup recommendation mechanism so explicit overrides and YB-Master behavior remain unchanged.
- Add RAM-boundary and override tests, and document the RAM-dependent default.

Closes #6331

## Upgrade/Rollback safety

Mixed-version clusters can temporarily run YB-TServers with different 2 GiB and 4 GiB defaults. Each server applies its cap independently; there are no wire-format or on-disk changes. Rolling back restores the 2 GiB default after restart, while explicit flag values remain unchanged.

## Test plan

- [x] `ServerMainUtilTest.GlobalMemstoreDefault`
- [x] `ServerMainUtilTest.GlobalMemstoreOverride`
- [x] `FlushITest.TestFlushHappens`


bmcnama-shopify:
trigger jenkins

Issue #32697, opened 2026-07-17
[DocDB] Split child loses retryable-request state after restart


egladysh:
Jira Link: [DB-22419](https://yugabyte.atlassian.net/browse/DB-22419)
### Description

  A split child tablet can lose its post-split retryable-request state after restart. When the split parent is already running, the child copies the parent’s frozen pre-split state and skips restoring its own persisted state and WAL history. Retrying an acknowledged write may then execute it twice.

  ### Steps to reproduce

  1. Create a table with one tablet and split it.
  2. Keep the split parent tablet running.
  3. Apply a write to one child, but simulate losing the client response.
  4. Persist the child’s retryable-request state.
  5. Restart the tablet server, ensuring the parent opens before the child.
  6. Retry the original write with the same client and request IDs.
  7. Observe that the write executes again instead of being deduplicated.

  Expected: the child restores its own retryable-request state and rejects the duplicate.

  Actual: the child copies the parent’s pre-split state, forgets the post-split request, and executes it again.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-22419]: https://yugabyte.atlassian.net/browse/DB-22419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #32726, opened 2026-07-17
[#32697] DocDB: Skip copying parent retryable requests when split child has persisted state

egladysh:
## Summary

When a split child tablet is opened on tserver restart, `TSTabletManager::OpenTablet` copies retryable requests from the split parent tablet when `enable_copy_retryable_requests_from_parent` is set. This copy happened whenever the tablet had a split parent, even when the child had already flushed its own retryable requests bootstrap state to disk.

Overwriting the child's persisted state with the parent's state regresses retryable-request tracking, so an already-applied retried write can be applied a second time after restart.

This change only copies retryable requests from the parent when the child has no bootstrap state file of its own (`!bootstrap_state_manager->has_file_on_disk()`). When the child has persisted state, that state is authoritative: it is loaded during bootstrap and WAL replay re-registers any newer requests.

## Test plan

- [x] New regression test `TabletSplitSingleServerITest.RetryableWriteAfterSplitChildRestart` passes with the fix (run in a debug build, where the test's sync points are active):

```
./yb_build.sh debug --cxx-test tablet-split-itest \
  --gtest_filter TabletSplitSingleServerITest.RetryableWriteAfterSplitChildRestart
```

### DIFF
--- a/src/yb/integration-tests/tablet-split-itest.cc
+++ b/src/yb/integration-tests/tablet-split-itest.cc
@@ -3057,6 +3057,91 @@ TEST_F(TabletSplitSingleServerITest, TestRetryableWriteWithPersistedStructure) {
   TestRetryableWrite();
 }
 
+TEST_F(TabletSplitSingleServerITest, RetryableWriteAfterSplitChildRestart) {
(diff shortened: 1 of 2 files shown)


bmcnama-shopify:
trigger jenkins

PR #32754, opened 2026-07-20
[BACKPORT 2026.1][#32550] YSQL: Fix wrong results from DISTINCT pushdown through subquery scan (#32571)

pao214:
## Summary

When distinct pushdown is enabled (`yb_enable_distinct_pushdown`),
`create_subqueryscan_path` propagated the subpath's `yb_uniqkeys` (a
distinctness proof) into the outer query verbatim. Those keys are
expressed in the subquery's namespace, so carrying them unchanged let
the outer query treat its rows as already-distinct when they were not
producing duplicate rows despite `SELECT DISTINCT`.

Example of the wrong result:

```sql
SELECT DISTINCT u.k FROM u, (SELECT DISTINCT a FROM t) s ORDER BY u.k;
```

returned duplicate `k` values instead of the distinct set.

This change adds `yb_convert_subquery_uniqkeys`, which translates each
inner uniqkey into the outer query's namespace:

- Match the inner uniqkey against the subquery's target list.
- Map the matched target entry to the `Var` that the subquery scan
actually emits (via `reltarget`).
- If any uniqkey cannot be translated, drop the entire proof
(`yb_uniqkeys = NIL`) so the outer `DISTINCT` remains responsible for
deduplication.

Original commit: ee367279a3e9a333544e9870603df84e64fa9f15 / #32571

## Test plan

Jenkins

Add regression and coverage cases to the yb.orig.distinct_pushdown_join
YSQL regress test:

- Correctness: DISTINCT over a join with a DISTINCT subquery (the
reported
bug) now returns the distinct set instead of duplicate rows; also
covered
through cross join, inner join, the nullable side of a left join, and a
  UNION subquery.
- Translation retained: nested subquery scans, and a domain-to-base
(RelabelType) relabel, keep distinct pushdown and elide the outer
Unique.
- Proof dropped correctly: subquery with extra distinct keys, a key not
  exposed by the subquery scan, an expression index, and a collation
  mismatch all retain the outer Unique.
- Each case asserts both the EXPLAIN plan and the result rows.

### DIFF
--- a/src/postgres/src/backend/optimizer/path/pathkeys.c
+++ b/src/postgres/src/backend/optimizer/path/pathkeys.c
@@ -2075,3 +2075,48 @@ yb_get_ecs_for_query_uniqkeys(PlannerInfo *root)
 
 	return ecs;
 }
+
+/*
+ * Convert a subquery's uniqkeys into the terms of the outer query.
+ */
+List *
+yb_convert_subquery_uniqkeys(RelOptInfo *rel, List *subquery_uniqkeys,
+							 List *subquery_tlist)
+{
+	List	   *retval = NIL;
+	ListCell   *i;
+
+	foreach(i, subquery_uniqkeys)
+	{
+		Expr	   *uniqkey = (Expr *) lfirst(i);
+		Var		   *outer_var = NULL;
+		ListCell   *j;
+
+		foreach(j, subquery_tlist)
+		{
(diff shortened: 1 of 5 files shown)


pao214:
Trigger Jenkins

---

pao214:
use internal tooling instead

Issue #32385, opened 2026-06-29
[YSQL] Possibly DDL-atomicity / OID-consistency defect exposed by killing the master leader mid-CREATE EXTENSION


archit-rastogi:
Jira Link: [DB-22042](https://yugabyte.atlassian.net/browse/DB-22042)
### Description

Description:

The extension creation was in-flight for ~7.4 s straddling a master-leader failover, and the duplicate-key error after the new leader came up.
```
Exception in thread "main" com.yugabyte.util.PSQLException: ERROR: duplicate key value violates unique constraint "pg_amop_oid_index"
	at com.yugabyte.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2725)
	at com.yugabyte.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2412)
	at com.yugabyte.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:371)
	at com.yugabyte.jdbc.PgStatement.executeInternal(PgStatement.java:502)
	at com.yugabyte.jdbc.PgStatement.execute(PgStatement.java:419)
	at com.yugabyte.jdbc.PgStatement.executeWithFlags(PgStatement.java:341)
	at com.yugabyte.jdbc.PgStatement.executeCachedSql(PgStatement.java:326)
	at com.yugabyte.jdbc.PgStatement.executeWithFlags(PgStatement.java:302)
	at com.yugabyte.jdbc.PgStatement.execute(PgStatement.java:297)
	at com.zaxxer.hikari.pool.ProxyStatement.execute(ProxyStatement.java:94)
	at com.zaxxer.hikari.pool.HikariProxyStatement.execute(HikariProxyStatement.java)
	at com.yugabyte.sample.apps.SqlPgVector.createTablesIfNeeded(SqlPgVector.java:429)
	at com.yugabyte.sample.Main.run(Main.java:180)
	at com.yugabyte.sample.Main.main(Main.java:89)

```

Stress run: https://stress.dev.yugabyte.com/stress_test/8f908932-b15f-436e-829d-1ed9cc751da1

Timeline:

```
|         Time | Event                                                                                                                                  | Source                    |
|--------------+----------------------------------------------------------------------------------------------------------------------------------------+---------------------------|
| 00:21:23.330 | CREATE DATABASE db_1 (db oid **16640**) — created fresh                                                                                | 172.151.17.211 PG log     |
[...]

PR #32798, opened 2026-07-22
[#32385] YSQL: Add CREATE EXTENSION master failover coverage

bmcnama-shopify:
## Summary

Add deterministic coverage for creating the pgvector extension while the master leader restarts.

The test pauses installation after the first operator class and verifies that the extension row, `pg_amop` rows, and database catalog version remain uncommitted. It then kills and restarts the leader with single-OID prefetching to force the reservation retry path, and validates one outer commit, catalog integrity, vector execution, and clean drop/recreate behavior.

## Test plan

- [x] `./yb_build.sh release --cxx-test pg_ddl_atomicity-test --gtest_filter PgDdlAtomicityMasterFailoverTest.CreateVectorExtensionDuringLeaderRestart -n 10 --sj --skip-pg-parquet --no-odyssey --no-ybc`

### DIFF
--- a/src/yb/yql/pgwrapper/pg_ddl_atomicity-test.cc
+++ b/src/yb/yql/pgwrapper/pg_ddl_atomicity-test.cc
@@ -40,6 +40,7 @@
 #include "yb/util/async_util.h"
 #include "yb/util/backoff_waiter.h"
 #include "yb/util/monotime.h"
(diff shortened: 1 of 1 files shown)


iSignal:
Does it repro #32385, which was an intermittent failure? Ideally the new test would reproduce that failure deterministically?

PR #32818, opened 2026-07-22
[#32817] DocDB: Clean tablet snapshots asynchronously

### LINKED ISSUE #32817: [DocDB] Clean tablet snapshots asynchronously
Jira Link: [DB-22530](https://yugabyte.atlassian.net/browse/DB-22530)
## Description

Today, tablet snapshot directory deletion is performed synchronously during Raft apply. The apply path recursively deletes the snapshot directory tree while holding `ReplicaState::update_lock_` (and `Tablet::create_checkpoint_lock_`). For large snapshots this can block Raft replication and apply processing for the duration of the recursive filesystem deletion.

### Proposal

Move snapshot directory deletion to a two-phase, best-effort process:

1. **Logical delete (during Raft apply):** atomically rename the active snapshot directory to a reserved tombstone name and sync the parent directory. The operation returns success from Raft apply even if the rename or parent sync fails, preserving existing best-effort availability semantics. The recursive traversal no longer runs under the Raft locks.
2. **Physical delete (asynchronous):** delete successfully tombstoned directories on a dedicated, process-wide cleanup pool, outside `ReplicaState::update_lock_` and `Tablet::create_checkpoint_lock_`. Transient failures are retried in the background (bounded, without sleeping worker threads) while the tablet is alive and the path still exists.
3. **Recovery:** recognize and clean up tombstones left behind after a restart by their reserved directory-name suffix.

### Rollout

Gated behind a new runtime flag `enable_async_snapshot_directory_cleanup` (default `false`). When disabled, the existing synchronous best-effort recursive deletion is used, while still recognizing and cleaning any tombstones left by prior enablement. The asynchronous path never uses synchronous recursive deletion as a fallback.

No wire-format or tablet-metadata schema change is introduced; mixed-version and rollback safety are preserved.

### Issue Type

kind/enhancement

I confirm this issue does not contain any sensitive information.

[DB-22530]: https://yugabyte.atlassian.net/browse/DB-22530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

craigsoules:
## Summary

Tablet snapshot directory deletion previously ran synchronously during Raft apply, recursively deleting the snapshot directory tree while holding `ReplicaState::update_lock_` (and `Tablet::create_checkpoint_lock_`). For large snapshots this blocks Raft replication and apply processing for the full duration of the recursive filesystem deletion.

This change replaces that with a two-phase, best-effort deletion process, gated behind a new runtime flag `enable_async_snapshot_directory_cleanup` (default `false`).

### Logical delete (Raft apply)

`TabletSnapshots::Delete()` atomically renames the active snapshot directory to a reserved tombstone name and syncs the parent directory. The recursive traversal no longer runs under the Raft locks. The operation returns success from Raft apply even if the rename or parent sync fails — preserving the existing best-effort availability semantics — and never falls back to synchronous recursive deletion.

### Physical delete (asynchronous)

Successfully tombstoned directories are deleted on a dedicated, process-wide cleanup pool (`snapshot_cleanup_pool_size`, default `4`), outside `ReplicaState::update_lock_` and `Tablet::create_checkpoint_lock_`. A per-tablet serial token prevents concurrent cleanup passes for a single tablet, while the pool bounds cleanup concurrency across the process. Transient logical and physical failures are retried in the background — bounded and without sleeping worker threads — while the tablet is alive and the relevant path still exists. Epoch-based tracking guards against duplicate deletion requests racing with an active cleanup pass.

### Recovery

Startup scanning recognizes tombstones by their reserved directory-name suffix and reconciles them: it removes durable tombstones and recovers the active path where appropriate. Work queued before the cleanup pool is installed is preserved and drained once the pool is available.

### Consumers audited

Snapshot-directory consumers were audited so temporary (`*.tmp`) and logically-deleted (tombstoned) directories are not treated as live snapshots — in particular remote bootstrap does not advertise them, and the tserver snapshots web page reflects the new states.

### Observability

Adds tablet-level metrics: tombstone successes/failures, parent-sync failures, cleanup successes/failures/retries, pending logical and physical deletion gauges, and a gauge for the age of the oldest pending deletion.

No wire-format or tablet-metadata schema change is introduced.

Fixes #32817.

## Upgrade/Rollback safety

No protobuf, wire-format, or tablet-metadata schema change is introduced, so there is no cross-version RPC or on-disk metadata compatibility concern. The upgrade-relevant surface is the new behavior gated by flags plus the on-disk tombstone directories.

**Flags**
- `enable_async_snapshot_directory_cleanup` (runtime, default `false`): ships disabled. With the default, behavior is unchanged from today — snapshot deletion uses the existing synchronous best-effort recursive deletion. Intended rollout is: land disabled, enable in canary environments, monitor Raft latency / pending-deletion metrics / filesystem errors / disk usage, then flip the default in a follow-up.
- `snapshot_cleanup_pool_size` (non-runtime, default `4`): bounds process-wide cleanup concurrency; requires a restart to change.

**Forward (enable):** when the flag is turned on, deletes create reserved-suffix tombstone directories and clean them asynchronously. This is safe on a mixed-version cluster because the behavior is local to each node and involves no replicated format change.

**Rollback (disable / downgrade):** turning the flag back off (or downgrading to a build without this change) resumes synchronous deletion. Tombstone directories already created while the flag was enabled are still recognized and cleaned: when the flag is disabled the code continues to recognize and remove tombstones, and an older build ignores reserved-suffix/`*.tmp` directories as non-live snapshots. No manual cleanup step is required, and no tombstone can be mistaken for a live snapshot by remote bootstrap or restore.

**On-disk state:** tombstones are plain directories with a reserved name suffix under the tablet's top snapshots directory; they contain no new metadata format. A crash mid-deletion leaves either the tombstone (removed on next startup scan) or, in the rare rename-succeeded-but-sync-failed case, the original active directory (best-effort semantics, may remain) — matching the documented best-effort guarantees.

## Test plan

- [ ] New unit tests in `src/yb/tablet/tablet_snapshots-test.cc`:
  - [ ] `TabletSnapshotPathTest.AsyncCleanupDisabledByDefault`, `RecoverActivePath`
  - [ ] `TabletSnapshotsTest.RetainsDeletionUntilCleanupPoolIsInstalled`
  - [ ] `TabletSnapshotsTest.RetriesDirectoryCheckAndRenameFailures`, `RetriesParentSyncFailure`, `RetriesPhysicalDeletionFailure`, `RetriesFinalParentSyncFailure`
  - [ ] `TabletSnapshotsTest.CoalescesDuplicateDeletionRequests`
  - [ ] `TabletSnapshotsTest.ScanRecoversActiveAndTombstonePair`
  - [ ] `TabletSnapshotsTest.SharedPoolBoundsCleanupConcurrency`
  - [ ] `TabletSnapshotsTest.ShutdownWaitsForRunningCleanup`
- [ ] Integration coverage `SnapshotTest.DeleteSnapshotTombstoneCompletedBeforePhysicalCleanup` — verifies snapshot deletion completes once tombstoned without waiting for physical cleanup, that writes continue while physical cleanup is paused (`TEST_pause_after_tombstoning_snapshot`), and that tombstones are physically removed once cleanup is unpaused.
- [ ] Remote-bootstrap coverage extended to confirm temporary (`*.tmp`) and logically-deleted (tombstoned) snapshot directories are not advertised.
- [ ] Existing snapshot create/restore/delete regression suites pass with the flag both disabled (default) and enabled.


craigsoules:
trigger jenkins

bmcnama-shopify:
trigger jenkins

src/yb/tablet/tablet_snapshots.cc
```diff
@@ -71,18 +80,72 @@ DEFINE_test_flag(int32, delay_tablet_export_metadata_ms, 0,
...
 namespace yb::tablet {
 
 namespace {
 
 const std::string kTempSnapshotDirSuffix = ".tmp";
+const std::string kDeletedSnapshotDirSuffix = ".deleted.tmp";
 const std::string kTabletMetadataFile = "tablet.metadata";
 const std::string kLastSnapshotPrefix = "last_snapshot.";
 
+Result<bool> IsSnapshotDirectory(Env& env, const std::string& path) {
```
yamen-haddad: There is nothing in this function specific to snapshots.
Should we call it `IsDirectory`

craigsoules: Agreed there is nothing snapshot-specific here. Renamed in b22cfa918f, but to `DirectoryExists` rather than `IsDirectory`: `Env::IsDirectory` already exists with different semantics, while this helper maps NotFound to `false` and returns an error only when the check fails or the path exists but is not a directory.

---
_automated - Claude (claude-fable-5, pi)_

yamen-haddad:
Thanks for updating the PR. @craigsoules 
I am making another pass, and I hope to submit a full review tomorrow.

src/yb/tablet/tablet_snapshots.h
```diff
@@ -130,9 +157,53 @@ class TabletSnapshots : public TabletComponent {
 
   Result<docdb::CotableIdsMap> GetCotableIdsMap(const std::string& snapshot_dir);
 
+  enum class TombstoneState {
+    kComplete,
+    kTombstoned,
+    kBothPresent,
+    kRetry,
+  };
```
yamen-haddad, line 171: Can we add a one-liner comment explaining the meaning of each state? Especially the first 2 states.

src/yb/tablet/tablet_snapshots.cc
```diff
@@ -658,32 +846,367 @@ Result<std::string> TabletSnapshots::RestoreToTemporary(
...
+  }
+
+  if (*snapshot_dir_exists) {
+    const Status rename_status = env().RenameFile(deletion->active_dir, deletion->tombstone_dir);
+    if (PREDICT_FALSE(!rename_status.ok())) {
+      if (metrics_) {
+        metrics_->tombstone_failures->Increment();
+      }
+      LOG_WITH_PREFIX(WARNING) << "Cannot tombstone snapshot dir " << deletion->active_dir << " as "
+                               << deletion->tombstone_dir << ": " << rename_status;
```
yamen-haddad: << deletion->tombstone_dir << ", Rename status: " << rename_status;

src/yb/tablet/tablet_snapshots.h
```diff
@@ -130,9 +157,53 @@ class TabletSnapshots : public TabletComponent {
...
+
+  struct PendingSnapshotDeletion {
+    std::string active_dir;
+    std::string tombstone_dir;
+    MonoTime first_pending_at = MonoTime::Now();
+    bool logical_pending = true;
+    bool physical_pending = false;
+    bool parent_sync_pending = false;
```
yamen-haddad, line 186: This field tracks parent syncing for the cleanup phase only. Can you write a small comment explaining that with a small explanation how the tombstone parent syncing state can be inferred, so there is no need to be tracked?

I am just concerned in general about new readers of the code and the extra effort they need to figure out the state management. So some comments/clarifications can be very handy.

src/yb/tablet/tablet_snapshots.cc
```diff
@@ -658,32 +846,367 @@ Result<std::string> TabletSnapshots::RestoreToTemporary(
...
+      return TombstoneState::kRetry;
+    }
+    deletion->logical_pending = false;
```
yamen-haddad: This assignment is dead code, and it is misleading for readers. The decision to finish logical deletion is based on the durability phase below. So let's remove it.

src/yb/tablet/tablet_snapshots.cc
```diff
@@ -658,32 +846,367 @@ Result<std::string> TabletSnapshots::RestoreToTemporary(
...
+                               << deletion->tombstone_dir << ": " << sync_status;
+      return false;
+    }
+    deletion->parent_sync_pending = false;
+    deletion->physical_pending = false;
+    PublishPendingDeletionState(*deletion);
+    return !deletion->logical_pending;
+  }
+
+  const auto tombstone_state = TryTombstoneSnapshotDir(deletion);
```
yamen-haddad: This call is unconditional, and it acquires the `create_checkpoint_lock ` and also does an fsync unconditionally.
This means that in the happy path, we will call `TryTombstoneSnapshotDir` twice for every snapshot.

Should we guard retrying to tombstone only when `deletion->logical_pending` is true?
i.e.:
```
if (deletion->logical_pending) {
    const auto tombstone_state = TryTombstoneSnapshotDir(deletion);
    PublishPendingDeletionState(*deletion);
    if (tombstone_state == TombstoneState::kComplete) {
      return true;
    }
    if (tombstone_state == TombstoneState::kRetry) {
      return false;
    }
  } 
```

I tracked the state machine and it looks like we will only set `deletion->logical_pending` to false when the tombstone already succeeded.
Otherwise, deletion->logical_pending will always be true and thus we will not miss calling `TryTombstoneSnapshotDir`.

src/yb/tablet/tablet_snapshots.h
yamen-haddad: Example for the `parent_sync_pending`.
// Tracks whether an fsync is owned for the parent directory after recursive delete of the tombstone.

src/yb/tablet/tablet_snapshots.cc
craigsoules: Good catch on the redundant lock and fsync - adopted in dd0064d3a9, with one addition beyond the suggested guard. Skipping `TryTombstoneSnapshotDir` also skips its `kComplete` detection, which today is what handles a tombstone that is already gone. That case is reachable: if a coalesced duplicate request bumps the entry's epoch while a cleanup pass is mid-deletion, the pass's completion is discarded, and the next pass (with `logical_pending == false`) would call `DeleteRecursively` on a nonexistent directory. The `CoalescesDuplicateDeletionRequests` test caught exactly this with the guard applied verbatim. So the skip path now does a cheap lock-free existence check (no checkpoint lock, no fsync) and completes immediately when the tombstone is already gone, and `DeleteRecursively` treats NotFound as already-removed as defense in depth. Also added a test asserting the happy path performs exactly two parent syncs.

---
_automated - Claude (claude-fable-5, pi)_

craigsoules:
@yamen-haddad Pushed new changes to address your review comments.  I agreed with the AI's assessment on your last comment, so I let it adjust accordingly.  That would be the one thing I'd double check the thinking on.

src/yb/tablet/tablet_snapshots.cc
```diff
@@ -658,32 +846,384 @@ Result<std::string> TabletSnapshots::RestoreToTemporary(
...
+    return;
+  }
+  const auto delay = FLAGS_TEST_snapshot_cleanup_retry_delay_ms > 0
+                         ? FLAGS_TEST_snapshot_cleanup_retry_delay_ms * 1ms
+                         : 1s * (1 << std::min(retry_attempt_++, 6U));
+  retry_scheduled_ = true;
+  if (metrics_) {
+    metrics_->cleanup_retries->Increment();
+  }
+  UpdatePendingMetricsUnlocked();
```
yamen-haddad: Why do we need to update metrics here? We didn't change any state related to `pending_deletions_`

craigsoules: Think you're right... looks like it didn't get cleaned up from an earlier commit 👍

yamen-haddad:
Thanks @craigsoules ! I am doing a final pass now, and things look good overall for the workflow.

src/yb/tablet/tablet_snapshots-test.cc
```diff
@@ -0,0 +1,546 @@
...
+  ASSERT_OK(DeleteSnapshot("snapshot.rename.retry", 2));
+  ASSERT_OK(WaitFor(
+      [this] { return test_env_->active_deletes() == 1; }, 10s,
+      "Wait for retried snapshot deletion"));
+  ASSERT_EQ(MetricValue<AtomicGauge<uint64_t>>(METRIC_snapshot_pending_logical_deletions), 0);
+  ASSERT_EQ(MetricValue<AtomicGauge<uint64_t>>(METRIC_snapshot_pending_physical_deletions), 1);
+  test_env_->ReleaseDeletes();
+  ASSERT_OK(WaitForRemoved(paths));
+  ASSERT_GE(MetricValue<Counter>(METRIC_snapshot_tombstone_failures), 4);
+  ASSERT_GE(MetricValue<Counter>(METRIC_snapshot_cleanup_retries), 1);
```
yamen-haddad: Any reason the assertion here is loose (greater than 1)? In theory, we should have at least 3 snapshot_cleanup_retries. (4 failure injections - 1 TryTombstone called as part of Delete())

src/yb/tablet/tablet_snapshots-test.cc
```diff
@@ -0,0 +1,546 @@
...
+  ASSERT_EQ(MetricValue<AtomicGauge<uint64_t>>(METRIC_snapshot_pending_physical_deletions), 1);
+
+  test_env_->ReleaseDeletes();
+  ASSERT_OK(WaitForRemoved(paths));
+  ASSERT_OK(WaitFor(
+      [this] {
+        return MetricValue<AtomicGauge<uint64_t>>(METRIC_snapshot_pending_physical_deletions) == 0;
+      },
+      10s, "Wait for duplicate snapshot deletion to complete"));
+  ASSERT_EQ(test_env_->delete_calls(), 1);
```
yamen-haddad, line 412: Should we add an assert that a cleanup has been rescheduled for this deletion as the epoch was bumped?
`  ASSERT_EQ(MetricValue<Counter>(METRIC_snapshot_cleanup_retries), 1); `

src/yb/tablet/tablet_snapshots-test.cc
```diff
@@ -0,0 +1,546 @@
...
+    operation.AllocateRequest()->CopyFrom(request);
+    operation.set_op_id(OpId(1, op_index));
+    ASSERT_OK(tablet_harness->tablet()->snapshots().Delete(operation));
+    harnesses.push_back(std::move(tablet_harness));
+  }
+
+  ASSERT_OK(WaitFor(
+      [this] { return test_env_->active_deletes() == 2; }, 10s,
+      "Wait for snapshot cleanup pool saturation"));
+  ASSERT_EQ(test_env_->max_active_deletes(), 2);
```
yamen-haddad: Nit: I would define a constant `kCleanupThreadPoolSize = 2` and use it in the Setup() of the test, and use it here in the assertion.

craigsoules:
@yamen-haddad Accepted all of the changes.  Thanks!

Just a heads up that once this is merged I will be following up with a PR to improve snapshot creation 😃

src/yb/tablet/tablet_snapshots.cc
```diff
@@ -658,32 +846,383 @@ Result<std::string> TabletSnapshots::RestoreToTemporary(
...
+  }
+
+  TEST_PAUSE_IF_FLAG(TEST_pause_after_tombstoning_snapshot);
+  {
+    SCOPED_WAIT_STATUS(Snapshot_CleanupSnapshotDir);
+    // NotFound means the tombstone is already gone (e.g. a previous pass removed it but its
+    // completion was discarded due to an epoch bump); treat it as removed and continue to the
+    // parent sync.
+    const Status deletion_status = env().DeleteRecursively(deletion->tombstone_dir);
+    if (PREDICT_FALSE(!deletion_status.ok() && !deletion_status.IsNotFound())) {
```
yamen-haddad: Went into a rabbit hole following this last change. My understanding is that `Env::DeleteRecursively` never returns `NotFound` in Posix, so this check is dead, and the comment describes behavior that doesn't exist. I can see why this has been added to avoid DeleteRecursively walking into a directory that might not exist.
So what about dropping the comment and the new check, as this is dead code?
The only side effect in this rare case is that we will increment `cleanup_successes` once for a no-op, which I feel is tolerable noise compared to the code complexity we might introduce on top of already complex branches.

craigsoules: Makes sense.  Pushed the change 👍

src/yb/integration-tests/snapshot-test.cc
```diff
@@ -560,9 +588,65 @@ TEST_F(SnapshotTest, CreateSnapshot) {
   ASSERT_OK(cluster_->RestartSync());
 }
 
+// Verifies that physical cleanup only starts after the snapshot directory is tombstoned.
+TEST_F(SnapshotTest, DeleteSnapshotTombstonesBeforePhysicalCleanup) {
```
yamen-haddad: I feel the test purpose here is not expressed in the test name and the first comment. The test main point is to assert that delete snapshots completes after tombstoning.
What about the following name: `DeleteSnapshotCompletesBeforePhysicalCleanup` or `LogicalSnapshotDeleteDoesNotWaitForPhysicalCleanup` or something along these lines.

src/yb/integration-tests/snapshot-test.cc
```diff
@@ -560,9 +588,65 @@ TEST_F(SnapshotTest, CreateSnapshot) {
...
+  ASSERT_OK(delete_status);
+
+  workload.Start();
+  workload.WaitInserted(100);
+  workload.StopAndJoin();
+  ASSERT_GE(workload.rows_inserted(), 100);
+
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_pause_after_tombstoning_snapshot) = false;
+  delete_thread_holder.JoinAll();
+}
```
yamen-haddad, line 649: Optional: should we close the loop and check that tombstoned directories have been cleaned up?

src/yb/integration-tests/snapshot-test.cc
craigsoules: `DeleteSnapshotTombstoneCompletedBeforePhysicalCleanup`?

src/yb/tablet/tablet_bootstrap.cc
```diff
@@ -1998,12 +1998,15 @@ class TabletBootstrap {
           const string snapshot_dir = JoinPathSegments(top_snapshots_dir, dir_name);
 
           if (TabletSnapshots::IsTempSnapshotDir(snapshot_dir)) {
-            LOG_WITH_PREFIX(INFO) << "Deleting old temporary snapshot directory " << snapshot_dir;
+            const auto snapshot_dir_type =
+                TabletSnapshots::IsDeletedSnapshotDir(snapshot_dir) ? "deleted" : "temporary";
```
yamen-haddad: Nit: tombstoned instead of deleted as `snapshot_dir_type`

src/yb/integration-tests/snapshot-test.cc
```diff
@@ -118,6 +124,28 @@ using master::TableIdentifierPB;
...
+      snapshot_dir);
+  ASSERT_EQ(
+      ASSERT_RESULT(
+          tablet::TabletSnapshots::ActiveSnapshotDirFromDeletedSnapshotDir(
+              "/tmp/snapshots/snapshot.id.with.dots.7.1234.deleted.tmp")),
+      "/tmp/snapshots/snapshot.id.with.dots");
+  ASSERT_FALSE(tablet::TabletSnapshots::IsDeletedSnapshotDir(snapshot_dir + ".tmp"));
+  ASSERT_FALSE(tablet::TabletSnapshots::IsDeletedSnapshotDir(
+      snapshot_dir + ".invalid.1234.deleted.tmp"));
+}
```
yamen-haddad: This test repeats the test `RecoverActivePath` in `tablet_snapshots-test.cc`. Let's remove this test and keep the unit test there. Maybe add `IsDeletedSnapshotDir` asserts there, as this file is more for integration tests.

yamen-haddad (approved):
The diff LGTM. Please take a look at the minor comments from the last pass.
@craigsoules  Thanks for working on the async delete snapshot. Looking forward the create snapshot diff 😄

src/yb/integration-tests/snapshot-test.cc
yamen-haddad: `DeleteSnapshotTombstoneCompletedBeforePhysicalCleanup` Is fine.

rahulb-yb:
trigger jenkins

PR #32828, opened 2026-07-23
[#10492] DocDB: Support maximum replicas per placement block

### LINKED ISSUE #10492: [DocDB] Support max_num_replicas in placement config
Jira Link: [DB-763](https://yugabyte.atlassian.net/browse/DB-763)
### Description

Feature request; Currently, placement constraints are [specified as follows:](https://github.com/yugabyte/yugabyte-db/blob/10f931bc103158f355d3b8152d0cc09c75823c93/src/yb/master/master.proto#L216)

```
Overall RF
List [
   (region/AZ, min_num_replicas in that region/AZ)
]
```

This doesn't allow expressing situations like creating RF3 tables on 4 nodes across 4 AZs and letting the master choose some 3 out of the 4 nodes. Instead, we need to pick 3 AZs explicitly and set min_num_replicas = 1 for each of them, causing the 4th node to be unused instead of distributing tablets evenly.

```
RF3
[
    AZ1, min=1
    AZ2, min=1
    AZ3, min=1
]
```

Supporting a max_num_replicas param could possibly allow this to be specified as 
```
RF3
[
      AZ1, min=0, max=1,
      AZ2, min=0, max=1
      AZ3, min=0, max=1
      AZ4, min=0, max=1
]
```

Or, more generally
```
RF=N
[ 
       AZi, min=0, max=floor(N/2),
       ....
]
```

bmcnama-shopify:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/32828/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

## Summary

Add an optional `max_num_replicas` constraint to placement blocks used by
cluster, table, and tablespace policies. When omitted, the effective maximum
is the placement replication factor, preserving existing behavior and metadata.

Parse and validate the constraint for primary and read-replica placements,
extend `yb-admin` placement syntax to `cloud.region.zone[:min[:max]]`, and
enforce the bound during initial replica assignment and ongoing balancing.
The maximum is a hard cap on placement decisions: table creation fails if the
caps make a quorum infeasible, and the balancer never adds a replica above a
block's maximum, except transiently during same-block replacement (for
example, draining a blacklisted server), where the add precedes the remove.

A placement that specifies any maximum must consist solely of fully-qualified,
non-duplicate placement blocks, so every tserver is unambiguously attributed
to one block; wildcard blocks cannot be combined with maxima.

The balancer treats under-replication, over-replication, and blocks above
their maximum as mutually exclusive handling states: under-replication is
repaired first, blocks above their maximum are repaired add-before-remove, and
removals for over-replicated tablets are steered to blocks above their
maximum. The load balancer's time-to-balance estimate does not account for
maxima and can be low when they are binding.

Closes #10492

## Upgrade/Rollback safety

`PlacementBlockPB.max_num_replicas` is an additive optional proto2 field. No
catalog migration or metadata backfill is required; absent values continue to
mean the placement replication factor.

Explicit maxima can only be written by new binaries (the new `yb-admin`
syntax and tablespace JSON field); old binaries preserve and ignore the
unknown field. In a mixed-version cluster an old master leader does not
enforce stored maxima, so placement can transiently exceed a cap until a
new-version leader repairs it on a subsequent balancer run — degraded but
self-healing, with no correctness impact. Rollback remains wire-compatible:
older binaries ignore the field and revert to legacy placement behavior
without rewriting metadata.

## Test plan

- [x] `build-support/lint.sh --rev upstream/master`
- [x] `TablespaceParserTest.MaxNumReplicasParsing`
- [x] `TablespaceParserTest.MaxNumReplicasValidation`
- [x] `CatalogManagerUtilTest.MaxNumReplicasValidation`
- [x] `TestPlacementInfoContainsPlacementInfo.TestMaxReplicas`
- [x] `LoadBalancerRF5MaxReplicasMockedTest.RepairPlacementAboveMaximum`
- [x] `LoadBalancerRF5MaxReplicasMockedTest.UnderReplicationPrioritizedOverMaxPlacement`
- [x] `LoadBalancerMaxReplicasBlacklistMockedTest.BlacklistedMoveAllowedAtMaxPlacement`
- [x] `LoadBalancerRF5MaxReplicasManyTabletsMockedTest.BalancedLoadStillRepairsOverMaxPlacement`
- [x] `OptimalLoadDistributionTest.Slack`
- [x] `OptimalLoadDistributionTest.SlackManyTservers`
- [x] `load_balancer_mocked-test` (all tests)
- [x] `load_balancer_placement_policy-test` (all tests, including `CreateTableRespectsMaxNumReplicas` and `MaxNumReplicasIsAHardCap`)
- [x] `yb-admin_client-test` (all tests)
- [x] `common_net-test` (all tests)
- [x] `AdminCliTest.TestModifyPlacementPolicy`


bmcnama-shopify:
trigger jenkins

src/yb/common/tablespace_parser-test.cc
```diff
@@ -586,6 +588,84 @@ TEST(TablespaceParserTest, ReadReplicaPlacementParsingErrors) {
...
+      R"({"num_replicas":1,"placement_blocks":[
+        {"cloud":"c","region":"r","zone":"z","min_num_replicas":1,
+         "max_num_replicas":2}]})",
+      ""));
+}
+
+TEST(TablespaceParserTest, MaxNumReplicasDisabledBeforeAutoFlagPromotion) {
+  google::FlagSaver flag_saver;
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_enable_placement_block_max_num_replicas) = false;
+  const auto replication_info = ASSERT_RESULT(TablespaceParser::FromString(
```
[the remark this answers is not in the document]
bmcnama-shopify: Confirmed: the test as written could not pass, and the test-plan checkmark was stale. Fixed by flipping the test's expectation rather than gating the aggregate on the AutoFlag.

Validation of explicit maxima is intentionally always-on, matching `CatalogManagerUtil::IsPlacementInfoValid` (`CatalogManagerUtilTest.MaxNumReplicasValidation` and `LoadBalancerRF5MaxReplicasMockedTest.RejectInvalidMaximumsOnActivation` both assert rejection with the flag off); only enforcement in placement decisions is gated on promotion via `GetEffectiveMaxNumReplicas`. Gating the parser's aggregate check instead would let `CREATE TABLESPACE` persist a placement that table creation and every balancer run would then reject.

The test now parses a feasible placement to show explicit maxima are preserved but not effective before promotion, and asserts the infeasible aggregate is rejected regardless of flag state. Full affected unit suites re-ran green locally against the rebased head.

---
_automated · OpenCode (claude-fable-5)_

bmcnama-shopify:
trigger jenkins

neera-mital:
Thanks @bmcnama-shopify  for the PR, and sorry for the delayed response. @basavaraj29  is taking the review and aiming to get you initial feedback by Friday.  Appreciate your patience, feel free to tag me if anything needs unblocking.

iSignal:
@bmcnama-shopify : Can we rebase and retrigger unit tests in the mean time? 

```
[yb-build/alma8-gcc15-fastdebug](https://jenkins.dev.yugabyte.com/job/yugabyte-db-phabricator/491296/)
yb-build/alma8-gcc15-fastdebug — Rebase or Build failure
[yb-csi/tests-all-ran](https://csiweb.dev.yugabyte.com/pull/32828)
yb-csi/tests-all-ran — Test runs were incomplete
[yb-required](https://jenkins.dev.yugabyte.com/job/csi-diff-summary/146330/)
yb-required — Failed 8 required checks
```

src/yb/common/common_flags.cc
```diff
@@ -41,6 +41,9 @@ DEFINE_RUNTIME_AUTO_bool(enable_automatic_tablet_splitting, kExternal, false, tr
     "If false, disables automatic tablet splitting driven from the yb-master side, and in this "
     "case the value of tserver's ysql_num_tablets is recommended to be set to -1.");
 
+DEFINE_RUNTIME_AUTO_bool(enable_placement_block_max_num_replicas, kExternal, false, true,
```
basavaraj29: I don't see a valid reason to have an auto flag for this feature. in mixed mode (during upgrades) too, I don't see the existing load balancing algorithm disturbing the ideal state achieved by the new algorithm.

additionally, auto flags are meant to be untouched by end users. if we really see the need to protect this feature under a flag, it should be a normal runtime gflag.

src/yb/common/common_net.cc
```diff
@@ -162,6 +171,33 @@ bool PlacementInfoContainsPlacementInfo(const PlacementInfoPB& lhs, const Placem
...
 
+  const auto has_explicit_max = [](const PlacementInfoPB& placement) {
+    return std::ranges::any_of(
+        placement.placement_blocks(),
+        [](const auto& block) { return block.has_max_num_replicas(); });
+  };
+  if (has_explicit_max(lhs) || has_explicit_max(rhs)) {
+    if (lhs.num_replicas() != rhs.num_replicas() ||
+        lhs.placement_blocks_size() != rhs.placement_blocks_size()) {
+      return false;
```
basavaraj29, line 190: counter example
```
lhs = num_replicas: 2, placement_blocks: { a.*:1:2 }
rhs = num_replicas: 2, placement_blocks: { a.b*:1:1,  a.c*1:1}
```

the current logic is totally wrong. refer the comment below on how the matching needs to be done.

src/yb/common/common_net-test.cc
```diff
@@ -239,6 +243,28 @@ TEST(TestPlacementInfoContainsPlacementInfo, TestUUID) {
...
+            { .min_replicas = 1, .placement = "c1.r1.z2", .max_replicas = 2 }} },
+      { .replicas = 3,
+        .blocks = {
+            { .min_replicas = 1, .placement = "c1.r1.z2", .max_replicas = 2 },
+            { .min_replicas = 1, .placement = "c1.r1.z1", .max_replicas = 2 }} }));
+  ASSERT_TRUE(PlacementInfoContains(
+      { .replicas = 3,
+        .blocks = {{ .min_replicas = 1, .placement = "c1.r1.z1" }} },
+      { .replicas = 3,
+        .blocks = {{ .min_replicas = 1, .placement = "c1.r1.z1", .max_replicas = 3 }} }));
```
basavaraj29, line 260: if we swap lhs rhs here, what do you expect to happen? add that as a case.

also requires more testing with wildcard placements etc

src/yb/master/catalog_manager.cc
```diff
@@ -2555,6 +2555,21 @@ Status CatalogManager::ValidateTableReplicationInfo(
...
+    }
+    return Status::OK();
+  };
+  if (replication_info.has_live_replicas()) {
+    RETURN_NOT_OK(validate_explicit_max(replication_info.live_replicas()));
+  }
+  for (const auto& read_replicas : replication_info.read_replicas()) {
+    RETURN_NOT_OK(validate_explicit_max(read_replicas));
+  }
+
```
basavaraj29: earlier, we weren't checking `IsPlacementInfoValid` over here. looks like that was intentional. can remove this

src/yb/master/catalog_manager.cc
```diff
@@ -5308,13 +5323,16 @@ Status CatalogManager::CheckValidPlacementInfo(const PlacementInfoPB& placement_
...
+      size_t max_num_replicas =
+          GetEffectiveMaxNumReplicas(pb, narrow_cast<int32_t>(num_replicas));
       // For every placement block, we can only satisfy upto the number of
       // tservers present in that particular placement block.
       total_feasible_replicas += min(allowed_ts_size, min_num_replicas);
       // Extra tablet servers beyond min_num_replicas will be used to place
       // the extra replicas over and above the minimums.
-      if (allowed_ts_size > min_num_replicas) {
-        total_extra_servers += allowed_ts_size - min_num_replicas;
+      size_t max_feasible_replicas = min(allowed_ts_size, max_num_replicas);
```
basavaraj29, line 5341: this will make the policy stricter.

earlier, we allow create table to go through despite not having `min_replicas` number of tservers in the placement info. the only check being done here was if we are able to place majority peers.

now, something like a:1:2, b:1:2, c:1:1 would fail when the cluster currently just has a(3). I think it would still be ok to allow a(3) now, and then load balance to a(2), b(2), c(1) once the nodes are available.

basavaraj29: can update the nearby comments accordingly.

src/yb/master/catalog_manager.cc
```diff
@@ -12332,19 +12350,38 @@ Status CatalogManager::HandlePlacementUsingPlacementInfo(const PlacementInfoPB&
...
+      SelectReplicas(
+          {selected}, 1, config, &already_selected_ts, member_type, per_table_state, global_state);
+      for (size_t i = 0; i != implicit_cast<size_t>(placement_info.placement_blocks_size()); ++i) {
+        if (selected->MatchesCloudInfo(
+                placement_info.placement_blocks(narrow_cast<int>(i)).cloud_info())) {
+          ++selected_per_block[i];
+          break;
+        }
+      }
+      --replicas_left;
```
basavaraj29: below approach seems more intuitive and simpler 
```
size_t max_tservers_left = all_allowed_ts.size() - already_selected_ts.size();
// Upper bounded by the tservers left.
replicas_left = min(replicas_left, max_tservers_left);
DCHECK_GE(replicas_left, 0);
// TSDescriptorVector candidates;
// remaining_ts = set_difference{all_allowed_ts, already_selected_ts}
// sort remaining_ts by load
// for it : remaining_ts:
//   ts = *it
//   if selected_per_block[ts->placement_info] <= placement_info_map[ts->placement_info].max_num_replicas || replicas_left >= remaining_ts.end() - it
       candidates.emplace_back(ts);
       selected_per_block[ts->placement_info]++;
       replicas_left--;
// SelectReplicas(candidates, candidates.size(), config, &already_selected_ts, member_type, per_table_state, global_state)
```

This will also cover the above mentioned case of placing a(3) when b,c isn't available and the policy is a:1:2, b:1:2, c:1:2

src/yb/master/catalog_manager.cc
```diff
@@ -13190,6 +13227,20 @@ Result<int32_t> CatalogManager::GetClusterConfigVersion() {
...
+            [](const auto& block) { return block.has_max_num_replicas(); })) {
+      return CatalogManagerUtil::IsPlacementInfoValid(placement_info);
+    }
+    return Status::OK();
+  };
+  RETURN_NOT_OK(validate_explicit_max(replication_info.live_replicas()));
+  for (const auto& read_replicas : replication_info.read_replicas()) {
+    RETURN_NOT_OK(validate_explicit_max(read_replicas));
+  }
+
```
basavaraj29, line 13323: remove, seems like we skip the check here intentionally

src/yb/master/cluster_balance.cc
```diff
@@ -275,6 +275,11 @@ Status ClusterLoadBalancer::PopulateReplicationInfo(
   if (state_->placement_.num_replicas() == 0 && has_read_replicas) {
     state_->placement_.set_num_replicas(FLAGS_replication_factor);
   }
+  if (std::ranges::any_of(
+          state_->placement_.placement_blocks(),
+          [](const auto& block) { return block.has_max_num_replicas(); })) {
+    RETURN_NOT_OK(CatalogManagerUtil::IsPlacementInfoValid(state_->placement_));
+  }
```
basavaraj29: remove

src/yb/master/cluster_balance_util.h
```diff
@@ -75,6 +75,12 @@ struct CBTabletMetadata {
   // Set of placement ids that have less replicas available than the configured minimums.
   std::unordered_set<CloudInfoPB, cloud_hash, cloud_equal_to> under_replicated_placements;
 
+  // Current running and starting replica counts per placement block.
+  std::unordered_map<CloudInfoPB, size_t, cloud_hash, cloud_equal_to> placement_replica_counts;
+
+  // Replicas in placement blocks that exceed their configured maximum.
+  std::set<TabletServerId> over_max_replicated_tablet_servers;
```
basavaraj29: `over_max_placement_tablet_servers` seems like a better name

src/yb/master/cluster_balance.cc
```diff
@@ -1041,6 +1079,11 @@ Result<bool> ClusterLoadBalancer::HandleAddReplicas(
 
   // Missing placements / under-replicated tablets are handled in ProcessUnderReplicatedTablets.
 
+  if (VERIFY_RESULT(
+          HandleAddIfOverMaxPlacement(out_tablet_id, out_from_ts, out_to_ts))) {
+    return true;
+  }
+
```
basavaraj29: this should be lower priority compared to `HandleAddIfWrongPlacement`. should go between `HandleAddIfWrongPlacement` and `normal cluster balancing`.

src/yb/master/cluster_balance_util.cc
basavaraj29: modify this to `placement_to_min_max_replicas` may be?

src/yb/master/cluster_balance_util.cc
```diff
@@ -412,24 +419,47 @@ Status PerTableLoadState::UpdateTablet(TabletInfo *tablet) {
...
+    std::set<TabletServerId> generic_removal_candidates;
     // Loop over the data and populate extra replica as well as missing replica information.
     for (const auto& [cloud_info, replicas] : placement_to_replicas) {
       const size_t min_num_replicas = placement_to_min_replicas[cloud_info];
+      const auto placement_block = std::find_if(
+          placement_.placement_blocks().begin(), placement_.placement_blocks().end(),
+          [&cloud_info](const auto& block) {
+            return cloud_equal_to()(block.cloud_info(), cloud_info);
+          });
+      DCHECK(placement_block != placement_.placement_blocks().end());
```
basavaraj29: remove this, wouldn't need it after you address the above comments.

src/yb/master/cluster_balance_util.cc
```diff
@@ -412,24 +419,47 @@ Status PerTableLoadState::UpdateTablet(TabletInfo *tablet) {
...
         VLOG(3) << "Placement " << cloud_info.ShortDebugString() << " is under-replicated by"
                 << " " << min_num_replicas - replicas.size() << " count";
         // Placements that are under-replicated should be handled ASAP.
         tablet_meta.under_replicated_placements.insert(cloud_info);
+      }
+      if (replicas.size() > max_num_replicas) {
+        VLOG(3) << "Placement " << cloud_info.ShortDebugString() << " exceeds its maximum by "
+                << replicas.size() - max_num_replicas << " replicas";
+        for (const auto& [ts_uuid, _] : replicas) {
+          tablet_meta.over_max_replicated_tablet_servers.insert(ts_uuid);
```
basavaraj29: `over_max_replicated_tablet_servers` name is misleading, and could give a wrong impression of over replicated

src/yb/master/cluster_balance_util.cc
basavaraj29, line 346: track `tablet_meta.is_over_max_placements` too may be?

src/yb/master/cluster_balance_util.cc
```diff
@@ -412,24 +419,47 @@ Status PerTableLoadState::UpdateTablet(TabletInfo *tablet) {
...
       } else if (tablet_meta.is_over_replicated && min_num_replicas < replicas.size()) {
         // If this tablet is over-replicated, consider all the placements that have more than the
         // minimum number of tablets, as candidates for removing a replica.
         VLOG(3) << "Placement " << cloud_info.ShortDebugString() << " is over-replicated by"
                 << " " << replicas.size() - min_num_replicas << " count";
         for (const auto& [ts_uuid, _] : replicas) {
-          tablet_meta.over_replicated_tablet_servers.insert(ts_uuid);
+          generic_removal_candidates.insert(ts_uuid);
         }
       }
```
basavaraj29, line 462: this ordering seems problematic, I think it is better we switch it to
```
if (min_num_replicas > replicas.size()) {
  ...
} else if (tablet_meta.is_over_replicated) {
  if (tablet_meta.is_over_max_placements) {
    if (replicas.size() > max_num_replicas) {
      // add to over replicated list
      // this will make HandleRemoveReplicas consider placements with over max replicas alone
    }
  } else if (min_num_replicas < replicas.size()) {
    // add to over replicated list
    // if there's no placement with over max replicas, then remove based on load by considering all candidates with > min_replicas
  }
  ...
} else if (tablet_meta.is_over_max_placements) {
  ...
  // this will add a replica before triggering the over-replication case
}
```

so say you had A(3), B(1), C(1). the suggested order would move us to A(3), B(2), C(1) first, and then take just the over replication path and kick a peer out from A, giving us A(2), B(2), C(1). 

we need not take the max placements violation path when invoking the over-replication path. makes sense?

src/yb/master/cluster_balance_util.cc
```diff
@@ -412,24 +419,47 @@ Status PerTableLoadState::UpdateTablet(TabletInfo *tablet) {
...
     }
+    if (!tablet_meta.over_max_replicated_tablet_servers.empty()) {
+      tablet_meta.over_replicated_tablet_servers =
+          tablet_meta.over_max_replicated_tablet_servers;
+      tablets_over_max_placements_.insert(tablet_id);
```
basavaraj29: why are we doing this? `over_max_placements` doesn't imply over replication, right?

say you had rf5 with policy A:1:3, B:1:3, C:1:3, and the placement was A(3), B(1), C(1), and then you changed it to *:1:2.

Now, since A has > max num placements, your implementation would probably kick out a peer from A before adding another replica because you are wrongly adding the tserver to `over_replicated_tablet_servers`

neither imply the other. we can have over replication but may or may not violate max placements in a zone. conversely, we can have max placements violation but may or may not have over replication.

src/yb/master/cluster_balance_util.cc
```diff
@@ -412,24 +419,47 @@ Status PerTableLoadState::UpdateTablet(TabletInfo *tablet) {
...
+    } else {
+      tablet_meta.over_replicated_tablet_servers = std::move(generic_removal_candidates);
```
basavaraj29: don't think we need to track things this way. under replication, over replication and exceeding max num placements are different things, and would need different handling in the above if else chain

src/yb/master/cluster_balance_util.cc
basavaraj29, line 491: add a similar chain for over max placements case.

src/yb/master/cluster_balance_util.cc
```diff
@@ -539,7 +569,7 @@ void PerTableLoadState::UpdateTabletServer(std::shared_ptr<TSDescriptor> ts_desc
 }
 
 Result<bool> PerTableLoadState::CanAddTabletToTabletServer(
-    const TabletId& tablet_id, const TabletServerId& to_ts) {
+    const TabletId& tablet_id, const TabletServerId& to_ts, const TabletServerId& from_ts) {
```
basavaraj29, line 586: please revert changes to this function. we shouldn't be restricting moves at this level just because we are exceeding max_num_replicss. this is used in the blacklist path, and we would essentially be blocking node removal etc just because we didn't want to move a replica to another node since the zone had max_num_replicas already.

src/yb/master/cluster_balance_util.cc
```diff
@@ -718,8 +773,9 @@ Result<bool> PerTableLoadState::CanSelectWrongPlacementReplicaToMove(
       VLOG(3) << out.str();
     }
     for (const auto& to_uuid : sorted_load_) {
-      if (VERIFY_RESULT(CanAddTabletToTabletServer(tablet_id, to_uuid))) {
-        *out_from_ts = *tablet_meta.wrong_placement_tablet_servers.begin();
+      const auto& from_uuid = *tablet_meta.wrong_placement_tablet_servers.begin();
+      if (VERIFY_RESULT(CanAddTabletToTabletServer(tablet_id, to_uuid, from_uuid))) {
```
basavaraj29, line 797: it is ok for us to resolve the wrong placement first and then resolve over placements

src/yb/master/cluster_balance_util.cc
basavaraj29, line 715: shouldn't be blocking movement over `max_num_replicas` for blacklist as stated above

src/yb/master/cluster_balance_util.cc
```diff
@@ -760,6 +816,7 @@ Status PerTableLoadState::RemoveReplica(const TabletId& tablet_id, const TabletS
   // It's possible that the tablet is over-replicated multiple times, but we won't remove multiple
   // replicas in one run anyways because we only iterate over the over-replicated tablets once.
   tablets_over_replicated_.erase(tablet_id_key);
+  tablets_over_max_placements_.erase(tablet_id_key);
```
basavaraj29, line 843: think we won't have the tablet here if you address my previous comments.

src/yb/master/cluster_balance_util.cc
basavaraj29, line 1138: I'm not reviewing the changes in this function. the logic should be similar to what I described earlier here https://github.com/yugabyte/yugabyte-db/pull/32828/changes#r3738941726

src/yb/master/cluster_balance_util.cc
```diff
@@ -1089,51 +1153,32 @@ Result<TsTableLoadMap> CalculateOptimalLoadDistribution(
...
-    //             those tservers could be assigned without exceeding the load on tserver i.
-    //  prefix_load = the ACTUAL sum of the load of tservers 0 to i-1.
-    //
-    // If i * load > prefix_load + slack, then we can put both the existing load of tservers 0 to
-    // i-1 PLUS all the slack into this rectangle.
-    // In the above chart, if 2 <= slack <= 4 then the algorithm halts at i == 3 because we can add
-    // the slack to T0, T1, and T2 without their load exceeding T3's load of 3.
-    // However if slack is 5 or greater than we cannot fit these 5 additional tablet replicas into
-    // this rectangle, which means we cannot evenly distribute slack across T0, T1, and T2 without
-    // their load exceeding that of T3's. So the loop iterates past i == 3.
```
basavaraj29, line 1129: please try to rework this function with minimal changes to the logic in here. also, why are we trying to delete this comment? It explains the current algorithm in detail and is useful.

src/yb/master/load_balancer_mocked-test.cc
```diff
@@ -909,6 +912,67 @@ class LoadBalancerRF5MockedTest : public LoadBalancerMockedTest {
   int NumReplicas() const override { return 5; }
 };
 
+class LoadBalancerRF5MaxReplicasMockedTest : public LoadBalancerRF5MockedTest {
```
basavaraj29, line 915: needs more targeted testing covering the gaps I highlighted as part of the review comments. they might go in `load_balancer_mini_clustert-test` I suppose.

1. test table creation that violates max_num_replicas when desired placements are unavailable but majority are available. assert that the cluster gets back to normal once nodes are added/come back up (also can have a case where max isn't violated).
2. blacklist shouldn't get blocked due to max_num_replicas
3. assert that add/move as part of fixing max placements violating isn't leading to under-replication in any way i.e. assert we first go to over-replicated state, and then kick the peer out from the zone which violates max_num_replicas.
4. under-replication should be prioritized over fixing max placement violations

and other relevant scenarios that you can think of.

basavaraj29 (changes requested):

bmcnama-shopify:
trigger jenkins

src/yb/common/common_flags.cc
craigsoules: @iSignal Can we implement this feature without a flag?  I'd be excited to do so 😃

src/yb/master/catalog_manager.cc
craigsoules: @iSignal @basavaraj29 From the design doc you guys shared:

"max is intended to be a hard cap, always respected, even if other copies cannot be placed due to other AZs being down etc."

Which would imply that the proposed implementation is correct here and we should not allow a(3).  If we want to change that then let's think through the trade-offs and then update the design doc with motivation around this.  I would argue that it's preferable to fail in this case since we aren't able to construct a valid quorum across our defined fault domains.  If we're in a situation where b and c have both become unavailable, we've lost quorum and the cluster is in a bad state.  Failing at that point is the least of our problems, and probably better than succeeding in a situation where you can't actually guarantee correctness?

Maybe let's get on a call because it feels like there's a number of feedbacks along these lines.

docs/content/stable/admin/yb-admin.md
```diff
@@ -1546,7 +1546,7 @@ yb-admin \
 ```
 
 * *master-addresses*: Comma-separated list of YB-Master hosts and ports. Default is `localhost:7100`.
-* *placement-info*: Comma-delimited list of placements for *cloud*.*region*.*zone*. Optionally, after each placement block, you can also specify a minimum replica count separated by a colon. This count indicates how many minimum replicas of each tablet we want in that placement block. Its default value is 1. It is not recommended to repeat the same placement multiple times but instead specify the total count after the colon. However, if you specify a placement multiple times, the total count from all mentions is taken.
+* *placement-info*: Comma-delimited list of placements in the form *cloud*.*region*.*zone*`[:min[:max]]`. The minimum defaults to 1. The optional maximum limits how many replicas of one tablet can be placed in the block; when omitted, it defaults to the replication factor. It is not recommended to repeat the same placement multiple times. Repeated placements sum their explicit minimums and maximums; if any occurrence omits the maximum, the aggregate maximum is omitted.
```
basavaraj29: why is the description `This count indicates how many minimum replicas of each tablet we want in that placement block` relevant to min_replicas being removed? need more clarification on what happens to min/max with repeated placement blocks

docs/content/stable/admin/yb-admin.md
```diff
@@ -1556,14 +1556,14 @@ yb-admin \
...
 
-This will place a minimum of:
+This permits:
 
-1. 2 replicas in aws.us-west.us-west-2a
-2. 2 replicas in aws.us-west.us-west-2b
-3. 1 replica in aws.us-west.us-west-2c
+1. Between 1 and 2 replicas in aws.us-west.us-west-2a
+2. Between 1 and 2 replicas in aws.us-west.us-west-2b
+3. Between 1 and 2 replicas in aws.us-west.us-west-2c
```
basavaraj29, line 1581: let the exiting example be and add another example of case of min + max

docs/content/stable/admin/yb-admin.md
basavaraj29: also explicitly state that max isn't supported in combination with wildcard

src/yb/common/common_net-test.cc
basavaraj29: ignore the latter, but the former is still relevant

src/yb/common/common_net-test.cc
```diff
@@ -239,6 +243,28 @@ TEST(TestPlacementInfoContainsPlacementInfo, TestUUID) {
...
+  ASSERT_FALSE(PlacementInfoContains(
+      { .replicas = 3,
+        .blocks = {{ .min_replicas = 1, .placement = "c1.r1.z1", .max_replicas = 1 }} },
+      { .replicas = 3,
+        .blocks = {{ .min_replicas = 3, .placement = "c1.r1.z1", .max_replicas = 3 }} }));
```
basavaraj29, line 265: add a case with multiple blocks where min is matching, but max differs when comparing corresponding blocks in lhs and rhs

src/yb/common/common_net.cc
```diff
@@ -162,6 +171,33 @@ bool PlacementInfoContainsPlacementInfo(const PlacementInfoPB& lhs, const Placem
     return false;
   }
 
+  const auto has_explicit_max = [](const PlacementInfoPB& placement) {
+    return std::ranges::any_of(
+        placement.placement_blocks(),
+        [](const auto& block) { return block.has_max_num_replicas(); });
+  };
+  if (has_explicit_max(lhs) || has_explicit_max(rhs)) {
+    if (lhs.num_replicas() != rhs.num_replicas() ||
```
basavaraj29: why?
```
lhs = num_replicas: 5, placement_blocks: { a.r1.z1:1:2, a.r1.z2:1:2, a.r1.z3:1:2 }
rhs = num_replicas: 3, placement_blocks: { a.r1.z1:1:1, a.r1.z2:1:1, a.r1.z3:1:1 }
```
lhs contains rhs

src/yb/common/common_net.cc
```diff
@@ -162,6 +171,33 @@ bool PlacementInfoContainsPlacementInfo(const PlacementInfoPB& lhs, const Placem
...
+                                                 : rhs.num_replicas())) {
+        return false;
+      }
+    }
+    return true;
+  }
+
   // Basic idea is to try to match replicas of RHS to as specific as possible replicas of LHS.
   // For each placement block, we have exactly min_num_replicas replicas of that placement, and then
   // we have "slack" replicas that can be in of the placement blocks for the remaining replicas.
```
basavaraj29, line 232: refer the comment on how we validate `PlacementInfoContainsPlacementInfo`. since we don't support wildcards with placements, I think something like this would do (in additional to the existing checks)
```
for each matching placement block, compute sum(min(max_num_replicas(lhs, rhs)))
if this sum < rhs num_replicas, lhs doesn't contain rhs.
also would need to check 
```

for now, we can early return false if we ever encounter a combination of max + wildcard in the combined lhr + rhs view

src/yb/common/tablespace_parser.cc
```diff
@@ -113,13 +115,29 @@ Status ValidateReplicationInfo(const ReplicationInfoPB& replication_info) {
...
+          return STATUS_FORMAT(
+              Corruption, "max_num_replicas ($0) must be > 0", i->max_num_replicas());
+        }
+        if (i->max_num_replicas() < i->min_num_replicas()) {
+          return STATUS_FORMAT(
+              Corruption,
+              "max_num_replicas ($0) must be greater than or equal to min_num_replicas ($1)",
+              i->max_num_replicas(), i->min_num_replicas());
+        }
+      }
```
basavaraj29: just `if (i->has_max_num_replicas() && i->max_num_replicas() < i->min_num_replicas()) ` should do

src/yb/integration-tests/load_balancer_placement_policy-test.cc
```diff
@@ -320,6 +321,31 @@ TEST_F(LoadBalancerPlacementPolicyTest, CreateTableWithNondefaultMinNumReplicas)
...
+
+  vector<int> counts_per_ts;
+  GetLoadOnTservers(table_name, 6, &counts_per_ts);
+  ASSERT_EQ(std::accumulate(counts_per_ts.begin(), counts_per_ts.end(), 0), 5 * num_tablets());
+  for (const auto& [first, second] :
+       {std::pair(0, 3), std::pair(1, 4), std::pair(2, 5)}) {
+    ASSERT_LE(counts_per_ts[first] + counts_per_ts[second], 2 * num_tablets());
+  }
+}
+
```
basavaraj29, line 397: what if the load balancer ends up with some form of 2,2,1 even without the new changes? the test wouldn't add any value then. to assert the changes, we need to artificially force 3,1,1 for multiple tables such that the nodes still seem balanced, and verify that the LB resolves it to some form of 2,2,1

src/yb/master/catalog_manager.cc
basavaraj29, line 5288: add a check where `sum(max) >= num_replicas` may be?

basavaraj29, line 5326: update this comment accordingly.

src/yb/master/catalog_manager.cc
basavaraj29: still relevant, just ignore the latter part in the `if` clause - `|| replicas_left >= remaining_ts.end() - it`.

src/yb/master/cluster_balance_util.cc
basavaraj29: that is, change the value to `pair<int, int>` tracking {min, max}

src/yb/master/cluster_balance_util.cc
basavaraj29: still relevant, assume we are changing the policy from A:1:3,B:1:3,C:1:3 -> A:1:2,B:1:2,C:1:2

src/yb/master/cluster_balance_util.cc
basavaraj29: updated the comment, still relevant

src/yb/master/cluster_balance_util.cc
basavaraj29: still relevant

src/yb/master/cluster_balance_util.cc
basavaraj29, line 586: if we are making this change, then let's not leave the new arg at default. better we force the caller to make a decision and provide the arg if they want to hard restrict on max.

this function is being used in 
- normal load balancing `GetLoadToMove`
- under replication `HandleAddIfMissing...`
- wrong placements (blacklist etc) `HandleAddIfWrong...`

for now, we can restrict it all places. but I would feel more comfortable if we also add tests along the following lines
- have A:1:1, B:1:2, C:1:2 and initiate a full move. 

the test would assert that we would be temporarily violating max_num_replicas and letting the load move through.

src/yb/master/cluster_balance_util.cc
basavaraj29: still relevant

src/yb/master/cluster_balance_util.cc
basavaraj29: still relevant since I updated the comment

basavaraj29 (changes requested):
agreed that we should treat max_num_replicas as a hard cap in most cases, so marked the now irrelevant comments as resolved. remaining ones should still be relevant.

src/yb/common/common_flags.cc
craigsoules: Went ahead and removed the flag entirely. The reasoning: explicit maxima can only be written by new binaries (new yb-admin syntax / tablespace JSON), and old binaries preserve-and-ignore the proto2 field, so the worst mixed-version outcome is an old leader transiently placing above a cap during a rolling upgrade — which a new-version leader repairs on its next balancer run. That's degraded-but-self-healing rather than a correctness issue, which didn't seem to meet the bar for an AutoFlag. It also removed the awkward "stored but not enforced until promotion" split and the per-balancer-run revalidation machinery. If you'd prefer a runtime gflag as an operational kill switch, that's a mechanical re-add — let me know.

src/yb/common/common_net-test.cc
craigsoules: Done — the RF5/RF3 containment case now has an explicit companion assertion with the arguments swapped (false, caught by the num_replicas check).

src/yb/master/catalog_manager.cc
craigsoules: Agreed that the full `IsPlacementInfoValid` doesn't belong here — it enforces duplicate/prefix/overlap rules that historic table-level placements were never held to. But dropping validation entirely would let an invalid maximum (`max < min`, `sum(max) < RF`) be persisted at the table level and then poison the balancer. This site now calls a new narrow `CatalogManagerUtil::ValidateMaxNumReplicasFields` that checks only the constraints this PR introduces and is a strict no-op for placements without explicit maxima, so legacy behavior is exactly preserved. A comment documents why the full check is intentionally skipped.

src/yb/master/catalog_manager.cc
craigsoules: Reworked per your sketch (minus the withdrawn overflow clause): the slack phase is now a single pass over tservers in least-loaded-first order — the same order `SelectReplica` used — skipping any tserver whose block is at its effective maximum. Since each tserver hosts at most one replica of a tablet, per-block counts rise by at most one per pick, so the maximums are hard caps; if the caps prevent placing every replica, the tablet starts under-replicated. Also drops the repeated `FindTServersForPlacementBlock` calls per iteration.

src/yb/master/catalog_manager.cc
craigsoules: Same treatment as the other `IsPlacementInfoValid` call site: replaced with the narrow `ValidateMaxNumReplicasFields`, which validates only the new max-specific constraints and is a no-op for placements without explicit maxima — so the intentional skip of the full check is preserved for legacy placements.

src/yb/master/cluster_balance.cc
craigsoules: Removed. With the AutoFlag gone, all write paths (tablespace parser, SetClusterConfig via IsPlacementInfoValid, table-level via the narrow validator) reject invalid maxima at write time, so the per-run revalidation had no remaining purpose — and failing there would have disabled balancing for the whole table.

src/yb/master/cluster_balance_util.h
craigsoules: Renamed to `over_max_placement_tablet_servers`, and added it (plus the new `is_over_max_placements`) to `CBTabletMetadata::ToString` for debuggability.

src/yb/master/cluster_balance.cc
craigsoules: Done — `HandleAddIfOverMaxPlacement` now runs after `HandleAddIfWrongPlacement` and before normal balancing, with the rationale in the header comment.

src/yb/master/cluster_balance_util.cc
craigsoules: Done — the map now holds `{min, effective max}` pairs, populated once up front.

src/yb/master/cluster_balance_util.cc
craigsoules: Done — falls out of the min/max map change above.

src/yb/master/cluster_balance_util.cc
craigsoules: Renamed to `over_max_placement_tablet_servers`.

src/yb/master/cluster_balance_util.cc
craigsoules: Done — added `CBTabletMetadata::is_over_max_placements`, set during analysis and used to populate `tablets_over_max_placements_`.

src/yb/master/cluster_balance_util.cc
craigsoules: Done as part of the restructure — see the detailed reply on the "`over_max_placements` doesn't imply over replication" thread. The chain now handles under-replication, over-replication (with removals preferring blocks above their maximum), and the over-max add-first case independently, matching the ordering you sketched: in the A(3),B(1),C(1) example the balancer adds to B first, and only the resulting over-replication triggers the remove from A.

src/yb/master/cluster_balance_util.cc
craigsoules: You're right, and this was the most important feedback on the PR — the counter-example is valid: with the policy changed to `*:1:2` and layout A(3),B(1),C(1), the tablet is not over-replicated yet A's servers landed in `over_replicated_tablet_servers`. Restructured so the three conditions are tracked independently: `under_replicated_placements` drives adds (highest priority); `over_replicated_tablet_servers` is populated **only** when the tablet actually has more than num_replicas replicas; and the renamed `over_max_placement_tablet_servers` drives the add-before-remove path (`HandleAddIfOverMaxPlacement`). In your example the balancer now adds to B first (add-before-remove), the tablet becomes over-replicated, and the removal is then steered to A. When a tablet is both over-replicated and above a block maximum, removal candidates are restricted to the offending block(s); otherwise anything above its minimum is fair game. Covered by `RepairPlacementAboveMaximum` (which now asserts no bare remove fires first) and the new mocked tests.

src/yb/master/cluster_balance_util.cc
craigsoules: Done as part of the restructure — the three conditions are now tracked and handled independently; see the detailed reply on the "`over_max_placements` doesn't imply over replication" thread.

src/yb/master/cluster_balance_util.cc
craigsoules: Done — the restructured loop logs the over-max condition symmetrically with under/over-replication, and a block comment documents how the three conditions interact.

src/yb/master/cluster_balance_util.cc
craigsoules: Kept it, with a comment explaining why: a tablet that is *both* over-replicated and above a block maximum legitimately reaches `RemoveReplica` while still in `tablets_over_max_placements_` (the removal is steered to the offending block), so the erase is the same invariant maintenance as the `tablets_over_replicated_` erase above it — it prevents double-handling in one run. Happy to convert to a DCHECK-absent if you'd rather assert it can't happen, but I don't think that invariant holds in the both-conditions case.

src/yb/master/cluster_balance_util.cc
craigsoules: Reworked — see the reply on the comment-deletion thread just below: the original water-filling algorithm is restored verbatim for placements without explicit maxima, and the capacity-aware path only runs when maxima exist.

src/yb/master/cluster_balance_util.cc
craigsoules: Restored the original water-filling algorithm — including the full bar-graph comment — as the slack-distribution path whenever no block has an explicit maximum, so behavior and documentation for legacy placements are byte-identical to master. The per-replica priority-queue path now only runs when explicit maxima exist, where water-filling can't be applied directly because a block may exhaust its capacity before the load levels out; that's called out in a comment on the branch.

src/yb/master/load_balancer_mocked-test.cc
craigsoules: Added, mapped to your list:

1. `MaxNumReplicasIsAHardCap` (mini-cluster): per the design-doc discussion on the other thread, the max is a hard cap — creation *fails* when the caps make a quorum infeasible even though enough tservers exist (asserted on the cap-aware feasibility error text, so it can't pass via the raw quorum checks). It then verifies creation succeeds once a z3 tserver is added, and that the LB heals the under-replicated table to full RF within the caps as capacity appears.
2. `BlacklistedMoveAllowedAtMaxPlacement`: blacklist drains are not blocked by maxima — the same-block move goes through by temporarily exceeding the maximum, then the blacklisted replica is removed.
3. `RepairPlacementAboveMaximum` now asserts `HandleRemoveReplicas` returns false before the add — no transient under-replication.
4. `UnderReplicationPrioritizedOverMaxPlacement`: a tablet at A(3),B(1),C(0) with caps of 2 gets the missing-placement add into C first, then the over-max move, then the remove from A.

docs/content/stable/admin/yb-admin.md
craigsoules: Rewrote the bullet: the minimum's original description is restored, repeated-block aggregation is spelled out with concrete equivalences (`z:1:2,z:1:2` is equivalent to `z:2:4`; `z:1:2,z:1` is equivalent to `z:2`), and it now states that maxima require every block in the policy to be fully qualified — no wildcards anywhere in a policy that uses maxima. That's also enforced: master-side in validation (an overlapping uncapped wildcard could otherwise absorb a capped block's tservers and bypass the cap) and client-side in yb-admin for a friendlier error.

docs/content/stable/admin/yb-admin.md
craigsoules: Done — the original min-only RF5 example is back as the primary example, with the min+max variant added after it, including notes on the hard-cap semantics and the transient same-block replacement exception.

src/yb/common/common_net-test.cc
craigsoules: Done, in both directions: a tighter rhs cap under a looser lhs cap is contained; a looser rhs cap over a tighter lhs cap is not (rhs could place two replicas in a block lhs caps at one).

src/yb/common/common_net.cc
craigsoules: You're right — the strict-equality shortcut was testing equality, not containment. Reworked per your suggestion on the other thread; this RF5/RF3 example is now an explicit passing test case, along with the swapped direction returning false.

src/yb/common/common_net.cc
craigsoules: Implemented: for each matching block we compute `sum(min(max_lhs, max_rhs))` and return false if it's below rhs's num_replicas, and any combination of explicit maxima and wildcard blocks returns false early (that combination is now also rejected at validation time, so it can't be persisted within a single placement). One addition on top of your sketch: a per-block check that rhs's *achievable* maximum (`min(max_rhs, rhs_RF - sum(other rhs mins))`) fits under lhs's cap. The sum check alone accepts e.g. lhs `{z1:1:1, z2:1:2}` containing rhs `{z1:1:2, z2:1:2}` (sum = 3 >= 3) even though rhs may legally place 2 in z1, which lhs forbids — that case is in the tests. The generic min-matching logic below is untouched and still runs after the max checks.

src/yb/common/tablespace_parser.cc
craigsoules: Done — `min_num_replicas > 0` is enforced immediately above, so `max < min` subsumes `max <= 0`; collapsed to the single condition with a note and updated the test expectation.

src/yb/integration-tests/load_balancer_placement_policy-test.cc
craigsoules: Agreed on both counts. The creation test was vacuous — with two tservers per zone, a two-per-zone bound holds structurally — so z0 now has three tservers, making a 3-in-z0 placement otherwise possible. For the balancing half, forcing artificial layouts is much easier in the mocked harness, so the "nodes seem balanced" construction went there: `BalancedLoadStillRepairsOverMaxPlacement` builds 10 tservers (6/2/2) and 4 tablets each shaped 3-1-1 such that *every tserver holds exactly two replicas of the table*. The test first asserts the balancer does nothing without maxima (add and remove both return false — plain load balancing can't detect this), then sets maxima of 2 and asserts a move out of zone a for every tablet.

src/yb/master/catalog_manager.cc
craigsoules: Done — `CheckValidPlacementInfo` now fails early with "Sum of maximum replicas per placement (X) is less than num_replicas (Y)" instead of falling through to the misleading "not enough tablet servers" quorum error.

craigsoules: Done — the slack comment now notes that slack replicas can go into any valid placement block as long as the block stays within its effective maximum.

src/yb/master/cluster_balance_util.cc
craigsoules: Done on both counts. The `from_ts` default is removed and all three call sites pass it explicitly (empty at the no-source add in `HandleAddIfMissingPlacement`, with a comment); the same-block exemption semantics are documented on the declaration. Added `BlacklistedMoveAllowedAtMaxPlacement` along the lines you described: policy `a:1:1,b:1:1,c:1:1`, blacklist a's only serving node — asserts a regular add to the fresh a-node is rejected (cap enforced), the blacklist-driven move goes through by temporarily exceeding the maximum, and the blacklisted replica is then removed. Also documented the transient exception in the yb-admin docs so the "hard cap" description matches this behavior.

src/yb/master/catalog_manager.cc
```diff
@@ -12390,19 +12418,54 @@ Status CatalogManager::HandlePlacementUsingPlacementInfo(const PlacementInfoPB&
...
+      }
+      const auto& pb = placement_info.placement_blocks(narrow_cast<int>(block_idx));
+      if (selected_per_block[block_idx] >=
+          implicit_cast<size_t>(GetEffectiveMaxNumReplicas(pb, narrow_cast<int32_t>(nreplicas)))) {
+        continue;
+      }
+      SelectReplicas(
+          {*ts_it}, 1, config, &already_selected_ts, member_type, per_table_state, global_state);
+      ++selected_per_block[block_idx];
+      --replicas_left;
```
basavaraj29, line 12468: why can't we just do it as proposed earlier?
```
size_t max_tservers_left = all_allowed_ts.size() - already_selected_ts.size();
// Upper bounded by the tservers left.
replicas_left = min(replicas_left, max_tservers_left);
DCHECK_GE(replicas_left, 0);
// TSDescriptorVector candidates;
// remaining_ts = set_difference{all_allowed_ts, already_selected_ts}
// sort remaining_ts by load
// for it : remaining_ts:
//   ts = *it
//   if selected_per_block[ts->placement_info] <= placement_info_map[ts->placement_info].max_num_replicas || replicas_left >= remaining_ts.end() - it
       candidates.emplace_back(ts);
       selected_per_block[ts->placement_info]++;
       replicas_left--;
// SelectReplicas(candidates, candidates.size(), config, &already_selected_ts, member_type, per_table_state, global_state)
```

- replace vector with unordered_map
- call `SelectReplicas` outside of the for loop just once. wouldn't need the following then 
```
    // Copy the sorted load order, as selecting a replica below re-sorts it. Selections only
    // change the load of tservers that cannot be selected again for this tablet, so the copied
    // order remains correct.
```

what am I missing?

src/yb/master/cluster_balance_util.cc
basavaraj29: not done. please read the first comment again.

> we need not take the max placements violation path when invoking the over-replication path. makes sense?

src/yb/master/cluster_balance_util.cc
```diff
@@ -405,31 +417,60 @@ Status PerTableLoadState::UpdateTablet(TabletInfo *tablet) {
...
+    tablet_meta.is_over_max_placements =
+        !tablet_meta.over_max_placement_tablet_servers.empty();
+    if (tablet_meta.is_over_replicated) {
+      // If a placement block exceeds its maximum, restrict removal candidates to the offending
+      // block(s) so the remove fixes the maximum violation first. Otherwise, any placement above
+      // its minimum is a valid removal candidate.
+      tablet_meta.over_replicated_tablet_servers = tablet_meta.is_over_max_placements
+          ? tablet_meta.over_max_placement_tablet_servers
+          : std::move(generic_removal_candidates);
+    }
```
basavaraj29, line 473: you shouldn't need this if you follow the suggested psuedo code above

src/yb/master/load_balancer_mocked-test.cc
```diff
@@ -909,6 +912,201 @@ class LoadBalancerRF5MockedTest : public LoadBalancerMockedTest {
...
+  // Shape the tablet to a(3), b(1), c(0): 4 replicas, under-replicated in c and above the
+  // maximum in a.
+  RemoveReplica(tablets_[0].get(), tservers[4]);
+  RemoveReplica(tablets_[0].get(), tservers[5]);
+  ASSERT_OK(ResetLoadBalancerAndAnalyzeTablets());
+
+  // The first action must be an add into the under-replicated placement c, not the over-max
+  // move out of a.
+  std::string tablet_id, from_ts, to_ts;
+  ASSERT_TRUE(ASSERT_RESULT(HandleOneAddIfMissingPlacement(tablet_id, to_ts)));
```
basavaraj29, line 986: this doesn't assert that we prioritize wrong placements over violating max replication right? shouldn't we call `HandleAddReplicas` instead and verify the desired `from`, `to` ts?

src/yb/master/load_balancer_mocked-test.cc
```diff
@@ -909,6 +912,201 @@ class LoadBalancerRF5MockedTest : public LoadBalancerMockedTest {
...
+  // exactly 2 replicas of this table):
+  //   t0: a000, a001, a002, b000, c000
+  //   t1: a003, a004, a005, b000, c000
+  //   t2: a000, a001, a002, b001, c001
+  //   t3: a003, a004, a005, b001, c001
+  const std::vector<std::vector<size_t>> keep_indexes = {
+      {0, 1, 2, 6, 8}, {3, 4, 5, 6, 8}, {0, 1, 2, 7, 9}, {3, 4, 5, 7, 9}};
+  for (size_t i = 0; i < tablets_.size(); ++i) {
+    for (size_t j = 0; j < tservers.size(); ++j) {
+      if (std::find(keep_indexes[i].begin(), keep_indexes[i].end(), j) ==
```
basavaraj29, line 1080: `keep_indexes[i]` - `i` is bound by `tablets_.size()` which isn't local to this test. this should be configurable

src/yb/master/cluster_balance_util.cc
basavaraj29: I don't get this

basavaraj29 (changes requested):

src/yb/master/cluster_balance_util.cc
craigsoules: You're right — the previous change only did half of this. Restructured so the three conditions are mutually exclusive handling states, decided at analysis time in `UpdateTablet`:

- a tablet that is **missing replicas** is owned by the under-replication path;
- a tablet that is **over-replicated** is owned by the removal path, with removal candidates restricted to the offending block(s) if any block is above its maximum (else any block above its minimum, as before);
- only a tablet that is **neither** goes into `tablets_over_max_placements_` for the add-before-remove repair.

Over-max is now tracked as a set of placements (`over_max_placements` / `has_over_max_placements()`), mirroring the existing `under_replicated_placements` / `is_missing_replicas()` pair, rather than a stored flag plus a tserver set. That removed the over-replicated `continue` in `HandleAddIfOverMaxPlacement` and the erase in `RemoveReplica` you flagged — both were reconciliation glue between structures that shouldn't have been separate.

One precision: membership is fixed at analysis time and not updated as adds happen within a run. Re-handling within a run is guarded by `tablets_added_` at the top of `HandleAddIfOverMaxPlacement` — the same check `CanAddTabletToTabletServer` already makes, hoisted so an already-handled tablet doesn't re-scan every destination. The old `is_over_replicated` skip was doing this job only by accident (via `AddStartingTablet` flipping the flag), so this names the actual condition.

src/yb/master/cluster_balance_util.cc
craigsoules: Gone. Classification is now a single `if / else if` chain per block (under-min → `under_replicated_placements`; over-max → `over_max_placements`), and removal candidates are chosen from those two sets only when the tablet is over-replicated. The post-loop flag and the ternary this comment was on no longer exist.

src/yb/master/cluster_balance_util.cc
craigsoules: Fair — and I retract the earlier defense of it. It was there for the case where a tablet was both over-replicated and above a block maximum at analysis time. With the restructure above, that case can no longer put the tablet in `tablets_over_max_placements_` (over-replication owns it), so the erase was dead code. Deleted, along with the comment.

src/yb/master/catalog_manager.cc
craigsoules: You weren't missing anything — took your shape: one pass over the load-sorted tservers collecting cap-respecting candidates into an `unordered_map`-tracked per-block capacity, then apply them all after the loop. The "copy the sorted order" subtlety and its comment are gone, since nothing mutates the order during collection.

One adjustment to the apply step: instead of `SelectReplicas(candidates, candidates.size(), …)`, the candidates go through a small `AddReplicaToConfig` helper extracted from `SelectReplicas`'s loop body (the bookkeeping + peer add). Reason: `SelectReplica` linear-searches its input vector on every pick, so feeding it the K-candidate vector would re-search K times for tservers we've already chosen. Every candidate is selected, so there's nothing left to choose — applying directly is the same result without the redundant search, and `SelectReplicas` itself now just calls the same helper.

src/yb/master/load_balancer_mocked-test.cc
craigsoules: Correct — it didn't, and it couldn't have: the priority lived only in the run-loop call order (`ProcessUnderReplicatedTablets` before `HandleAddReplicas`), which the mocked harness can't exercise, and `HandleAddIfOverMaxPlacement` would in fact have produced an over-max move on that state.

Made it structural: a tablet that is missing replicas is excluded from `tablets_over_max_placements_` (consistent with your chain, where per-block under-replication is the first branch). The test now does what you asked — `ASSERT_FALSE(HandleAddReplicas(...))` on the a(3),b(1),c(0) state, then the missing-placement add with the destination asserted in c, then the over-max move with from/to asserted.

One behavior consequence worth calling out explicitly: a tablet whose missing placement is chronically unfixable (e.g. a policy zone with no tservers) now also has its over-max repair deferred until the minimum becomes satisfiable. I think that's the right priority — the cluster is already degraded and the minimum is the higher-priority invariant — but flagging it since it's a real choice. Happy to discuss if you see it differently.

src/yb/master/load_balancer_mocked-test.cc
craigsoules: Fixed. Added a `KeepOnlyReplicasOn(tablet, {uuids})` helper to the test base so each test declares its layouts by tserver uuid instead of index arithmetic, and the multi-tablet test now asserts `layouts.size() == tablets_.size()` so the coupling to the fixture's tablet count is explicit and fails loudly rather than indexing out of range. All three max-placement tests use the helper.

src/yb/master/cluster_balance_util.cc
```diff
@@ -1141,6 +1217,36 @@ Result<TsTableLoadMap> CalculateOptimalLoadDistribution(
...
+        continue;
+      }
+      ++optimal_load_distribution[index].second;
+      --remaining_block_capacities[block_index];
+      --slack;
+      candidates.emplace(load + 1, current_load_rank, index);
+    }
+    if (slack > 0) {
+      return STATUS(InvalidArgument, "Placement maximums cannot accommodate all tablet replicas");
+    }
```
basavaraj29: don't see an easy way to estimate instead of trying to place each slack replica. this function is purely used for reporting metrics and estimated data yet to move. let's skip these changes and add a note at the place of invocation stating that this doesn't fully factor in `max_placements`

basavaraj29 (approved):

src/yb/master/cluster_balance_util.cc
craigsoules: Reverted the function to match master — the per-replica placement is gone entirely, and the call site in `AnalyzeTablets` now notes that the estimate can be low when maximums are binding. Dropped the `MaxNumReplicas` estimate test with it.

craigsoules:
trigger jenkins

craigsoules:
trigger jenkins

Issue #25, opened 2018-01-27
Support local indexes


OmerCerrahoglu:

kmuthukk:
moving this to your queue Robert for now.

karmakaze:
Suggestion in PR https://github.com/yugabyte/yugabyte-db/pull/32877 to have an index track PK balanced nodes on best effort basis using syntax `SPLIT FOLLOWING TABLE` where the index and PK have same HASH parts. That is a much lower bar than this issue is requesting with the hope that _further work can **optimistically** improve the distributed transactions_ that use the feature.

PR #32877, opened 2026-07-26
Eventual near-local index

karmakaze:
This PR is to chip-away at issue https://github.com/yugabyte/yugabyte-db/issues/25 (Support local indexes)

That PR hasn't had much progress as it is a very large ask that affects many and deep parts of the system.

This PR aims to make the implementation closer to the goal of https://github.com/yugabyte/yugabyte-db/issues/25 without actually solving it. Specifically this PR:
- introduces new syntax `SPLIT FOLLOWING TABLE` for an index to specify best effort locating a secondary index with the same HASH as the PK to balance to the same node(s)
- the feature (`yb_enable_follow_table_index`) must be enabled before use
- it does this by setting desired balance targets matching current PK's
- these targets may never be realized as there could be constraints on the index which have very different data distributions than the PK, but that's ok there's still much potential benefit here

So with the above limitations the most we can expect from this PR is:
- network communications on same host more often where both PK and index operations are involved _(significant only if inter-node network is saturated or in different availability zone/region/cloud-provider)_

#### With performance benefits being marginal, why this PR?

Issue https://github.com/yugabyte/yugabyte-db/issues/25 is asking for 100% transactional index balancing tracking PK which I believe to be too high a bar (at least at present). I also don't believe that's really what users of YugabyteDB always want--they want to have good P95/P99 etc latencies, which we could possibly get by **mostly** _(rather than absolutely)_ locating indexes with PK. Further limiting this to when the index and PK share same HASH parts should make it more likely in practice than not.

The hope/expectation is that after this PR is in the main line, that someone who is specifically interested in the distributed transaction aspects may carry it forward to **optimistically** combine operations when conditions are met.

The PR comment below has a **review of the initial 2 commits**, along with descriptions of the **3rd** and **4th commits that address them**.

### DIFF
--- a/src/postgres/src/backend/commands/indexcmds.c
+++ b/src/postgres/src/backend/commands/indexcmds.c
@@ -1256,6 +1256,21 @@ DefineIndex(Oid relationId,
 				ereport(ERROR,
 						(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
 						 errmsg("cannot use TABLEGROUP with SPLIT")));
+
+			/*
+			 * YB: 'SPLIT FOLLOWING TABLE' is a prototype feature gated by the
+			 * yb_enable_follow_table_index GUC (mirrors the master gflag
+			 * --ysql_yb_enable_follow_table_index). Reject early with a friendly
+			 * error when the feature is disabled; the master enforces this
+			 * authoritatively as well.
+			 */
+			if (stmt->split_options->split_type == FOLLOW_TABLE &&
+				!yb_enable_follow_table_index)
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("SPLIT FOLLOWING TABLE is not enabled"),
+						 errhint("Set yb_enable_follow_table_index to true to use "
+								 "follow-table indexes.")));
 		}
 
 		colocation_id = YbGetColocationIdFromRelOptions(stmt->options);
@@ -1424,8 +1439,14 @@ DefineIndex(Oid relationId,
 	 * YB: Keep the SPLIT clause and the yb_presplit reloption in sync so that
 	 * the index is created with the right pre-split values and yb_presplit is
 	 * persisted for REINDEX and dump/restore.
+	 *
+	 * SPLIT FOLLOWING TABLE carries no num_tablets/split_points to synchronize
+	 * (the index derives its partitioning from the base table at the master),
+	 * so it is excluded here and rides the split_options plumbing untouched.
 	 */
-	if (IsYugaByteEnabled())
+	if (IsYugaByteEnabled() &&
+		!(stmt->split_options &&
+		  stmt->split_options->split_type == FOLLOW_TABLE))
 		YbSyncSplitOptionsAndPresplit(&stmt->split_options, &stmt->options);
 
 	/*
--- a/src/postgres/src/backend/commands/indexcmds.c
+++ b/src/postgres/src/backend/commands/indexcmds.c
@@ -1256,21 +1256,6 @@ DefineIndex(Oid relationId,
 				ereport(ERROR,
 						(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
 						 errmsg("cannot use TABLEGROUP with SPLIT")));
-
-			/*
-			 * YB: 'SPLIT FOLLOWING TABLE' is a prototype feature gated by the
-			 * yb_enable_follow_table_index GUC (mirrors the master gflag
-			 * --ysql_yb_enable_follow_table_index). Reject early with a friendly
-			 * error when the feature is disabled; the master enforces this
-			 * authoritatively as well.
-			 */
-			if (stmt->split_options->split_type == FOLLOW_TABLE &&
-				!yb_enable_follow_table_index)
-				ereport(ERROR,
-						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						 errmsg("SPLIT FOLLOWING TABLE is not enabled"),
-						 errhint("Set yb_enable_follow_table_index to true to use "
-								 "follow-table indexes.")));
 		}
 
 		colocation_id = YbGetColocationIdFromRelOptions(stmt->options);
@@ -1438,16 +1423,30 @@ DefineIndex(Oid relationId,
 	/*
 	 * YB: Keep the SPLIT clause and the yb_presplit reloption in sync so that
 	 * the index is created with the right pre-split values and yb_presplit is
-	 * persisted for REINDEX and dump/restore.
-	 *
-	 * SPLIT FOLLOWING TABLE carries no num_tablets/split_points to synchronize
-	 * (the index derives its partitioning from the base table at the master),
-	 * so it is excluded here and rides the split_options plumbing untouched.
+	 * persisted for REINDEX and dump/restore.  This covers SPLIT FOLLOWING
+	 * TABLE too: it serializes to the operand-less "FOLLOWING TABLE", which is
+	 * what lets pg_get_indexdef re-emit the clause instead of falling back to
+	 * the index's current tablet count.
 	 */
-	if (IsYugaByteEnabled() &&
-		!(stmt->split_options &&
-		  stmt->split_options->split_type == FOLLOW_TABLE))
+	if (IsYugaByteEnabled())
 		YbSyncSplitOptionsAndPresplit(&stmt->split_options, &stmt->options);
+
+	/*
+	 * YB: 'SPLIT FOLLOWING TABLE' is a prototype feature gated by the
+	 * yb_enable_follow_table_index GUC (mirrors the master gflag
+	 * --ysql_yb_enable_follow_table_index).  The check lives here, after the
+	 * sync above, so that it also covers a split_options derived from a
+	 * yb_presplit reloption rather than from an explicit SPLIT clause.  The
+	 * master enforces this authoritatively as well.
+	 */
+	if (stmt->split_options &&
+		stmt->split_options->split_type == FOLLOW_TABLE &&
+		!yb_enable_follow_table_index)
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("SPLIT FOLLOWING TABLE is not enabled"),
+				 errhint("Set yb_enable_follow_table_index to true to use "
+						 "follow-table indexes.")));
 
 	/*
 	 * Parse AM-specific options, convert to text array form, validate.
--- a/src/postgres/src/backend/commands/yb_cmds.c
+++ b/src/postgres/src/backend/commands/yb_cmds.c
@@ -23,6 +23,7 @@
 
 #include "postgres.h"
 
+#include "access/genam.h"
 #include "access/heapam.h"
 #include "access/htup_details.h"
 #include "access/nbtree.h"
@@ -553,6 +554,18 @@ CreateTableHandleSplitOptions(YbcPgStatement handle, TupleDesc desc,
 				YBTransformPartitionSplitPoints(handle, split_options->split_points, attrs, attr_count);
 				break;
 			}
+
+		case FOLLOW_TABLE:
+			/*
+			 * YB: The SPLIT clause grammar is shared between CREATE TABLE and
+			 * CREATE INDEX, so SPLIT FOLLOWING TABLE parses here too. There is
+			 * no table for a table to follow, though: the clause only means
+			 * something for a secondary index.
+			 */
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+					 errmsg("SPLIT FOLLOWING TABLE is only valid for a secondary index")));
+			break;
 
 		default:
 			ereport(ERROR,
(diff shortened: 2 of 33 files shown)


karmakaze:
Since there hasn't been anyone looking at this PR, I got an initial review (by an AI):
<details>
<summary>Review of first 2 commits (by Claude Code)</summary>
I reviewed the two commits against the PR description. Here's what I found.

#### Does it do what the description says?

Mostly, with one substantive exception. The syntax, the feature gate, creation-time partition inheritance, the split reconciler, and the soft placement preference are all there and wired end-to-end (PG grammar → pggate → pg_client proto → master). But two of the description's specific claims don't hold up:

- "an index to specify... the same HASH as the PK" — this is not enforced for YSQL. ValidateFollowTableEligibility (catalog_manager.cc:4585) only compares hash-column counts; the column-identity check is guarded by index_info.indexed_hash_column_ids_size(), which is YCQL-only. So CREATE INDEX ON t(b) SPLIT FOLLOWING TABLE on t(a PRIMARY KEY) is accepted and inherits the base's boundaries even though index entries hash on b. Same for a base table with no PK — the hidden ybrowid hash column counts as 1 and passes. The comment is candid about this, but the check is doable in DefineIndex, where both the index key and the base relation's PK are in hand.
- "balance to the same node(s)" — see gap #7 below; replica co-placement without leader affinity delivers much less than the framing suggests.

#### Correctness issues

1. Dump/restore and REINDEX silently drop the property. indexcmds.c:1443 skips YbSyncSplitOptionsAndPresplit for FOLLOW_TABLE, so no yb_presplit reloption is persisted. pg_get_indexdef (ruleutils.c:1679) then emits SPLIT INTO %N TABLETS for any hash index — so ysql_dump/restore turns a follow-table index into a plain hash index, with no warning. Note the skip is also unnecessary: YbSplitOptionsToPresplitString (pg_yb_utils.c:7945) already returns NULL for an unrecognized type, so the sync would be a no-op anyway; the only thing the skip changes is that a yb_presplit='' sentinel no longer gets stripped. The fix is the opposite of what's here — add a FOLLOW_TABLE case returning "FOLLOWING TABLE" (which yb_presplit_value already parses, thanks to the new SplitClause production) plus a matching ruleutils branch.

2. The reconciler bypasses the split disable-lists and the outstanding-split throttle. ReconcileFollowTableIndexSplits calls DoSplitTablet(..., ManualSplit::kTrue, ...), which sets IgnoreDisabledList, skipping ValidateTableAgainstDisabledLists / ValidateTabletAgainstDisabledList (tablet_split_manager.cc:291,422). Those lists are exactly what backup, DisableSplittingForBackfillingTable, and the DisableTabletSplitting RPC rely on to guarantee no splits happen. It also doesn't go through TabletSplitManager::MaybeDoSplitting, so FLAGS_outstanding_tablet_split_limit doesn't apply — a base table that gains many boundaries at once produces one forced split per index tablet in a single pass, unthrottled. The comment explains why ManualSplit::kTrue is used (to get past the new follow-table exclusion), but that flag carries more than intended.

3. Turning the flag off strands existing follow-table indexes. tablet_split_manager.cc:273 disables automatic splitting purely on the persisted follow_table_mode, with no flag check, while ReconcileFollowTableIndexSplits returns early when the flag is false. Set --ysql_yb_enable_follow_table_index=false at runtime and those indexes get neither automatic nor follow-driven splitting — they grow unbounded. Gate both or neither.

4. The GUC gate is bypassable via the reloption. The check at indexcmds.c:1266 sits inside if (stmt->split_options), but split_options can be derived from yb_presplit later at line 1443. CREATE INDEX i ON t(a) WITH (yb_presplit='FOLLOWING TABLE') reaches the FOLLOW_TABLE case with the GUC off. The master gflag still catches the default configuration, so impact is limited to a session-level SET, but the gate belongs in CreateIndexHandleSplitOptions (after the sync) instead.

5. Unvalidated enum cast across the proto boundary. pg_create_table.cc:218 does static_cast<master::YbFollowTableMode>(req_.follow_table_mode()). proto3 accepts any int for the tserver mirror; assigning an out-of-range value to a proto2 enum field trips protobuf's debug assert. Validate the value, and add a static_assert pinning the two enums together since the comment says they're kept in sync by hand.

#### Design gaps vs. the stated goal

6. No leader affinity. Co-placing replicas doesn't get you same-node index/base access — that needs the index tablet's leader on the base tablet's leader node. Nothing in the leader-balancing path was touched, so at RF=3 you land there ~1/3 of the time. This is the main reason the claimed benefit ("network communications on same host more often") is weaker than the mechanism implies. Preferring the base leader's tserver first inside the SelectReplica preference set, plus a leader-move preference, would be the natural follow-up.

7. Placement following is purely opportunistic. GetTabletToMove uses the preference only as a tiebreak among moves the balancer already wanted to make; nothing initiates a move for co-placement. Consistent with "best effort," but combined with #6 convergence is weak — worth saying plainly in the flag description.

#### Efficiency

8. O(index_tablets × base_tablets) with an allocation per iteration, in both cluster_balance.cc:924 and catalog_manager.cc:12209: GetFollowTablePreferredReplicas calls GetTableInfo (takes mutex_) and base_table->GetTablets() (allocates the whole vector) once per index tablet, then linear-scans. Build a partition_key_start → replicas map once per table. Same shape in the reconciler's inner covering-tablet scan.

9. Log spam. A persistently failing split logs a WARNING every bg-task pass (~1s default). Rate-limit it.

#### Smaller

- TableInfo::follow_table_mode() / follows_table() are dead code — every call site reads l->pb.follow_table_mode() directly.
- CREATE TABLE ... SPLIT FOLLOWING TABLE now parses (the new production is shared via YbOptSplit) and fails with the generic "invalid split options". A targeted error would be better; YbValidatePresplitForRelation likewise has no FOLLOW_TABLE case.
- No master-side guard against a colocated follow-table index. PG blocks it at indexcmds.c:1250, so it's only defense-in-depth, but the reconciler would then warn once per pass forever.
- Creating the index while the base is mid-split can fail with "Partitions does not cover the full partition keyspace" from CreatePartitions. Fails safe, but the error is opaque.
- No tests and no docs. Nothing covers the accept/reject matrix, partition inheritance, reconciliation, or the placement preference.

#### Verified as correct

Worth recording, since several of these are the easy places to get wrong:

- No new grammar conflicts — I ran bison on both master's and this branch's gram.y; both clean (%expect 0 holds).
- Proto field numbers 43 / 36 / 28 are all free; master_ddl.proto already imports catalog_entity_info.proto; SysTablesEntryPB's only reserved is 15.
- ClusterLoadBalancer and CatalogManagerBgTasks are already friends of CatalogManager, so the new private methods are reachable.
- The lock-ordering claims in the comments hold: the preference precompute in ProcessPendingAssignmentsPerTable runs before the tablet write locks, and the reconciler drops each tablet lock before DoSplitTablet takes mutex_.
- SelectReplica's preference path iterates the already-placement-filtered ts_descs, and sorted_replica_load_ is initialized from that same set (catalog_manager.cc:13996) — so it doesn't widen eligibility; no blacklisted or dead tserver can slip through.
- req.set_num_tablets(partitions_size()) correctly hits priority 1 in CalculateNumTabletsForTableCreation and routes into the explicit-partitions branch of CreatePartitions.
- follow_table_mode is set on the dirty PB before the sys_catalog_->Upsert at catalog_manager.cc:5153, so it persists.
- GetEncodedPartitionKey produces exactly the doc-key prefix encoding the hash split path expects.
- The reconciler does converge without re-issuing: split children replace the parent in TableInfo::partitions_, so the boundary appears in index_starts on the next pass.
</details>

### The 3rd commit addresses the reviewed points:

<details>
<summary>3rd commit changes</summary>

#### PostgreSQL layer

SPLIT FOLLOWING TABLE now survives dump/restore and REINDEX (it previously did not — the index came back as an ordinary hash index, silently):
- pg_yb_utils.c — YbSplitOptionsToPresplitString serializes FOLLOW_TABLE to the operand-less "FOLLOWING TABLE"; YbParsePresplitString accepts it (it would otherwise have been mangled into INTO FOLLOWING TABLE TABLETS); YbValidatePresplitForRelation rejects it on non-index or non-hash relations.
- indexcmds.c — dropped the FOLLOW_TABLE skip of YbSyncSplitOptionsAndPresplit, and moved the yb_enable_follow_table_index gate to after the sync, so WITH (yb_presplit='FOLLOWING TABLE') no longer bypasses it.
- ruleutils.c — pg_get_indexdef emits SPLIT FOLLOWING TABLE from the reloption instead of the index's current tablet count. Also fixes a single-tablet follow index losing the clause under the old num_tablets > 1 guard.
- pg_yb_utils.c — YbGetSplitOptions returns FOLLOW_TABLE when the reloption says so, closing the REINDEX / table-rewrite / TRUNCATE hole in one place (4 call sites).

One deliberate consequence: restoring a dump onto a universe with --ysql_yb_enable_follow_table_index=false now errors clearly rather than silently downgrading the index.

Hash-key identity is now enforced (yb_cmds.c, new YbValidateFollowTableHashKey). The master's identity check only fires for YCQL — the PGSQL create path never populates IndexInfoPB.columns/indexed_hash_column_ids, so for YSQL it degraded to comparing hash counts. An index on (b HASH) following a PK of (a HASH) passed. Now the index's leading HASH attnums must equal the base PK's, in order, and a base table with no PK is rejected outright. CREATE TABLE ... SPLIT FOLLOWING TABLE gets a targeted error instead of "invalid split options".

#### Master

- Reconciler no longer overrides operator intent (catalog_manager.cc). It issued ManualSplit::kTrue, which waives disabled lists, PITR coverage, TTL and replica limits. Added IgnoreFollowTableExclusion to ValidateSplitCandidateTable so the reconciler can pre-check "splittable on every other ground?" before forcing. Also skips colocated indexes, caps splits per pass (new runtime flag follow_table_index_max_splits_per_pass, default 8 — these forced splits aren't covered by outstanding_tablet_split_limit), and rate-limits its warnings.
- Automatic-split exclusion is now flag-gated (tablet_split_manager.cc). Without this, turning the feature off froze existing follow-table indexes: no reconciler, and no automatic splitting either.
- Colocated/vector indexes rejected at CreateTable — they have no tablets of their own to follow.
- O(index_tablets × base_tablets) → one map per table: GetFollowTablePreferredReplicas replaced by GetFollowTableReplicasByPartitionKey, used by both the balancer and ProcessPendingAssignmentsPerTable.
- Proto boundary (pg_create_table.cc): static_asserts pinning the proto2/proto3 YbFollowTableMode values together, plus a validity check — proto3 enums are open, and the old unchecked static_cast was UB for an unknown value.
- Dead accessors: TableInfo::follow_table_mode()/follows_table() were unused and took a second read lock; folded onto PersistentTableInfo and adopted at the four existing l->pb.follow_table_mode() sites.

Nothing is built or tested — this repo needs a full YB build, which I didn't run.

Still open, as you asked, for the next pass: more assertive balancer influence (the current preference is only a tiebreak in GetTabletToMove) and leader affinity — without which co-located replicas still don't guarantee same-host reads.
</details>

<details>
<summary>4th commit changes</summary>

Leader affinity, plus the reconciler and flag-description follow-ups from review. Seven files.

Data supply (catalog_manager.h/.cc) — GetFollowTableReplicasByPartitionKey became GetFollowTableBasePlacements, returning { std::set<TabletServerId> replicas; TabletServerId leader; } per base-tablet partition key. The same single pass over the base table's tablets now also records which tserver holds each base tablet's leader — that's the placement that actually matters, since an index entry's base-row read is served by the base leader.

State (cluster_balance_util.h, AnalyzeTablets) — new PerTableLoadState::follow_table_preferred_leader_ts_ (tablet → wanted leader tserver), populated alongside the existing replica-preference map.

Candidate ordering (GetLeaderToMove, cluster_balance.cc:1492) — the list from GetLeadersOnTSToMove is materialized and partitioned three ways: base leader on the destination (moving co-locates it), no opinion, base leader on the source (already co-located, moving breaks it). Every candidate produces an identical load change, so this trades away nothing. The third group is the part that makes co-placement accumulate under churn instead of drifting — without it, a forced move is as likely to destroy co-location as create it. Also changed .reason = std::move(reason) to a copy: a pre-existing bug left any candidate reached after the stepdown-backoff continue with an empty reason string, and reordering makes that path easier to hit.

Proactive pass (GetLeaderToMoveForFollowTable, cluster_balance.cc:1717) — runs from HandleLeaderMoves after both load-driven passes decline, so co-placement never delays a move that's correcting load. Guards: destination known to this table's state, not leader-blacklisted, same affinitized priority, hosts a running peer (reusing GetLeadersOnTSToMove with a singleton set to get the drive-aware path in the same call), stepdown backoff respected, and strictly fewer leaders than the source both per-table and cluster-wide. Deterministic selection (largest load gain, ties by tablet id) since unordered_map order isn't meaningful.

Flag — follow_table_index_leader_affinity, runtime bool, default true, gating both the pass and the ordering. Only has effect when ysql_yb_enable_follow_table_index is on.

Reconciler (ReconcileFollowTableIndexSplits) — two review items. It now calls ValidateSplitCandidateTablet with both ignore flags false before splitting (catalog_manager.cc:3799): ManualSplit::kTrue is needed to get past the follow-table exclusion, but it was also waiving the per-tablet disabled list and the TTL check, not just the table-level lists already pre-checked. This picks up the colocated / tablet-RUNNING / parent-not-yet-deleted checks too, so an unsplittable tablet is skipped with a rate-limited log line instead of being handed to DoSplitTablet to fail. Separately, the quadratic covering-tablet scan is gone (catalog_manager.cc:3759): index tablets are keyed by partition_key_start in the pass that already walks them, and the covering tablet is the greatest start ≤ boundary, since index tablets tile the hash key space contiguously and the end never needs consulting. The same map answers "does the index already have this boundary", so it replaces the old set rather than adding to it.

Flag description (common_flags.cc, guc.c) — the old text claimed the index tracks the base table's "split boundaries and placement". Boundaries are actively reconciled; placement is only a preference the balancer applies when it costs no load balance. Reworded to say that plainly.

Two things to weigh:

The global-load guard makes the proactive pass inert at rest. Requiring strictly-lower load on both axes is what makes the move provably un-undoable — allowing equal global load leaves a real 2-cycle with GetLeaderToMove's global-balancing branch. The cost is that once leader load is exactly even, no single move can improve co-placement without creating variance, so the pass stops firing. Steady state stays around 1-in-3 for RF3: leaders are Raft-elected, not placed, and leader balancing spreads index leaders independently of base leaders. Worth being explicit that the fully co-located configuration is also perfectly load-balanced — this isn't a tradeoff, it's a reachability problem, and closing it needs a swap primitive (trade two leaders between two nodes, net-zero load change) that LeaderMoveDetails/MoveLeader can't express today.
</details>

#### Changes left out from the original review

1. Replica placement is still opportunistic (#7). GetTabletToMove uses the preference only as a tiebreak among moves the balancer already wanted to make; nothing initiates a move for co-placement. HandleRemoveReplicas also picks victims purely on LoadComparator, so a co-placement-motivated add can be undone on a later iteration. Deferred deliberately — it's the same design question as the leader swap (soft preference vs. real placement constraint) and worth settling in one pass rather than accreting tiebreaks.
2. FLAGS_outstanding_tablet_split_limit still doesn't apply to reconciler splits (#2 residual). GetOutstandingSplitCount() is a method on a helper class local to tablet_split_manager.cc, built fresh per DoSplitting pass; nothing on TabletSplitManager exposes a live count, so it isn't reachable from the reconciler. follow_table_index_max_splits_per_pass (default 8) bounds the burst, not the outstanding total. Same for _per_tserver (default 3). The rest of #2 is closed.
3. Leader affinity's outcome, not its mechanism (#6). The leader-balancing path is no longer untouched, but the headline number is unchanged pending swaps. See above.
4. Opaque error when the index is created while the base is mid-split. Still surfaces CreatePartitions' "Partitions does not cover the full partition keyspace". Fails safe; message untouched.
5. No tests, no docs. Nothing covers the accept/reject matrix, partition inheritance, reconciliation, or the placement preference — and the surface is larger now than when the review was written. This is the biggest remaining gap on the branch.

PR #32885, opened 2026-07-27
[#32235] YSQL: Detect and stop creating stale index opclasses that break major upgrades

### LINKED ISSUE #32235: [YSQL] `pg_restore` fails with wrong timestamp_ops assigned during PG15 upgrade.
Jira Link: [DB-21918](https://yugabyte.atlassian.net/browse/DB-21918)
### Description

`pg_restore` fails with wrong timestamp_ops assigned during PG15 upgrade.

```
pg_restore: error: could not execute query: ERROR:  operator class "timestamp_ops" does not accept data type timestamp with time zone
Command was: CREATE INDEX NONCONCURRENTLY "foo_idx_v" ON "foo"."foo_v" USING "lsm" ("id" ASC, id2" ASC, "created_at" "timestamp_ops" DESC);
```

### Issue Type

kind/bug

[DB-21918]: https://yugabyte.atlassian.net/browse/DB-21918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bmcnama-shopify:
## Summary

Fixes #32235 (Jira: DB-21918)

`pg_restore` fails with `ERROR: operator class "timestamp_ops" does not accept data type timestamp with time zone` during the PG11 -> PG15 YSQL major version upgrade (`upgrade_ysql_major_version_catalog`), aborting the upgrade mid-restore.

**Root cause.** With `yb_enable_alter_table_rewrite = false` (legacy ALTER flow), `ALTER TABLE ... ALTER COLUMN ... TYPE` skips `ATPostAlterTypeCleanup` for YB relations, assuming the table clone recreated all dependent indexes. The clone, however, only runs when the type change requires a rewrite (`tab->rewrite > 0`). For a no-rewrite change (e.g. `timestamp -> timestamptz` under UTC), the column type is updated in place and dependent indexes keep the old type's operator class in `pg_index.indclass`. PG11 tolerates this at runtime, but `pg_get_indexdef` prints the now non-default opclass explicitly in dumps, and PG15's `ResolveOpClass` (no legacy-name shim since upstream PG12) rejects the dumped `CREATE INDEX` during the upgrade's pg_restore.

**Fixes (two independent layers):**

1. **Pre-flight check** - new `yb_check_index_opclass_consistency()` in the pg_upgrade YB check battery (`src/postgres/src/bin/pg_upgrade/check.c`). Flags index key columns whose operator class input type no longer accepts the table column's type, mirroring `ResolveOpClass` acceptance rules (exact match, binary-coercible cast, domain base type, polymorphic/pseudo-type opclasses). Existing stale state now fails `pg_upgrade --check` with a per-index report and remediation instead of aborting mid-upgrade.
2. **Creation-side fix** - `src/postgres/src/backend/commands/tablecmds.c`: under the legacy flow, run `ATPostAlterTypeCleanup` when no clone actually happened (`tab->rewrite == 0`). Dependent indexes/constraints are then re-resolved exactly as the default rewrite flow already does: incompatible indexes are rebuilt with the new type's default operator class; compatible ones take the existing in-place `YbATExecAlterIndexAttributeType` route (metadata-only update, relfilenode preserved).

The restore-side option (reinstating the pre-PG12 legacy-name shim under binary upgrade) was deliberately not taken: it would silently mask genuine mismatches during upgrade.

## Upgrade/Rollback safety

- The new pre-flight check only affects `pg_upgrade --check` / the check phase of the YSQL major version upgrade. It reads catalog state and reports; it performs no writes. Clusters without stale index opclasses see no behavior change. Clusters with stale state now fail fast in `--check` (previously they failed mid-restore with a rollback) - strictly better operator experience, no new failure class.
- The creation-side fix changes no catalog schema, no on-disk format, and no wire format. It only makes the legacy (`yb_enable_alter_table_rewrite = false`) no-rewrite ALTER TYPE path perform the same dependent-object cleanup the default path already performs, so newly-altered indexes carry correct metadata going forward. Existing stale metadata is not rewritten by this change (that is what the pre-flight check plus drop/recreate remediation is for).
- Rollback: reverting the commit restores prior behavior with no migration needed in either direction.

## Test plan

- [x] `./yb_build.sh release --cxx-test ysql_major_upgrade_check-test --gtest_filter 'YsqlMajorUpgradeCheckTest.IndexOpclassMismatch'` - new upgrade-check test (old version 2024.2.4.0/PG11): manufactures the stale `indclass` state via a catalog update under `yb_non_ddl_txn_for_sys_tables_allowed` (simulating pre-fix clusters), asserts `pg_upgrade --check` fails naming the index and the type mismatch, asserts drop/recreate recovery passes, and asserts legitimate opclass/type differences (domains, enums, binary-coercible varchar/text) are not flagged. Red run (without the check.c fix): `pg_upgrade --check` reported `*Clusters are compatible*` against the stale cluster and the test failed with `Expected pg_upgrade to fail with error(s)`. Green run (with the fix): passed.
- [x] `./yb_build.sh release --java-test 'org.yb.pgsql.TestPgRegressTable'` - extended `yb.orig.alter_table_rewrite` regress test: with `yb_enable_alter_table_rewrite = OFF`, a no-rewrite `timestamp -> timestamptz` leaves no explicit opclass in `pg_get_indexdef`, `indclass`/index `pg_attribute` match the new type, `yb_index_check` passes on the rebuilt indexes, a hinted index scan returns correct rows, the UNIQUE-backing index is rebuilt and still enforced, and a compatible `varchar` widening takes the in-place route (OID/relfilenode reused, `atttypmod` updated). Red run (without the tablecmds.c fix): `pg_get_indexdef` printed `created_at timestamp_ops DESC` and index attrs stayed `timestamp without time zone`/`timestamp_ops`. Green run (with the fix): passed.
- [x] `./build-support/lint.sh --rev origin/master` - 0 errors (20 warnings, all pre-existing unmarked-YB-hunk/style findings in untouched regions of the modified upstream-owned files).

### DIFF
--- a/src/postgres/src/backend/commands/tablecmds.c
+++ b/src/postgres/src/backend/commands/tablecmds.c
@@ -5610,12 +5610,21 @@ ATRewriteCatalogs(List **wqueue, LOCKMODE lockmode,
 			 * ATExecAlterColumnType since it should be done only once if
 			 * multiple columns of a table are altered).
 			 *
-			 * Do not perform this work if it is a YB relation because we will
-			 * have created an entirely new table, so there is no need for
-			 * cleanup.
+			 * YB Note: skip the cleanup when the legacy ALTER TYPE flow
+			 * (yb_enable_alter_table_rewrite disabled) cloned the table onto
+			 * an entirely new relation, which recreated all dependent
+			 * objects. The clone only happens when the type change requires
+			 * a rewrite (see yb_clone_table in ATExecAlterColumnType). If no
+			 * rewrite was required (tab->rewrite == 0), the column type was
+			 * updated in place and the cleanup must still run to re-resolve
+			 * dependent indexes and constraints: skipping it used to leave
+			 * indexes with a stale operator class, making the dumped index
+			 * DDL fail to restore during a YSQL major version upgrade
+			 * (GH issue #32235).
 			 */
 			if (pass == AT_PASS_ALTER_TYPE &&
-				(!IsYBRelation(tab->rel) || yb_enable_alter_table_rewrite))
+				(!IsYBRelation(tab->rel) || yb_enable_alter_table_rewrite ||
+				 tab->rewrite == 0))
 				ATPostAlterTypeCleanup(wqueue, tab, lockmode);
 
 			if (tab->rel)
--- a/src/postgres/src/bin/pg_upgrade/check.c
+++ b/src/postgres/src/bin/pg_upgrade/check.c
@@ -49,6 +49,7 @@ static void yb_check_user_attributes(PGconn *old_cluster_conn,
 static void yb_check_yugabyte_user(PGconn *old_cluster_conn);
 static void yb_check_old_cluster_user(PGconn *old_cluster_conn);
 static void yb_check_invalid_indexes();
+static void yb_check_index_opclass_consistency();
 static void yb_check_installed_extensions();
 static void yb_check_pgcrypto_schema();
 static void yb_check_yb_role_prefix();
@@ -235,6 +236,7 @@ check_and_dump_old_cluster(bool live_check)
 		old_9_3_check_for_line_data_type_usage(&old_cluster);
 
 	yb_check_invalid_indexes();
+	yb_check_index_opclass_consistency();
 	yb_check_installed_extensions();
 	yb_check_pgcrypto_schema();
 	yb_check_yb_role_prefix();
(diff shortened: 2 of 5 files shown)


bmcnama-shopify:
trigger jenkins

PR #32895, opened 2026-07-27
[#16670] dist-trace: add TraceContextPB and RequestHeader.trace_context

ellabaron-code:
Define TraceContextPB (W3C trace-context) in common.proto and add the
trace_context field to the RPC RequestHeader. This is the wire type shared by
the RPC-header and shared-memory trace-propagation transports; no code
references it yet.

Upgrade/Downgrade Safety: Additive protobuf-only change. It adds a new message
(TraceContextPB) and a new optional field (RequestHeader.trace_context, field
8); no code reads the field yet, so runtime behavior is unchanged. Older
binaries ignore the unknown field, so mixed-version clusters and rollback are
safe. No special upgrade or downgrade procedure is required.

### DIFF
--- a/src/yb/rpc/rpc_header.proto
+++ b/src/yb/rpc/rpc_header.proto
@@ -54,6 +54,21 @@ message RemoteMethodPB {
   required string method_name = 2;
 };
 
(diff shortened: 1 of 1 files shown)


ellabaron-code:

PR #32896, opened 2026-07-27
[#16670] dist-trace: span/scope API and its direct callers

ellabaron-code:
Rework the dist_trace library into the span+scope API the rest of the stack
builds on, and update every in-tree caller of the changed API in lockstep so the
tree keeps building:

  - SpanWithScope (a span bundled with an activated thread-local scope), plus
    Start{Client,Server}SpanWithScope and ActivateParentScope.
  - GetActiveSpanContext / GetActiveTraceparent for cross-boundary propagation.
  - InitDistTrace now takes a service name; CleanupDistTrace -> ShutdownDistTrace
    (mirrored in ybc_dist_trace and its pg_yb_utils.c caller).
  - Replace the RpcSpanAttrs pending-attr buffer with a plain string-pair vector
    drained internally by StartClientSpanWithScope; drop the now-unused public
    GetPendingRpcAttrPairs/ClearPendingRpcAttrs.

Callers updated:
  - outbound_call: client span via StartClientSpanWithScope, write trace_context
    into the RPC RequestHeader, restore the parent around the completion callback.
  - pg_client / pg_session: shared-memory client span and pending RPC attributes.
  - dist_trace-test: new InitDistTrace signature and ShutdownDistTrace rename.

The shared-memory trace_context propagation itself (PgClient::PrepareAndSend)
lands in a later PR.

### DIFF
--- a/src/postgres/src/backend/utils/misc/pg_yb_utils.c
+++ b/src/postgres/src/backend/utils/misc/pg_yb_utils.c
@@ -1098,7 +1098,7 @@ YBInitPostgresBackend(const char *program_name, const YbcPgInitPostgresInfo *ini
 			hex_encode((const char *) YbGetLocalTServerUuid(), UUID_LEN, hex_uuid);
 			hex_uuid[2 * UUID_LEN] = '\0';
 
-			YBCInitDistTrace(MyProcPid, hex_uuid);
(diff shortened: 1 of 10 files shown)


ellabaron-code:

PR #32898, opened 2026-07-27
[#16670] dist-trace: RPC trace-context propagation (wire codec + client/server spans)

ellabaron-code:
End-to-end propagation of a distributed trace across the RPC transport:
decode the context off the wire, create the server span on the callee,
and hand a local call's client span to its in-process inbound span. The
client and server halves reference each other's context accessors, so
they ship together.

Wire codec (serialization.{cc,h}):
- ParseTraceContext(Slice): decode the RequestHeader.trace_context wire
  field; ToSpanContext(TraceContextPB): the shared-memory equivalent;
  both via BuildSpanContext, failing on a zero/invalid context.
- ParseHeader captures RequestHeader.trace_context into ParsedRequestHeader.

Server / inbound span (yb_rpc.{cc,h}):
- YBInboundCall::ParseFrom parses the header trace_context into
  parent_span_context_ (best-effort: a bad context is logged, never fails
  the RPC).
- CreateServerSpan starts a StartServerSpanWithScope child of that parent
  (remote) or of an explicitly passed context (local); RespondSuccess/
  Failure set status and end it; DropServerSpanScope releases the scope on
  the handler thread while the span ends later.

Client local-call handoff (local_call.{cc,h}, rpc_context.cc):
- LocalOutboundCall tags rpc.local_call and drops its client-span scope in
  the ctor (a local call never hops threads to drop it later).
- otel_span_context() exposes the outbound span's context so the local
  inbound span can parent under it (no wire header exists locally).
- RpcContext creates the server span: from the wire header for remote
  calls, from the outbound call's context for local calls.

### DIFF
--- a/src/yb/rpc/local_call.cc
+++ b/src/yb/rpc/local_call.cc
@@ -20,6 +20,7 @@
 #include "yb/rpc/rpc_controller.h"
 #include "yb/rpc/rpc_header.messages.h"
 
+#include "yb/util/dist_trace.h"
 #include "yb/util/format.h"
 #include "yb/util/memory/memory.h"
 #include "yb/util/result.h"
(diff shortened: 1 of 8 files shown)


ellabaron-code:

PR #32899, opened 2026-07-27
[#16670] dist-trace: carry trace context across task and Raft-apply thread hops

ellabaron-code:
Same capture/re-activate pattern as the async scope-carry infra, applied
to the higher-level task types whose work runs on a different thread than
the one that triggered it. Each snapshots dist_trace::GetActiveSpanContext()
where the scope is still live and re-establishes it with ActivateParentScope
around the deferred run, so the work's RPCs nest under the triggering trace.

- tablet/operation_driver: OperationDriver captures at construction,
  re-activates in ApplyTask (APPLYING/APPLIED UpdateTransaction RPCs on
  the Raft apply thread).
- master/async_rpc_tasks_base: RetryingRpcTask captures at construction,
  re-activates at the top of Run().
- master/multi_step_monitored_task: captures at construction, re-activates
  in RunInternal before each step.
- tserver/tablet_validator: captures per scheduled index (carried from the
  CreateTablet RPC into OpenTablet), re-activates the first batched index's
  context around the GetBackfillStatus poll.

Purely additive; all no-ops when the captured context is invalid.

### DIFF
--- a/src/yb/master/async_rpc_tasks_base.cc
+++ b/src/yb/master/async_rpc_tasks_base.cc
@@ -116,7 +116,9 @@ RetryingRpcTask::RetryingRpcTask(
     : master_(master),
       callback_pool_(callback_pool),
       async_task_throttler_(async_task_throttler),
-      deadline_(UnresponsiveDeadline()) {}
+      deadline_(UnresponsiveDeadline()),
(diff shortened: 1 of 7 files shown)


ellabaron-code:

PR #32900, opened 2026-07-27
[#16670] dist-trace: propagate trace context over the PG<->tserver shared memory exchange

ellabaron-code:
Carry the distributed-trace parent across the shared-memory Perform /
AcquireObjectLock fast path, mirroring what the RPC header path already
does. The read-side codec (rpc::ToSpanContext) shipped earlier; this adds
the write side and wires both ends of the exchange.

- ash/rpc_wait_state: TraceContextSerializer encodes the active span's
  SpanContext as a standalone length-prefixed TraceContextPB blob (always
  emits the length prefix, zero when unset, so the framing stays
  parseable). Big-endian layout matches what rpc::ToSpanContext reads back.
- pggate/pg_client (PrepareAndSend): start the outbound shared-memory span,
  serialize the trace_context blob into the exchange buffer after the ASH
  metadata, size the obtain accordingly, and end the span on the early-exit
  path so it is not leaked.
- tserver/pg_client_session: parse the trace_context blob out of the
  exchange, and when present start a server span (StartServerSpanWithScope)
  as a child of the propagated context, with rpc.system=inbound_shmem and
  matching service/method attributes; set Ok/Error status around the
  handler.

### DIFF
--- a/src/yb/ash/rpc_wait_state.cc
+++ b/src/yb/ash/rpc_wait_state.cc
@@ -12,6 +12,8 @@
 //
 #include "yb/ash/rpc_wait_state.h"
 
+#include "yb/gutil/endian.h"
+
 #include "yb/rpc/inbound_call.h"
 #include "yb/rpc/lightweight_message.h"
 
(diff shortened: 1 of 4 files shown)


ellabaron-code:

PR #32901, opened 2026-07-27
[#16670] dist-trace: initialize distributed tracing in the master/tserver process

ellabaron-code:
DbServerBase is the shared base of both Master and TabletServer, so
DbServerBase::Init is the single place that initializes the process-wide
dist_trace tracer (service name = server name, node id = permanent uuid)
for every DB server process -- master, tserver, and the PG backend's
server -- with DbServerBase::Shutdown tearing it down. Gated by
IsDistTraceEnabled, so it is a no-op when tracing is off.


ellabaron-code:

PR #32902, opened 2026-07-27
[#16670] dist-trace: propagate traceparent to the PG backend via the startup packet

ellabaron-code:
Carry a W3C traceparent from the tserver into the PG backend over the
connection startup packet, so the backend's init (and subsequent queries)
root under the originating trace. The tserver-side connection already has
an active trace context; this threads it to the backend.

libpq client (fe-connect.c, fe-protocol3.c, libpq-int.h):
- new yb_dist_traceparent conninfo option on pg_conn, freed with the conn,
  emitted as a startup-packet option when set.

libpq_utils (.cc/.h):
- PGConnSettings gains a traceparent field; BuildConnectionString appends
  yb_dist_traceparent; internal connections populate it from
  dist_trace::GetActiveTraceparent().

backend (postmaster.c, postinit.c, libpq-be.h):
- postmaster parses the yb_dist_traceparent startup param into
  Port::yb_dist_traceparent;
- InitPostgres turns it into the pre-existing yb_dist_tracecontext GUC
  (traceparent='...') so backend init activates that root span.

### DIFF
--- a/src/postgres/src/backend/postmaster/postmaster.c
+++ b/src/postgres/src/backend/postmaster/postmaster.c
@@ -2489,6 +2489,15 @@ retry1:
 							 errhint("Value must be one of the registered "
 									 "YbInternalConnKind wire names.")));
 			}
+			else if (YBIsEnabledInPostgresEnvVar()
+					 && strcmp(nameptr, "yb_dist_traceparent") == 0)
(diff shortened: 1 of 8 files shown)


ellabaron-code:

PR #32903, opened 2026-07-27
[#16670] dist-trace: end-to-end distributed-tracing tests

ellabaron-code:
pgwrapper/dist_trace-test.cc exercises the full propagation path across
the layers in this stack -- process init, RPC header and shared-memory
context propagation, span parent/child nesting, and startup-packet
traceparent -- asserting the expected trace/span relationships end to end.


ellabaron-code:

PR #32929, opened 2026-07-28
[DOC-1425] Change table column name to SQL standard

ddhodge:
Change table column name to SQL standard from Postrgresql isolation

@netlify /stable/explore/transactions/isolation-levels/#isolation-levels-in-postgresql-and-ysql


aishwarya24 (approved):
LGTM

Issue #29306, opened 2025-11-10
[DocDB] Fix code that holds the `CatalogManager::mutex_` for too long


druzac:
Jira Link: [DB-19084](https://yugabyte.atlassian.net/browse/DB-19084)
### Description

Currently the `CatalogManager::mutex_`, a spinlock, is held for too long by some codepaths. For example, `AlterTable` and `CreateTable`. `AlterTable` in particular holds the `CatalogManager::mutex_` exclusively while doing IO. Since most RPCs to the master leader acquire the `CatalogManager::mutex_` at least non-exclusively, any exclusive acquisition can p95 latency spikes or even brief master leader availability loss.

At the very least we should not do any IO while holding the `CatalogManager::mutex_`.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-19084]: https://yugabyte.atlassian.net/browse/DB-19084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #19066, opened 2023-09-11
[CDCSDK] Introduce batch AlterTable & use during creating CDCSDK stream


dr0pdb:
### Description

During the creation of a CDCSDK stream, we iteratively update the `wal_retention_time` of all tables using `AlterTable` RPC. This is required to stop the Raft WAL from GC'ed before CDC streams the changes.

The RPC is invoked once for each table within the namespace. This is a one-time operation but can be very costly if there are many tables.

Consider introducing a batched RPC method in Master DDL service which updates the WAL retention time for all tables within a namespace in a single RPC.

### Source connector version

N/A

### Connector configuration

N/A

### YugabyteDB version

Master

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

PR #32965, opened 2026-07-29
[#29306] DocDB: Release CatalogManager::mutex_ before AlterTable sys-catalog write

jordanwheeler:
## Summary

`CatalogManager::AlterTableWithBatchTracker` held `CatalogManager::mutex_` in exclusive mode across the sys-catalog Raft write that persists an altered table's schema. `mutex_` is a reader-writer spinlock, and nearly every master RPC takes it in shared mode, so each ALTER TABLE stalled the whole master for the length of a consensus commit. Issue #29306 calls out this exact site with "we should not do any IO while holding the CatalogManager::mutex_", and the code marks it with `// todo(GH29185): too much happens here under the cm mutex.` This change deletes that TODO.

This change takes `mutex_` in at most two short scopes, and never across the write or together with the table's copy-on-write lock. A rename checks its target name for collisions and reserves it before the copy-on-write lock is taken, then cleans up the old name after the alter commits. Only a YCQL rename needs those scopes. Every other alter, including every PGSQL alter and the CDC retention path, takes no exclusive `mutex_` at all. `CreateTable`, `CreateNamespace`, and `DoSplitTablet` already run their sys-catalog writes without holding `mutex_`. To measure the effect I swapped the master binary on a cluster with 7 masters and 280 tablet servers and created a CDCSDK stream over a 2000-table namespace, which issues one ALTER per table. During stream creation the control build logged over 8,000 slow-RPC warnings per run and the patched build logged 24, none of them from tserver heartbeats or catalog reads. Stream creation itself took the same time on both builds. The full numbers are in the table below. The new ordering also fixes a self-deadlock and two rollback bugs in the old failure path, and stops a failing rename from leaking its reserved name.

## Why it matters

CDCSDK stream creation calls AlterTable once per table in the namespace, through `SetAllCDCSDKRetentionBarriers`, to set `wal_retention_secs`. On a namespace with thousands of tables that means thousands of back-to-back exclusive holds of `mutex_`, each spanning a Raft commit. While each hold lasts, every shared acquirer waits. That includes catalog lookups, connection setup, and tserver heartbeat processing. Bulk ALTERs from any source create the same pattern. Stream creation is just the densest case I measured.

The cost only appears when exclusive holds arrive back to back. A single-ALTER micro-benchmark will not show it.

## What changed

`AlterTableWithBatchTracker` now takes `mutex_` in at most two short scopes, and the sys-catalog write runs with no catalog-wide lock held. `mutex_` is never held together with the table's copy-on-write lock, so no path through the function can invert the order of those two locks.

A YCQL rename checks its target name for collisions and reserves it in `table_names_map_` in one short exclusive scope before the table's copy-on-write lock is taken. The reservation points at the table, so a concurrent CREATE or rename aiming at the same name fails against it. A cleanup guard declared before the copy-on-write lock erases the reservation if the alter fails on any later step, and the declaration order makes the guard run only after that lock is released. If the write succeeds, the guard is dismissed, the new name is the table's live name, and the old name is removed in a second short scope after the in-memory commit. If the write fails, the staged schema changes are discarded and the guard drops the reserved name. The old name was never touched, so the table stays reachable under it and there is nothing else to roll back.

Every other alter has no name-map state to maintain and takes no exclusive `mutex_` at all. PGSQL alters are all in this group, including PGSQL renames, because PGSQL enforces name uniqueness in its own layer and its tables are not in the by-name map. The CDC `wal_retention_secs` path is in this group too.

## Latent bugs in the old failure path

Moving the rollback out of `UpdateSysCatalogWithNewSchema` also fixes three bugs in the old code.

The first is a self-deadlock. When the write failed, `UpdateSysCatalogWithNewSchema` re-acquired the non-recursive `mutex_` to clean up the name map, while its caller still held it. For a YCQL table the cleanup ran whenever a namespace id or a new table name was passed in, and the caller always passed the namespace id, so every failing YCQL alter hung the master, rename or not. The rollback now lives in the caller, runs after the COW lock is released, and takes `mutex_` fresh, the same way `AbortTableCreation` cleans up after a failed create.

The second bug was masked by the first. `new_table_name` defaults to the table's current name, so a failing alter that renamed nothing would have erased the table's live name entry, even though nothing had been reserved. The rollback is now gated exactly like the reservation, on a YCQL alter that actually carries a new name.

The failure path also never restored the old name, which the old code had already erased before the write. A failing rename would have left the table unreachable by either name until a leader change rebuilt the map. Deferring the erase to after the commit removes that state to roll back.

The old failure path ended in a `CHECK_EQ` that required the erase to remove exactly one entry. That assertion was unreachable, because the thread deadlocked on the lock acquisition two lines above it. Even if it had run, a failing alter with no rename would have passed the check by erasing the live name entry. The check could only fire when the entry was absent, and once `mutex_` is released across the write an absent entry is a legal state. A leader change rebuilds the by-name map from the sys-catalog, so a rollback that runs after the master loses and regains leadership can find the reservation already gone. Crashing the master over a stale entry in a map that self-heals is the wrong response, so the rollback and the post-commit erase log a warning and continue. The old code ignored the result of its pre-write old-name erase entirely. The delete path already logs a warning when an erase from this map finds nothing.

With the rollback moved, the `new_namespace_id` and `new_table_name` parameters of `UpdateSysCatalogWithNewSchema` are dead. They are removed here, along with the matching arguments in `YsqlDdlTxnAlterTableHelper`, whose PGSQL call sites could never reach the erased branch.

## Why releasing the lock is safe

The write itself touches nothing that `mutex_` protects. `mutex_` guards the catalog-wide lookup maps, and the only map mutations on this path are the name reservation and the old-name removal after commit. Each runs under its own short `mutex_` acquisition.

The table's own metadata is guarded by its copy-on-write lock, and that lock stays held across the write. Concurrent operations can still find the table through the maps, but they read its last committed state. Another writer to the same table blocks on the COW lock until this alter commits or rolls back. The newly observable rename window is described in the behavior change section below.

## Behavior change during YCQL rename

The old code erased the old name from the by-name map before the write and held `mutex_` until after the in-memory commit, so concurrent RPCs blocked and then saw only the final state. With the lock released during the write, the table resolves under both names until the rename settles. Only catalog operations on the master can observe this window. Row reads and writes route by table ID, never through the by-name map, so the data path does not change.

Operations that use the new name can resolve the table before the rename is durable. If the write then fails, a DROP or ALTER that used the new name has acted on a name that never committed. A CREATE or rename that targets either name fails with AlreadyPresent during the window. For the old name that used to block and then succeed once the rename was done. For the new name a failed write makes the rejection spurious, and a retry succeeds once the rollback drops the reservation.

Operations that use the old name keep resolving the table during the window instead of blocking. Metadata reads such as `GetTableSchema` return the last committed state, which is still correct because the rename is not durable yet. A DROP that resolves the old name during the window and then waits on the COW lock deletes the table after the rename commits. That race already exists on master for a DROP that resolves the name just before the alter takes `mutex_`, so the window widens rather than opens. After the rename commits, the old name is erased under a fresh `mutex_` acquisition, so it stays visible for a moment before it becomes free.

The window is narrow. Only YCQL renames are affected, because PGSQL tables are not in the by-name map and non-rename alters reserve nothing. An operation that wins the race still serializes on the table's COW write lock, so no inconsistent state is ever persisted, and the table stays reachable by its old name the whole time. `CreateTable` already has the same window today. It reserves a name, releases `mutex_`, and then writes the sys-catalog, and name lookups do not filter uncommitted entries. This change widens an existing hazard rather than introducing a new kind. YCQL table rename is also reachable only through the client API, not through CQL.

The collision check now runs before the table-state checks, which changes one error code. A rename that targets a taken name on a table that is not yet ready returns AlreadyPresent for the name instead of TryAgain for the table state. A rename that goes on to fail those state checks also holds its reservation until the cleanup guard drops it on the failure return, so a CREATE racing such a doomed alter can see a transient AlreadyPresent for the target name and succeeds on retry. `CreateTable`'s uncommitted reservations expose the same transient today.

The complete fix for this problem is to revalidate a name-resolved table against its committed metadata after taking the COW lock. That changes DROP semantics globally and covers the pre-existing races of the same shape, so it belongs in a follow-up.

## Measured impact

I swapped only the master binary on a test cluster with 7 masters and 280 tablet servers, kept the tservers and the master leader fixed, and created one CDCSDK stream over a namespace with 2000 tables. Both master binaries are built from the upstream 2025.2.5.0-b31 tag. Neither build contains the CDC stream-creation batching from #19066. That change landed after the b31 tag, so both arms dispatch the stream's ALTERs without pacing. The control is that source unmodified and the fix is the same source plus the first commit of this PR. The reservation rework in the second commit was not re-measured. For the non-rename alters driven here it removes the one remaining short `mutex_` hold. Three runs per arm. Every run reproduced its arm's result.

The master logs an `RPC took a long time` warning from `ScopedLeaderSharedLock` when handling an RPC takes longer than one second (`master_log_lock_warning_ms`). The warning rows in the table below count those log lines during stream creation, and the p99 figures come from the durations the warnings report.

Throughout each run, a loop on one of the cluster's nodes called `yb-admin list_tables` and `yb-admin list_namespaces` and recorded each call's latency. Every call is a fresh process, so nothing is cached and each call performs a full catalog read against the master leader. The idle baseline is the p99 of the same calls while no stream creation is running, roughly 130 to 170 ms across runs.

| signal during stream creation | control (n=3) | fix (n=3) |
|---|---|---|
| master slow-RPC warnings, all RPCs | 8,042-10,155 | 24 in every run (over 300x fewer) |
| slow-RPC warnings for catalog reads (ListTables/ListNamespaces/GetTableSchema) | 178-272 (p99 3.6-4.4 s) | 0 |
| slow-RPC warnings for tserver heartbeats | 3,791-4,128 (p99 82-98 s) | 0 |
| `yb-admin` catalog reads, p99 | 1.4-2.6 s (11-16x that run's idle baseline) | 133-173 ms (1.0-1.1x) |
| stream create duration | ~152-164 s | ~151-159 s (unchanged) |

When running against the first commit in this branch, the same 24 warnings appear in each run. Twenty-three come from the master's `catalog_manager_bg_tasks` loop and one is the CreateCDCStream call itself, which runs as long as the whole create and reports once on both builds. Each pass of the loop holds the leader lock in shared mode for about 12 seconds, and this change does not touch that code. So the two builds match whenever ALTERs are not running and the gains above are not a general speedup.

However, during stream creation, that background task loop queued behind the same back-to-back exclusive holds and took as long as 162 seconds to complete without this fix. It never took longer than 13 seconds when running against the first commit in this branch.

Stream create duration is unchanged. This fix removes the collateral blocking, not the create work itself.

## Pre-existing issues

Three issues here predate this change. The reservation rework fixes the first two.

1. Fixed here. On master, an error between the name reservation and the write leaks the reservation. A CREATE of that name then fails with AlreadyPresent until a leader change or a master restart rebuilds the by-name map, and dropping the table does not clear the entry. The cleanup guard now covers every failure return after the reservation. Tests pin the rollback for a failed option validation, a failed alter step, a table deleted in the reservation window, and a failed sys-catalog write.
2. Fixed here. On master, a second alter on the same table holds `mutex_` while it waits for the first alter's COW lock, so one slow alter can stall every master RPC through a queued neighbor. `mutex_` is no longer held when the COW lock is taken, so a queued alter blocks nothing else. A new test pins it.
3. Still present. The [`!has_changes && force_send_alter_request` branch](https://github.com/yugabyte/yugabyte-db/blob/3ca634624044b28f55b064ee3088a41d5f0f20b8/src/yb/master/catalog_manager.cc#L8315-L8321) runs `SendAlterTableRequest` while holding the table's copy-on-write lock, including a synchronous sys-catalog write for transactional PGSQL alters. No code in the repository sets that field.

## Testing

Two test flags inject failures. The new `TEST_fail_alter_table_sys_catalog_write` fails the sys-catalog write at the Upsert call. The existing `TEST_fail_alter_table_after_commit` fails the alter right after its in-memory commit. Two new stage values for the existing `TEST_block_alter_table` flag park an alter mid-flight. `sys_catalog_write` parks an alter just before the write starts, holding the table's copy-on-write lock, which is the state concurrent RPCs observe while a real write is in flight. `post_name_reservation` parks a rename after it reserves its new name and before it takes that lock, when it holds no lock at all.

- YCQL `alter_table-test`. The existing rename test now also asserts that the old name stops resolving after a successful rename. A new rename-failure test asserts that the alter returns an error instead of deadlocking, that the reserved new name is released and free for a fresh CREATE, and that the old name still resolves. A new add-column-failure test asserts that a failing non-rename alter leaves the live name mapping intact. The rename test runs under both of the suite's tablet-report batch-size parameterizations. The failure tests run once each, on a fixture instantiated at a single batch size, because the batch size does not affect them.
- Four tests cover the reservation rework's failure paths. The first combines a rename with a `wal_retention_secs` change. That combination fails validation after the new name is reserved, so the test pins the leak fix by asserting that the old name still resolves, that the target name does not, and that a CREATE of the target name then succeeds. The second fails inside schema-change application instead, by combining a rename to a free name with a drop of a column that does not exist, and asserts the same rollback. The third renames to a taken name while also carrying an invalid alter step and asserts that the collision is reported, which pins the new check order deliberately. The fourth renames a table to its own current name, which collides with the table itself before anything is reserved, and asserts that the table stays reachable and alterable afterwards, so a failure that reserved nothing never erases a live name. All four run on the single-batch-size fixture.
- One test injects a failure right after the alter's in-memory commit. The rename is committed at that point, so the failure must not roll it back. The alter returns the error to the client, the new name keeps resolving, the old name is free for a fresh CREATE, and the renamed table accepts a further alter. A rollback that outlived the commit would have erased the table's live name.
- Five tests park an alter at the `sys_catalog_write` stage and drive concurrent operations through the window. While an add-column alter is parked, a schema read of the altering table returns the committed schema without the new column, a listing shows the table once, and a CREATE of an unrelated table completes. The alter is still parked after those operations finish, so they overlapped the write window instead of following it. Each of these operations acquires `mutex_`, which the old code held across the write. While a rename is parked, both names resolve to the same table ID, a lookup through the reserved name returns the committed old name, a listing shows the table exactly once under the old name, a CREATE targeting either name fails with AlreadyPresent, and once the rename commits a CREATE reusing the old name succeeds. The third test sends a DROP through the old name while a rename is parked. The drop blocks on the table's copy-on-write lock, and an unrelated CREATE acquires `mutex_` and completes while the drop stays blocked, so the blocked drop is not holding `mutex_`. Once the rename commits, the drop deletes the renamed table, and afterwards neither name resolves and a CREATE of each name succeeds, so neither map entry leaked. The fourth queues a second rename behind a parked first rename. The queued rename reserves its own name and then waits for the table lock with no lock held, so its reserved name resolves, all three names resolve to the one table, a listing shows the table once, and an unrelated CREATE completes while it waits. Once both renames commit, only the final name resolves, both earlier names are free for fresh CREATEs, and a further alter succeeds. That pins the queued-alter fix, and it pins that a rename erases the name that was committed when it acquired the lock rather than the name its request carried. The fifth renames a second table to the name a parked rename has reserved. The second rename fails with AlreadyPresent while the first is still parked, so its collision check does not wait behind the parked rename, and the contested name resolves to the renaming table rather than the second table. Once the parked rename commits, the contested name belongs to the renamed table and its vacated old name is free for a fresh CREATE. These tests run on the same single-batch-size fixture.
- Two tests park a rename at the `post_name_reservation` stage and send a DROP through the open window, one through the committed old name and one through the reserved new name. The parked rename holds no lock, so both drops run to completion while it is parked. The rename then finds the table deleted and fails, and both tests assert that the failure releases the reservation and that a CREATE of each name succeeds, so a drop landing in the reservation window leaks nothing.
- The whole `alter_table-test` binary passes on this branch, all 50 runnable test instances across its parameterizations. The two instances of one upstream-disabled test are skipped.
- PGSQL coverage. The alter-related tests of `pg_ddl_atomicity-test`, including `AlterTableRollbackOnMasterCrash`, and the rename test of `pg_ddl_transaction-test` exercise the changed helper signature. All 27 runs across three sweeps pass.
- PGSQL concurrent DDL stress. I ran all 20 instances of `PgDdlAtomicityStressTest.Main` on this branch, covering plain, colocated, partitioned, and colocated partitioned tables. Each instance runs five threads against one table. Two threads alternate ALTER TABLE ADD COLUMN and DROP COLUMN, one of them with defaults. One thread creates and drops an index, one builds a concurrent index, and one updates rows throughout. Sixteen of the instances inject failures into the master's DDL verification and rollback paths at 10 percent probability. The same four `TEST_ysql_` flags repeat across each table shape. PGSQL alters flow through the same AlterTable path, which now takes no exclusive `mutex_` for them, and each DDL on a partitioned table expands into one statement per partition plus the parent, which multiplies the concurrent DDL through that path. All 20 pass.
- The failure-injection tests catch the old bug. I applied only the test flag and the injection to master and left the old locking in place. Both failure-injection tests deadlock the master there. The AlterTable worker spins in the failure path trying to re-acquire `mutex_`, heartbeats and catalog reads time out behind it, and the run has to be killed at the test timeout. On this branch the same tests pass in under a second each, and the injected failure surfaces as a clean error.

### DIFF
--- a/src/yb/integration-tests/alter_table-test.cc
+++ b/src/yb/integration-tests/alter_table-test.cc
@@ -76,6 +76,7 @@
 #include "yb/tserver/ts_tablet_manager.h"
 
 #include "yb/util/atomic.h"
+#include "yb/util/backoff_waiter.h"
 #include "yb/util/faststring.h"
 #include "yb/util/logging_test_util.h"
 #include "yb/util/metrics.h"
@@ -101,6 +102,7 @@ DECLARE_int32(replication_factor);
 DECLARE_int32(log_min_seconds_to_retain);
 DECLARE_int32(catalog_manager_report_batch_size);
 DECLARE_string(TEST_block_alter_table);
+DECLARE_bool(TEST_fail_alter_table_after_commit);
 DECLARE_bool(TEST_fail_alter_table_sys_catalog_write);
 
 METRIC_DECLARE_counter(sys_catalog_peer_write_count);
@@ -334,8 +336,9 @@ class ReplicatedAlterTableTest : public AlterTableTest {
   virtual int num_replicas() const override { return 3; }
 };
 
-// Subclass for the tests that target the sys-catalog write step of an alter. The tablet-report
-// batch size that parameterizes AlterTableTest does not affect these tests, so they run at a
+// Subclass for the tests that target the master-side windows of an alter, from the name
+// reservation through the sys-catalog write and the in-memory commit. The tablet-report batch
+// size that parameterizes AlterTableTest does not affect these tests, so they run at a
 // single value.
 class AlterTableSysCatalogWriteTest : public AlterTableTest {};
 
@@ -687,6 +690,9 @@ TEST_P(AlterTableSysCatalogWriteTest, TestRenameSysCatalogWriteFailureDoesNotDea
 
   ASSERT_TRUE(ASSERT_RESULT(TableNameResolves(kTableName)));
   ASSERT_FALSE(ASSERT_RESULT(TableNameResolves(new_name)));
+
+  // The rolled-back name is free for a fresh table.
+  ASSERT_OK(CreateTable(new_name));
 }
 
 // The non-rename variant of the sys-catalog write-failure test above. A column add reserves no
@@ -706,11 +712,13 @@ TEST_P(AlterTableSysCatalogWriteTest, TestAddColumnSysCatalogWriteFailureKeepsNa
   ASSERT_OK(AddNewI32Column(kTableName, "added-after-add-failure"));
 }
 
-// The three tests below open the window that the sys-catalog write for an alter runs in.
-// FLAGS_TEST_block_alter_table = "sys_catalog_write" parks the alter after it has released
-// CatalogManager::mutex_ and before it starts the write, holding the table's COW write lock.
-// That is the exact state a concurrent RPC observes while a real write is in flight. Each test
-// drives client operations through the open window and asserts what they see.
+// The window tests below park an alter mid-flight with FLAGS_TEST_block_alter_table and drive
+// concurrent client operations through the open window. "sys_catalog_write" parks the alter
+// after it has released CatalogManager::mutex_ and before it starts the sys-catalog write,
+// holding the table's COW write lock. That is the exact state a concurrent RPC observes while a
+// real write is in flight. "post_name_reservation" parks a rename after it has reserved its new
+// name and released mutex_, before it takes the COW write lock. That is the state a concurrent
+// RPC observes between the reservation and the lock, when the rename holds no lock at all.
 
 // Catalog operations must not block while an alter's sys-catalog write is in flight, because
 // mutex_ is released across the write. While the alter is parked, a schema read of the altering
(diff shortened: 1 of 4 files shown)


jordanwheeler:
I added three tests that turn the claims about the write window into assertions.

A new stage value for the existing `TEST_block_alter_table` flag, `sys_catalog_write`, parks an alter after it has released `mutex_` and before the sys-catalog write starts. A parked alter holds the table's COW lock, so concurrent RPCs observe the same state they would while a real write is in flight. Each test opens that window, drives concurrent client operations through it, and asserts what they see.

- `TestCatalogOpsProceedWhileAlterWritesSysCatalog` parks an add-column alter. A schema read of the altering table, a listing, and a CREATE of an unrelated table all complete during this. Each of those operations acquires `mutex_`, which the old code held across the write. The schema read returns the committed schema without the new column.
- `TestRenameWindowBothNamesResolve` parks a rename. Both names resolve to the same table ID, a listing shows the table exactly once under the old name, and a CREATE targeting either name fails with AlreadyPresent. Once the rename commits, a CREATE reusing the old name succeeds.
- `TestDropByOldNameWaitsForRenameCommit` sends a DROP through the old name while a rename is parked. The drop resolves the table and then blocks on the COW lock until the rename commits. It then deletes the renamed table. Afterwards neither name resolves, and a CREATE of each name succeeds, which proves neither by-name map entry leaked.

The third test asserts behaviour that predates this change. The delete handler resolves the name to a table first and takes its locks later, so a drop that resolves just before a rename commits deletes the renamed table under the old locking too. This change widens that window from an instant to the length of the write, as described in the rename section of the PR description. Re-validating the name at the point where the delete takes the table lock would close this race. That is a separate change with its own blast radius, because it alters DROP semantics for every caller, so it is not part of this PR.

druzac:
Trigger Jenkins

src/yb/master/catalog_manager.cc
```diff
@@ -8351,12 +8356,8 @@ Status CatalogManager::AlterTableWithBatchTracker(
...
   }
 
-  // Remove the old name. Not present if PGSQL.
-  if (table->GetTableType() != PGSQL_TABLE_TYPE && req->has_new_table_name()) {
-    TRACE("Removing (namespace, table) combination ($0, $1) from by-name map",
-          namespace_id, table_name);
-    table_names_map_.erase({namespace_id, table_name});
-  }
+  // The old name stays in the by-name map until the sys-catalog write commits. A failed write
+  // then leaves the table reachable by its original name with nothing to roll back.
```
druzac: This is one of those comments that is a complete non sequitur when reading the code after this PR has been merged. Remove.

src/yb/master/catalog_manager.cc
```diff
@@ -8464,18 +8504,19 @@ Status CatalogManager::UpdateSysCatalogWithNewSchema(
   for (const auto& entry : ddl_log_entries) {
     ddl_log_entry_pointers.push_back(&entry);
   }
-  Status s = sys_catalog_->Upsert(epoch, ddl_log_entry_pointers, table);
+  // This function does not roll back the by-name map on failure. Callers hold the table's COW
+  // lock and not mutex_, so the rollback happens in the caller after it releases the COW lock.
```
druzac: Again, this comment is unnecessary to merge. This function doesn't even have the right arguments to remove the new name from the by-name map anymore, and there is nothing about renaming when reading  this function in isolation. This comment will probably confuse more than help a year or two from now.

devansh5398:
Hi @jordanwheeler! Thanks for the detailed write-up and the numbers. It's right that `AlterTableWithBatchTracker` holds `mutex_` across the sys-catalog write, and moving the sys_catalog write out from the mutex is the right long-term fix.

That said, we are not considering this a high priority from CDC perspective for a few reasons:
- The stream-creation pain is already mitigated: https://github.com/yugabyte/yugabyte-db/commit/ae340c54d91331c2d064cac81531dd2194b02d47 (present in >= 2025.2.4 and >= 2026.1.1) batches the per-table `AlterTable` calls during stream creation, adds a delay between batches so heartbeats and other shared acquirers of `mutex_` get through, and throttles the ALTER TABLE RPCs to tservers. Since the problem really shows up when exclusive holds come back to back, this pacing removes most of the practical impact.
- We're also cutting down the number of ALTERs. With upcoming feature of publication_scoped_cdc, stream creation will set retention barriers only on the tables that need them instead of every table in the namespace, so far fewer AlterTable calls on this path to begin with.
- This touches the generic ALTER TABLE path, not just CDC. Every YSQL and YCQL alter goes through the code being changed here. Locking changes in this path need careful review and regression coverage.

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -273,6 +282,29 @@ class AlterTableTest : public YBMiniClusterTestBase<MiniCluster>,
...
 
+  Result<bool> TableHasColumn(const YBTableName& table_name, const string& column_name) {
+    YBSchema schema;
+    dockv::PartitionSchema partition_schema;
+    RETURN_NOT_OK(client_->GetTableSchema(table_name, &schema, &partition_schema));
+    for (const auto& column : schema.columns()) {
+      if (column.name() == column_name) {
+        return true;
+      }
+    }
```
druzac: Prefer stuff from the STL `algorithm` library for this. `std::ranges::any_of` does this.

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -273,6 +282,29 @@ class AlterTableTest : public YBMiniClusterTestBase<MiniCluster>,
...
+
+  Result<bool> TableNameResolves(const YBTableName& table_name) {
+    YBSchema schema;
+    dockv::PartitionSchema partition_schema;
+    Status s = client_->GetTableSchema(table_name, &schema, &partition_schema);
+    if (s.IsNotFound()) {
+      return false;
+    }
+    RETURN_NOT_OK(s);
+    return true;
```
druzac: nit:
Using the `RETURN_NOT_OK` macro hurts readability here as you're already testing status codes.
```suggestion
    if (s.ok()) {
      return true;
    }
    if (s.IsNotFound()) {
      return false;
    }
    return s;
```

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +645,213 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
   ASSERT_OK(table_alterer->RenameTo(new_name)
             ->Alter());
 
+  // After a successful rename the old name must not resolve. The commit erased it.
+  YBSchema schema;
+  dockv::PartitionSchema partition_schema;
+  ASSERT_NOK(client_->GetTableSchema(kTableName, &schema, &partition_schema));
```
druzac: Isn't this the wrong function to invoke? Why aren't you using your new `TableNameResolves` helper here?

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +645,213 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
+// An ALTER TABLE whose sys-catalog write fails must surface an error instead of wedging the
+// master. If AlterTable held CatalogManager::mutex_ across the write, the failure-path rollback
+// would re-acquire that same non-recursive mutex and self-deadlock, so the write must run with
+// mutex_ released.
+//
+// This is the rename variant. The new name is reserved under mutex_ before the write. After the
+// injected failure the reservation must be rolled back, and the old name must still resolve
+// because only a commit erases it.
+TEST_P(AlterTableTest, TestRenameSysCatalogWriteFailureDoesNotDeadlock) {
+  DontVerifyClusterBeforeNextTearDown();
```
druzac: Curious, why did do you need to invoke this? It's not obvious to me why validation after this test should fail.

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +645,213 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
+  // Confirm mutex_ is not wedged. A subsequent alter acquires the same mutex and must complete.
+  // The old name must still resolve because a failure never erases it, and the reserved new name
+  // must be rolled back.
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_fail_alter_table_sys_catalog_write) = false;
+  ASSERT_OK(AddNewI32Column(kTableName, "added-after-rename-failure"));
+
+  YBSchema schema;
+  dockv::PartitionSchema partition_schema;
+  ASSERT_OK(client_->GetTableSchema(kTableName, &schema, &partition_schema));
+  ASSERT_NOK(client_->GetTableSchema(new_name, &schema, &partition_schema));
```
druzac: Again, why aren't you invoking `TableNameResolves` here?

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +645,213 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
+  Status alter_status;
+  threads.AddThreadFunctor([&] {
+    alter_status = AddNewI32Column(kTableName, "column-added-during-park");
+  });
+  ASSERT_OK(parked_sink.WaitFor(30s * kTimeMultiplier));
+
+  ASSERT_FALSE(ASSERT_RESULT(TableHasColumn(kTableName, "column-added-during-park")));
+
+  auto tables = ASSERT_RESULT(client_->ListTables());
+  ASSERT_EQ(1, std::count_if(tables.begin(), tables.end(), [](const YBTableName& name) {
```
druzac: nit: Can you use `ranges::count_if` instead, here?

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +645,213 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
+  });
+  Status rename_status;
+  threads.AddThreadFunctor([&] {
+    std::unique_ptr<YBTableAlterer> table_alterer(client_->NewTableAlterer(kTableName));
+    rename_status = table_alterer->RenameTo(new_name)->Alter();
+  });
+  ASSERT_OK(parked_sink.WaitFor(30s * kTimeMultiplier));
+
+  auto info_by_old_name = ASSERT_RESULT(client_->GetYBTableInfo(kTableName));
+  auto info_by_new_name = ASSERT_RESULT(client_->GetYBTableInfo(new_name));
```
druzac, line 785: You mention yourself in the PR writeup that this behaviour is accidental and will be fixed later. If we're going to assert on it in a test, we absolutely should add a comment to that effect. We don't want future maintainers to think we thought this behaviour was desirable.
```suggestion
  auto info_by_old_name = ASSERT_RESULT(client_->GetYBTableInfo(kTableName));
  // todo(GH...): Currently, it is possible to retrieve the table by the new name before AlterTable completes. This is not desirable and will be fixed later.
  auto info_by_new_name = ASSERT_RESULT(client_->GetYBTableInfo(new_name));
```

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +645,213 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
+    return name.table_name() == kTableName.table_name();
+  }));
+  ASSERT_EQ(0, std::count_if(tables.begin(), tables.end(), [&](const YBTableName& name) {
+    return name.table_name() == new_name.table_name();
+  }));
+
+  Status s = CreateTable(new_name);
+  ASSERT_TRUE(s.IsAlreadyPresent()) << s;
+  s = CreateTable(kTableName);
+  ASSERT_TRUE(s.IsAlreadyPresent()) << s;
```
druzac: These assertions are maddening when they fail with another status code because you just get `expected: true, actual: false`. Maintainers want to see what the actual status code was. I believe `MATCHER_P` can be used to do create something that prints much more helpful debug information on failure, I have an example in another file:
```
MATCHER_P(EqualsCode, expected_status, "") {
  return arg.code() == expected_status.code();
}
```

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +645,213 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
 
+// An ALTER TABLE whose sys-catalog write fails must surface an error instead of wedging the
+// master. If AlterTable held CatalogManager::mutex_ across the write, the failure-path rollback
+// would re-acquire that same non-recursive mutex and self-deadlock, so the write must run with
+// mutex_ released.
+//
+// This is the rename variant. The new name is reserved under mutex_ before the write. After the
+// injected failure the reservation must be rolled back, and the old name must still resolve
+// because only a commit erases it.
+TEST_P(AlterTableTest, TestRenameSysCatalogWriteFailureDoesNotDeadlock) {
```
druzac: Why did you subclass `AlterTableTest` for these tests? The test parameter controls the batch size of catalog writes when processing TS heartbeats. How is that relevant for the behaviour you're testing here? We do not have infinite test bandwidth.

druzac (commented):
Looks mostly good. Thanks for the work!

src/yb/integration-tests/alter_table-test.cc
jordanwheeler: Switched to the helper, here and at the remaining `GetTableSchema` checks in the new tests. `ASSERT_NOK` passes on any error, and the helper only returns false on `NotFound` and propagates everything else.

src/yb/integration-tests/alter_table-test.cc
jordanwheeler: I copied the call from `TestShutdownWithPendingTasks` when writing the tests, and that test has a real reason to skip verification. I removed it from both tests, and they pass with the full teardown verification running.

src/yb/integration-tests/alter_table-test.cc
jordanwheeler: The pre-existing item in the description is the drop through a stale name in the third test, not the double resolution here. Both names resolving inside the window seems to be part of the design of the reservation. It blocks CREATE collisions by being an ordinary by-name map entry, and lookups read through it.

I added a comment here explaining that, and a TODO on the drop test for a follow-up. That follow-up would reject a drop through the stale old name and flip that test's assertions.

src/yb/integration-tests/alter_table-test.cc
jordanwheeler: They were on `AlterTableTest` only to reuse its setup and helpers. The batch size doesn't affect any of them. I moved the five new tests onto a fixture instantiated at a single batch size, so each runs once.

druzac:
Relevant prior work here is 113fd2ba8bd4a1b8fd8a324965f6a9cb0ea92443.

jordanwheeler:
Trigger Jenkins

jordanwheeler:
The first commit on this branch takes the sys-catalog write out from under `mutex_`, which removes the IO under the lock. After that commit, `AlterTable` still took one short exclusive scope per alter for the pre-write checks and the by-name map bookkeeping.

After discussing with @druzac, I added a second commit that removes the remaining hold from the common path. A YCQL rename now takes `mutex_` in two short scopes. The first checks the new name for collisions and reserves it in the by-name map. The second erases the old name after the alter commits. Every other alter, including every PGSQL alter, takes no exclusive `mutex_` at all.

AlterTable used to hold `mutex_` in small scopes like this. https://github.com/yugabyte/yugabyte-db/commit/113fd2ba8bd4a1b8fd8a324965f6a9cb0ea92443 widened the hold because those scopes were acquired while the table's copy-on-write write lock was held, which inverted the lock order against every path that takes `mutex_` before locking a table. The second commit brings the small scopes back with the order fixed. The collision check and the reservation run before the table lock is taken. The guard that erases the reservation when an alter fails is declared before the table lock, so it fires after that lock is released. `mutex_` and the table lock are never held together.

This results in two behaviour changes. A rename to a name that is already taken, on a table that is not yet ready, returns `AlreadyPresent` where it used to return `TryAgain`, because the collision check now runs before the table state checks. More importantly, a failing rename no longer leaks its reservation. The old code inserted the new name into the by-name map and then ran the remaining validations. A rename combined with an option that failed validation, e.g. a bad `replication_info` in the same request, returned the error and left the new name in the map. The leaked entry blocked every later use of the name. Creating a table with that name failed with `AlreadyPresent`, and so did retrying the same rename after fixing the request, because the collision check does not exempt the table's own stale entry. Nothing cleared the entry short of a master leader change, which rebuilds the map from the sys catalog. The cleanup guard now erases the reservation on any failure, so a failed rename can be corrected and retried immediately.

The second commit also adds the first real test coverage for that rename process. `alter_table-test` had one rename test before this PR, which covered the happy path. The nine new tests cover a rename that fails, collides, races a drop, or queues behind another rename.

All of the name-map work is YCQL only. YSQL table names live in the PostgreSQL catalog, so a YSQL alter never touches these scopes. The PR description has the full design and the measured impact.

src/yb/master/catalog_manager.cc
```diff
@@ -8434,14 +8464,42 @@ Status CatalogManager::AlterTableWithBatchTracker(
...
+  // The alter is committed. If a rename reserved a new name, that name is the table's live name
+  // now, so dismiss the guard that would erase it. The dismissal must stay directly after the
+  // commit. A failure return between the two would run the guard and erase the live name.
+  name_reservation_guard.Cancel();
+
+  // A committed rename erases its old name here rather than before the write, so that a failed
+  // write leaves the table reachable by the old name. Tolerate an entry that is already absent.
+  // Both mutex_ and the COW lock were released above, so a concurrent mutation could have
+  // removed it.
+  if (table->GetTableType() != PGSQL_TABLE_TYPE && req->has_new_table_name()) {
```
druzac: nit: I'd prefer to use `reserved_new_name` here, instead of copying the predicate that gates setting `reserved_new_name` to `true`.
```suggestion
  if (reserved_new_name) {
```

druzac:
trigger jenkins

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +659,573 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
+  threads.AddThreadFunctor([&] {
+    std::unique_ptr<YBTableAlterer> table_alterer(client_->NewTableAlterer(kTableName));
+    // The second rename changes the name again before the tablets report, so polling for this
+    // alter to finish by name cannot succeed. Return once the master has committed the rename.
+    first_rename_status = table_alterer->RenameTo(first_name)->wait(false)->Alter();
+  });
+  ASSERT_OK(parked_sink.WaitFor(30s * kTimeMultiplier));
+
+  threads.AddThreadFunctor([&] {
+    std::unique_ptr<YBTableAlterer> table_alterer(client_->NewTableAlterer(kTableName));
```
druzac, line 902: This should work with either name, right, as both `kTableName` and `first_name` can both be used to resolve the table at the point at which the first rename request is blocked?

src/yb/integration-tests/alter_table-test.cc
```diff
@@ -613,9 +659,573 @@ TEST_P(AlterTableTest, TestRenameTableAndAdd) {
...
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_block_alter_table) = "post_name_reservation";
+  StringWaiterLogSink parked_sink("Blocking AlterTableWithBatchTracker");
+
+  // Declared before the holder. Its destructor joins the thread, which writes this status, so
+  // the status must outlive the holder on every return path.
+  Status rename_status;
+  TestThreadHolder threads;
+  auto release_alter = ScopeExit([] {
+    ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_block_alter_table) = "";
+  });
```
druzac: I think I mentioned `std::future` in a previous review pass, right? I see you've added these comments about declaration ordering and thread safety wherever you've used `TestThreadHolder`. You wouldn't need these comments, or this particular ordering, if you used `shared_ptr<promise>`s and `future`s instead. I don't understand why you resisted my suggestion. These kinds of comments are a code smell.

src/yb/master/catalog_manager.cc
```diff
@@ -8434,14 +8464,42 @@ Status CatalogManager::AlterTableWithBatchTracker(
...
+    SleepFor(kSleepFor);
+  }
+
+  Status update_status = UpdateSysCatalogWithNewSchema(table, ddl_log_entries, epoch, resp);
+  if (!update_status.ok()) {
+    // The write failed. Discard the staged COW changes. The reservation guard erases the
+    // reserved new name once the COW lock is released. The old name was never erased, so the
+    // table stays reachable by it.
+    return update_status;
+  }
```
druzac: This is a little too elaborate for what is standard catalog manager practice. We RETURN_NOT_OK on sys catalog IO all the time. 
```suggestion
  // If the write fails, the WriteLock destructor will discard the staged COW changes and the 
  // CancelableScopeExit will remove the name reservation in table_names_map_ if necessary.
  RETURN_NOT_OK(UpdateSysCatalogWithNewSchema(table, ddl_log_entries, epoch, resp));
```

druzac (approved):

druzac:
Looks good, @jordanwheeler! Thanks for the work. Really appreciate the care you've put in to validating your changes, given that YCQL table rename semantics are irrelevant to your use case.

I may have more work to do ensuring all necessary tests get run so we can get this landed.

src/yb/integration-tests/alter_table-test.cc
jordanwheeler: I misunderstood that the first time thinking the comment was only about the assertion helpers. These tests now use `std::future`. Much better.

src/yb/integration-tests/alter_table-test.cc
jordanwheeler: Yep, either name works here. The `GetYBTableInfo` assertions below check that all three names resolve to the same table. 

Using `kTableName` for the second rename is deliberate though. It makes the name from that request differ from the name that's committed by the time the rename applies. This way the assertions at the end can verify that the rename erased the committed name rather than the requested one.

jordanwheeler:
Trigger Jenkins

Issue #32293, opened 2026-06-22
Flaky test: PgAutoAnalyzeTest.AutoAnalyzeForMappedRelations regressed by #31844 (D53799)


spolitov:
## Summary

`PgAutoAnalyzeTest.AutoAnalyzeForMappedRelations` started failing (flaky, sanitizer-first) on `master` starting with the build that landed [#31844] / D53799 — *"YSQL: Route relcache-init connection through yb_internal_conn"* (commit `8cc33f7656`).

CSI: https://csiweb.dev.yugabyte.com/stability/test?branch=master&uid=tests-pgwrapper%2Fpg_auto_analyze-test%3A%3A%3APgAutoAnalyzeTest.AutoAnalyzeForMappedRelations

## Stability boundary

- **b167** (`88c7cddb`, 2026-06-15) — last build passing on all configurations.
- **b168** (`8c0cc39e`, 2026-06-16) — first failure (alma8-clang21-**tsan**); the test has flaked on tsan/asan ever since.

The `b167 -> b168` delta is only 9 commits; eight are docs/YBA/PLAT/vector-index changes that cannot affect `pg_auto_analyze`. The only functional YSQL catalog-machinery change in the range is `8cc33f7656` ([#31844]).

## Why this commit

`AutoAnalyzeForMappedRelations` creates tables and waits for the auto-analyze worker to update `reltuples` on the mapped catalog relation `pg_class`. [#31844] rewrites `YbUseMinimalCatalogCachesPreload()`:

```c
-  if (yb_is_internal_connection)   // OLD: every internal conn -> minimal preload
-      return true;
-  return false;
+  kind = YbLookupInternalConnKindByBackendType(MyBackendType);   // NEW: only relcache_init kind
+  return kind != YB_INTERNAL_CONN_KIND_NONE &&
+      YbInternalConnKindDescriptors[kind].use_minimal_preload;
```

Before the change every internal connection (including the auto-analyze background worker) ran with *minimal* catalog-cache preload. After it, only the relcache-init builder does; all other internal connections now do a *full* preload. That changes the catalog-load timing/behavior of the auto-analyze worker connection, which is exactly the path this test exercises. The tsan-first, intermittent signature is consistent with a timing-sensitive regression rather than a hard break.

## Suggested next step

Confirm by running `pg_auto_analyze-test` (tsan) at `88c7cddb` vs `8c0cc39e`, or by reverting [#31844] on a branch and checking the flake disappears.

PR #33006, opened 2026-07-30
[#31805] DocDB: Fix 60s libpq connect block in CheckYsqlLaggingCatalogVersions

### LINKED ISSUE #31805: [DocDB] CheckYsqlLaggingCatalogVersions blocks 60s on libpq connect after postmaster shutdown
Jira Link: [DB-21507](https://yugabyte.atlassian.net/browse/DB-21507)
## Context

While investigating #31257 (shutdown hangs in ASan tests), the daemon-side logs from a recent failing run of `PgCatalogVersionTest.InvalMessageMinimalRetention` (RP item 465570100, 2026-05-18) show:

```
[ts-3] I0518 05:14:08.428838 60578 libpq_utils.cc:623]
    Connect with "host=/tmp/.yb.127.234.35.2:15989 ... connect_timeout=60 ..." failed:
    No such file or directory ..., time taken: 60.000s
[ts-3] W0518 05:14:08.428929 60578 tablet_server.cc:2464]
    Could not get the set of lagging catalog versions: ...
[ts-3] I0518 05:14:08.429003 60578 tablet_server.cc:2455] ScheduleCheckLaggingCatalogVersions
```

After the tserver's local postmaster has been shut down (which happens early in `ShutdownServicesImpl`, well before `tablet_server->Shutdown()`), every `CheckYsqlLaggingCatalogVersions` invocation blocks for the full 60 s libpq connect timeout before failing — and then immediately re-schedules itself via `ScheduleCheckLaggingCatalogVersions`.

## The bug

`TabletServer::ScheduleCheckLaggingCatalogVersions` (`src/yb/tserver/tablet_server.cc:2452-2467`):

```cpp
void TabletServer::ScheduleCheckLaggingCatalogVersions() {
  LOG(INFO) << __func__;
  messenger()->scheduler().Schedule(
    [this](const Status& status) {
      if (!status.ok()) {
        LOG(INFO) << status;
        return;
      }
      auto s = CheckYsqlLaggingCatalogVersions();
      ...
      ScheduleCheckLaggingCatalogVersions();   // re-schedule
    },
    std::chrono::seconds(FLAGS_check_lagging_catalog_versions_interval_secs));
}
```

Two problems:

1. It schedules directly on `messenger()->scheduler()`, bypassing any `ScheduledTaskTracker` — so there's no shutdown coordination. The callback only bails on `!status.ok()`, which is only set when the scheduler aborts a *pending* task, not when one is already running.
2. The underlying `CheckYsqlLaggingCatalogVersions` opens a libpq connection with `connect_timeout=60`. When the postmaster is gone (which is the steady state during tserver shutdown), every call ties up an io_thread for 60 s. With `--check_lagging_catalog_versions_interval_secs=1` (the value used by the failing test), this happens every second.

## Impact

Eats one io_thread for 60 s per invocation during shutdown. This is one of the contributing factors to #31257's overall test-budget exhaustion under ASan.

## Suggested fix

Either or both:

- Wrap the recurring task in a `ScheduledTaskTracker` so `PgClientServiceImpl`-style `StartShutdown`/`CompleteShutdown` can bail out cleanly. Add a shutdown check at the top of the callback body so an already-firing task bails out before doing the libpq connect.
- Reduce `connect_timeout` on this connection to something appropriate for an internal heartbeat-style check (e.g. 2 s × `kTimeMultiplier`). 60 s is far too generous for a localhost connect.

The first option is the principled fix; the second is a cheap defensive band-aid.

## References

- Parent investigation: #31257
- Structural fix for the sequential-shutdown amplifier: #31782

[DB-21507]: https://yugabyte.atlassian.net/browse/DB-21507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bmcnama-shopify:
## Summary
Fixes #31805 (Jira: DB-21507).

`TabletServer::ScheduleCheckLaggingCatalogVersions()` scheduled the periodic lagging-catalog-versions check directly on the messenger scheduler with no `ScheduledTaskTracker`, and `CheckYsqlLaggingCatalogVersions()` opened its internal libpq connection with the general 60s client timeout (`tserver_yb_client_default_timeout_ms`). `PGConn::Connect` retries a failed connect until the deadline, so whenever the local postmaster is unreachable (stopped, restarting, or its unix socket removed) each invocation pinned a messenger io_thread for the full 60s, logged `Could not get the set of lagging catalog versions` only after a minute, and re-scheduled itself unconditionally - including during process shutdown, where the blocked io_thread stalls the `Messenger::Shutdown` reactor join: SIGTERM-to-exit took ~61s with a "dumping all thread stacks" report instead of ~1s. The `should_stop` predicate added by 2ef5075daa (#32293) only covers the window after `TabletServer::Shutdown()` has set `shutting_down_`; the runtime manifestation and the gap before `Shutdown()` remained.

This applies both fixes suggested in the issue:
- Use a short connect budget (`2s * kTimeMultiplier`) for this heartbeat-style check instead of `default_client_timeout()`. A missed cycle only delays invalidation-message garbage collection until the next interval. This also covers the non-shutdown manifestation, which the `should_stop` predicate cannot.
- Own the task with an `rpc::ScheduledTaskTracker` (`check_lagging_catalog_versions_task_`, the `PgClientServiceImpl` / `YsqlManager` pattern): the callback bails out once `shutting_down_` is set, `StartShutdown()` at the top of `TabletServer::Shutdown()` stops the re-schedule loop, and `CompleteShutdown()` runs before the base-class shutdowns join the messenger io_threads that execute the callback. The runtime re-read of `check_lagging_catalog_versions_interval_secs` on every cycle is preserved.

The new regression test drives the check at a 1s interval, removes the postmaster's unix socket directory (the exact post-postmaster-shutdown state), and asserts the check fails fast and that a subsequent graceful tserver shutdown is quick. On the unfixed tree the test fails: no check failure is logged within 20s of socket removal because the in-flight connect is still inside its 60s budget (`connect_timeout=59` visible in the conn string). With the fix, the failure is logged ~2.9s after socket removal and graceful tserver shutdown takes ~0.13s.

## Upgrade/Rollback safety
No wire-format, persisted-state, RPC-versioning, or flag-default changes. The change is node-local (an internal connect deadline for a periodic best-effort check, plus shutdown-path task tracking), so mixed-version clusters and rollback are unaffected.

## Test plan
- [x] `./yb_build.sh debug --cxx-test pg_catalog_version-test --gtest_filter PgCatalogVersionTest.CheckLaggingCatalogVersionsPostmasterDown` - new regression test; fails on the unfixed tree (log-waiter timeout after 20005ms), passes with the fix (failure logged 2.9s after socket removal; tserver shutdown 0.134s).
- [x] `./yb_build.sh debug --cxx-test pg_catalog_version-test --gtest_filter PgCatalogVersionTest.InvalMessageMinimalRetention` - the test from the issue report; passes with the fix.
- [x] `./build-support/lint.sh --rev origin/master` - 0 errors, 0 warnings.

### DIFF
--- a/src/yb/tserver/tablet_server.cc
+++ b/src/yb/tserver/tablet_server.cc
@@ -118,6 +118,7 @@
 #include "yb/util/status_format.h"
 #include "yb/util/status_log.h"
 #include "yb/util/string_util.h"
+#include "yb/util/tsan_util.h"
 
 #include "yb/yql/pggate/util/ybc_util.h"
 #include "yb/yql/pgwrapper/libpq_utils.h"
@@ -938,6 +939,12 @@ void TabletServer::Shutdown() {
   }
   LOG(INFO) << "TabletServer shutting down...";
 
+  // Stop the periodic lagging-catalog-versions check from re-scheduling itself. An
+  // already-running check aborts quickly, because shutting_down_ is set: the callback checks it
+  // before connecting, and the libpq connect retry loop observes it through the should_stop
+  // predicate of CreateInternalPGConn.
+  check_lagging_catalog_versions_task_.StartShutdown();
+
   // Best effort to give up our ysql lease.
   // todo(zdrudi): there's lifetime issues trying to access the pg_supervisor here through
   // callbacks. Probably due to the way the MiniCluster sets up the PgSupervisor.
@@ -997,6 +1004,11 @@ void TabletServer::Shutdown() {
   // Unblock PG backends still waiting on a relcache-init connection before we join reactors below;
   // otherwise their orphaned RPCs keep their connections open and wedge reactor shutdown.
   AbortInFlightRelcacheInitConnections();
+
+  // Wait out any in-flight lagging-catalog-versions check and abort the pending one. This must
+  // happen while the messenger io_threads that run the check are still alive; the base class
+  // shutdowns below join them.
+  check_lagging_catalog_versions_task_.CompleteShutdown();
 
   DbServerBase::Shutdown();
   RpcAndWebServerBase::Shutdown();
@@ -2139,7 +2151,13 @@ void TabletServer::DoGarbageCollectionOfInvalidationMessages(
 }
 
 Status TabletServer::CheckYsqlLaggingCatalogVersions() {
-  auto deadline = CoarseMonoClock::Now() + default_client_timeout();
+  // This is a periodic best-effort check against the local postmaster: use a short connect
+  // budget instead of the general client timeout. When the postmaster is unreachable (e.g. it
+  // is already stopped while the process shuts down), the full 60s budget would pin a
+  // messenger io_thread in the connect retry loop for the whole minute per invocation
+  // (#31805). A missed cycle merely delays invalidation message garbage collection until the
+  // next interval.
+  auto deadline = CoarseMonoClock::Now() + 2s * kTimeMultiplier;
   auto pg_conn = VERIFY_RESULT(
       CreateInternalPGConn("template1", kDefaultInternalPgUser, false, deadline));
   const std::string query = "SELECT datid, local_catalog_version FROM "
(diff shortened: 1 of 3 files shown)


bmcnama-shopify:
trigger jenkins

Issue #27835, opened 2025-07-01
[CDCSDK] Connector crash with 'ERROR: CDCSDK Trying to fetch already GCed intents' due to orphaned slot from dropped and recreated database


hariharasudhan-yb:
Jira Link: [DB-17444](https://yugabyte.atlassian.net/browse/DB-17444)
### Description

In CDC LRU, the PG connector stopped streaming with the following error , 

`Version - 2.27.0.0-b56`

```
2025-06-23 04:26:32,735 ERROR  Postgres|ybconnector_2|streaming|0  Producer failure   [io.debezium.pipeline.ErrorHandler]
com.yugabyte.util.PSQLException: ERROR: CDCSDK Trying to fetch already GCed intents for transaction 4ef7eb63-eb69-46be-86ae-6886d36933ae
  Where: Catalog Version Mismatch: A DDL occurred while processing this query. Try again.
	at com.yugabyte.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2675)
	at com.yugabyte.core.v3.QueryExecutorImpl.processCopyResults(QueryExecutorImpl.java:1263)
	at com.yugabyte.core.v3.QueryExecutorImpl.readFromCopy(QueryExecutorImpl.java:1163)
	at com.yugabyte.core.v3.CopyDualImpl.readFromCopy(CopyDualImpl.java:44)
	at com.yugabyte.core.v3.replication.V3PGReplicationStream.receiveNextData(V3PGReplicationStream.java:160)
	at com.yugabyte.core.v3.replication.V3PGReplicationStream.readInternal(V3PGReplicationStream.java:125)
	at com.yugabyte.core.v3.replication.V3PGReplicationStream.readPending(V3PGReplicationStream.java:82)
	at io.debezium.connector.postgresql.connection.PostgresReplicationConnection$1.readPending(PostgresReplicationConnection.java:669)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.searchWalPosition(PostgresStreamingChangeEventSource.java:427)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.execute(PostgresStreamingChangeEventSource.java:209)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.execute(PostgresStreamingChangeEventSource.java:44)
[...]

PR #33048, opened 2026-08-01
[BACKPORT 2024.2][#27835] CDC: Dropping associated streams during a database drop

devansh5398:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/33048/>) ⏳

Summary:
It is seen that when a database (having an active replication slot) is dropped and a node is restarted, then after `FLAGS_cdc_intent_retention_ms` time we start getting `Trying to fetch already GCed intents` error for every `GetChanges()` call.

The issue happens because:
* Currently the workflow of dropping the database don’t delete the associated streams and so such streams' metadata doesn’t get deleted.
* Now, when a master restarts, it tries to load all streams' metadata in the in-mem map `cdc_stream_map_` from the proto persisted on disk using its `CDCStreamLoader::Visit()` function. However, it is not able to do so since the underlying namespace for such stream is already deleted.
* Now, when UPAM tries to load the stream metadata (using `CDCServiceImpl::GetStream()` in `CDCServiceImpl::GetNamespaceMinRecordIdCommitTimeMap()`, it doesn’t find it and so `GetNamespaceMinRecordIdCommitTimeMap()` returns non-ok status.
* For CDC state table tablet-stream entries (unrelated to streams of dropped database), `CDCServiceImpl::PopulateTabletCheckPointInfo()` won’t be able to fill the `cdcsdk_min_checkpoint_map` because `GetNamespaceMinRecordIdCommitTimeMap()` had returned non-ok status. The `cdcsdk_min_checkpoint_map` is the important map taken into consideration for refreshing the retention barriers. If there  isn't an entry for a tablet in it, the UPAM can't refresh such tablet's retention barrier.
* Thus the retention barriers for such tablet eventually gets lifted (after `FLAGS_cdc_min_replicated_index_considered_stale_secs` as part of maintenance op `ResetStaleRetentionBarriersOp`) and we see `Trying to fetch already GCed intents` error.

To correct this:
* The diff adds the logic of marking the associated streams as `DELETING` in the workflow of namespace deletion.
* We let `CDCStreamLoader::Visit()` function to load the stream which don't have the valid namespace in `cdc_stream_map_` but rightaway mark them in `DELETING` state.
* This way Catalog manager's background task `CatalogManager::CleanUpDeletedXReplStreams()` eventually runs, deleting the stream metadata and entries associated with such stream from cdc state table.
* The UPAM will not see entries related to such stream and `CDCServiceImpl::GetNamespaceMinRecordIdCommitTimeMap()` will not return non-ok status. Thus UPAM won't stuck resolving GCed intents error.

**Upgrade/Rollback safety**
There can be scenario where a database was dropped on an older version where dropping of associated streams are not part of drop database workflow. That way such streams linger in the system (along with their slot entries in cdc_state table). Now when the cluster is upgraded to version which now supports streams drop in database drop workflow, the above mentioned solution approach takes care of loading such streams, marking them as `DELETING` and thus eventually clearing them from system.

Jira: DB-17444

Original commit: 58f9ec18f4be457983ecde08f0801df1d02e7c5e / D49225

Test Plan:
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestUPAMMovesRetentionBarriersOnDBDropAndMasterRestart
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestLogicalStreamsDroppedWhenMasterRestartBeforeStreamsMarkingInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestLogicalStreamsDroppedWhenMasterRestartBeforeStreamsDeletionInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestGRPCStreamsDroppedWhenMasterRestartBeforeStreamsMarkingInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestGRPCStreamsDroppedWhenMasterRestartBeforeStreamsDeletionInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestStreamWithoutNamespaceDrops

Reviewers: xCluster, hsunder, asrinivasan, sumukh.phalgaonkar, skumar, stiwary

Reviewed By: asrinivasan

Subscribers: ycdcxcluster, ybase

Tags: #jenkins-ready

Differential Revision: https://phorge.dev.yugabyte.com/D49225
## Merge conflicts

`2024.2` is missing several unrelated, un-backported commits that master/2025.2 already have in this area of the code, so most conflicts here are noise from that gap rather than real logic conflicts. One block (`DeleteYsqlDatabaseAsync`) required real adaptation to the branch's older code shape.

**Trivial** (branch simply lacked context from other, unrelated commits — kept only this commit's own additions):
- `src/yb/integration-tests/cdcsdk_ysql-test.cc:28` — accepted cherry-pick's added `#include "yb/master/catalog_manager.h"`; skipped the also-conflicting `sys_catalog_constants.h` include, which belongs to different, un-backported code and isn't used by this commit's changes.
- `src/yb/integration-tests/cdcsdk_ysql_test_base.h:148` — kept branch's existing flag declarations and added only this commit's new `TEST_cdcsdk_disable_stream_drop_during_db_drop`; skipped 4 unrelated flag declarations (`TEST_enable_table_rewrite_for_cdcsdk_table`, `TEST_delay_before_complete_expired_pg_sessions_shutdown_ms`, `cdcsdk_vwal_tablets_to_poll_batch_size`, `TEST_cdcsdk_vwal_getchanges_rpc_delay_ms`) that belong to other, un-backported commits.
- `src/yb/integration-tests/cdcsdk_ysql_test_base.h:893` — adjacent, non-overlapping method declarations added by both sides; kept both.
- `src/yb/master/catalog_manager.cc:612` — added only this commit's new `DEFINE_test_flag(bool, cdcsdk_disable_stream_drop_during_db_drop, ...)`; skipped 6 unrelated flag definitions (vector-index backend flags, `system_table_num_tablets`, etc.) and a `DECLARE_bool(create_initial_sys_catalog_snapshot)` that isn't needed here (already visible on this branch via a transitive include).

**Non-trivial:**
- `src/yb/integration-tests/cdcsdk_ysql-test.cc:12086-12324` — the incoming diff's context spanned ~10 pre-existing tests (`TestUPAMMovesRetentionBarriersForCatalogTablet`, the `TestOriginId*` family) that exist on master/2025.2 but were never backported to `2024.2`. Extracted and kept only this commit's own 6 new tests (`TestUPAMMovesRetentionBarriersOnDBDropAndMasterRestart`, `TestLogicalStreamsDroppedWhenMasterRestartBeforeStreams{Marking,Deletion}InDBDrop`, `TestGRPCStreamsDroppedWhenMasterRestartBeforeStreams{Marking,Deletion}InDBDrop`, `TestStreamWithoutNamespaceDrops`), discarding the unrelated pre-existing tests from the conflict region entirely.
- `src/yb/integration-tests/cdcsdk_ysql-test.cc:12144,12217,12297` — the new tests call `MakeCDCStateTable(test_client())`, a helper that doesn't exist in this branch's `cdc_test_util.h`/`.cc` (added later, unrelated commit). Replaced with `CDCStateTable(test_client())` directly, using the constructor overload that already exists on this branch and is already used elsewhere in this same file — functionally identical, since master's `MakeCDCStateTable` is itself just a thin wrapper around that same constructor.
- `src/yb/master/catalog_manager.cc:9176-9193` (`DeleteYsqlDatabaseAsync`) — real structural conflict, not noise. This branch has no `is_ysql_major_upgrade` concept in this function and uses an older 2-arg `DeleteYsqlDBTables(database, epoch)` signature (master/2025.2 use a 3-arg signature with `DeleteYsqlDBTablesType`, which doesn't exist here), and holds the namespace write lock (`l`/`metadata`) continuously from the top of the function rather than releasing and reacquiring it later as master does. Adapted the new CDCSDK-stream-drop block accordingly: dropped the `!is_ysql_major_upgrade &&` condition, kept the branch's existing early-return guard on `metadata.state()` and its existing `DeleteYsqlDBTables(database, epoch)` call, and reused the already-held `l`/`metadata` in the failure path instead of re-declaring/re-locking them (which would have re-acquired an already-held write lock).

### DIFF
--- a/src/yb/integration-tests/cdcsdk_ysql-test.cc
+++ b/src/yb/integration-tests/cdcsdk_ysql-test.cc
@@ -10,6 +10,8 @@
 // or implied.  See the License for the specific language governing permissions and limitations
 // under the License.
 
+#include <atomic>
+
 #include <gtest/gtest.h>
 
 #include "yb/cdc/cdc_service.pb.h"
@@ -23,12 +25,14 @@
 #include "yb/integration-tests/cdcsdk_test_base.h"
 #include "yb/integration-tests/cdcsdk_ysql_test_base.h"
 
+#include "yb/master/catalog_manager.h"
 #include "yb/master/tasks_tracker.h"
 
 #include "yb/tserver/tablet_server.h"
 #include "yb/tserver/ts_tablet_manager.h"
 
 #include "yb/util/test_macros.h"
+#include "yb/util/test_thread_holder.h"
 #include "yb/util/tostring.h"
 #include "yb/util/metric_entity.h"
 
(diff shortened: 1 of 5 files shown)


devansh5398:
Trigger Jenkins

PR #33049, opened 2026-08-01
[BACKPORT 2025.2][#27835] CDC: Dropping associated streams during a database drop

devansh5398:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/33049/>) ⏳

Summary:
It is seen that when a database (having an active replication slot) is dropped and a node is restarted, then after `FLAGS_cdc_intent_retention_ms` time we start getting `Trying to fetch already GCed intents` error for every `GetChanges()` call.

The issue happens because:
* Currently the workflow of dropping the database don’t delete the associated streams and so such streams' metadata doesn’t get deleted.
* Now, when a master restarts, it tries to load all streams' metadata in the in-mem map `cdc_stream_map_` from the proto persisted on disk using its `CDCStreamLoader::Visit()` function. However, it is not able to do so since the underlying namespace for such stream is already deleted.
* Now, when UPAM tries to load the stream metadata (using `CDCServiceImpl::GetStream()` in `CDCServiceImpl::GetNamespaceMinRecordIdCommitTimeMap()`, it doesn’t find it and so `GetNamespaceMinRecordIdCommitTimeMap()` returns non-ok status.
* For CDC state table tablet-stream entries (unrelated to streams of dropped database), `CDCServiceImpl::PopulateTabletCheckPointInfo()` won’t be able to fill the `cdcsdk_min_checkpoint_map` because `GetNamespaceMinRecordIdCommitTimeMap()` had returned non-ok status. The `cdcsdk_min_checkpoint_map` is the important map taken into consideration for refreshing the retention barriers. If there  isn't an entry for a tablet in it, the UPAM can't refresh such tablet's retention barrier.
* Thus the retention barriers for such tablet eventually gets lifted (after `FLAGS_cdc_min_replicated_index_considered_stale_secs` as part of maintenance op `ResetStaleRetentionBarriersOp`) and we see `Trying to fetch already GCed intents` error.

To correct this:
* The diff adds the logic of marking the associated streams as `DELETING` in the workflow of namespace deletion.
* We let `CDCStreamLoader::Visit()` function to load the stream which don't have the valid namespace in `cdc_stream_map_` but rightaway mark them in `DELETING` state.
* This way Catalog manager's background task `CatalogManager::CleanUpDeletedXReplStreams()` eventually runs, deleting the stream metadata and entries associated with such stream from cdc state table.
* The UPAM will not see entries related to such stream and `CDCServiceImpl::GetNamespaceMinRecordIdCommitTimeMap()` will not return non-ok status. Thus UPAM won't stuck resolving GCed intents error.

**Upgrade/Rollback safety**
There can be scenario where a database was dropped on an older version where dropping of associated streams are not part of drop database workflow. That way such streams linger in the system (along with their slot entries in cdc_state table). Now when the cluster is upgraded to version which now supports streams drop in database drop workflow, the above mentioned solution approach takes care of loading such streams, marking them as `DELETING` and thus eventually clearing them from system.

Jira: DB-17444

Original commit: 58f9ec18f4be457983ecde08f0801df1d02e7c5e / D49225

Test Plan:
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestUPAMMovesRetentionBarriersOnDBDropAndMasterRestart
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestLogicalStreamsDroppedWhenMasterRestartBeforeStreamsMarkingInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestLogicalStreamsDroppedWhenMasterRestartBeforeStreamsDeletionInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestGRPCStreamsDroppedWhenMasterRestartBeforeStreamsMarkingInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestGRPCStreamsDroppedWhenMasterRestartBeforeStreamsDeletionInDBDrop
yb_build.sh --cxx-test cdcsdk_ysql-test --gtest_filter CDCSDKYsqlTest.TestStreamWithoutNamespaceDrops

Reviewers: xCluster, hsunder, asrinivasan, sumukh.phalgaonkar, skumar, stiwary

Reviewed By: asrinivasan

Subscribers: ycdcxcluster, ybase

Tags: #jenkins-ready

Differential Revision: https://phorge.dev.yugabyte.com/D49225

## Merge conflicts

- `src/yb/integration-tests/cdcsdk_ysql-test.cc:30` — accepted cherry-pick's added includes (`master.h`, `master_replication.pb.h`, `master_replication.proxy.h`); branch had no code at that location.
- `src/yb/integration-tests/cdcsdk_ysql_test_base.h:159` — adjacent, non-overlapping flag declarations added by both sides; kept both.
- `src/yb/integration-tests/cdcsdk_ysql_test_base.h:930` — adjacent, non-overlapping method declarations added by both sides; kept both.
- `src/yb/master/catalog_manager.h:828` — adjacent, non-overlapping method declarations (`HandleDroppedTablesForCDCSDKStreams` pre-existing on branch, `DropAllCDCSDKStreams` newly cherry-picked); kept both.

### DIFF
--- a/src/yb/integration-tests/cdcsdk_ysql-test.cc
+++ b/src/yb/integration-tests/cdcsdk_ysql-test.cc
@@ -10,6 +10,8 @@
 // or implied.  See the License for the specific language governing permissions and limitations
 // under the License.
 
+#include <atomic>
+
 #include <gtest/gtest.h>
 
 #include "yb/cdc/cdc_service.pb.h"
@@ -37,6 +39,7 @@
 #include "yb/util/logging_test_util.h"
 #include "yb/util/metrics.h"
 #include "yb/util/test_macros.h"
+#include "yb/util/test_thread_holder.h"
 #include "yb/util/tostring.h"
 
 DECLARE_uint32(wait_for_ysql_backends_catalog_version_client_master_rpc_timeout_ms);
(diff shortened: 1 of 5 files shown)


devansh5398:
Trigger Jenkins

Issue #33075, opened 2026-08-03
[DocDB] Deadlock detector AddBlocker leaves an invalid rpc handle in handles_, crashing Send() on shutdown


ellabaron-code:
Jira Link: [DB-22794](https://yugabyte.atlassian.net/browse/DB-22794)
## Description

`LocalProbeProcessor::AddBlocker` in the deadlock detector pushes the result of
`rpc::Rpcs::Prepare()` into `handles_` *before* validating it against `InvalidHandle()`. Once the
detector's rpc registry has been shut down, `Prepare()` returns `InvalidHandle()`, the early return
leaves that past-the-end iterator in `handles_`, and `Send()` unconditionally dereferences every
element of `handles_` — crashing the tserver.

### Root cause

In [`src/yb/docdb/deadlock_detector.cc:245`](https://github.com/yugabyte/yugabyte-db/blob/ab23d4f1cb/src/yb/docdb/deadlock_detector.cc#L245):

```cpp
handles_.push_back(rpcs_->Prepare());
auto handle = handles_.back();
if (handle == rpcs_->InvalidHandle()) {
  LOG_WITH_PREFIX_AND_FUNC(WARNING) << "Shutting down. Cannot send probe.";
  return;      // <-- the invalid handle stays in handles_
}
```

`rpc::Rpcs::Prepare()` returns `InvalidHandle()` — which is `calls_.end()` on the registry's
`boost::container::stable_vector` — as soon as `shutdown_` is set. The guard correctly detects this
and returns early, but the push has already happened, so `handles_` now holds an end iterator.

`Send()` then does, at
[`deadlock_detector.cc:318`](https://github.com/yugabyte/yugabyte-db/blob/ab23d4f1cb/src/yb/docdb/deadlock_detector.cc#L318):

```cpp
for (auto& handle : handles_) {
  (**handle).SendRpc();
}
```

`**handle` dereferences the end iterator and then dereferences the garbage `RpcCommandPtr` it reads
out of it, so the process dies on the resulting bad access. The `handles_.size() == 0` short-circuit
at the top of `Send()` does not help — `handles_` is non-empty, it just contains an unusable entry.

### Failure scenario

The window opens whenever `AddBlocker` runs concurrently with detector shutdown, i.e. when a
transaction status tablet is being shut down (tablet split, load balancer move, tserver shutdown,
`DeleteTablet`) while a probe is being assembled.

[...]

PR #33076, opened 2026-08-03
[#33075] DocDB: Validate the rpc handle in LocalProbeProcessor::AddBlocker

ellabaron-code:
## Summary

`LocalProbeProcessor::AddBlocker` pushed the result of `Rpcs::Prepare()` into `handles_` before
checking it against `InvalidHandle()`. Once the deadlock detector's rpc registry is shut down,
`Prepare()` returns `InvalidHandle()`, which is `calls_.end()`. The early return then left that
past-the-end iterator in `handles_`, and `Send()` dereferences every element of `handles_` via
`(**handle).SendRpc()`, so the tserver crashed. The `handles_.size() == 0` check at the top of
`Send()` does not catch this, because `handles_` is not empty — it holds one entry that cannot be
used.

Only the `ProcessProbe` rpc path gets here. It holds no `ScopedRWOperation`, so a status tablet can
shut down while one of its probe handlers is still adding blockers. The poller-driven path does not
have this problem, because `Poller::Shutdown` waits for its callbacks to finish before the
coordinator shuts the detector down.

This checks the handle before pushing it, so an invalid handle never enters `handles_`. `Send()` then
takes its existing empty-`handles_` path and invokes the callback, so the probe rpc still gets an
answer instead of being dropped.

Also adds a regression test. It parks a probe handler at a new sync point in `GetProbesToForward` and
shuts that handler's status tablet down underneath it, so `AddBlocker` runs against an
already-shut-down `rpcs_`. The shutdown runs on its own thread, because if the parked handler is
serving a probe the detector sent to itself, `Rpcs::CompleteShutdown` waits for a call that only the
test thread can release.

## Test plan

New regression test, run in release on macOS arm64:

```
./yb_build.sh release --cxx-test pg_wait_on_conflict-test \
  --gtest-filter PgWaitQueuesProbeForwardingTest.ForwardProbeDuringDetectorShutdown -n 10
```

- [x] 10/10 iterations pass with the fix.
- [x] Every iteration logs `AddBlocker: Shutting down. Cannot send probe.` exactly once, confirming
      the guarded path is actually exercised rather than the test passing vacuously.
- [x] Without the fix, the same interleaving crashes the tserver in `Send()`.
- [x] Both shutdown orderings are covered across the 10 iterations: those where `TEST_Shutdown`
      drains immediately (<1ms), and the self-probe case where it blocks until the test releases the
      parked handler.

### DIFF
--- a/src/yb/docdb/deadlock_detector.cc
+++ b/src/yb/docdb/deadlock_detector.cc
@@ -245,8 +245,6 @@ class LocalProbeProcessor : public std::enable_shared_from_this<LocalProbeProces
     auto& blocking_subtxn_set = info.blocker_txn_info.blocking_subtxn_set;
     auto handle = rpcs_->Prepare();
     if (handle == rpcs_->InvalidHandle()) {
-      // Must not reach handles_: InvalidHandle() is calls_.end(), and Send() dereferences every
-      // element of handles_ unconditionally.
       LOG_WITH_PREFIX_AND_FUNC(WARNING) << "Shutting down. Cannot send probe.";
       return;
     }
@@ -1159,8 +1157,6 @@ class DeadlockDetector::Impl : public std::enable_shared_from_this<DeadlockDetec
         log_prefix_, detector_id, probe_num, req.min_probe_num(), probe_origin_txn_id, &rpcs_,
         &client(), nullptr /* probe_latency */);
 
-    // Lets tests shut the status tablet down while this handler is parked, so that AddBlocker below
-    // sees an already shut down rpcs_. The argument tells the test which tablet to shut down.
     TEST_SYNC_POINT_CALLBACK(
         "DeadlockDetector::GetProbesToForward:BeforeAddBlockers",
         const_cast<TabletId*>(&status_tablet_));
--- a/src/yb/docdb/deadlock_detector.cc
+++ b/src/yb/docdb/deadlock_detector.cc
@@ -45,6 +45,7 @@
 #include "yb/util/shared_lock.h"
 #include "yb/util/status_format.h"
 #include "yb/util/strongly_typed_uuid.h"
+#include "yb/util/sync_point.h"
 #include "yb/util/unique_lock.h"
 #include "yb/util/yb_pg_errcodes.h"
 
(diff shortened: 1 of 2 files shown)


src/yb/docdb/deadlock_detector.cc
```diff
@@ -242,12 +243,14 @@ class LocalProbeProcessor : public std::enable_shared_from_this<LocalProbeProces
     auto& blocker_id = info.blocker_txn_info.id;
     auto& blocker_status_tablet = info.blocker_txn_info.status_tablet;
     auto& blocking_subtxn_set = info.blocker_txn_info.blocking_subtxn_set;
-    handles_.push_back(rpcs_->Prepare());
-    auto handle = handles_.back();
+    auto handle = rpcs_->Prepare();
     if (handle == rpcs_->InvalidHandle()) {
+      // Must not reach handles_: InvalidHandle() is calls_.end(), and Send() dereferences every
+      // element of handles_ unconditionally.
```
basavaraj29: remove the comment

src/yb/docdb/deadlock_detector.cc
```diff
@@ -1156,6 +1159,12 @@ class DeadlockDetector::Impl : public std::enable_shared_from_this<DeadlockDetec
         log_prefix_, detector_id, probe_num, req.min_probe_num(), probe_origin_txn_id, &rpcs_,
         &client(), nullptr /* probe_latency */);
 
+    // Lets tests shut the status tablet down while this handler is parked, so that AddBlocker below
+    // sees an already shut down rpcs_. The argument tells the test which tablet to shut down.
```
basavaraj29: comment seems redundant, lets remove it.

basavaraj29 (commented):
thanks for the find, and the fix.

was there an existing test that caught the issue is asan/tsan or other modes? if so, I'm in favor of mentioning that in the test plan with a higher `-n` instead of checking in the new test. just don't see much value in the new test - unit tests are mostly for catching regressions and a test executing a pre determined path like this isn't of much help.

ellabaron-code:
@basavaraj29
To answer your question, this issue did not come from another test failure. It was one of the issues found by a Fable run against YB code. We are fixing some of those issues that might affect us.

I created this test first to reproduce the issue then fixed it. If you think we can do it as a part of another test or  we do not need this test, just let me know.

src/yb/docdb/deadlock_detector.cc
ellabaron-code: Fixed.

src/yb/docdb/deadlock_detector.cc
ellabaron-code: Fixed.

basavaraj29 (approved):

ellabaron-code:
trigger jenkins

rahulb-yb:
trigger jenkins

Issue #31301, opened 2026-04-25
[YSQL][DistributedTracing] Reduce the number of sequential scan spans


IshanChhangani:
Jira Link: [DB-21157](https://yugabyte.atlassian.net/browse/DB-21157)
### Description

It seems we are showing one sequential scan per processed row. This becomes unreadable quite quickly. Let us reduce the number of sequential scan spans to a reasonable limit.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-21157]: https://yugabyte.atlassian.net/browse/DB-21157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #23584, opened 2024-08-21
[YSQL] Optimize SELECT .. FOR UPDATE  SKIP LOCKED 


sushantrmishra:
Jira Link: [DB-12502](https://yugabyte.atlassian.net/browse/DB-12502)
### Description

As https://github.com/yugabyte/yugabyte-db/issues/20505 : We optimized the locking implementation by batching the rows for locking. 

when SKIP LOCKED option is provided then lock is still acquired one row at a time. We should implement a more optimized version of acquiring the locks when SKIP locked options is provided.

**Setup :**

```
create table lock_test (h int primary key, v1 int, v2 int);
insert into lock_test select i,i,i from generate_series(1,1000) as i;
```

**Lock 10 rows with SKIP locked** 

```
explain (analyze, dist) select * from lock_test where v1 >= 10 and v1 < 20 for update skip locked;
                                                   QUERY PLAN

---------------------------------------------------------------------------------------------------------------
-
 LockRows  (cost=0.00..115.00 rows=1000 width=44) (actual time=6.561..13.530 rows=10 loops=1)
   ->  Seq Scan on lock_test  (cost=0.00..105.00 rows=1000 width=44) (actual time=1.015..1.064 rows=10 loops=1)
         Storage Filter: ((v1 >= 10) AND (v1 < 20))
         Storage Table Read Requests: 1
         Storage Table Read Execution Time: 0.820 ms
         Storage Table Rows Scanned: 1000
 Planning Time: 0.077 ms
 Execution Time: 13.612 ms
 Storage Read Requests: 11 -----------> 1 Request for Main table read and 10 RPC for locking the  10 Rows
 Storage Read Execution Time: 12.722 ms
 Storage Rows Scanned: 1010
 Storage Execution Time: 12.722 ms
```

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-12502]: https://yugabyte.atlassian.net/browse/DB-12502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

markpyb:
+1

karthik-ramanathan-3006:
Fixed in master, 2026.1.

PR #33100, opened 2026-08-04
[BACKPORT 2025.2][#31301] DistributedTracing: Emit one span per executor node instead of one per tuple

IshanChhangani:
Summary:
The executor wrapped every ExecProcNode call in a distributed-tracing span,
producing one span per tuple (millions of spans for a large scan). This revision
changes the behaviour to emit exactly one span per executor plan node:

- Create the node's span on its first ExecProcNode call and keep it the current
  span for the duration of each call, so live RPC and shared-memory spans nest
  under the executor node that issued them.
- Since we cannot figure out if a particular RPC / row fetch is the last one for
  any execProcNode, we end all exec node spans together at ExecEndNode.
- Close any still-open node spans on query abort.

All previously emitted spans (root query, parse/rewrite/plan, the extended
protocol ext.* spans, RPC and shared-memory spans) are unchanged.

Original commit: 103f54edc9b5c2b86c5e55079c0ad6f9f89b9d2a / D54967

Test Plan:
./yb_build.sh --cxx-test dist_trace-test --gtest_filter DistTraceTest.TestNodeSpanPerNodeAndRpcNesting

Manual: ran traced Seq Scan, Nested Loop, and YB Batched Nested Loop queries on
a local cluster exporting to an OTLP collector + Jaeger. Confirmed exactly one
span per executor node regardless of row count , RPC spans
nested under the issuing node, and correct parent containment at every level.

Reviewers: asaha, aman.mangal

Reviewed By: asaha

Subscribers: swapnil.kasaliwal, jason, yql, ybase

Differential Revision: https://phorge.dev.yugabyte.com/D54967

## Merge conflicts

- src/postgres/src/include/nodes/execnodes.h:776 — this commit's own addition (bool yb_dist_trace_has_node_spans;) sits right after an unrelated field, yb_read_ahead_allowed, that exists on master (added separately by the SKIP LOCKED read-ahead-optimization work, #23584) but was never backported to 2025.2. Losing that anchor line made git dump both fields into the conflict. Resolved by keeping only yb_dist_trace_has_node_spans, matching this commit's actual diff; did not introduce the unrelated field.

### DIFF
--- a/src/postgres/src/backend/commands/portalcmds.c
+++ b/src/postgres/src/backend/commands/portalcmds.c
@@ -36,6 +36,7 @@
 
 /* YB includes */
 #include "pg_yb_utils.h"
+#include "yb_dist_trace.h"
 
 
 /*
@@ -330,6 +331,15 @@ PortalCleanup(Portal portal)
 			PG_END_TRY();
 
 			CurrentResourceOwner = saveResourceOwner;
+		}
+		else if (YBCIsDistTraceEnabled())
+		{
+			/*
+			 * YB: executor shutdown is skipped for failed portals, so
+			 * ExecEndNode never runs and node spans left open by a suspended
+			 * portal would leak.
+			 */
+			YbDistTraceEndNodeSpans(queryDesc->planstate, /* errored */ true);
 		}
 	}
 }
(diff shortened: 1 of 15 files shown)


IshanChhangani:
Trigger Jenkins

mangalaman93 (approved):

IshanChhangani:
trigger jenkins

Issue #33103, opened 2026-08-04
[DocDB] Use-after-free: remote bootstrap request reads TrackedPeer fields after releasing queue_lock_


ellabaron-code:
Jira Link: [DB-22825](https://yugabyte.atlassian.net/browse/DB-22825)
## Description

A `TrackedPeer*` selected as the remote bootstrap source inside `queue_lock_` is dereferenced after
the lock is released. A concurrent `Peer::Close()` (configuration change, step-down, or replica
eviction) can delete that same `TrackedPeer` between the unlock and the dereference, producing a
heap use-after-free on a routine consensus path.

**Location:** `PeerMessageQueue::GetRemoteBootstrapRequestForPeer` in `src/yb/consensus/consensus_queue.cc`.

### Root cause

Under `LockGuard lock(queue_lock_)` the code sets

```c++
rbs_source = rbs_from_leader_only ? local_peer_ : VERIFY_RESULT(FindClosestPeerForBootstrap(peer));
```

and then leaves the lock scope. Outside the lock it reads

```c++
req->set_bootstrap_source_peer_uuid(rbs_source->uuid);
*req->mutable_bootstrap_source_private_addr() = {rbs_source->last_known_private_addr.begin(),
                                                 rbs_source->last_known_private_addr.end()};
```

and likewise for `broadcast_addr` and `cloud_info`.

`FindClosestPeerForBootstrap` returns `it->second` — a remote follower's `TrackedPeer*` out of
`peers_map_`, not just `local_peer_`. `UntrackPeer` is the only deleter:

```c++
LockGuard lock(queue_lock_);
TrackedPeer* peer = EraseKeyReturnValuePtr(&peers_map_, uuid);
if (peer != nullptr) { delete peer; }
```

and it is reached via `Peer::Close() -> queue_->UntrackPeer(...)`, called from
`PeerManager::ClosePeersNotInConfig` and `PeerManager::Close`. No lock spans both the dereference
and the delete.

An in-function comment already states that peer fields must be accessed within `queue_lock_`. The
most recent change to this function (#31930 / #31241) added a `pending_config_op_id` copy inside the
lock but left every `rbs_source->...` read outside it.

`local_peer_` is never deleted while the queue is open, so the bug requires a remote-follower
`rbs_source`, i.e. `remote_bootstrap_from_leader_only == false`. That flag is
[...]

PR #33104, opened 2026-08-04
[#33103] DocDB: Copy the remote bootstrap source out of peers_map_ under queue_lock_

ellabaron-code:
## Summary

`PeerMessageQueue::GetRemoteBootstrapRequestForPeer` looked the remote bootstrap source up in
`peers_map_` under `queue_lock_`, kept the raw `TrackedPeer*`, and then read its `uuid`,
`cloud_info`, `last_known_private_addr` and `last_known_broadcast_addr` after releasing the lock.

`UntrackPeer` is the only deleter of those objects and runs under the same lock, reached from
`Peer::Close()` on routine configuration change, step-down and replica-eviction paths. Nothing keeps
the peer alive across the unlock, so the request could be populated from freed memory — a garbage
UUID or addresses in the RBS request, a crash, or heap corruption once the block is reused.

`local_peer_` is never deleted while the queue is open, so this needs a remote-follower source, i.e.
`remote_bootstrap_from_leader_only == false` — the auto-promoted default on finalized clusters.

Fix: copy the selected source into a `std::optional<TrackedPeer>` while the lock is held and read
the copy afterwards. `TrackedPeer` is copy-constructible and small, and the tail of the function
only formats the request, so the copy is cheap and keeps the post-unlock region intact. The one
remaining raw dereference, `local_peer_->uuid`, is replaced with the `const local_peer_uuid_` member,
which is fixed at construction and needs no lock.

## Test plan

- [x] New regression test
      `ConsensusQueueTest.TestRemoteBootstrapSourceUntrackedAfterQueueLockRelease`: tracks two
      followers in the same zone, brings one up to date so it is chosen as the closest bootstrap
      source, drives the other into remote bootstrap with a `TABLET_NOT_FOUND` response, then
      untracks the source from a sync point placed immediately after `queue_lock_` is released and
      asserts the resulting request still carries the source's uuid, address and cloud info.
- [x] Test fails on master without the fix (release build, macOS arm64 — the freed `TrackedPeer`
      yields a corrupted `bootstrap_source_peer_uuid`) and passes with it.
- [x] Full `consensus_queue-test` binary (14 cases) passes with the fix.

### DIFF
--- a/src/yb/consensus/consensus_queue.cc
+++ b/src/yb/consensus/consensus_queue.cc
@@ -71,6 +71,7 @@
 #include "yb/util/status_log.h"
 #include "yb/util/sync_point.h"
 #include "yb/util/tostring.h"
+#include "yb/util/unique_lock.h"
 #include "yb/util/url-coding.h"
 
 using namespace std::literals;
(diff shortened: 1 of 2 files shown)


rahulb-yb:
trigger jenkins

rahulb-yb:
trigger jenkins

src/yb/consensus/consensus_queue.cc
```diff
@@ -1089,15 +1090,14 @@ Result<const PeerMessageQueue::TrackedPeer*> PeerMessageQueue::FindClosestPeerFo
 
 Status PeerMessageQueue::GetRemoteBootstrapRequestForPeer(const string& uuid,
                                                           StartRemoteBootstrapRequestPB* req) {
-  TrackedPeer* peer = nullptr;
-  const TrackedPeer* rbs_source = nullptr;
+  std::optional<TrackedPeer> rbs_source;
   int64_t current_term;
   OpId pending_config_op_id;
   {
```
basavaraj29: may be just remove the scope around the lock?

src/yb/consensus/consensus_queue.cc
basavaraj29, line 1159: these should just be short operations, and we can do them under the lock instead of introducing an optional field.

basavaraj29 (changes requested):

src/yb/consensus/consensus_queue.cc
ellabaron-code: Done — the lock now covers the whole function, so the scope and the locals that only existed to carry values out of it (`current_term`, `pending_config_op_id`) are gone. It's a `yb::UniqueLock` rather than a `LockGuard` so it can be released explicitly before the test sync point at the end; plain `std::unique_lock` isn't thread-safety annotated, so clang flags every `queue_state_` access.

src/yb/consensus/consensus_queue.cc
ellabaron-code: Agreed — they're plain protobuf assignments with no I/O and no other lock, so the request is now populated under `queue_lock_` and the source is read through the raw pointer again. No copy, no optional.

src/yb/consensus/consensus_queue.cc
```diff
@@ -1089,59 +1091,53 @@ Result<const PeerMessageQueue::TrackedPeer*> PeerMessageQueue::FindClosestPeerFo
...
-    peer->needs_remote_bootstrap = false;
-    if (PREDICT_FALSE(FLAGS_TEST_assert_remote_bootstrap_happens_from_same_zone)) {
-      CHECK_EQ(
-          TablespaceParser::GetLocalityLevel(
-              rbs_source->cloud_info.value(), peer->cloud_info.value()),
-          LocalityLevel::kZone)
-          << "Expected rbs source to be in same zone as new peer";
-    }
+  // Acess/Edit peer's fields within queue_lock_'s scope to avoid race. For instance, this peer's
+  // information could be accessed while finding RBS source for another newly added peer.
```
basavaraj29: remove this comment now that we hold the lock for the whole scope

src/yb/consensus/consensus_queue.cc
```diff
@@ -1166,6 +1162,9 @@ Status PeerMessageQueue::GetRemoteBootstrapRequestForPeer(const string& uuid,
     req->set_is_served_by_tablet_leader(true);
   }
 
+  lock.unlock();
+  TEST_SYNC_POINT("PeerMessageQueue::GetRemoteBootstrapRequestForPeer:AfterQueueLockReleased");
```
basavaraj29: I agree that the unit test helps prove the failure, but it wouldn't make any more sense now to still have it. how do you feel about removing the test totally and just checking in the code fix? 

if so, we could remove the Unique lock and just have a plain `LockGuard`.

basavaraj29 (approved):
minor comments.

src/yb/consensus/consensus_queue.cc
ellabaron-code: Fixed.

src/yb/consensus/consensus_queue.cc
ellabaron-code: Removed the test and changed back to `LockGuard`.

basavaraj29 (approved):

ellabaron-code:
trigger jenkins

rahulb-yb:
trigger jenkins

Issue #30180, opened 2026-02-02
[YSQL] [Tablet Metadata] Add attributes in `yb_tablet_metadata.tablet_attrs`


GauravSingh-yb:
Jira Link: [DB-20041](https://yugabyte.atlassian.net/browse/DB-20041)

Description

The column `yb_tablet_metadata.tablet_attrs` is currently returned as NULL / empty JSON.
This column is intended to store miscellaneous tablet-level metadata such as:

- SST file size
- Active WAL size
- Other tablet runtime/storage attributes

However, as of now, the column is not populated with any values.

--------------------------------------------------

**_Query:_**

```plpgsql
CREATE TABLE test_range_table (  --range table
  a INT,
  b INT,
  c INT,
  d INT,
  e INT,
  PRIMARY KEY (a ASC, b ASC, c ASC)
)
SPLIT AT VALUES (
  (1,2,1),
  (1,4,5),
  (1,4,10)
);

SELECT *
FROM yb_tablet_metadata
WHERE relname = 'test_range_table'
  AND db_name = 'yugabyte';
``` 

The above query returns rows where the tablet_attrs column is empty (NULL).

--------------------------------------------------

**_Sample Output:_**
```
             tablet_id           |  oid  | db_name  |     relname      | start_hash_code | end_hash_code |     leader     |     replicas     | start_range | end_range  | tablet_attrs
---------------------------------+-------+----------+------------------+-----------------+---------------+----------------+------------------+-------------+------------+--------------
9329d0a5e32f4bd2bee390e17b9cbdd5 | 16384 | yugabyte | test_range_table |                 |               | 127.0.0.1:5433 | {127.0.0.1:5433} |             | [1, 2, 1]  |              
8f8d1bcd67d844378e7e44e37296f66f | 16384 | yugabyte | test_range_table |                 |               | 127.0.0.1:5433 | {127.0.0.1:5433} | [1, 2, 1]   | [1, 4, 5]  |              
03338fa7053d4184bf76654a198e9fad | 16384 | yugabyte | test_range_table |                 |               | 127.0.0.1:5433 | {127.0.0.1:5433} | [1, 4, 5]   | [1, 4, 10] |              
ef255607ddc648d58f04f216600df9f4 | 16384 | yugabyte | test_range_table |                 |               | 127.0.0.1:5433 | {127.0.0.1:5433} | [1, 4, 10]  |            |              

``` 
                        

[...]

PR #30861, opened 2026-03-28
[#30180] YSQL: Add attributes in yb_tablet_metadata.tablet_attrs

anstormx:
> [!NOTE]
> **Medium Risk**
> Adds a new `jsonb` output column backed by tserver-side fan-out RPCs to enrich tablet metadata, which could impact latency/robustness of the metadata call despite best-effort error handling. Changes touch cross-layer plumbing (tserver RPC -> pggate structs -> Postgres SRF/view) and require careful compatibility with existing callers.
> 
> **Overview**
> Exposes new per-tablet *disk size* metadata via `yb_tablet_metadata` by adding a `tablet_attrs` `jsonb` column (also returned from `yb_get_tablet_metadata`).
> 
> Implements end-to-end plumbing to populate these attributes: the tserver `TabletsMetadata` RPC now best-effort enriches master tablet metadata by fanning out `ListTablets` to all tservers (with a 5s timeout) to fetch SST/WAL/uncompressed SST sizes, and the Postgres-side SRF builds `tablet_attrs` JSONB from the returned sizes.
> 
> Adds a migration to update `pg_proc`/recreate the view, and extends regression tests to validate expected JSON keys and non-negative values.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 976b0f3fbe227ec86f50cd83554ca12a832b8a56. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #30294, opened 2026-02-12
YSQL: Add `attributes (jsonb)` to `yb_tablet_metadata`

keisku:
Closes https://github.com/yugabyte/yugabyte-db/issues/29665

---

**Testing**

Upgrade ysql.

```bash
./bin/yb-ctl --rf 3 create
./build/latest/bin/yb-admin upgrade_ysql
```

Able to see `active_ssts_size` and `wals_size` in the `yb_tablet_metadata`.

```bash
./bin/ysqlsh -c "\d yb_tablet_metadata"
            View "pg_catalog.yb_tablet_metadata"
     Column      |  Type   | Collation | Nullable | Default
-----------------+---------+-----------+----------+---------
 tablet_id       | text    |           |          |
 oid             | oid     |           |          |
 db_name         | text    |           |          |
 relname         | text    |           |          |
 start_hash_code | integer |           |          |
 end_hash_code   | integer |           |          |
 leader          | text    |           |          |
 replicas        | text[]  |           |          |
 attributes      | jsonb   |           |          |
```

Run this right after creating a cluster

```bash
./bin/ysqlsh -c "
SELECT *
FROM yb_tablet_metadata
WHERE relname = 'test_sizes';
"
 tablet_id | oid | db_name | relname | start_hash_code | end_hash_code | leader | replicas | attributes
-----------+-----+---------+---------+-----------------+---------------+--------+----------+------------
(0 rows)
```

Create a test table

```bash
./bin/ysqlsh -c "CREATE TABLE test_sizes (k INT PRIMARY KEY, v TEXT);"
CREATE TABLE
```

See values in `attributes`.

```bash
./bin/ysqlsh -P pager=off -c "
SELECT *
FROM yb_tablet_metadata
WHERE relname = 'test_sizes';
"
            tablet_id             |  oid  | db_name  |  relname   | start_hash_code | end_hash_code |     leader     |                    replicas                    |                                                                                                    attributes
----------------------------------+-------+----------+------------+-----------------+---------------+----------------+------------------------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #33108, opened 2026-08-04
[#30180] YSQL: Expose colocated tablet disk size via tablet_attrs and introduce yb_tablegroup_size() function

ajcaldera1:
Users in Aeon are unable to see the size of tables stored in the colocated tablegroup, it would be helpful to have the size for the tablet exposed via YSQL so that it can be monitored to not exceed our general guidance on colocated tablets.

Populate yb_tablet_metadata.tablet_attrs with leader SST/WAL sizes from master drive info, and add yb_tablegroup_size() for colocation parent footprint. Leave pg_table_size empty for colocated child tables to avoid overcounting shared storage.

## Upgrade/Rollback safety
Optional PgGetTableDiskSizeRequestPB.yb_table_id field is ignored by older tservers; tablet_attrs changes from always-NULL to populated JSON (additive). New yb_tablegroup_size arrives via YSQL migration V109.

---
_automated · Cursor Grok 4.5_

### DIFF
--- a/docs/content/stable/explore/observability/yb-tablet-metadata.md
+++ b/docs/content/stable/explore/observability/yb-tablet-metadata.md
@@ -39,7 +39,7 @@ The following table describes the columns of the `yb_tablet_metadata` view.
 | replicas | text[] | A list of replica IP addresses and port (includes leader) associated with the tablet. |
 | start_range | text | Starting range key (inclusive) for the tablet. (NULL for hash-sharded tables.) |
 | end_range | text | Ending range key (exclusive) for the tablet. (NULL for hash-sharded tables.) |
-| tablet_attrs | json | Reserved for future use. Currently empty. |
+| tablet_attrs | json | Leader-replica disk size fields when available: `sst_bytes`, `wal_bytes`, `uncompressed_sst_bytes`, `total_bytes` (SST+WAL), and optionally `vector_index_bytes`. NULL when sizes are unavailable or the row is privilege-masked. |
 | tablet_state | text | Current state of the tablet. <br>Possible tablet states are `PREPARING`, `CREATING`, `RUNNING`, `REPLACED`, or `DELETED`. |
 
 The `relname`, `start_range`, and `end_range` columns are masked for unprivileged users (shown as `<insufficient privilege>`). Only superusers or users with the `yb_db_admin` role can see the entire unmasked view.
--- a/docs/content/stable/explore/observability/yb-tablet-metadata.md
+++ b/docs/content/stable/explore/observability/yb-tablet-metadata.md
@@ -37,6 +37,10 @@ The following table describes the columns of the `yb_tablet_metadata` view.
 | end_hash_code | int | Ending hash code (exclusive) for the tablet. (NULL for range-sharded tables.) |
 | leader | text | IP address, port of the leader node for the tablet. |
 | replicas | text[] | A list of replica IP addresses and port (includes leader) associated with the tablet. |
+| start_range | text | Decoded start key for range-sharded tablets. (NULL for hash-sharded tables.) |
+| end_range | text | Decoded end key for range-sharded tablets. (NULL for hash-sharded tables.) |
+| tablet_attrs | jsonb | Leader-replica disk size fields when available: `sst_bytes`, `wal_bytes`, `uncompressed_sst_bytes`, `total_bytes` (SST+WAL), and optionally `vector_index_bytes`. NULL when sizes are unavailable or the row is privilege-masked. |
+| tablet_state | text | Master-side tablet lifecycle state (for example, RUNNING). |
 
 ## Examples
 
--- a/docs/content/stable/explore/observability/yb-tablet-metadata.md
+++ b/docs/content/stable/explore/observability/yb-tablet-metadata.md
@@ -39,7 +39,7 @@ The following table describes the columns of the `yb_tablet_metadata` view.
 | replicas | text[] | A list of replica IP addresses and port (includes leader) associated with the tablet. |
 | start_range | text | Starting range key (inclusive) for the tablet. (NULL for hash-sharded tables.) |
 | end_range | text | Ending range key (exclusive) for the tablet. (NULL for hash-sharded tables.) |
-| tablet_attrs | json | Leader-replica disk size fields when available: `sst_bytes`, `wal_bytes`, `uncompressed_sst_bytes`, `total_bytes` (SST+WAL), and optionally `vector_index_bytes`. NULL when sizes are unavailable or the row is privilege-masked. |
+| tablet_attrs | json | Tablet leader disk size fields when available: `sst_bytes`, `wal_bytes`, `uncompressed_sst_bytes`, `total_bytes` (SST+WAL), and optionally `vector_index_bytes`. NULL when sizes are unavailable or the row is privilege-masked. |
 | tablet_state | text | Current state of the tablet. <br>Possible tablet states are `PREPARING`, `CREATING`, `RUNNING`, `REPLACED`, or `DELETED`. |
 
 The `relname`, `start_range`, and `end_range` columns are masked for unprivileged users (shown as `<insufficient privilege>`). Only superusers or users with the `yb_db_admin` role can see the entire unmasked view.
--- a/src/postgres/src/backend/utils/misc/pg_yb_utils.c
+++ b/src/postgres/src/backend/utils/misc/pg_yb_utils.c
@@ -86,6 +86,7 @@
 #include "catalog/pg_yb_logical_client_version.h"
 #include "catalog/pg_yb_profile.h"
 #include "catalog/pg_yb_role_profile.h"
+#include "catalog/pg_yb_tablegroup.h"
 #include "catalog/yb_catalog_version.h"
 #include "catalog/yb_logical_client_version.h"
 #include "catalog/yb_type.h"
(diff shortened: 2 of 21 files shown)


src/postgres/src/backend/utils/misc/pg_yb_utils.c
```diff
@@ -9396,8 +9437,42 @@ yb_get_tablet_metadata(PG_FUNCTION_ARGS)
...
-		/* TODO(#30180): Populate tablet_attrs. */
-		nulls[12] = true;
+		/*
+		 * tablet_attrs: leader SST/WAL sizes. Omit when privilege-masked or
+		 * when master did not report drive info for the leader.
+		 */
+		if (show_real && tablet->has_disk_size)
+		{
+			char		attrs_buf[256];
+			int64		total_bytes =
```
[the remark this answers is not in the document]
ajcaldera1: won't fix — same as GetTableDiskSize in master; overflow not reachable for disk sizes

docs/content/stable/explore/observability/yb-tablet-metadata.md
```diff
@@ -39,7 +39,7 @@ The following table describes the columns of the `yb_tablet_metadata` view.
 | replicas | text[] | A list of replica IP addresses and port (includes leader) associated with the tablet. |
 | start_range | text | Starting range key (inclusive) for the tablet. (NULL for hash-sharded tables.) |
 | end_range | text | Ending range key (exclusive) for the tablet. (NULL for hash-sharded tables.) |
-| tablet_attrs | json | Reserved for future use. Currently empty. |
+| tablet_attrs | json | Leader-replica disk size fields when available: `sst_bytes`, `wal_bytes`, `uncompressed_sst_bytes`, `total_bytes` (SST+WAL), and optionally `vector_index_bytes`. NULL when sizes are unavailable or the row is privilege-masked. |
```
iSignal: minor: Tablet leader disk size fields reads better than Leader-replica

src/postgres/src/backend/utils/misc/pg_yb_utils.c
```diff
@@ -6101,6 +6102,45 @@ yb_is_database_colocated(PG_FUNCTION_ARGS)
...
+	int64_t		size = 0;
+	int32_t		num_missing_tablets = 0;
+	HeapTuple	tuple;
+
+	tuple = SearchSysCache1(YBTABLEGROUPOID, ObjectIdGetDatum(tablegroup_oid));
+	if (!HeapTupleIsValid(tuple))
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("tablegroup with OID %u does not exist", tablegroup_oid)));
+
```
iSignal, line 6125: minor: we can release the tuple here. we only need it for the error msg below and can copy it out here.

src/yb/tserver/pg_client.proto
```diff
@@ -1068,6 +1068,9 @@ message PgDeleteDBSequencesResponsePB {
 message PgGetTableDiskSizeRequestPB {
   PgObjectIdPB table_id = 1;
   reserved 2; // ASH metadata is now passed in RequestHeader PB
+  // DocDB table UUID. When set, takes precedence over table_id. Used for
+  // colocation parent tables which have no PG oid mapping.
+  optional string yb_table_id = 3;
```
iSignal, line 1073: minor: we could omit this yb table id and let table_id above carry the tablegroup id. If pg client service fails to look up the docdb table id from the table id, it can fallback to assuming this is a tablegroup and looking up the colocation or tablegroup suffix? that avoids the pggate plumbing.

iSignal (commented):
would be good to also have a test case for colocation + tablespaces, though that should work with current code.

@hbhanawat : Given we had a couple of older PRs that you looked at, let us know if you have any thoughts on other ways to expose this info? 

https://github.com/yugabyte/yugabyte-db/pull/30861
https://github.com/yugabyte/yugabyte-db/pull/30294

src/postgres/src/include/catalog/pg_proc.dat
```diff
@@ -12361,4 +12361,10 @@
   proargmodes => '{i,i,i,o,o,o}',
   proargnames => '{hba_path,guc_path,ident_path,hba_error,guc_error,ident_error}',
   prosrc => 'yb_pg_validate_conf_file'},
+
+{ oid => '8116',
```
iSignal: maybe rebase because this may be taken already.

docs/content/stable/explore/observability/yb-tablet-metadata.md
ajcaldera1: Updated to \"Tablet leader disk size fields\" in `0bf9d43996e`.

---
_automated · Cursor Grok 4.6_

src/postgres/src/backend/utils/misc/pg_yb_utils.c
ajcaldera1: Good catch — we now copy `grpname` and `ReleaseSysCache` before the disk-size RPC in `0bf9d43996e`, so the cache pin is not held across the network call.

---
_automated · Cursor Grok 4.6_

src/yb/tserver/pg_client.proto
ajcaldera1: Prefer keeping the explicit `yb_table_id` field.

`PgObjectId::GetYbTableIdFromPB` only converts the OID to a UUID — it does not fail — so a tablegroup oid would produce the bare tablegroup id (no `.colocation.parent` / `.tablegroup.parent` suffix). Resolving the parent needs `MyDatabaseColocated` (or trying both suffixes). Doing that as a NotFound fallback on the shared `GetTableDiskSize` path would either mis-handle missing tables or still need a flag. Building the parent UUID at the call site where colocation is known keeps the RPC unambiguous and upgrade-safe as an optional field.

Happy to revisit if you still prefer the heuristic approach.

---
_automated · Cursor Grok 4.6_

src/postgres/src/include/catalog/pg_proc.dat
ajcaldera1: Rebased onto latest `upstream/master`. OID `8116` is still free there (upstream ends at `8115`), and the migration remains `V109`.

---
_automated · Cursor Grok 4.6_

GauravSingh-yb (commented):
I am trying to understand extra functionality the new function `yb_tablegroup_size()` is adding. We could get leader SST+WAL file sizes via a SQL command on `yb_tablet_metadata`. 

```
SELECT SUM((tablet_attrs->>'total_bytes')::bigint) AS bytes
FROM yb_tablet_metadata
WHERE db_name = current_database()
  AND oid = 'table_name'::regclass::oid;
```
 Am I missing some some additional functionality?

src/postgres/src/include/catalog/pg_proc.dat
```diff
@@ -12361,4 +12361,10 @@
...
+  descr => 'Disk size in bytes of a tablegroup colocation parent tablet (leader SST+WAL)',
+  proname => 'yb_tablegroup_size', provolatile => 'v',
+  prorettype => 'int8', proargtypes => 'oid',
```
GauravSingh-yb, line 12368: I think the function defaults to `proparallel = 's'` if not mentioned explicitly. This should be 'r' restricted like other master lookup functions. Will need to fix the migration file as well.

ajcaldera1: Set `proparallel => 'r'` in `pg_proc.dat` and in V109 (INSERT plus an UPDATE so a re-run still lands restricted if the row was inserted as safe). Shipped in `6e6f6120b04`.

---
_automated · Cursor Grok 4.6_

ajcaldera1:
> I am trying to understand extra functionality the new function `yb_tablegroup_size()` is adding. We could get leader SST+WAL file sizes via a SQL command on `yb_tablet_metadata`.
> 
```
SELECT SUM((tablet_attrs->>'total_bytes')::bigint) AS bytes
FROM yb_tablet_metadata
WHERE db_name = current_database()
AND oid = 'table_name'::regclass::oid;
```
> 
> Am I missing some some additional functionality?

@GauravSingh-yb  -

I'll admit it's a bit of syntax sugar, but it's important to note that colocation parents have no `pgclass` entry and therefore no `oid`. This would be more correct:

```
SELECT (tablet_attrs->>'total_bytes')::bigint
FROM yb_tablet_metadata
WHERE db_name = current_database()
  AND relname LIKE '%.colocation.parent.tablename';
```

src/postgres/src/include/catalog/pg_proc.dat
GauravSingh-yb: We should bump up the oid. I see the last used oid as 8117 in `catalog.h`
```
#define YB_LAST_USED_OID 8117 /* corresponds to pg_authid.yb_global_views_user */
```

GauravSingh-yb: I think we should bump up oid.
```
#define YB_LAST_USED_OID 8117 /* corresponds to pg_authid.yb_global_views_user */
```

ajcaldera1: Agreed. 8116 is already `GRAPHIDOID`, and `YB_LAST_USED_OID` is 8117 (`pg_authid.yb_global_views_user`). `yb_tablegroup_size` is now OID 8118, with `YB_LAST_USED_OID` bumped to match (`643c2a175ce`).

---
_automated · Cursor Grok 4.6_

GauravSingh-yb:
LGTM!

PR #33111, opened 2026-08-04
[#16670] DocDB: dist-trace: span/scope API and its direct callers

ellabaron-code:
## Summary

Part 1 of 5. This stack replaces the earlier 9-PR split (#32895–#32903) with five
themed PRs, each of which carries the tests for what it adds and passes on its own
commit. The content is otherwise the same change, re-partitioned.

Rework the `dist_trace` library into the span+scope API the rest of the stack builds
on, and update every in-tree caller of the changed API in lockstep so the tree keeps
building:

- `SpanWithScope` (a span bundled with an activated thread-local scope), plus
  `Start{Client,Server}SpanWithScope` and `ActivateParentScope`.
- `GetActiveSpanContext` for reading the current span's identity.
- `InitDistTrace` now takes a service name; `CleanupDistTrace` → `ShutdownDistTrace`
  (mirrored in `ybc_dist_trace` and its `pg_yb_utils.c` caller).
- Replace the `RpcSpanAttrs` pending-attr buffer with a plain string-pair vector
  drained internally by `StartClientSpanWithScope`; drop the now-unused public
  `GetPendingRpcAttrPairs`/`ClearPendingRpcAttrs`.

Callers updated:

- `outbound_call`: client span via `StartClientSpanWithScope`, restore the parent
  around the completion callback, and report the RPC's user-facing status message on
  the span.
- `pg_client` / `pg_session`: shared-memory client span and pending RPC attributes.
- `dist_trace-test`: new `InitDistTrace` signature and `ShutdownDistTrace` rename,
  plus the span-name and status-message expectations that follow from the two caller
  changes above (`"shmem req X"` → `"shmem X"`; RPC span status now carries
  `Status::ToUserMessage()` text rather than the error code's name). Keeping these
  with the change that causes them is what makes this commit green on its own — in
  the previous split they lived six PRs later, which is why every intermediate PR
  was red.

Nothing here writes a trace context onto the wire: that needs `TraceContextPB`,
which arrives with the RPC-boundary PR, so the header write and the matching reader
land together there.

### Batching flag changes and why they diverge from the OTel SDK defaults

- `otel_batch_max_queue_size` 2048 → 16384. A tserver under load produces spans far
  faster than a single PG backend does; at 2048 the queue overflows and drops spans,
  which shows up as holes in a trace rather than as slowness.
- New `otel_ysql_batch_max_queue_size`, defaulting to the previous 2048, used only by
  the ysql process. A backend is one session's worth of spans and there is one
  process per connection, so the larger queue would be paid per connection for no
  benefit.
- `otel_batch_schedule_delay_ms` 5000 → 500. The 5 s default means a span can sit in
  the exporter for 5 s after the operation finished; that is longer than most of the
  queries being traced, and makes interactive debugging (and tests that wait for a
  span) needlessly slow.

`StartSpanWithScope` has no caller outside `dist_trace.cc` yet —
`StartClientSpanWithScope` is built on it — but it is the primitive the two public
variants are defined in terms of, so it stays public rather than being inlined into
its one use.

Also drops a comment on `StartClientSpanWithScope` claiming the span may be "an
optional root gated by `otel_rpc_sampling_ratio`": no such flag exists, and the
function returns `nullptr` when there is no active context.

## Test plan

      commit (the point of pulling the test-literal fixes back into this PR).

### Stack

1. #33111 — span/scope API and its direct callers

Together these replace #32895–#32903, which will be closed.

### DIFF
--- a/src/postgres/src/backend/utils/misc/pg_yb_utils.c
+++ b/src/postgres/src/backend/utils/misc/pg_yb_utils.c
@@ -1098,7 +1098,7 @@ YBInitPostgresBackend(const char *program_name, const YbcPgInitPostgresInfo *ini
 			hex_encode((const char *) YbGetLocalTServerUuid(), UUID_LEN, hex_uuid);
 			hex_uuid[2 * UUID_LEN] = '\0';
 
-			YBCInitDistTrace(MyProcPid, hex_uuid);
+			YBCInitDistTrace(hex_uuid);
 
 			/* Hooks that close node spans left open by a query abort. */
 			YbDistTraceInstallExecutorHooks();
@@ -1110,7 +1110,7 @@ void
 YBOnPostgresBackendShutdown()
 {
 	if (YBCIsDistTraceEnabled())
-		YBCCleanupDistTrace();
+		YBCShutdownDistTrace();
 
 	YBCDestroyPgGate();
 }
--- a/src/yb/rpc/outbound_call.cc
+++ b/src/yb/rpc/outbound_call.cc
@@ -151,13 +151,11 @@ bool FinishedState(RpcCallState state) {
   return false;
 }
 
-void SetSpanStatus(opentelemetry::trace::Span& span, RpcCallState state) {
+void SetSpanStatus(opentelemetry::trace::Span& span, RpcCallState state, const Status& status) {
   switch (state) {
     case TIMED_OUT:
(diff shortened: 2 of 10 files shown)


ellabaron-code:
Superseded by #33116, which is the same commit opened from the upstream branch so the whole stack (#33116 → #33112 → #33113 → #33114 → #33115) references one copy of each branch. Closing this one; no review has happened here yet.

PR #33112, opened 2026-08-04
[#16670] DocDB: dist-trace: Propagate trace context across the RPC boundary

ellabaron-code:
## Summary

Carry a distributed trace from an RPC caller to its callee. The request header gains an
optional trace-context field; the caller writes its active span's context into it while
the header is being written, and the callee decodes it and opens a server span parented
under the caller's span, so one trace spans both processes. Short-circuited local calls
take the parent directly from the outbound call, since no header exists. Decoding is
best-effort — a malformed context is logged and the call proceeds untraced. Master and
tserver also start their process-wide tracer here, without which none of these spans are
exported. The wire field, its writer and its reader land together because none of them
is observable on its own.

## Upgrade/Rollback safety

`src/yb/rpc/rpc_header.proto` gains one new optional field:
`RequestHeader.trace_context` (tag 12), of a new message type `TraceContextPB`. No
existing field is renumbered, retyped, renamed, or removed, and no default changes.

**Forward compatibility (new client → old server).** A new node only writes the field
when distributed tracing is enabled (`otel_collector_traces_endpoint` set) *and* a
trace is active for the call; otherwise the header bytes are exactly what they are
today. When it is written, an old server's protobuf parser does not know tag 12 and
`SkipField`s it into unknown fields, then handles the RPC normally. The field is
purely observability metadata — no request semantics depend on it — so the old server
simply serves an untraced call.

**Backward compatibility (old client → new server).** The field is absent.
`ParsedRequestHeader.trace_context` stays empty, `YBInboundCall::ParseFrom` leaves
`parent_span_context_` invalid, and `CreateServerSpan` starts no server span. Same
behavior as tracing being disabled.

**Malformed input.** `ParseTraceContext` / `ToSpanContext` reject a zero or otherwise
invalid context. Parse failure is logged and the call proceeds untraced — a bad or
hostile trace context can never fail an RPC.

**Rollback.** Nothing is persisted: the field exists only in in-flight RPC headers, so
there is no on-disk or catalog state to migrate or clean up. Rolling a node back to a
build without this change makes it fall into the "old server" case above from the
first RPC onward; rolling the whole cluster back leaves no trace of the field.
Enabling or disabling `otel_collector_traces_endpoint` is likewise a pure runtime
toggle with no persistent effect.

## Test plan

      commit, including the new `TestRpcSpanReachesTabletServer`.

`TestRpcSpanReachesTabletServer` runs a `SELECT` under a known traceparent and asserts
that the tserver's inbound `rpc yb.tserver.PgClientService.Perform` span lands in that
trace as a child of the ysql backend's outbound span, with the expected
`rpc.service`/`rpc.method` attributes. The new `WaitForRemoteChildSpan` collector
helper does the caller/callee pairing check.

Mixed-version wire behavior is argued from the encoding rather than exercised by a
test — see the Upgrade/Rollback section. An old peer `SkipField`s the unknown tag, and
the field is only written when tracing is on and a trace is active.

### Stack

### DIFF
--- a/src/yb/rpc/local_call.h
+++ b/src/yb/rpc/local_call.h
@@ -50,6 +50,16 @@ class LocalOutboundCall : public OutboundCall {
     return req_;
   }
 
+  // Span context of this outbound call's trace span, to parent the matching local inbound span
+  // (local calls carry no wire header). nullopt when tracing is off. Safe after the scope is
+  // dropped.
+  std::optional<opentelemetry::trace::SpanContext> otel_span_context() const {
+    if (const auto& span = otel_span(); span) {
+      return span->GetContext();
+    }
+    return std::nullopt;
+  }
+
   bool is_local() const override { return true; }
 
   ThreadPoolTag pool_tag() const;
@@ -134,6 +144,7 @@ auto HandleCall(InboundCallPtr call, F f) {
       f(req, resp, std::move(rpc_context));
     }
   }
+  yb_call->DropServerSpanScope();
 }
 
 class LocalYBInboundCallTracker : public InboundCall::CallProcessedListener {
--- a/src/yb/rpc/outbound_call.cc
+++ b/src/yb/rpc/outbound_call.cc
@@ -277,19 +277,18 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
   IncrementCounter(rpc_metrics_->outbound_calls_created);
   IncrementGauge(rpc_metrics_->outbound_calls_alive);
 
-  // Capture this call's parent before StartClientSpanWithScope makes otel_span_ current;
-  // InvokeCallbackSync restores it around the callback so its follow-on work nests as a sibling.
+  // Capture this call's parent; InvokeCallbackSync restores it around the callback so its
+  // follow-on work nests as a sibling.
   trace_parent_ = dist_trace::GetActiveSpanContext();
 
-  otel_span_ = dist_trace::StartClientSpanWithScope(Format("rpc $0", remote_method_.ToString()));
+  // Not attached: the wire header and the local inbound call both take this span from GetContext().
+  otel_span_ = dist_trace::StartClientSpanWithScope(
+      Format("rpc $0", remote_method_.ToString()), /*attach=*/false);
   if (otel_span_) {
     otel_span_->SetAttribute("rpc.system", "outbound_rpc");
     otel_span_->SetAttribute("rpc.service", remote_method_.service_name());
     otel_span_->SetAttribute("rpc.method", remote_method_.method_name());
     otel_span_->SetAttribute("rpc.call_id", call_id_);
-    // Detach right away, while the token is still top of stack. The span needs no ambient context:
-    // both the wire header and the local inbound call take it from GetContext().
-    otel_span_->DropScope();
   }
 }
 
--- a/src/yb/rpc/outbound_call.cc
+++ b/src/yb/rpc/outbound_call.cc
@@ -378,6 +378,33 @@ Status OutboundCall::SetRequestParam(
     metadata_size += 1; // add tag size of RequestHeader::kMetadataFieldNumber
   }
 
+  // Distributed-trace context: extract the outbound span's SpanContext (if any) and pre-compute the
+  // serialized size of the TraceContextPB submessage so it can be folded into the header length.
+  size_t trace_context_size = 0;
+  size_t trace_context_message_size = 0;
+  uint32_t version_and_flags = 0;
+  opentelemetry::trace::TraceId trace_id;
+  opentelemetry::trace::SpanId span_id;
+  if (otel_span_) {
+    auto span_context = otel_span_->GetContext();
+    if (span_context.IsValid()) {
+      trace_id = span_context.trace_id();
+      span_id = span_context.span_id();
+      auto trace_flags = span_context.trace_flags();
+      // TraceContextPB submessage layout:
+      // - trace_id_hi: 1 byte tag + 8 bytes fixed64
+      // - trace_id_lo: 1 byte tag + 8 bytes fixed64
+      // - span_id:     1 byte tag + 8 bytes fixed64
+      // - version_and_flags: 1 byte tag + varint (size depends on the combined value)
+      constexpr uint32_t kVersion = 0;
+      version_and_flags = (kVersion << 8) | trace_flags.flags();
+      size_t version_and_flags_varint_size = Output::VarintSize32(version_and_flags);
+      trace_context_message_size = 3 * 9 + 1 + version_and_flags_varint_size;
+      trace_context_size =
+          1 + Output::VarintSize64(trace_context_message_size) + trace_context_message_size;
+    }
+  }
+
   auto use_crc = FLAGS_rpc_enable_crc;
   size_t header_pb_len = 1 + call_id_size + // int32 call_id = 1
                          serialized_remote_method.size() + // RemoteMethodPB remote_method = 2
@@ -386,6 +413,7 @@ Status OutboundCall::SetRequestParam(
   if (pool_tag) {
     header_pb_len += 1 + Output::VarintSize64(pool_tag); // uint64 pool_tag = 7
   }
+  header_pb_len += trace_context_size; // TraceContext trace_context = 8
   if (use_crc) {
     header_pb_len += 1 + sizeof(uint32_t); // fixed32 crc = 15
   }
@@ -435,6 +463,33 @@ Status OutboundCall::SetRequestParam(
   if (pool_tag) {
     dst = CodedOutputStream::WriteTagToArray(RequestHeader::kPoolTagFieldNumber << 3, dst);
     dst = Output::WriteVarint64ToArray(pool_tag, dst);
+  }
+
+  if (trace_context_size > 0) {
+    // Write the TraceContextPB submessage. Field numbers match yb.rpc.TraceContextPB in
+    // rpc_header.proto.
+    // The 16-byte trace id splits into two big-endian 64-bit halves; the span id is one big-endian
+    // 64-bit.
+    dst = Output::WriteTagToArray(
+        (RequestHeader::kTraceContextFieldNumber << 3) | WireFormatLite::WIRETYPE_LENGTH_DELIMITED,
+        dst);
+    dst = Output::WriteVarint32ToArray(narrow_cast<uint32_t>(trace_context_message_size), dst);
+
+    dst = Output::WriteTagToArray(
+        (TraceContextPB::kTraceIdHiFieldNumber << 3) | WireFormatLite::WIRETYPE_FIXED64, dst);
+    dst = Output::WriteLittleEndian64ToArray(BigEndian::Load64(trace_id.Id().data()), dst);
+
+    dst = Output::WriteTagToArray(
+        (TraceContextPB::kTraceIdLoFieldNumber << 3) | WireFormatLite::WIRETYPE_FIXED64, dst);
+    dst = Output::WriteLittleEndian64ToArray(BigEndian::Load64(trace_id.Id().data() + 8), dst);
+
+    dst = Output::WriteTagToArray(
+        (TraceContextPB::kSpanIdFieldNumber << 3) | WireFormatLite::WIRETYPE_FIXED64, dst);
+    dst = Output::WriteLittleEndian64ToArray(BigEndian::Load64(span_id.Id().data()), dst);
+
+    dst = Output::WriteTagToArray(
+        (TraceContextPB::kVersionAndFlagsFieldNumber << 3) | WireFormatLite::WIRETYPE_VARINT, dst);
+    dst = Output::WriteVarint32ToArray(version_and_flags, dst);
   }
 
   // CRC should be at the end of header, otherwise adjust CRC filling logic below.
--- a/src/yb/rpc/outbound_call.cc
+++ b/src/yb/rpc/outbound_call.cc
@@ -287,6 +287,8 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
     otel_span_->SetAttribute("rpc.service", remote_method_.service_name());
     otel_span_->SetAttribute("rpc.method", remote_method_.method_name());
     otel_span_->SetAttribute("rpc.call_id", call_id_);
+    // Detach right away, while the token is still top of stack. The span needs no ambient context:
+    // both the wire header and the local inbound call take it from GetContext().
     otel_span_->DropScope();
   }
 }
--- a/src/yb/rpc/rpc_context.cc
+++ b/src/yb/rpc/rpc_context.cc
@@ -147,11 +147,16 @@ RpcContext::RpcContext(std::shared_ptr<YBInboundCall> call,
     RespondRpcFailure(ErrorStatusPB::ERROR_INVALID_REQUEST, s);
     return;
   }
+  call_->CreateServerSpan();
   TRACE_EVENT_ASYNC_BEGIN1("rpc_call", "RPC", this, "call", call_->ToString());
 }
 
 RpcContext::RpcContext(std::shared_ptr<LocalYBInboundCall> call)
     : call_(call), params_(call.get(), boost::null_deleter()) {
+  // Inbound (server-side) span for this local call.
+  if (auto outbound_call = call->outbound_call(); outbound_call) {
+    call_->CreateServerSpan(outbound_call->otel_span_context());
+  }
   TRACE_EVENT_ASYNC_BEGIN1("rpc_call", "RPC", this, "call", call_->ToString());
 }
 
--- a/src/yb/rpc/rpc_header.proto
+++ b/src/yb/rpc/rpc_header.proto
@@ -54,6 +54,21 @@ message RemoteMethodPB {
   required string method_name = 2;
 };
 
+// Distributed-trace context (W3C trace-context; https://www.w3.org/TR/trace-context/) propagated
+// across a process boundary. Defined here so a single message and wire format serves both
+// transports: it is carried in the RPC RequestHeader below (alongside AshMetadataPB), and as an
+// independent length-prefixed slot in the shared-memory PG<->tserver exchange.
+message TraceContextPB {
+  // 8 bytes (high 64 bits of trace_id)
+  optional fixed64 trace_id_hi = 1;
+  // 8 bytes (low 64 bits of trace_id)
+  optional fixed64 trace_id_lo = 2;
+  // 8 bytes
+  optional fixed64 span_id = 3;
+  // Low byte: flags, High byte: version (1-2 bytes on wire)
+  optional uint32 version_and_flags = 4;
+}
+
 // The header for the RPC request frame.
 message RequestHeader {
   reserved 6;
@@ -81,6 +96,9 @@ message RequestHeader {
   optional AshMetadataPB metadata = 5;
 
   optional uint64 pool_tag = 7;
+
+  // Distributed-trace context, propagating the trace from caller to the server handling this RPC.
+  optional TraceContextPB trace_context = 8;
 
   // CRC32C for message body and header without CRC field. Use 15 as max 1 byte tag.
   optional fixed32 crc = 15;
--- a/src/yb/rpc/serialization.cc
+++ b/src/yb/rpc/serialization.cc
@@ -229,6 +229,9 @@ Status ParseHeader(
       case RequestHeader::kMetadataFieldNumber:
         parsed_header->metadata = VERIFY_RESULT(ParseString(buf, "metadata", in));
         break;
+      case RequestHeader::kTraceContextFieldNumber:
+        parsed_header->trace_context = VERIFY_RESULT(ParseString(buf, "trace_context", in));
+        break;
       default: {
         if (!SkipField(tag & 7, in)) {
           return STATUS_FORMAT(Corruption, "Unable to skip: $0", tag);
@@ -372,6 +375,93 @@ Result<ParsedRemoteMethod> ParseRemoteMethod(const Slice& buf) {
     }
   }
   return result;
+}
+
+namespace {
+
+// Rebuilds a remote SpanContext from the wire fields. Fails if the trace/span ids are zero.
+Result<opentelemetry::trace::SpanContext> BuildSpanContext(
+    uint64_t trace_id_hi, uint64_t trace_id_lo, uint64_t span_id, uint32_t version_and_flags) {
+  // 16-byte trace id from its high/low 64-bit halves; 8-byte span id (big-endian on the wire).
+  uint8_t trace_id_bytes[16];
+  BigEndian::Store64(trace_id_bytes, trace_id_hi);
+  BigEndian::Store64(trace_id_bytes + 8, trace_id_lo);
+  uint8_t span_id_bytes[8];
+  BigEndian::Store64(span_id_bytes, span_id);
+
+  // Flags are the low byte; the high byte is the (currently unused) version.
+  uint8_t flags = version_and_flags & 0xFF;
+
+  auto span_context = opentelemetry::trace::SpanContext(
+      opentelemetry::trace::TraceId(trace_id_bytes),
+      opentelemetry::trace::SpanId(span_id_bytes),
+      opentelemetry::trace::TraceFlags(flags),
+      /* is_remote= */ true);
+
+  if (!span_context.IsValid()) {
+    return STATUS(Corruption, "Invalid trace context: trace_id_hi/lo or span_id is zero");
+  }
+  return span_context;
+}
+
+}  // namespace
+
+Result<opentelemetry::trace::SpanContext> ParseTraceContext(const Slice& buf) {
+  CodedInputStream in(buf.data(), narrow_cast<int>(buf.size()));
+  in.PushLimit(narrow_cast<int>(buf.size()));
+  uint64_t trace_id_hi = 0, trace_id_lo = 0, span_id = 0;
+  uint32_t version_and_flags = 0;
+  bool has_trace_id_hi = false, has_trace_id_lo = false, has_span_id = false,
+       has_version_and_flags = false;
+  while (in.BytesUntilLimit() > 0) {
+    auto tag = in.ReadTag();
+    auto field = tag >> 3;
+    switch (field) {
+      case TraceContextPB::kTraceIdHiFieldNumber:
+        if (!in.ReadLittleEndian64(&trace_id_hi)) {
+          return STATUS(Corruption, "Unable to decode trace_id_hi");
+        }
+        has_trace_id_hi = true;
+        break;
+      case TraceContextPB::kTraceIdLoFieldNumber:
+        if (!in.ReadLittleEndian64(&trace_id_lo)) {
+          return STATUS(Corruption, "Unable to decode trace_id_lo");
+        }
+        has_trace_id_lo = true;
+        break;
+      case TraceContextPB::kSpanIdFieldNumber:
+        if (!in.ReadLittleEndian64(&span_id)) {
+          return STATUS(Corruption, "Unable to decode span_id");
+        }
+        has_span_id = true;
+        break;
+      case TraceContextPB::kVersionAndFlagsFieldNumber:
+        if (!in.ReadVarint32(&version_and_flags)) {
+          return STATUS(Corruption, "Unable to decode version_and_flags");
+        }
+        has_version_and_flags = true;
+        break;
+      default: {
+        if (!SkipField(tag & 7, &in)) {
+          return STATUS_FORMAT(Corruption, "Unable to skip: $0", tag);
+        }
+      }
+    }
+  }
+  if (!has_trace_id_hi || !has_trace_id_lo || !has_span_id || !has_version_and_flags) {
+    return STATUS(Corruption, "Incomplete trace context: missing field");
+  }
+  return BuildSpanContext(trace_id_hi, trace_id_lo, span_id, version_and_flags);
+}
+
+Result<opentelemetry::trace::SpanContext> ToSpanContext(const TraceContextPB& trace_context) {
+  if (!trace_context.has_trace_id_hi() || !trace_context.has_trace_id_lo() ||
+      !trace_context.has_span_id() || !trace_context.has_version_and_flags()) {
+    return STATUS(Corruption, "Incomplete trace context: missing field");
+  }
+  return BuildSpanContext(
+      trace_context.trace_id_hi(), trace_context.trace_id_lo(), trace_context.span_id(),
+      trace_context.version_and_flags());
 }
 
 std::string ParsedRequestHeader::RemoteMethodAsString() const {
--- a/src/yb/rpc/yb_rpc.cc
+++ b/src/yb/rpc/yb_rpc.cc
@@ -29,7 +29,9 @@
 #include "yb/rpc/serialization.h"
 
 #include "yb/util/debug/trace_event.h"
+#include "yb/util/dist_trace.h"
 #include "yb/util/format.h"
+#include "yb/util/logging.h"
 #include "yb/util/result.h"
 #include "yb/util/status_format.h"
 
@@ -305,6 +307,18 @@ Status YBInboundCall::ParseFrom(const MemTrackerPtr& mem_tracker, CallData* call
   DVLOG(4) << "Parsed YBInboundCall header: " << header_.call_id;
   UpdateWaitStateInfo();
 
+  // Extract the propagated distributed-trace parent from the header, if present. The span itself is
+  // created later (CreateServerSpan) once the request params have been parsed. Tracing is
+  // best-effort: a malformed context is logged and dropped, never fails the RPC.
+  if (dist_trace::IsDistTraceEnabled() && !header_.trace_context.empty()) {
+    auto parsed = ParseTraceContext(header_.trace_context);
+    if (parsed.ok()) {
+      parent_span_context_ = std::move(*parsed);
+    } else {
+      YB_LOG_EVERY_N_SECS(WARNING, 5) << "Failed to parse RPC trace context: " << parsed.status();
+    }
+  }
+
   consumption_ = ScopedTrackedConsumption(mem_tracker, call_data->size());
   request_data_ = std::move(*call_data);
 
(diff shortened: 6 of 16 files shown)


src/yb/yql/pgwrapper/dist_trace-test.cc
```diff
@@ -1774,6 +1817,25 @@ TEST_F(DistTraceRpcTest, TestRpcSpans) {
       "RPC span to appear in trace"));
 }
 
+// Cross-boundary: the Perform RPC's trace must reach the tserver. The inbound server span on
+// TabletServer should be a child of the PG backend's (ysql) outbound client span.
+TEST_F(DistTraceRpcTest, TestRpcSpanReachesTabletServer) {
```
IshanChhangani: This is an end to end test for the happy path, is it possible to add a test for the error path too? Ex. Error message in the span when RPC fails. Also should this have master spans too? if yes we can test them also.

IshanChhangani (changes requested):

src/yb/util/dist_trace.cc
```diff
@@ -38,12 +38,17 @@ DEFINE_NON_RUNTIME_string(otel_ssl_ca_cert_string, "",
     "CA certificate bundle contents for OTLP HTTPS trace export. Used only when no CA certificate "
     "bundle path is configured.");
 
-DEFINE_NON_RUNTIME_uint32(otel_batch_max_queue_size, 2048,
+DEFINE_NON_RUNTIME_uint32(otel_batch_max_queue_size, 16384,
```
IshanChhangani, line 44: Can you share the rationale behind these values? The earlier values were otel standards.

secondly,
Also if we are changing flag values due to the amount of spans that tserver can generate, should we have separate values for master? Or do we expect both of them to be generating equal amount of spans for a generic query?

src/yb/rpc/local_call.h
```diff
@@ -50,6 +50,16 @@ class LocalOutboundCall : public OutboundCall {
...
   }
 
+  // Span context of this outbound call's trace span, to parent the matching local inbound span
+  // (local calls carry no wire header). nullopt when tracing is off. Safe after the scope is
+  // dropped.
+  std::optional<opentelemetry::trace::SpanContext> otel_span_context() const {
+    if (const auto& span = otel_span(); span) {
+      return span->GetContext();
+    }
+    return std::nullopt;
```
IshanChhangani, line 59: why not return GetInvalid()?

src/yb/tserver/db_server_base.cc
```diff
@@ -125,6 +131,10 @@ void DbServerBase::Shutdown() {
   if (txn_manager) {
     txn_manager->Shutdown();
   }
+
+  if (dist_trace::IsDistTraceEnabled()) {
```
IshanChhangani: Master::Shutdown() does not seem to call DbServerBase::Shutdown(). How are the master spans ending?

src/yb/tserver/db_server_base.cc
```diff
@@ -125,6 +131,10 @@ void DbServerBase::Shutdown() {
...
+    dist_trace::ShutdownDistTrace();
```
IshanChhangani: `RpcAndWebServerBase::Shutdown()` has not happened yet. So our trace provider is set on none. Any inbound RPC at this point may try to GetTracer()?

src/yb/rpc/rpc_context.cc
```diff
@@ -147,11 +147,16 @@ RpcContext::RpcContext(std::shared_ptr<YBInboundCall> call,
     RespondRpcFailure(ErrorStatusPB::ERROR_INVALID_REQUEST, s);
```
IshanChhangani, line 148: Is my understanding correct that a client sending a request body the server cannot deserialize gets an ERROR_INVALID_REQUEST response whose caller-side client span shows an error with no corresponding server span in the trace. If yes can this be misleading to a network error.

src/yb/rpc/rpc_context.cc
```diff
@@ -147,11 +147,16 @@ RpcContext::RpcContext(std::shared_ptr<YBInboundCall> call,
...
   }
+  call_->CreateServerSpan();
   TRACE_EVENT_ASYNC_BEGIN1("rpc_call", "RPC", this, "call", call_->ToString());
 }
 
 RpcContext::RpcContext(std::shared_ptr<LocalYBInboundCall> call)
     : call_(call), params_(call.get(), boost::null_deleter()) {
+  // Inbound (server-side) span for this local call.
+  if (auto outbound_call = call->outbound_call(); outbound_call) {
+    call_->CreateServerSpan(outbound_call->otel_span_context());
```
IshanChhangani, line 157: `YBInboundCall` already has `parent_span_context_` for the remote path, why does the local path need a second channel for the same thing?

src/yb/rpc/serialization.cc
```diff
@@ -374,6 +377,93 @@ Result<ParsedRemoteMethod> ParseRemoteMethod(const Slice& buf) {
...
+  }
+  if (!has_trace_id_hi || !has_trace_id_lo || !has_span_id || !has_version_and_flags) {
+    return STATUS(Corruption, "Incomplete trace context: missing field");
+  }
+  return BuildSpanContext(trace_id_hi, trace_id_lo, span_id, version_and_flags);
+}
+
+Result<opentelemetry::trace::SpanContext> ToSpanContext(const TraceContextPB& trace_context) {
+  if (!trace_context.has_trace_id_hi() || !trace_context.has_trace_id_lo() ||
+      !trace_context.has_span_id() || !trace_context.has_version_and_flags()) {
```
IshanChhangani, line 459: Wouldn't unsampled traces have version_and_flags = 0?

src/yb/rpc/serialization.cc
```diff
@@ -374,6 +377,93 @@ Result<ParsedRemoteMethod> ParseRemoteMethod(const Slice& buf) {
...
+
+  if (!span_context.IsValid()) {
+    return STATUS(Corruption, "Invalid trace context: trace_id_hi/lo or span_id is zero");
+  }
+  return span_context;
+}
+
+}  // namespace
+
+Result<opentelemetry::trace::SpanContext> ParseTraceContext(const Slice& buf) {
```
IshanChhangani: why are we separately serializing for rpc vs shmem?
```
Result<opentelemetry::trace::SpanContext> ParseTraceContext(const Slice& buf) {
  TraceContextPB pb;
  if (!pb.ParseFromArray(buf.data(), narrow_cast<int>(buf.size()))) {
    return STATUS(Corruption, "Unable to parse trace context");
  }
  return ToSpanContext(pb);
}
```
doesn't something like this work?

src/yb/rpc/outbound_call.cc
```diff
@@ -436,6 +464,33 @@ Status OutboundCall::SetRequestParam(
     dst = Output::WriteVarint64ToArray(pool_tag, dst);
   }
 
+  if (trace_context_size > 0) {
+    // Write the TraceContextPB submessage. Field numbers match yb.rpc.TraceContextPB in
+    // rpc_header.proto.
+    // The 16-byte trace id splits into two big-endian 64-bit halves; the span id is one big-endian
+    // 64-bit.
+    dst = Output::WriteTagToArray(
```
IshanChhangani, line 470: How about we use generated protobuf facility for it?

Check `MetadataSerializer::MetadataSerializer(rpc::MetadataSerializationMode mode)` for reference

src/yb/rpc/serialization.cc
IshanChhangani: Also maybe instead of `const Slice&` Just `Slice` works too.

src/yb/tserver/db_server_base.cc
```diff
@@ -53,6 +54,11 @@ DbServerBase::~DbServerBase() {
 Status DbServerBase::Init() {
   RETURN_NOT_OK(RpcAndWebServerBase::Init());
 
+  // Initialize distributed tracing once per process here
```
IshanChhangani: Redundant comment.

IshanChhangani (changes requested):

src/yb/yql/pgwrapper/dist_trace-test.cc
ellabaron-code: Added testing for RPC failure and added test for master span.

src/yb/util/dist_trace.cc
ellabaron-code: On the values — the SDK defaults are tuned for a single application process, and a tserver isn't one:

 - `otel_batch_max_queue_size` 2048 → 16384: a loaded tserver produces spans far faster than one PG backend does, and at 2048 the queue overflows and drops spans, which shows up as holes in a trace rather than as slowness.

 - `otel_batch_schedule_delay_ms` 5000 → 500: tserver and master generate a lot of spans, so we want to drain the queue faster to keep space free for the next batch. It also means a span isn't held for 5 s after the operation it describes finished — longer than most of the queries being traced — which makes interactive debugging and any test that waits for a span needlessly slow.

On master: it's a separate process with its own flags, so it can be tuned independently and defaults to the same value as the tserver, which is fine. ysql is the one that can't work that way — the postgres backends are started by the tserver and inherit its flags, so without a separate flag name every backend would get the tserver-sized queue, once per connection.

src/yb/rpc/local_call.h
ellabaron-code: It would work — `GetInvalid()` is the all-zero sentinel and `IsValid()` distinguishes it — so this is a style choice, and I went with `std::optional` for two reasons. "Tracing is off" is the absence of a parent, not a parent with a zero id, and the caller's contract is then enforced by the type instead of by remembering to call `IsValid()`. And returning `std::nullopt` doesn't construct a trace context at all, where `GetInvalid()` builds and returns a zeroed one for the callee to inspect and throw away — on a path that is taken on every local call when tracing is off.

`YBInboundCall::parent_span_context_` is `std::optional `for the same reason, so both parenting channels read alike.

src/yb/rpc/rpc_context.cc
ellabaron-code: Correct on both counts: `ParseParam` failing responds before CreateServerSpan(), so the trace has an erroring client span with nothing under it, and the shape is indistinguishable from a call that never arrived — only the status text tells them apart (a remote parse error vs. a NetworkError).

The parse failure and its `ERROR_INVALID_REQUEST` response are pre-existing; the span placement is what this PR added.

src/yb/rpc/rpc_context.cc
ellabaron-code: `parent_span_context_` is populated in ParseFrom from the wire header, and a local call never goes through that path — there is no header, the request object is handed over in-process — so nothing sets that member for the short-circuited call. The parent has to come from the outbound call it's bypassing, which is what the second channel is.

src/yb/rpc/serialization.cc
ellabaron-code: No — `version_and_flags` is written unconditionally, including when it's zero: the writer emits the tag and varint every time it emits a trace context, and that byte is part of the pre-computed submessage size. Zero parses fine on the way in too — `BuildSpanContext` masks the low byte into `TraceFlags` and only rejects a zero trace id or span id, so an unsampled context arrives as unsampled rather than as an error.

These fields started out as required, precisely so presence couldn't be in question. They were changed to optional on review, so the presence checks are there to cover a case the writer never actually produces — every field is always set, and an absent one means a corrupt or foreign frame.

src/yb/rpc/serialization.cc
ellabaron-code: The shared-memory path uses the generated protobuf serialization, like everything else that goes through the exchange. The RPC path follows what the RPC header already does: `ParseRequestHeader` walks the header field by field off a `CodedInputStream` and never materializes a `RequestHeader` proto, so `trace_context` comes out of that walk as a `Slice`. `ParseTraceContext` is written the same way as every other field there — `ParseFromArray` would reintroduce message construction and parsing on the per-RPC header path.

On `Slice` by value: I'd keep the reference. Passing by value copies the pointer and length at every call site; the reference doesn't, and there's nothing here that benefits from having its own copy.

src/yb/rpc/outbound_call.cc
ellabaron-code: The header encoding is hand-rolled for speed. On the read side `ParsedRequestHeader` holds slices into the receive buffer and the walk never materializes a message; on the write side the size is pre-computed and the fields are laid straight into the already-sized buffer. Going through the generated serializer for the trace context means building a `TraceContextPB` and a `ByteSizeLong()` pass on every traced call, so I kept it in the header's own style.

`MetadataSerializer` is a slightly different situation: the payload there is owned by ash, which fills its own `AshMetadataPB` from the current wait state and is injected into the RPC layer through the serializer factory. The trace context is written by the RPC layer itself, so there's no interface to cross.

src/yb/tserver/db_server_base.cc
ellabaron-code: Good catch, and it was worse than the spans: `Master::Shutdown()` overrides `Shutdown()` and calls `RpcAndWebServerBase::Shutdown()` directly, so master never reached the teardown at all — it left the provider installed and never flushed at a defined point. (The spans themselves were always ending at their own call sites; what master missed was the provider teardown.)

Both init and teardown now live in `RpcAndWebServerBase`, which master and tserver do go through. Init runs before `RpcServerBase::Init()`, so the tracer is ready before the first call; teardown runs at the end of `RpcAndWebServerBase::Shutdown()`. Since CQL and Redis derive from the same base and share the tserver process, the first server to initialize sets tracing up and the last one to shut down tears it down.

src/yb/tserver/db_server_base.cc
ellabaron-code: Right, and that ordering is fixed by the same move: the teardown now runs at the end of `RpcAndWebServerBase::Shutdown()`, after `RpcServerBase::Shutdown()` has taken the messenger down, so no inbound call can still be looking up the tracer.

src/yb/tserver/db_server_base.cc
ellabaron-code: Dropped — that block moved into `RpcAndWebServerBase::Init()` along with the teardown, so master gets both, and the comment there says why the placement matters rather than restating the call.

IshanChhangani:
Same description comments as given on PR1 and PR3 on all PRs.

src/yb/util/dist_trace.h
```diff
@@ -30,15 +31,19 @@
...
 
+namespace context = opentelemetry::context;
 namespace nostd = opentelemetry::nostd;
 namespace trace = opentelemetry::trace;
 
-// Bundles a span with an activated (thread-local) scope so work started after it inherits it as
-// parent; DropScope on the constructing thread before hopping threads. End() is safe from any
-// thread.
+// A span plus a context token attached to the constructing thread, so later work inherits the span.
 struct SpanWithScope {
```
amitanandaiyer: I think SpanWithScope is confusing me.

Span is something that may stay with the rpc/incoming-call and work across threads. 

Scope would be something equivalent of enabling and disabling the span for adoption on a specific thread. Can this not be kept separate. 

Similar to ScopedAdoptXXX (WaitState/Trace) which handle the RAII for setting the corresponding Scope.

The name is very similar to what's already used for WaitState/Trace but seems to be working differently, causing confusion.
Perhaps, if not in the current stack; it will be good to unify/clean up the naming/structure so they are somewhat similar if/when they are doing similar things.

ellabaron-code: The span and its scope logically belong together: the scope exists only to make that one span the
current context, its lifetime is bounded by the span's, and pairing an attach with the wrong span or
failing to detach is the whole class of bug the type prevents. Keeping them in one object makes that
pairing structural rather than something every call site has to get right.

`ScopedAdoptTrace` / `ScopedAdoptWaitState` are pure stack RAII over YB's own thread-locals — a
`Trace*` and an ASH `WaitStateInfoPtr` — and neither owns or terminates what it adopts. The
constraints differ here: the thread-local is OTel's `RuntimeContext`, whose detach pops a per-thread
stack, so it has to happen on the attaching thread, and the span itself may outlive the frame that
created it.

Destruction still does both — the destructor ends the span and the token's destructor detaches — which
is all a frame-scoped span needs, and most call sites rely on exactly that. `DropScope()` and `End()`
are also callable separately for the cases where those two points come apart. On the RPC path they do:
`CreateServerSpan` attaches during call setup, `HandleCall` drops the scope once the handler returns so
the handler's work nests under the span but the async continuation doesn't, and
`RespondSuccess`/`RespondFailure` ends the span later from whichever thread responds. Both operations
are idempotent, so the destructor remains a safe backstop.

src/yb/server/server_base.cc
```diff
@@ -167,6 +169,11 @@ struct CommonMemTrackers {
 
 std::unique_ptr<CommonMemTrackers> common_mem_trackers;
 
+// Several servers (tserver, CQL, Redis) can share a process, and tracing is process-wide: it is set
+// up by the first of them to initialize and torn down by the last one to shut down.
+std::mutex dist_trace_mutex;
```
amitanandaiyer, line 174: would an atomic work here?

src/yb/util/dist_trace.cc
```diff
@@ -58,11 +66,35 @@ DEFINE_NON_RUNTIME_string(otel_internal_log_level, "info",
...
+REGISTER_CALLBACK(
+    otel_collector_traces_endpoint, "UpdateDistTraceEnabled", &UpdateDistTraceEnabled);
+
 namespace yb::dist_trace {
 
+namespace internal {
+
+bool g_dist_trace_enabled = false;
+
+// Service name for the tracing resource and tracer (e.g. "ysql", "Master", "TabletServer").
```
amitanandaiyer, line 93: what would this be in a mini-cluster like set up?

src/yb/util/dist_trace.h
amitanandaiyer: Is the plan to avoid having a single span be collecting events/get activated on different threads. But create more spans (using span-context) as we go across threads? If so, this sounds reasonable. 

Would you have to worry about generating too many spans? Existing util/trace.h has a mechanism to create child traces. But we avoid creating too many child traces; and instead reuse the trace across threadpools.

src/yb/rpc/serialization.cc
IshanChhangani: nit: Slice here is just 2 pointers. `const Slice&` would have minimum one register for the address and then one per load. `slice.h` follows the use of direct `Slice` instead of const reference.

src/yb/util/dist_trace.h
ellabaron-code: Yes — that's the design. A live span is never activated on a second thread. What
crosses the hop is only the `SpanContext` (trace id, span id, flags), captured on
the submitting thread by `GetActiveSpanContext()`. On the worker,
`ActivateParentScope` wraps it in a `trace::DefaultSpan`, which is non-recording:
no span is started, ended, or exported. Its only job is to make that context
current, so anything created on that thread attaches to the right parent.

So the hops themselves add no spans. The span count is driven by RPC boundaries —
one client span per outbound call, one server span on the receiving side —
independent of how many threads or queues the work passes through.

On volume: nothing is created unless the request is already traced. A client span
is only started when there is an active context, and a server span only when the
inbound call carried a parent context on the wire. A request with no traceparent
produces no spans at all, so this isn't an always-on cost across the fleet the way
child traces would be.

src/yb/server/server_base.cc
ellabaron-code: An atomic wouldn't help here — what needs guarding isn't the counter, it's the
multi-step setup of the process-wide OTel provider, and there's no single atomic
object standing in for it. With a plain `fetch_add`, a second server in the same
process gets back 1, skips initialization, and can reach `GetDistTracer()` while
the first one is still inside `InitDistTrace`. The mutex makes the check and the
call one step, so the second server waits until tracing is actually ready, and it
keeps the last shutdown from tearing the provider down while another server is
still coming up — which in-process MiniCluster does, starting and stopping
individual servers while others are running.

src/yb/util/dist_trace.cc
ellabaron-code: In an in-process MiniCluster it's whichever server initializes first — masters
start before tservers, so it ends up "Master" for every server in the process,
and the node UUID in the resource comes from that same first server. Tracing's
OTel provider is process-wide, and `RpcAndWebServerBase::Init` only initializes
it for the first server to come up (guarded by the refcount in
`server_base.cc`), so spans from the in-process tservers are attributed to that
one service name too.

That doesn't affect the tracing tests — `dist_trace-test` runs on
ExternalMiniCluster, where every master and tserver is its own process and gets
its own service name, same as production. The shared name only matters if
someone enables tracing under an in-process MiniCluster, and there neither
`service.name` nor the node UUID tells you which server a span came from.

src/yb/rpc/rpc_context.cc
IshanChhangani: Isn't it better to have a server side span show error instead of not having a server side span and showing error in the client side? I am fine if you wish to address this as part of a different PR.

src/yb/rpc/serialization.cc
ellabaron-code: Ok, made the change.

src/yb/rpc/outbound_call.cc
```diff
@@ -277,17 +277,22 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
...
+    trace_parent_ = dist_trace::GetActiveSpanContext();
+
+    // Not attached: the wire header and the local inbound call both take this span from
+    // GetContext().
+    otel_span_ = dist_trace::StartClientSpanWithScope(
+        Format("rpc $0", remote_method_.ToString()), /*attach=*/false);
+    if (otel_span_) {
+      otel_span_->SetAttribute("rpc.system", "outbound_rpc");
+      otel_span_->SetAttribute("rpc.service", remote_method_.service_name());
+      otel_span_->SetAttribute("rpc.method", remote_method_.method_name());
```
IshanChhangani: Isn't the opname basically rpc.service + rpc.method? Why do we need separate attributes. (Avoiding redundant attributes is a good idea as we would be having 100s of these rpc spans for even simple txns)

src/yb/rpc/rpc_context.cc
ellabaron-code: Made the change — the server span is now created before the params are parsed, so a request the server can't deserialize gets a server span with the parse error on it instead of leaving a hole under the caller's span.

Everything the span needs comes from the request header, which is parsed earlier, so nothing depended on the params. The early-return path was already covered for teardown: the scope is dropped after the handler returns regardless of whether it ran, and `RespondFailure` sets the error status and ends the span.

src/yb/rpc/outbound_call.cc
ellabaron-code: Removed. `rpc.service` and `rpc.method` are gone from all four span-creation sites — outbound RPC, inbound RPC, outbound shared memory, and inbound shared memory — along with the two helpers that existed only to supply the method name. The span name already carries both, minus the transport prefix (`rpc ` / `shmem `).

The test assertions that read them are dropped as well; the wait helpers match on the span name, so they cover the same ground.

src/yb/rpc/outbound_call.cc
```diff
@@ -277,15 +277,20 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
...
-  // Capture this call's parent before StartClientSpanWithScope makes otel_span_ current;
-  // InvokeCallbackSync restores it around the callback so its follow-on work nests as a sibling.
-  trace_parent_ = dist_trace::GetActiveSpanContext();
-
-  otel_span_ = dist_trace::StartClientSpanWithScope(Format("rpc $0", remote_method_.ToString()));
-  if (otel_span_) {
-    otel_span_->SetAttribute("rpc.system", "yb_rpc");
-    otel_span_->SetAttribute("rpc.call_id", call_id_);
-    otel_span_->DropScope();
+  // Guarded so that with tracing off the span name below is never formatted.
```
IshanChhangani: Can be avoided.

ellabaron-code: Fixed.

src/yb/util/dist_trace.h
amitanandaiyer: I find these contradictary:

"A live span is never activated on a second thread."

"The span count is driven by RPC boundaries —"

A single write rpc (on the server) for example will be handled by many different threads asynchronously; generally anytime the execution jumps from one thread to another thread.  -- the callback is responsible for doing teh appropriate ADOPT_TRACE/ADOPT_WAIT_STATE such that the events on the subsequent threads are also captured correctly. 

Are we then only tracing teh "first thread" and not tracing the events that may happen on other threads? If we are tracing all threads, and would be creating a fresh span each time, I'd expect to see O(rpcs * num_thread_hops) spans.

src/yb/util/dist_trace.h
```diff
@@ -30,15 +31,19 @@
...
-  explicit SpanWithScope(nostd::shared_ptr<trace::Span> s)
-      : span(std::move(s)), scope(span) {}
+  explicit SpanWithScope(nostd::shared_ptr<trace::Span> s, bool attach = true)
+      : span(std::move(s)) {
+    if (attach) {
+      token = context::RuntimeContext::Attach(
+          context::RuntimeContext::GetCurrent().SetValue(trace::kSpanKey, span));
```
amitanandaiyer: Pls do split this up. Keep the scope as a local/stack variable on the thread where we wish to enable the scope. 

based on analysis from opus:

```
The fusion is already costing you a real bug
End() guards on IsRecording():


void End() {
  if (span && span->IsRecording()) { ... scope.reset(); span->End(); }
}
ActivateParentScope builds a SpanWithScope around a trace::DefaultSpan, and DefaultSpan::IsRecording() returns false unconditionally (default_span.h:29). So for that object the destructor never resets the scope. At [outbound_call.cc:569-574](vscode-webview://0rk2dcvpi3lvj7g1jeruaml6t3ms1v1s60duh22kl1n8lhjjqc9d/src/yb/rpc/outbound_call.cc#L569-L574):


// ... parent_scope drops at block end so it can't leak.
{
  auto parent_scope = dist_trace::ActivateParentScope(trace_parent_);
  callback_();
}
It does leak. The Attach token is never detached, so every RPC callback pushes one more entry onto that reactor thread's ThreadLocalContextStorage stack and never pops it — unbounded growth, and GetCurrent() keeps returning the stale parent, so every subsequent span on that reactor thread nests under a dead call's parent. The comment asserts the opposite of what happens. This is exactly the class of bug the fusion invites: the guard makes sense for the span, not for the scope, and one method owns both.

The split
Return the span alone, and make activation a scope-only stack type:


// Long-lived, thread-agnostic. Members hold this.
nostd::shared_ptr<trace::Span> StartClientSpan(std::string_view op_name);

// Stack-only. Deleted copy/move and deleted operator new make the misuse unrepresentable.
class [[nodiscard]] ActiveScope {
 public:
  explicit ActiveScope(const nostd::shared_ptr<trace::Span>& span);
  explicit ActiveScope(const trace::SpanContext& parent);  // replaces ActivateParentScope
  ~ActiveScope() = default;  // trace::Scope does the right thing, unconditionally
  ActiveScope(const ActiveScope&) = delete;
  ActiveScope(ActiveScope&&) = delete;
  static void* operator new(size_t) = delete;
 private:
  std::optional<trace::Scope> scope_;
};
Then otel_span_ becomes a plain shared_ptr<Span>, DropScope() and owner_thread disappear, the DCHECK disappears, and InvokeCallbackSync becomes dist_trace::ActiveScope parent_scope(trace_parent_); which actually detaches at block end.

Worth noting: no existing call site keeps the scope alive past its creating frame. Both StartClientSpanWithScope callers call DropScope() on the very next lines ([outbound_call.cc:290](vscode-webview://0rk2dcvpi3lvj7g1jeruaml6t3ms1v1s60duh22kl1n8lhjjqc9d/src/yb/rpc/outbound_call.cc#L290), [pg_client.cc:442](vscode-webview://0rk2dcvpi3lvj7g1jeruaml6t3ms1v1s60duh22kl1n8lhjjqc9d/src/yb/yql/pggate/pg_client.cc#L442)), and only SetAttribute runs in between — which never consults the context. So those two sites don't need a scope at all; they can just call StartClientSpan. The one place activation is genuinely wanted is InvokeCallbackSync, and that's a pure stack frame.

The one case stack RAII can't cover is keeping a context active across an async hop — but trace::Scope can't do that either (the token is bound to the attaching thread's stack). The correct tool there is what you're already doing: carry the SpanContext and re-attach on the far thread, which is itself a stack object in the continuation's frame.
```

src/yb/util/dist_trace.h
ellabaron-code: Thread hops aren't part of this PR — #33114 covers them, and the span-per-hop question is
answered by that PR rather than this one.

Worth flagging on process: this stack was split into one PR per layer at review request, so each
piece is deliberately incomplete on its own — the wire format lands without the thread-hop
propagation, which lands without the PG backend, and so on. Reviewing a single PR as though it
were the whole feature will keep surfacing gaps that a later PR fills. #33115 is the top of the
stack, so the full picture is there if it's easier to judge end-to-end behavior in one place.

src/yb/util/dist_trace.h
ellabaron-code: No leak here — the detach doesn't live in `End()`. It's in `~Token()`, which calls
`RuntimeContext::Detach` unconditionally (`runtime_context.h:186-189`), and `token` is a member,
so it runs whenever the object is destroyed. `IsRecording()` only decides whether the span gets
ended, which is right — a `DefaultSpan` has nothing to end. The snippet above has `End()` doing
`scope.reset()`; there's no `scope` member in this code, and that's where the conclusion comes
from. I also had my AI check the leak claim against the tree, and it did not confirm it.

On the split itself: I'm not going to do it. There's no bug to fix, so it comes down to which
API looks better, and I don't think two objects is better than one — keeping the span and its
scope together is what makes it impossible to attach one and forget to detach the other. Every
call site would now have to get both right.

src/yb/rpc/local_call.h
```diff
@@ -50,6 +50,16 @@ class LocalOutboundCall : public OutboundCall {
     return req_;
   }
 
+  // Span context of this outbound call's trace span, to parent the matching local inbound span
```
IshanChhangani: Can be avoided.

ellabaron-code: Same question here — what specifically can be avoided, the doc comment or the accessor? Can you clarify?

src/yb/rpc/rpc_header.proto
```diff
@@ -54,6 +54,21 @@ message RemoteMethodPB {
   required string method_name = 2;
 };
 
+// Distributed-trace context (W3C trace-context; https://www.w3.org/TR/trace-context/) propagated
+// across a process boundary. Defined here so a single message and wire format serves both
+// transports: it is carried in the RPC RequestHeader below (alongside AshMetadataPB), and as an
```
IshanChhangani: `(alongside AshMetadataPB)` 
Not relevant.

src/yb/rpc/yb_rpc.cc
```diff
@@ -316,6 +330,28 @@ Status YBInboundCall::ParseFrom(const MemTrackerPtr& mem_tracker, CallData* call
   return Status::OK();
 }
 
+void YBInboundCall::CreateServerSpan(std::optional<opentelemetry::trace::SpanContext> parent) {
+  // Remote calls supply no `parent` and fall back to the context parsed from the wire header; local
```
IshanChhangani: parent -> traceparent

ellabaron-code: Fixed.

src/yb/rpc/rpc_header.proto
ellabaron-code: Fixed

src/yb/util/dist_trace.cc
```diff
@@ -343,19 +384,35 @@ SpanWithScopePtr StartClientSpanWithScope(std::string_view op_name) {
...
-  if (!IsDistTraceEnabled() || !parent_context.IsValid()) {
+SpanWithScopePtr StartServerSpanWithScope(
+    std::string_view op_name, const trace::SpanContext& parent_context) {
+  if (!IsDistTraceEnabled()) {
+    return nullptr;
+  }
+  trace::StartSpanOptions options;
+  options.kind = trace::SpanKind::kServer;
+  options.parent = parent_context;
+  return std::make_unique<SpanWithScope>(GetDistTracer()->StartSpan(
```
IshanChhangani: Lets use the StartSpan wrapper instead, as done in `StartClientSpanWithScope`.  This may require updating `DCHECK(HasActiveContext()) ` in it to something like:
`  DCHECK(HasActiveContext() || std::holds_alternative<trace::SpanContext>(options.parent));` or drop the DCHECK altogether.

src/yb/util/dist_trace.h
ellabaron-code: Fixed as discussed.

src/yb/util/dist_trace.h
ellabaron-code: Fixed.

src/yb/rpc/local_call.h
IshanChhangani: Comment.

ellabaron-code: Fixed.

src/yb/util/dist_trace.cc
```diff
@@ -229,8 +264,9 @@ class TraceparentCarrier : public context::propagation::TextMapCarrier {
 
 }  // namespace
 
-bool IsDistTraceEnabled() {
-  return !FLAGS_otel_collector_traces_endpoint.empty();
+void TEST_SetOtelCollectorEndpoint(const std::string& endpoint) {
```
IshanChhangani, line 267: `google::FlagSaver` restores `otel_collector_traces_endpoint` at scope exit but should not re run the registered callback. so g_dist_trace_enabled stays true and every later test in the process runs with tracing on and an empty endpoint?

IshanChhangani (approved):

src/yb/util/dist_trace.cc
ellabaron-code: `StartSpan` is for starting a child of whatever span is currently active — that's what the `DCHECK` checks, and the pggate callers depend on it. The server span is different: it has no active span, its parent comes explicitly from the wire. If I push it through `StartSpan` I either have to weaken the `DCHECK` into an "either-or" check or drop it entirely, and then the callers that do need it lose the protection. So I'd rather leave `StartServerSpan` calling the tracer directly.

src/yb/util/dist_trace.cc
ellabaron-code: Fixed.

PR #33113, opened 2026-08-04
[#16670] DocDB: dist-trace: Propagate trace context over the PG-tserver shared memory exchange

ellabaron-code:
## Summary

Extend the same propagation to the shared-memory fast path between a PG backend and its
tserver, so requests that bypass the RPC stack still join the caller's trace. The backend
serializes its active span's context into the exchange alongside the ASH metadata it
already carries, and the tserver opens a server span under it. The encoding reuses the
wire type and decoder introduced for the RPC header rather than adding a second
representation of the same context.

## Test plan

      commit, including the new `TestSharedMemorySpansReachTabletServer` and
      `TestSharedMemoryFallbackToRpc`.

`TestSharedMemorySpansReachTabletServer` covers both request types the exchange carries,
`Perform` and `AcquireObjectLock`: each inbound span on the tserver must join the query's
trace as a child of the ysql backend's outbound span, with `rpc.system=yb_shmem`.
`TestSharedMemoryFallbackToRpc` sends a request too large for the exchange, asserting the
shared-memory span is left childless and the RPC that replaces it has its own inbound span
on the far side.

### Stack

### DIFF
--- a/src/yb/ash/rpc_wait_state.cc
+++ b/src/yb/ash/rpc_wait_state.cc
@@ -12,6 +12,8 @@
 //
 #include "yb/ash/rpc_wait_state.h"
 
+#include "yb/gutil/endian.h"
+
 #include "yb/rpc/inbound_call.h"
 #include "yb/rpc/lightweight_message.h"
 
@@ -129,6 +131,36 @@ uint8_t* MetadataSerializer::SerializeToArray(uint8_t* out) {
   }
 
   return metadata_.SerializeWithCachedSizesToArray(out);
+}
+
+void TraceContextSerializer::SetTraceContext(
+    const opentelemetry::trace::SpanContext& span_context) {
+  if (!span_context.IsValid()) {
+    return;
+  }
+  const auto trace_id = span_context.trace_id();
+  const auto span_id = span_context.span_id();
+  // Split the 16-byte trace id into two big-endian 64-bit halves; the span id is one big-endian
+  // 64-bit value -- the layout rpc::ToSpanContext reads back on the tserver.
+  trace_context_.set_trace_id_hi(BigEndian::Load64(trace_id.Id().data()));
+  trace_context_.set_trace_id_lo(BigEndian::Load64(trace_id.Id().data() + 8));
+  trace_context_.set_span_id(BigEndian::Load64(span_id.Id().data()));
+  // Low byte: flags; high byte: the (currently unused) version.
+  constexpr uint32_t kVersion = 0;
+  trace_context_.set_version_and_flags((kVersion << 8) | span_context.trace_flags().flags());
+  serialized_size_ = trace_context_.ByteSizeLong();
+}
+
+size_t TraceContextSerializer::SerializedSize() const {
+  return Output::VarintSize32(static_cast<uint32_t>(serialized_size_)) + serialized_size_;
+}
+
+uint8_t* TraceContextSerializer::SerializeToArray(uint8_t* out) const {
+  out = Output::WriteVarint32ToArray(static_cast<uint32_t>(serialized_size_), out);
+  if (serialized_size_ == 0) {
+    return out;
+  }
+  return trace_context_.SerializeWithCachedSizesToArray(out);
 }
 
 } // namespace yb::ash
--- a/src/yb/tserver/pg_client_session.cc
+++ b/src/yb/tserver/pg_client_session.cc
@@ -58,6 +58,8 @@
 
 #include "yb/rpc/lightweight_message.h"
 #include "yb/rpc/rpc_context.h"
+#include "yb/rpc/rpc_header.pb.h"
+#include "yb/rpc/serialization.h"
 #include "yb/rpc/sidecars.h"
 #include "yb/rpc/scheduler.h"
 
@@ -81,6 +83,7 @@
 #include "yb/util/cast.h"
 #include "yb/util/cgroups.h"
 #include "yb/util/debug-util.h"
+#include "yb/util/dist_trace.h"
 #include "yb/util/enums.h"
 #include "yb/util/logging.h"
 #include "yb/util/lw_function.h"
@@ -841,6 +844,30 @@ using PerformQueryTraits = QueryTraits<PgPerformRequestMsg, PgPerformResponseMsg
 using ObjectLockQueryTraits =
     QueryTraits<const PgAcquireObjectLockRequestMsg, PgAcquireObjectLockResponseMsg>;
 
+// Name of the inbound shared-memory span.
+template <QueryTraitsType T>
+std::string_view SharedMemHandlerSpanName();
+template <>
+std::string_view SharedMemHandlerSpanName<PerformQueryTraits>() {
+  return "shmem yb.tserver.PgClientService.Perform";
+}
+template <>
+std::string_view SharedMemHandlerSpanName<ObjectLockQueryTraits>() {
+  return "shmem yb.tserver.PgClientService.AcquireObjectLock";
+}
+
+// Method name attribute for the inbound shared-memory span.
+template <QueryTraitsType T>
+const char* SharedMemMethodName();
+template <>
+const char* SharedMemMethodName<PerformQueryTraits>() {
+  return "Perform";
+}
+template <>
+const char* SharedMemMethodName<ObjectLockQueryTraits>() {
+  return "AcquireObjectLock";
+}
+
 using ResponseSender = std::function<void()>;
 
 template <QueryTraitsType T>
@@ -1118,6 +1145,8 @@ class SharedExchangeQuery : public std::enable_shared_from_this<SharedExchangeQu
   // 8 bytes - timeout in milliseconds.
   // next - size of serialized AshMetadataPB protobuf (say 'x').
   // next 'x' bytes - serialized AshMetadataPB protobuf.
+  // next - size of serialized TraceContextPB protobuf (say 'y').
+  // next 'y' bytes - serialized TraceContextPB protobuf.
   // remaining bytes - serialized PgPerformRequestPB protobuf.
   template <class... Args>
   Result<RequestInfo> ParseRequest(
@@ -1129,6 +1158,8 @@ class SharedExchangeQuery : public std::enable_shared_from_this<SharedExchangeQu
     input += sizeof(uint64_t);
     RETURN_NOT_OK(rpc::ParseMetadataFromSharedMemory(
         &input, end - input, rpc::AnyMessagePtr(&ash_metadata_)));
+    RETURN_NOT_OK(rpc::ParseMetadataFromSharedMemory(
+        &input, end - input, rpc::AnyMessagePtr(&trace_context_)));
     RETURN_NOT_OK(req_.ParseFromSlice(Slice(input, end)));
     data_.emplace(
         std::forward<Args>(args)..., session_id, arena_, req_, resp_, sidecars_,
@@ -1149,6 +1180,7 @@ class SharedExchangeQuery : public std::enable_shared_from_this<SharedExchangeQu
   }
 
   const AshMetadataPB& ash_metadata() const { return ash_metadata_; }
+  const rpc::TraceContextPB& trace_context() const { return trace_context_; }
 
  private:
   void SendResponse() {
@@ -1215,6 +1247,7 @@ class SharedExchangeQuery : public std::enable_shared_from_this<SharedExchangeQu
   std::remove_const_t<typename T::ReqPB> req_;
   typename T::RespPB resp_;
   AshMetadataPB ash_metadata_;
+  rpc::TraceContextPB trace_context_;
   rpc::Sidecars sidecars_;
   std::weak_ptr<PgClientSession> session_;
   SharedExchange& exchange_;
@@ -2817,7 +2850,34 @@ class PgClientSession::Impl {
     auto track_guard = wait_state
         ? TrackSharedMemoryPgMethodExecution<T>(wait_state, query.ash_metadata())
         : std::nullopt;
-    return DoHandleSharedExchangeQuery(precondition_waiter, std::move(data), deadline);
+
+    // Fix span attibutes for shared memory exchange.
+    dist_trace::SpanWithScopePtr trace_scope;
+    const auto& trace_context = query.trace_context();
+    if (dist_trace::IsDistTraceEnabled() && trace_context.has_span_id()) {
+      auto parent_context = rpc::ToSpanContext(trace_context);
+      if (parent_context.ok()) {
+        trace_scope = dist_trace::StartServerSpanWithScope(
+            SharedMemHandlerSpanName<T>(), *parent_context);
+        if (trace_scope) {
+          // Mirror the attributes the RPC inbound span carries (yb_rpc.cc), minus the ones with no
+          // shared-memory analog (rpc.call_id).
+          trace_scope->SetAttribute("rpc.system", "inbound_shmem");
+          trace_scope->SetAttribute("rpc.service", "yb.tserver.PgClientService");
+          trace_scope->SetAttribute("rpc.method", SharedMemMethodName<T>());
+        }
+      }
+    }
+
+    const auto status = DoHandleSharedExchangeQuery(precondition_waiter, std::move(data), deadline);
+    if (trace_scope) {
+      if (status.ok()) {
+        trace_scope->SetStatus(opentelemetry::trace::StatusCode::kOk);
+      } else {
+        trace_scope->SetStatus(opentelemetry::trace::StatusCode::kError, status.ToUserMessage());
+      }
+    }
+    return status;
   }
 
   template <QueryTraitsType T, class... Args>
--- a/src/yb/tserver/pg_client_session.cc
+++ b/src/yb/tserver/pg_client_session.cc
@@ -2851,7 +2851,7 @@ class PgClientSession::Impl {
         ? TrackSharedMemoryPgMethodExecution<T>(wait_state, query.ash_metadata())
         : std::nullopt;
 
-    // Fix span attibutes for shared memory exchange.
+    // Start the inbound span for this shared memory request, as a child of the propagated context.
     dist_trace::SpanWithScopePtr trace_scope;
     const auto& trace_context = query.trace_context();
     if (dist_trace::IsDistTraceEnabled() && trace_context.has_span_id()) {
--- a/src/yb/yql/pggate/pg_client.cc
+++ b/src/yb/yql/pggate/pg_client.cc
@@ -1130,9 +1130,16 @@ class PgClient::Impl : public BigDataFetcher {
   ResultFuture<Data> PrepareAndSend(Method method, Args&&... args) {
     auto data = std::make_shared<Data>(std::forward<Args>(args)...);
     if (session_shared_mem_ && session_shared_mem_->exchange().ReadyToSend()) {
+      // Start the outbound shared-memory span before sizing the request.
+      data->StartSharedMemorySpan();
       ash::MetadataSerializer metadata(rpc::MetadataSerializationMode::kWriteOnZero);
+      ash::TraceContextSerializer trace_context;
+      if (data->otel_span) {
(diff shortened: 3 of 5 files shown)


src/yb/yql/pgwrapper/dist_trace-test.cc
```diff
@@ -1836,6 +1836,25 @@ TEST_F(DistTraceRpcTest, TestRpcSpanReachesTabletServer) {
   ASSERT_EQ(server_span.str_attrs["rpc.method"], "Perform");
 }
 
+// Cross-boundary over shared memory: the Perform shmem exchange's trace must reach the tserver.
+// The inbound server span on TabletServer should be a child of the ysql outbound client span.
+TEST_F(DistTraceTest, TestSharedMemoryPerformReachesTabletServer) {
```
IshanChhangani: How about we add some more tests? 
- AcquireObjectLock over shmem
- Force the shmem→TCP fallback

Let me know if any of these are nontrivial.

IshanChhangani (changes requested):

src/yb/tserver/pg_client_session.cc
```diff
@@ -2817,7 +2850,34 @@ class PgClientSession::Impl {
     auto track_guard = wait_state
         ? TrackSharedMemoryPgMethodExecution<T>(wait_state, query.ash_metadata())
         : std::nullopt;
-    return DoHandleSharedExchangeQuery(precondition_waiter, std::move(data), deadline);
+
+    // Fix span attibutes for shared memory exchange.
```
IshanChhangani: attributes

IshanChhangani: Also the block below does more than just add span attributes.

IshanChhangani:
> including the new TestSharedMemoryPerformSpans and
TestSharedMemoryPerformSpanReportsAsyncError.

Neither is there in this PR.


> The
span is instead handed to SharedExchangeQuery, which sets its status from the response
in SendResponse (the single choke point for success, cache-hit and error replies); the
handler only drops the thread-local scope, and the span ends when the last reference goes
away.

Confused here, does SharedExchangeQuery store a span?

IshanChhangani (changes requested):

src/yb/ash/rpc_wait_state.cc
```diff
@@ -131,4 +133,34 @@ uint8_t* MetadataSerializer::SerializeToArray(uint8_t* out) {
   return metadata_.SerializeWithCachedSizesToArray(out);
 }
 
+void TraceContextSerializer::SetTraceContext(
```
IshanChhangani: Am I right to say that this is the 3rd place we hand write hi/lo split. Can you explain me why are all 3 required?
Also their names are generic and does not represent the differences.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -1130,26 +1130,34 @@ class PgClient::Impl : public BigDataFetcher {
   ResultFuture<Data> PrepareAndSend(Method method, Args&&... args) {
     auto data = std::make_shared<Data>(std::forward<Args>(args)...);
     if (session_shared_mem_ && session_shared_mem_->exchange().ReadyToSend()) {
+      // Start the outbound shared-memory span before sizing the request.
+      data->StartSharedMemorySpan();
```
IshanChhangani, line 1092: What happens to this span when Obtain() returns null and we fall back to the RPC path?

src/yb/tserver/pg_client_session.cc
```diff
@@ -2817,7 +2850,34 @@ class PgClientSession::Impl {
...
+    dist_trace::SpanWithScopePtr trace_scope;
+    const auto& trace_context = query.trace_context();
+    if (dist_trace::IsDistTraceEnabled() && trace_context.has_span_id()) {
```
IshanChhangani: Why gate on has_span_id() when ToSpanContext requires all four fields? If those two ever disagree, would we notice?

IshanChhangani:
The description can be updated based on the feedback from the previous PRs. We can keep only the information that's relevant to this PR. Also, instead of a file-by-file summary, it would be more useful to explain the rationale behind the design decisions.

src/yb/ash/rpc_wait_state.cc
ellabaron-code: Three places, but only the write side is duplicated.

`BuildSpanContext` (`serialization.cc`) is the single reader — both `ParseTraceContext` and `ToSpanContext` funnel into it, so wire → `SpanContext` exists once.

The write side happens twice because the two transports serialize differently. The RPC header is laid out by hand into a pre-sized buffer, so there is no `TraceContextPB` object to populate — the fields are written straight out at `outbound_call.cc:479-487`. Shared memory uses the generated serialization like the rest of the exchange, so `TraceContextSerializer` fills in a real `TraceContextPB`. The 16-byte trace id → two 64-bit halves conversion is what sits between an OTel `SpanContext` and the four proto fields on both paths, so each writer does it.

On the names: the signatures distinguish them — `ParseTraceContext` takes a `Slice` of header bytes, `ToSpanContext` takes a `TraceContextPB`. The parameter type says which input it reads at every call site.

src/yb/yql/pggate/pg_client.cc
ellabaron-code: It gets ended right there with `kOk` and the request continues on the RPC path, which creates its own outbound span.

The span records the shared-memory attempt, which did happen — `ReadyToSend()` passed and `Obtain()` came back empty because the exchange was busy or the request didn't fit. So the trace shows a short s`hmem yb.tserver.PgClientService.Perform` span followed by the `rpc` one, which is what actually occurred. It ends `kOk` rather than an error because nothing failed.

src/yb/tserver/pg_client_session.cc
ellabaron-code: `has_span_id()` is a cheap presence check that keeps us from calling ToSpanContext at all on the untraced path, which is every request when tracing is off.

The two can't disagree: `TraceContextSerializer::SetTraceContext` sets all four fields or returns without setting any, and it is the only writer of this blob. A partial context would mean the exchange contents were corrupted, at which point the request protobuf that follows it wouldn't parse either.

src/yb/yql/pgwrapper/dist_trace-test.cc
ellabaron-code: Both are covered now.

`TestSharedMemorySpansReachTabletServer` (renamed from `TestSharedMemoryPerformReachesTabletServer`) now asserts both exchange request types: a traced `SELECT` already drives `AcquireObjectLock` calls over the exchange, so it checks for an inbound server span for each of `Perform` and `AcquireObjectLock`, both remote children of the `ysql` client span. The fixture turns object locking on, since it defaults off in debug builds.

The fallback needed no test flag — the exchange is a single page, so a 1 MB value in an `INSERT` cannot be obtained and the request takes the RPC path. `TestSharedMemoryFallbackToRpc` asserts the trace survives that: the abandoned shared-memory client span with no children under it, plus the RPC span whose inbound counterpart on the `tserver` is its remote child.

src/yb/tserver/pg_client_session.cc
ellabaron-code: Fixed.

ellabaron-code:
Rewrote the description. Dropped the two test names that don't exist, dropped the paragraph about the span being stored in `SharedExchangeQuery` — that isn't implemented here — and removed the `Upgrade/Rollback` section: it came from the parent PR's `rpc_header.proto` change showing up in this branch's diff against master, not from anything this PR does. The file-by-file summary is gone too, replaced with the design rationale for the layout, the shared encoder/decoder, and the span naming.

IshanChhangani:
> Rewrote the description. Dropped the two test names that don't exist, dropped the paragraph about the span being stored in `SharedExchangeQuery` — that isn't implemented here — and removed the `Upgrade/Rollback` section: it came from the parent PR's `rpc_header.proto` change showing up in this branch's diff against master, not from anything this PR does. The file-by-file summary is gone too, replaced with the design rationale for the layout, the shared encoder/decoder, and the span naming.

This did not get pushed I think.

ellabaron-code:
Dropped the Upgrade/Rollback section: the `rpc_header.proto` change is the parent PR's
and only shows up here because the diff is against master. Fixed the test plan to name
the two tests this PR really adds, `TestSharedMemorySpansReachTabletServer` and
`TestSharedMemoryFallbackToRpc`. Removed the paragraph about the span being handed to
`SharedExchangeQuery` — that isn't implemented here, so the question was fair. And the
file-by-file summary is gone in favour of a short description of what the change does.

IshanChhangani:
PR Description is not updated. It states `rpc.system=inbound_shmem` while this has been changed to `yb_shmem` recently.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -1112,26 +1112,34 @@ class PgClient::Impl : public BigDataFetcher {
...
       ash::MetadataSerializer metadata(rpc::MetadataSerializationMode::kWriteOnZero);
+      ash::TraceContextSerializer trace_context;
```
IshanChhangani: TraceContextSerializer has no dependency on ASH. It Should not be part of ash namespace. This can be dist_trace:: etc
Further `ToSpanContext()` function is relevant only for shmem, it shouldn't be part of rpc namespace.

src/yb/yql/pggate/pg_client.cc
IshanChhangani: How about we start span within `if (out)`? Since 2 spans one with shmem and next with RPC doesn't add any relevant info in this case. Or alternatively Shmem span can show some error?

src/yb/tserver/pg_client_session.cc
```diff
@@ -869,6 +872,18 @@ using PerformQueryTraits = QueryTraits<PgPerformRequestMsg, PgPerformResponseMsg
 using ObjectLockQueryTraits =
     QueryTraits<const PgAcquireObjectLockRequestMsg, PgAcquireObjectLockResponseMsg>;
 
+// Name of the inbound shared-memory span.
+template <QueryTraitsType T>
+std::string_view SharedMemHandlerSpanName();
+template <>
+std::string_view SharedMemHandlerSpanName<PerformQueryTraits>() {
```
IshanChhangani: pg_client.cc already have the exact same names:
```
std::string_view GetSharedMemSpanName(tserver::PgSharedExchangeReqType req_type) {
  switch (req_type) {
    case tserver::PgSharedExchangeReqType::PERFORM:
      return "shmem yb.tserver.PgClientService.Perform";
    case tserver::PgSharedExchangeReqType::ACQUIRE_OBJECT_LOCK:
      return "shmem yb.tserver.PgClientService.AcquireObjectLock";
    case tserver::PgSharedExchangeReqType_INT_MIN_SENTINEL_DO_NOT_USE_: [[fallthrough]];
    case tserver::PgSharedExchangeReqType_INT_MAX_SENTINEL_DO_NOT_USE_: break;
  }
  FATAL_INVALID_ENUM_VALUE(tserver::PgSharedExchangeReqType, req_type);
```
Can we move this function to a relevant header and use both places? Possibly in dist_trace-test.cc too.
Since addition of another in one of the place and not other silently breaks client-server pairing.

src/yb/tserver/pg_client_session.cc
```diff
@@ -2839,7 +2866,33 @@ class PgClientSession::Impl {
...
+          // shared-memory analog (rpc.call_id).
+          trace_span->SetAttribute("rpc.system", "yb_shmem");
+        }
+      }
+    }
+
+    dist_trace::ScopedAdoptSpan trace_scope(trace_span);
+    const auto status = DoHandleSharedExchangeQuery(precondition_waiter, std::move(data), deadline);
+    if (trace_span) {
+      if (status.ok()) {
```
IshanChhangani: This is identical to `PgClientData::EndSharedMemorySpan`
We can move this to dist_trace.cc/.h
```
void EndSpanWithStatus(const nostd::shared_ptr<trace::Span>& span, const Status& status) {
  if (!span) {
    return;
  }
  if (status.ok()) {
    span->SetStatus(trace::StatusCode::kOk);
  } else {
    span->SetStatus(trace::StatusCode::kError, status.ToUserMessage());
  }
  span->End();
}
```

src/yb/yql/pgwrapper/dist_trace-test.cc
```diff
@@ -1890,6 +1911,60 @@ TEST_F(DistTraceRpcTest, TestRpcPerformSpanReportsAsyncError) {
...
+      "SET yb_dist_tracecontext = 'traceparent=''$0'''", tp.full));
+  ASSERT_OK(conn_->FetchFormat("SELECT * FROM $0", kTableName));
+
+  auto perform_span = ASSERT_RESULT(collector_.WaitForRemoteChildSpan(
+      tp.trace_id, kSharedMemoryPerformSpanName,
+      "ysql" /* client_service */, "TabletServer" /* server_service */));
+
+  ASSERT_EQ(perform_span.str_attrs["rpc.system"], "yb_shmem");
+
+  ASSERT_OK(collector_.WaitForRemoteChildSpan(
```
IshanChhangani, line 1933: With `enable_object_lock_fastpath` on, what makes this SELECT's ACCESS SHARE lock go over the exchange instead of the local fastpath? 

Seems like this works because CreateTable's INSERT ... SELECT commit disabled the fastpath. 
Can we force the slow path in the test itself, or at least comment the dependency?

src/yb/tserver/pg_client_session.cc
```diff
@@ -2839,7 +2866,33 @@ class PgClientSession::Impl {
...
+        if (trace_span) {
+          // Mirror the attributes the RPC inbound span carries (yb_rpc.cc), minus the ones with no
+          // shared-memory analog (rpc.call_id).
+          trace_span->SetAttribute("rpc.system", "yb_shmem");
+        }
+      }
+    }
+
+    dist_trace::ScopedAdoptSpan trace_scope(trace_span);
+    const auto status = DoHandleSharedExchangeQuery(precondition_waiter, std::move(data), deadline);
```
IshanChhangani: Should the span live on `SharedExchangeQuery` and end in SendResponse, the way `YBInboundCall` ends its span in RespondSuccess/RespondFailure?
This span seems to only cover dispatch, and a failure later in FlushDone is marked ok. 
Async calls (FlushAsync, AcquireObjectLocksGlobalAsync) are not part of the span.

src/yb/tserver/pg_client_session.cc
```diff
@@ -1157,6 +1174,13 @@ class SharedExchangeQuery : public std::enable_shared_from_this<SharedExchangeQu
     input += sizeof(uint64_t);
     RETURN_NOT_OK(rpc::ParseMetadataFromSharedMemory(
         &input, end - input, rpc::AnyMessagePtr(&ash_metadata_)));
+    // A zero-length blob is a single 0 byte; skip it without building the protobuf or the stream.
+    if (input < end && *input == 0) {
+      ++input;
+    } else {
```
IshanChhangani: Tserver builds TraceContextPB then parses it through 2 coded streams then extracts the 4 fields in ToSpanContext. Can we avoid ToSpanContext by using `ParseTraceContext`?
Ex:
```
// ParseRequest
uint32_t trace_len = VERIFY_RESULT(ReadVarint32(&input, end));  // small helper next to ParseString
if (trace_len > 0) {
  auto ctx = rpc::ParseTraceContext(Slice(input, trace_len));
  if (ctx.ok()) {
    trace_context_ = *ctx;
  } else {
    YB_LOG_EVERY_N_SECS(WARNING, 1) << "Bad trace context in shared memory request: " << ctx.status();
  }
  input += trace_len;
}

// member
std::optional<opentelemetry::trace::SpanContext> trace_context_;

// HandleSharedExchangeQuery
if (dist_trace::IsDistTraceEnabled() && query.trace_context()) {
  trace_span = dist_trace::StartServerSpan(SharedMemHandlerSpanName<T>(), *query.trace_context());
  ...
}
```

This makes the code similar for both RPC and shmem. Why does the above code not work for shmem?

src/yb/tserver/pg_client_session.cc
ellabaron-code: Done — moved it to a shared header (`pg_shared_mem_trace.h`), and `pg_client.cc`, `pg_client_session.cc`, and `dist_trace-test.cc` all use it now.

src/yb/tserver/pg_client_session.cc
ellabaron-code: Done — `ParseRequest` now parses it with `rpc::ParseTraceContext` directly (the serializer/parser pair lives in `rpc/serialization.h`), no intermediate `TraceContextPB` or `ToSpanContext`. With the span moved onto `SharedExchangeQuery`, it's started right where the context is parsed.

src/yb/tserver/pg_client_session.cc
ellabaron-code: Yes, that's better — the span now lives on `SharedExchangeQuery`, started in `ParseRequest` and ended in `SendResponse` with the response status, so async completions (`FlushDone` for both `Perform` and `AcquireObjectLock`) are covered and failures there are marked as errors.

src/yb/tserver/pg_client_session.cc
ellabaron-code: Done — extracted a shared `EndSharedMemSpan` helper used by both sides. I put it in `pg_shared_mem_trace.h` next to the span-name helper rather than `dist_trace`, since only the shmem exchange uses this pattern.

src/yb/yql/pggate/pg_client.cc
ellabaron-code: The span has to exist before `Obtain` since its context is serialized into the request buffer (and its size counts toward the requested size), so it can't start inside `if (out)`. Went with the alternative: on the RPC fallback the shmem span ends with an error (request too large for the exchange).

src/yb/yql/pggate/pg_client.cc
ellabaron-code: Moved it to `rpc::` in `rpc/serialization.h` next to `ParseTraceContext`, so the writer and reader of the wire format sit together (`dist_trace` lives below `rpc` in the layering, so it can't hold it). `ToSpanContext` is gone entirely — the shmem path now uses `rpc::ParseTraceContext`.

src/yb/yql/pgwrapper/dist_trace-test.cc
ellabaron-code: Done — the fixture now forces `--enable_object_lock_fastpath=false` for all variants, so the lock always goes over the exchange.

ellabaron-code:
Updated — the description now says `rpc.system=yb_shmem`.

src/yb/yql/pggate/pg_client.cc
IshanChhangani: > The span has to exist before Obtain since its context is serialized into the request buffer (and its size counts toward the requested size), so it can't start inside if (out)

The size is fixed in our case so we don't need the span for that. So we can do the size check first but create the span only when we know the request fits.
I think with the current scheme we are also missing the span attributes (table_name, read and write counts etc) in the RPC span as they got struck to the shmem span.

src/yb/rpc/serialization.h
```diff
@@ -45,6 +45,7 @@
 #include "opentelemetry/trace/span_context.h"
 
 #include "yb/rpc/rpc_fwd.h"
+#include "yb/rpc/rpc_header.pb.h"
```
IshanChhangani: No longer needed.

ellabaron-code: Fixed.

src/yb/yql/pggate/pg_client.cc
ellabaron-code: Good catch — the slot size is actually constant, so no span is needed for sizing. Reworked: `Obtain` runs first and the span is only started once the request is known to fit, so a fallback to RPC keeps its staged attributes.

IshanChhangani (approved):

PR #33114, opened 2026-08-04
[#16670] DocDB: dist-trace: Carry trace context across thread hops

ellabaron-code:
## Summary

Carry the active trace context across thread boundaries, so work handed to a worker,
reactor, scheduler or Raft apply thread continues the trace that submitted it instead of
appearing as an orphan. Covers the general-purpose hand-off primitives and the
higher-level task types built on them. Purely additive, and inert when no trace is
active.

## Test plan

- [x] `./yb_build.sh release --cxx-test thread_pool-test`
- [x] `./yb_build.sh release --cxx-test scheduler-test`
- [x] `./yb_build.sh release --cxx-test periodic-test`
- [x] `./yb_build.sh release --cxx-test rpc_retrier-test`
- [x] `./yb_build.sh release --cxx-test async_rpc_tasks-test`
- [x] `./yb_build.sh release --cxx-test multi_step_monitored_task-test`
- [x] `./yb_build.sh release --cxx-test dist_trace-test`

Each instrumented hop gets a unit test that activates a trace context, hands work off, and
asserts that context is active where the work runs — plus a case with no active context, so
the assertions can fail:

| binary | hop | cases |
|---|---|---|
| `thread_pool-test` | `YBThreadPool::Enqueue` → worker; `Strand::Enqueue` → `Strand::Task::Done` | 4 |
| `scheduler-test` | `ScheduledTask::Run`, `ScheduledTaskWithId::Run` | 3 |
| `periodic-test` | `PeriodicTimer::Callback` | 2 |
| `rpc_retrier-test` (new binary) | `RpcRetrier::DoRetry` | 2 |
| `async_rpc_tasks-test` (new binary) | `RetryingRpcTask::Run` | 3 |
| `multi_step_monitored_task-test` | `MultiStepMonitoredTask::RunInternal`, both steps | 2 |

These need no cluster and no exporter: `ActivateParentScope` makes a context current with a
non-recording span, so the fixtures only set `otel_collector_traces_endpoint` to switch
tracing on. The contexts they compare against come from `MakeTestSpanContext` in the new
`util/dist_trace_test_util.h`.

`dist_trace-test` covers the hops end to end against a live cluster:

- `TestDdlRpcReachesMaster` — a `CREATE TABLE` under a known traceparent must reach the
  master, with the master's inbound RPC span joining the query's trace as a child of a
  tserver outbound span. That holds only if the context survives the intervening
  pool/scheduler/task hops. The fixture now pushes the OTEL flags to the masters as well,
  since the asserted span is exported by a master.
- `TestApplyTaskCarriesTraceContextToMaster` — covers `OperationDriver::ApplyTask`
  specifically. A DDL's transaction coordinator handles `UpdateTransaction(COMMITTED)` on an
  RPC handler thread, then sends `UpdateTransaction(APPLYING)` to the participants from the
  Raft apply thread; the sys catalog is one of those participants, so that second call
  reaches the master. The new `WaitForLocalHopToRemoteSpan` collector helper matches the
  resulting three spans as one chain — tserver inbound → tserver outbound → master inbound —
  so an unrelated `UpdateTransaction` pair elsewhere in the trace does not satisfy it.

### Stack


src/yb/master/async_rpc_tasks_base.cc
```diff
@@ -116,7 +116,9 @@ RetryingRpcTask::RetryingRpcTask(
     : master_(master),
       callback_pool_(callback_pool),
       async_task_throttler_(async_task_throttler),
-      deadline_(UnresponsiveDeadline()) {}
+      deadline_(UnresponsiveDeadline()),
+      // Snapshot the active trace context on the creating thread.
```
IshanChhangani: The comment doesn't convey anything extra then the next line. Can be avoided.

src/yb/master/async_rpc_tasks_base.cc
```diff
@@ -132,6 +134,9 @@ std::string RetryingRpcTask::LogPrefix() const {
 // Send the subclass RPC request.
 Status RetryingRpcTask::Run() {
   VLOG_WITH_PREFIX(1) << "Start Running";
+  // Re-establish the creating thread's trace context.
+  auto trace_scope = dist_trace::ActivateParentScope(trace_parent_);
```
IshanChhangani: How about we centralize the pattern:
- Create member variable.
- capture in the constructor (must run on the submitting thread)
- activate in the run method (executing thread)

This is currently copied in 8 different classes.

With just the member initialization also doing the capture (default constructor does the capture). This hard codes the dependency that the capture happens on the submitting thread.

src/yb/tserver/tablet_validator.cc
```diff
@@ -422,18 +427,24 @@ bool TabletMetadataValidator::Impl::ScheduleTabletPropertiesValidation(
     return false;
   }
 
+  // Active OTel context, live here because the threadpool carries it from the CreateTablet RPC into
+  // OpenTablet. Stored per index so the later GetBackfillStatus poll nests under the triggering
+  // trace.
+  const auto trace_parent = dist_trace::GetActiveSpanContext();
```
IshanChhangani: Does the trace_parent we capture here is ever valid?
This runs from  `open_tablet_pool_` which is the old pool(threadpool.h) while we have instrumented the newer `YBThreadPool`.
Having the capture coded into the constructor could have saved here too.

src/yb/util/strand.cc
```diff
@@ -153,6 +157,10 @@ void Strand::Task::Done(const Status& status) {
 
       auto running = active_enqueues_.load(std::memory_order_acquire) < Strand::kStopMark;
       const auto& actual_status = running ? status : StrandAbortedStatus();
+      // Re-activate the context captured at Enqueue for this task only, so its RPCs nest under the
```
IshanChhangani: Lets document the capture / activate comment once instead of all places. (Better to have a central class to facilitate this).

src/yb/util/thread_pool-test.cc
```diff
@@ -633,4 +636,141 @@ TEST_F(ThreadPoolTest, TestCgroupTaggedPools) {
...
+namespace dist_trace_hops {
+
+using dist_trace::MakeTestSpanContext;
+
+// Enables distributed tracing, pointed at an unreachable collector.
+class ThreadPoolTraceTest : public ThreadPoolTest {
+ protected:
+  void SetUp() override {
+    ThreadPoolTest::SetUp();
+    dist_trace::TEST_SetOtelCollectorEndpoint("http://127.0.0.1:1/v1/traces");
```
IshanChhangani: Setup is common along all these tests how about we pull this into `dist_trace_test_util.h`?

src/yb/util/thread_pool.h
```diff
@@ -22,6 +22,7 @@
 #include "yb/gutil/port.h"
 
 #include "yb/util/debug/leakcheck_disabler.h"
+#include "yb/util/dist_trace.h"
```
IshanChhangani, line 25: yb::ThreadPool is not instrumented. Is this intentional?

src/yb/rpc/scheduler.h
```diff
@@ -31,17 +32,21 @@ constexpr ScheduledTaskId kUninitializedScheduledTaskId = 0;
 class ScheduledTaskBase {
  public:
   explicit ScheduledTaskBase(ScheduledTaskId id, const SteadyTimePoint& time)
-      : id_(id), time_(time) {}
+      : id_(id), time_(time), trace_parent_(dist_trace::GetActiveSpanContext()) {}
```
IshanChhangani: In PeriodicTimer::Callback, the timer schedules its own next tick while the current tick's trace scope is still active. When the next tick runs, it activates the trace and then the next tick does the same. So if one tick ever runs with a trace attached, doesn't every tick after that keep inheriting it, even when the original request is done.

src/yb/util/strand.cc
```diff
@@ -153,6 +157,10 @@ void Strand::Task::Done(const Status& status) {
...
+      // triggering trace.
+      auto parent_scope =
+        dist_trace::ActivateParentScope(task->trace_parent());
```
IshanChhangani: - Thread A has trace X active and enqueues a task on a strand
- The strand submits itself to the thread pool, and that wrapper task gets stamped with trace X
- The worker activates trace X around the strand's entire drain loop
- Meanwhile thread B (no trace) enqueues its own task, and it drains in the same loop
- B's task has an invalid context, so ActivateParentScope is a no-op and trace X stays active
- So B's task runs under trace X and its RPC spans land in an unrelated query's trace

correct?

IshanChhangani (changes requested):

src/yb/util/strand.cc
IshanChhangani: nit: may fit in single line

src/yb/tserver/tablet_validator.cc
ellabaron-code: It is valid. `yb::ThreadPool` in `threadpool.h` is a wrapper around
`YBThreadPool` — it holds one as `underlying_`, and `SubmitFunc` / `Submit` /
`SubmitClosure` all funnel through `DoSubmit`, which enqueues via
`underlying_->EnqueueFunctor`. That's the instrumented path: `EnqueueFunctor`
stamps the submitting thread's context onto the task and the worker activates it
before running it. So there's no separate old pool to instrument — the old class
has no threads of its own.

For this line specifically: `OpenTablet` is submitted from `CreateNewTablet` /
`CreatePeerAndOpenTablet` on the CreateTablet RPC path, so the pool worker runs it
with that RPC's context active. `OpenTablet` then calls `ScheduleValidation`
inline, which calls `ScheduleTabletPropertiesValidation`, and the capture here
picks up the context that is still active on that same thread. It's stored per
index because the `GetBackfillStatus` poll happens later on the validator's own
background thread, which has no context of its own.

The colocated path arrives on a different thread but the same way — a
CHANGE_METADATA apply calls `Tablet::AddTable` -> `AddTableInMemory` -> the
schedule callback inline, and `OperationDriver` re-activates the submitting
context around apply — so the capture is valid there too.

src/yb/util/strand.cc
ellabaron-code: Consolidated. The task-carrier side is documented once on
`ThreadPoolTask::trace_parent()`, which is where both the stamp and the activation
read it from, and the class side is documented on the new `dist_trace::TraceParent`
that replaces the hand-written capture in the eight classes. All the per-site
comments are gone, which also collapses the activate call here onto one line.

src/yb/util/thread_pool-test.cc
ellabaron-code: Done. `dist_trace_test_util.h` now has a `ScopedTestDistTrace` class that sets the collector endpoint in its constructor and holds a `google::FlagSaver` to restore it.

All six trace fixtures (`ThreadPoolTraceTest`, `RetryingRpcTaskTraceTest`, `CatalogEntityTaskTraceTest`, `PeriodicTimerTraceTest`, `SchedulerTraceTest`, `RpcRetrierTraceTest`) now just hold one as a member, and five of them lose their `SetUp` override entirely. As a member the flag is set at fixture construction rather than inside `SetUp`, which is strictly earlier and fine for all of them.

src/yb/util/thread_pool.h
ellabaron-code: It is instrumented — `yb::ThreadPool` has no threads of its own, it is a thin wrapper that holds a `YBThreadPool` as `underlying_`. `SubmitClosure`, `Submit` and `SubmitFunc` all funnel into `DoSubmit`, and both branches of it (`threadpool_use_current_trace_for_tasks` on or off) call `underlying_->EnqueueFunctor(...)`. That is `TaskRecipient::EnqueueFunctor`, which runs the task through `StampTraceContext` to capture the submitting thread's context, and the worker activates it around `Run` in `thread_pool.cc`. So the capture and activate happen one layer down, for both pool APIs.

src/yb/rpc/scheduler.h
ellabaron-code: Neither happens. `trace_parent_` is `const` and captured once in the `PeriodicTimer` constructor, so no tick ever re-captures — a tick cannot acquire a context and hand a new one to its successor. And `parent_scope` lives only inside the `if (run_task)` block, so it is already destroyed by the time the next tick is scheduled; the `ScheduledTask` created there captures whatever is active on the reactor thread, which is nothing.

So a timer's ticks are parented under the context that existed where the timer was constructed, and nothing else. If that construction happened inside a request, the ticks stay attached to that request's trace for the life of the timer — which is the actual lifetime relationship, and worth being able to see.

src/yb/util/strand.cc
ellabaron-code: Correct — real leak, thanks. `YBThreadPool::Enqueue` stamped the strand's *wrapper* task too, and the worker activates the wrapper's context around `Done()`, which is the drain loop. An untraced task pushes nothing, so X stayed active.

Fixed by not giving the wrapper a context at all: `set_trace_parent` is now virtual and `Strand::Task` ignores it, so the only context active in the loop is the one the current task pushes.

Added `StrandDoesNotLeakContextToUntracedTask`, which blocks the traced task so the untraced one lands in the same drain, then asserts it observes no context. Fails without the fix.

src/yb/master/async_rpc_tasks_base.cc
ellabaron-code: Fixed.

src/yb/master/async_rpc_tasks_base.cc
ellabaron-code: Done — added `dist_trace::TraceParent` in `dist_trace.h`. Default construction captures the active context, `Activate()` returns the scope, and there is a `Capture()` for the one holder that is constructed off the submitting thread. `Activate()` is `[[nodiscard]]` since discarding the scope silently makes it a no-op.

All eight classes now hold one of these, which drops their constructor initializer entries and the `ActivateParentScope(trace_parent_)` calls, and gives the capture/activate contract a single place to be documented instead of eight comment copies.

src/yb/util/strand.cc
ellabaron-code: Fixed.

IshanChhangani:
>Purely additive, and inert when no trace is active.

Can we add a test for this? For example, run a representative set of queries that exercises most of the span-generation paths, then verify that no spans are generated after the workload completes.

src/yb/rpc/scheduler.h
IshanChhangani: The 4000WH TPCC run shows a degradation which seems to be coming from here. Any self-rescheduling task ( Raft consensus rpc, PeriodicTimer) that runs once under a traced context stays traced forever.

ellabaron-code: The behavior you're describing is a design decision, not an accident of this PR stack: a trace is expected to show every RPC a query triggers, including the ones that reach the wire only after the work has been handed to background machinery. Dropping the context at those hops makes those RPCs invisible and leaves gaps in the trace.

Where propagation should stop — and what the cost ceiling is for a self-rescheduling task that inherits a context — is a broader question than this PR stack, since it decides what the feature is able to show. It needs a wider discussion with everyone who has a stake in the trace contents, rather than being settled in this thread.

ellabaron-code:
That invariant doesn't hold, and not because of tracing: a query legitimately triggers RPCs that outlive it. Transaction apply, intent cleanup and the status-update calls to the master are dispatched asynchronously and reach the wire after the statement has already returned to the client. Spans for those are correct trace content — they are exactly the RPCs the query caused — so "no spans after the workload completes" would be asserting that YugabyteDB does no asynchronous follow-on work, which it does.

src/yb/master/multi_step_monitored_task.cc
```diff
@@ -26,7 +26,8 @@ namespace yb::master {
 
 MultiStepMonitoredTask::MultiStepMonitoredTask(
     ThreadPool& async_task_pool, rpc::Messenger& messenger)
-    : messenger_(messenger), async_task_pool_(async_task_pool) {}
```
IshanChhangani, line 29: Unrelated change.

src/yb/consensus/consensus_peers.cc
```diff
@@ -257,6 +259,9 @@ void Peer::DumpToHtml(std::ostream& out) const {
 }
 
 void Peer::SendNextRequest(RequestTriggerMode trigger_mode) {
+  // Not traced yet: a follow-up gives consensus its own root trace.
```
IshanChhangani: // TODO (#github issue no): .....

src/yb/consensus/multi_raft_batcher.cc
```diff
@@ -136,6 +138,9 @@ void MultiRaftHeartbeatBatcher::SendBatchRequest(std::shared_ptr<MultiRaftConsen
   if (!data) {
     return;
   }
+  // Not traced yet: a follow-up gives consensus its own root trace.
```
IshanChhangani: ditto

src/yb/tserver/tablet_validator.cc
```diff
@@ -94,6 +95,10 @@ struct IndexTableInfo {
   TableId  index_table_id;
   TableId  group_id; // Generally contains indexed_table_id.
 
+  // OTel context active when this index was scheduled (the CreateTablet RPC's trace, carried into
```
IshanChhangani: CreateTablet RPCs are sent from CatalogManagerBgTasks, which does not have the tracecontext. Does this trace_parent ever hold a value in that case? If yes then
` dist_trace::ScopedAdoptSpan otel_scope(otel_parent);` within SyncWithMaster is dead code then?

src/yb/util/dist_trace_test_util.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include "yb/util/dist_trace.h"
+#include "yb/util/flags.h"
+
+namespace yb::dist_trace {
+
+// Enables distributed tracing for the lifetime of the object, pointed at an unreachable collector.
+class ScopedTestDistTrace {
+ public:
+  ScopedTestDistTrace() { TEST_SetOtelCollectorEndpoint("http://127.0.0.1:1/v1/traces"); }
```
IshanChhangani: Can you confirm that this will get properly reset for the next test? I am doubtful that google::FlagSaver would run YB's REGISTER_CALLBACK.

src/yb/rpc/rpc.cc
```diff
@@ -228,6 +228,7 @@ void RpcRetrier::DoRetry(RpcCommand* rpc, const Status& status) {
   if (new_status.ok()) {
     controller_.Reset();
     VTRACE_TO(1, rpc->trace(), "Sending Rpc");
```
IshanChhangani, line 232: This covers the `rpc->SendRpc()` case when status is ok. Should we also cover deadline case `rpc->Finished(new_status)`

src/yb/util/thread_pool.h
```diff
@@ -176,18 +188,25 @@ class TaskRecipient {
...
   // Matches the interface of yb::ThreadPool::Submit.
   template <NonReferenceType F>
   Status Submit(const F& f) {
-    return EnqueueWithStatus(MakeFunctorThreadPoolTask<F, TaskType>(f));
+    return EnqueueWithStatus(StampTraceContext(MakeFunctorThreadPoolTask<F, TaskType>(f)));
+  }
+
+ private:
+  template <class Task>
+  Task* StampTraceContext(Task* task) {
```
IshanChhangani: Each call to EnqueueFunctor should capture trace context twice. First with StampTraceContext and then with YBThreadPool::Enqueue. 
Can we remove StampTraceContext and keep only the capture in Enqueue?

src/yb/consensus/consensus_peers.cc
```diff
@@ -257,6 +259,9 @@ void Peer::DumpToHtml(std::ostream& out) const {
...
+  auto detach_token =
+      opentelemetry::context::RuntimeContext::Attach(opentelemetry::context::Context{});
```
IshanChhangani: Can we skip it when IsDistTraceEnabled is false.

src/yb/rpc/scheduler.h
```diff
@@ -35,13 +36,15 @@ class ScheduledTaskBase {
...
   SteadyTimePoint time() const { return time_; }
+  const dist_trace::TraceParent& trace_parent() const { return trace_parent_; }
 
   virtual ~ScheduledTaskBase() {}
   virtual void Run(const Status& status) = 0;
 
  private:
   ScheduledTaskId id_;
   SteadyTimePoint time_;
+  dist_trace::TraceParent trace_parent_;
```
IshanChhangani, line 47: HeartbeatDone can reschedule itself from the RPC callback, where parent context is active. Does the scheduled task now capture that context and keep the heartbeat loop traced for the life of the transaction? Similar to the leak with consensus spans we saw before?

src/yb/master/async_rpc_tasks_base.h
```diff
@@ -245,6 +246,10 @@ class RetryingRpcTask : public server::RunnableMonitoredTask {
 
   virtual int num_max_retries();
   virtual int max_delay_ms();
+
+  // Trace context active when this task was created, re-activated at the top of Run() so the task's
+  // RPCs nest under it.
+  dist_trace::TraceParent trace_parent_;
```
IshanChhangani, line 252: does a traced CREATE INDEX end up tracing every BackfillIndex RPC and every index write on the tserver? This would mean millions of spans in a trace.

src/yb/rpc/scheduler.h
ellabaron-code: Correct — the scheduler carries its caller's context by design, so the heartbeat inheriting the query's trace doesn't make this code incorrect, though for a long-lived transaction that trace does keep growing every heartbeat interval. If we don't want to see transaction heartbeats for now, the right fix is to detach at the heartbeat call site the same way we do for consensus; on the other hand, if we do want them, it would be strange to see them disjointed from the query that started them.

src/yb/master/async_rpc_tasks_base.h
ellabaron-code: Yes, and that's desired behavior — we want to trace pretty much everything the query causes, backfill RPCs and index writes included

src/yb/master/multi_step_monitored_task.cc
ellabaron-code: Fixed.

src/yb/consensus/consensus_peers.cc
ellabaron-code: Done, referenced the stack's tracking issue #16670.

src/yb/consensus/multi_raft_batcher.cc
ellabaron-code: Done, referenced the stack's tracking issue #16670.

src/yb/tserver/tablet_validator.cc
ellabaron-code: You're right that the capture is dead for the `CatalogManagerBgTasks` path; the tablet
validator tracing wasn't pulling its weight here, so I removed it from this PR and
deferred it to follow-up work.

src/yb/util/dist_trace_test_util.h
ellabaron-code: Good catch — it wouldn't; `google::FlagSaver` restores the value without rerunning the
callback, leaving the cached bool stale. Restore now goes through
`TEST_SetOtelCollectorEndpoint`.

src/yb/rpc/rpc.cc
ellabaron-code: Done — one scope at the top of `DoRetry` now covers all three exits, with new tests for
the deadline and abort paths (both failed before the fix).

src/yb/util/thread_pool.h
ellabaron-code: Done — every concrete `Enqueue` already stamps, so the helper stamp was redundant.

src/yb/consensus/consensus_peers.cc
ellabaron-code: Done, via a `DetachTraceContext` helper that no-ops when tracing is disabled.

src/yb/rpc/scheduler.h
IshanChhangani: So if I understand correctly, In the current state:
BEGIN;

// queries for 5 min

COMMIT;
with `transaction_heartbeat_usec` = 500 ms (default) we would see ~2400 heartbeat spans. 
This case seems similar to consensus and may reproduce the regression with saw with it. It makes sense to `DetachTraceContext` for now.

ellabaron-code: Done — the context is now detached at the point where each heartbeat sequence starts, and there's a new test, `TestTxnHeartbeatsNotTraced`, that fails if heartbeat spans ever join a query trace again.

IshanChhangani (approved):

src/yb/client/transaction.cc
```diff
@@ -1919,6 +1919,8 @@ class YBTransaction::Impl final : public internal::TxnBatcherIf {
     } else {
       std::weak_ptr<YBTransaction> weak_transaction = transaction;
       auto send_new_heartbeat = [this, weak_transaction, status, promoting](const Status&) {
+        // TODO(#16670): start an independent span here, linked to the originating span.
```
amitanandaiyer, line 1922: we may need this same trick around

ts_local_lock_manager.cc:906
```
 ptr->ScheduleReleaseForLostMessages(client, lock_manager_weak, txn_id, subtxn_id, release_req);
```

 If not, all ddl operations appear as if they take 10 minutes to complete.

ok to defer this to a later per as well.

ellabaron-code: Agreed — deferring to a follow-up PR; I have a fair amount of work stacked on top of this one, so I'd rather not grow it here. Will apply the same treatment to `ScheduleReleaseForLostMessages` there so it doesn't stretch DDL spans to the 10-minute backstop.

PR #33115, opened 2026-08-04
[#16670] DocDB: dist-trace: Propagate traceparent to the PG backend via the startup packet

ellabaron-code:
## Summary

Carry the trace into PG backends the tserver spawns for internal connections, over the
connection startup packet, so a backend's initialization and the statements it runs root
under the originating trace instead of starting their own. A malformed value leaves the
backend untraced rather than failing the connection.

## Upgrade/Rollback safety

No proto, catalog, on-disk format, or gflag default changes, and nothing is persisted.

The one wire-visible change is a new startup-packet parameter, `yb_dist_traceparent`, which libpq
emits only when the matching conninfo option is set. PostgreSQL treats an unrecognized startup
parameter as a GUC name, so a backend built without this change would reject such a connection with
"unrecognized configuration parameter". That cannot arise here: the option is set only by
`BuildConnectionString` for internal connections, where a tserver connects to the postgres process
it spawned from its own binary, so client and backend are always the same build. With the option
unset the startup packet is byte-identical to today's, so external clients are unaffected.

Rollback is a plain binary swap — a node on an older build simply stops emitting and consuming the
parameter, and there is no state to migrate or clean up.

## Test plan

- [x] `./yb_build.sh release --cxx-test dist_trace-test`

`TestBackfillBackendJoinsQueryTrace` drives the production path that spawns a backend this
way: `CREATE INDEX` on a YSQL table, where `Tablet::BackfillIndexesForYsql` opens an internal
PG connection while the `BackfillIndex` server span is active. It asserts that the
`BACKFILL INDEX ...` statement the spawned backend runs appears in the statement's own trace,
as a `ysql` span parented under the tserver's `TabletServerAdminService.BackfillIndex` span.

The new `WaitForCrossServiceChildSpan` collector helper does the pairing, matching on the
parent's op name and the child's query text. Both are needed: `CREATE INDEX` opens a second
internal connection as well — the `pg_stat_activity` poll behind
`WaitForYsqlBackendsCatalogVersion` — and matching on services alone would be satisfied by
either, so the assertion has to name which one it means. `WaitForRemoteChildSpan` cannot be
reused here because the two sides share no op-name convention.

### Stack

### DIFF
--- a/src/postgres/src/backend/postmaster/postmaster.c
+++ b/src/postgres/src/backend/postmaster/postmaster.c
@@ -2489,6 +2489,15 @@ retry1:
 							 errhint("Value must be one of the registered "
 									 "YbInternalConnKind wire names.")));
 			}
+			else if (YBIsEnabledInPostgresEnvVar()
+					 && strcmp(nameptr, "yb_dist_traceparent") == 0)
+			{
+				/*
+				 * Stash traceparent for InitPostgres's backend-init span;
+				 * consumed here, turned into a corresponding GUC.
+				 */
+				port->yb_dist_traceparent = pstrdup(valptr);
+			}
 			else if (strncmp(nameptr, "_pq_.", 5) == 0)
 			{
 				/*
--- a/src/postgres/src/backend/utils/init/postinit.c
+++ b/src/postgres/src/backend/utils/init/postinit.c
@@ -1130,6 +1130,22 @@ InitPostgresImpl(const char *in_dbname, Oid dboid,
 	/* Connect to YugaByte cluster. */
 	YBInitPostgresBackend("postgres", yb_init_info);
 
+	if (IsYugaByteEnabled() && !bootstrap && MyProcPort != NULL &&
+		MyProcPort->yb_dist_traceparent != NULL &&
+		MyProcPort->yb_dist_traceparent[0] != '\0')
+	{
+		char		tracecontext[128];
+
+		snprintf(tracecontext, sizeof(tracecontext), "traceparent='%s'",
+				 MyProcPort->yb_dist_traceparent);
+
+		/* Warn instead of failing the connection when the value is rejected. */
+		(void) set_config_option("yb_dist_tracecontext", tracecontext,
+								 PGC_BACKEND, PGC_S_CLIENT, GUC_ACTION_SET,
+								 true /* changeVal */, WARNING,
+								 false /* is_reload */);
+	}
+
 	if (IsYugaByteEnabled() && !bootstrap)
 	{
 		HandleYBStatus(YBCPgTableExists(Template1DbOid,
--- a/src/yb/util/dist_trace.cc
+++ b/src/yb/util/dist_trace.cc
@@ -225,8 +225,8 @@ Status InitDistTraceProvider(const resource_sdk::Resource& resource_attrs) {
 // supplied (through GUC or comment) traceparent header.
 class TraceparentCarrier : public context::propagation::TextMapCarrier {
  public:
-  explicit TraceparentCarrier(nostd::string_view traceparent)
-      : traceparent_(traceparent) {}
+  explicit TraceparentCarrier(nostd::string_view traceparent = {})
+      : traceparent_(traceparent.data(), traceparent.size()) {}
 
   nostd::string_view Get(nostd::string_view key) const noexcept override {
     if (key == trace::propagation::kTraceParent) {
@@ -235,10 +235,16 @@ class TraceparentCarrier : public context::propagation::TextMapCarrier {
     return {};
   }
 
-  void Set(nostd::string_view, nostd::string_view) noexcept override {}
+  void Set(nostd::string_view key, nostd::string_view value) noexcept override {
+    if (key == trace::propagation::kTraceParent) {
+      traceparent_.assign(value.data(), value.size());
+    }
+  }
+
+  const std::string& traceparent() const { return traceparent_; }
 
  private:
-  nostd::string_view traceparent_;
+  std::string traceparent_;
 };
 
 }  // namespace
@@ -309,6 +315,17 @@ trace::SpanContext GetTraceparentSpanContext(const char* traceparent) {
 
   // Return the SpanContext from the parent context.
   return trace::GetSpan(parent_context)->GetContext();
+}
+
+std::string GetActiveTraceparent() {
+  if (!HasActiveContext()) {
+    return {};
+  }
+  TraceparentCarrier carrier;
+  static const auto propagator =
+      context::propagation::GlobalTextMapPropagator::GetGlobalPropagator();
+  propagator->Inject(carrier, context::RuntimeContext::GetCurrent());
+  return carrier.traceparent();
 }
 
 bool HasActiveContext() {
--- a/src/yb/yql/pgwrapper/dist_trace-test.cc
+++ b/src/yb/yql/pgwrapper/dist_trace-test.cc
@@ -603,6 +603,42 @@ class OtlpHttpCollector {
     return downstream_span;
   }
 
+  // Waits for a span from child_service in trace_id whose query text starts with
+  // child_query_prefix and whose parent is a span from parent_service with an op name starting
+  // with parent_op_prefix. Returns the child span.
+  Result<Span> WaitForCrossServiceChildSpan(
+      std::string_view trace_id, std::string_view parent_service,
+      std::string_view parent_op_prefix, std::string_view child_service,
+      std::string_view child_query_prefix) const EXCLUDES(mutex_) {
+    Span child_span;
+    RETURN_NOT_OK(WaitFor(
+        [&]() -> Result<bool> {
+          std::lock_guard lock(mutex_);
+          auto it = traces_.find(std::string(trace_id));
+          if (it == traces_.end()) return false;
+          const auto& spans = it->second.spans;
+          for (const auto& child : spans) {
+            if (child.service_name != child_service || child.parent_span_id.empty() ||
+                !child.query_text.starts_with(child_query_prefix)) {
+              continue;
+            }
+            for (const auto& parent : spans) {
+              if (parent.service_name == parent_service &&
+                  parent.op_name.starts_with(parent_op_prefix) &&
+                  parent.span_id == child.parent_span_id) {
+                child_span = child;
+                return true;
+              }
+            }
+          }
+          return false;
+        },
+        kOtelBatchScheduleDelayMs * kTimeMultiplier * 50ms,
+        Format("Span on '$0' for '$1*' parented by a '$2' span '$3*' in trace '$4'",
+               child_service, child_query_prefix, parent_service, parent_op_prefix, trace_id)));
+    return child_span;
+  }
+
  private:
   void HandleTraceRequest(const Webserver::WebRequest& req, Webserver::WebResponse* resp) {
     if (req.request_method != "POST") {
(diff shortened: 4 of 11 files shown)


src/postgres/src/backend/utils/init/postinit.c
```diff
@@ -1130,6 +1130,22 @@ InitPostgresImpl(const char *in_dbname, Oid dboid,
 	/* Connect to YugaByte cluster. */
 	YBInitPostgresBackend("postgres", yb_init_info);
 
+	if (IsYugaByteEnabled() && !bootstrap && MyProcPort != NULL &&
```
IshanChhangani: `IsYugaByteEnabled()` is unnecessary now, can be skipped.

src/postgres/src/backend/postmaster/postmaster.c
```diff
@@ -2489,6 +2489,15 @@ ProcessStartupPacket(Port *port, bool ssl_done, bool gss_done)
...
 									 "YbInternalConnKind wire names.")));
 			}
+			else if (YBIsEnabledInPostgresEnvVar()
+					 && strcmp(nameptr, "yb_dist_traceparent") == 0)
+			{
+				/*
+				 * Stash traceparent for InitPostgres's backend-init span;
+				 * consumed here, turned into a corresponding GUC.
+				 */
+				port->yb_dist_traceparent = pstrdup(valptr);
```
IshanChhangani, line 2502: Isn't it easier to pass yb_dist_tracecontext using generic startup GUC path `options='-c yb_dist_tracecontext=...'`?

src/postgres/src/backend/postmaster/postmaster.c
```diff
@@ -2489,6 +2489,15 @@ ProcessStartupPacket(Port *port, bool ssl_done, bool gss_done)
 							 errhint("Value must be one of the registered "
 									 "YbInternalConnKind wire names.")));
 			}
+			else if (YBIsEnabledInPostgresEnvVar()
+					 && strcmp(nameptr, "yb_dist_traceparent") == 0)
```
IshanChhangani, line 2495: postmaster.c:2367 :
```
	if (!YbIsAuthPassthroughInProgress(port))
		oldcontext = MemoryContextSwitchTo(TopMemoryContext);
```

So in case of connection manager, we might access memory from the transaction's context, which gets freed during abort/commit. Wouldn't that mean that our port->yb_dist_traceparent continue pointing to freed memory?

src/postgres/src/backend/utils/init/postinit.c
```diff
@@ -1130,6 +1130,22 @@ InitPostgresImpl(const char *in_dbname, Oid dboid,
...
+		MyProcPort->yb_dist_traceparent != NULL &&
+		MyProcPort->yb_dist_traceparent[0] != '\0')
+	{
+		char		tracecontext[128];
+
+		snprintf(tracecontext, sizeof(tracecontext), "traceparent='%s'",
+				 MyProcPort->yb_dist_traceparent);
+
+		/* Warn instead of failing the connection when the value is rejected. */
+		(void) set_config_option("yb_dist_tracecontext", tracecontext,
```
IshanChhangani: `set_config_option` with `PGC_S_CLIENT` would change the default value of the session with currently set value. That means even RESET yb_dist_tracecontext would not set it back to empty string.

src/postgres/src/backend/utils/init/postinit.c
```diff
@@ -1130,6 +1130,22 @@ InitPostgresImpl(const char *in_dbname, Oid dboid,
 	/* Connect to YugaByte cluster. */
 	YBInitPostgresBackend("postgres", yb_init_info);
 
+	if (IsYugaByteEnabled() && !bootstrap && MyProcPort != NULL &&
+		MyProcPort->yb_dist_traceparent != NULL &&
+		MyProcPort->yb_dist_traceparent[0] != '\0')
+	{
+		char		tracecontext[128];
```
IshanChhangani: We can reuse the `YB_TRACEPARENT_*` constants.

src/yb/util/dist_trace.cc
```diff
@@ -311,6 +317,17 @@ trace::SpanContext GetTraceparentSpanContext(const char* traceparent) {
   return trace::GetSpan(parent_context)->GetContext();
 }
 
+std::string GetActiveTraceparent() {
```
IshanChhangani, line 343: Can we just use `GetActiveSpanContext()`?

src/postgres/src/backend/utils/init/postinit.c
ellabaron-code: Fixed.

src/yb/yql/pgwrapper/dist_trace-test.cc
```diff
@@ -603,6 +603,42 @@ class OtlpHttpCollector {
     return downstream_span;
   }
 
+  // Waits for a span from child_service in trace_id whose query text starts with
+  // child_query_prefix and whose parent is a span from parent_service with an op name starting
+  // with parent_op_prefix. Returns the child span.
+  Result<Span> WaitForCrossServiceChildSpan(
```
IshanChhangani, line 641: Seems very similar to `WaitForRemoteChildSpan`, can we refactor to extract out the shared part?

IshanChhangani (changes requested):
Description can be updated just like the previous PRs.

src/postgres/src/backend/postmaster/postmaster.c
ellabaron-code: If this were only about the queries the client runs on the connection, `options` would work — the
GUC would be in place well before the first statement. It has to cover the backend's own
initialization as well. `process_startup_options` runs at the end of `InitPostgres`
(postinit.c:1384), long after `YBInitPostgresBackend` (:1131), and a follow-up change instruments
the catalog queries a backend runs between those two points. Passed via `options`, the GUC only
takes effect once all of them have already run, so that path cannot capture them. That is why the
traceparent is stashed on the `Port` and turned into the GUC right after the pggate connection is
established.

There is a failure-mode difference too: `options` goes through `process_postgres_switches` →
`SetConfigOption`, which resolves `PGC_S_CLIENT` at ERROR elevel. `check_yb_dist_tracecontext`
rejects a malformed traceparent, and also rejects any traceparent when `YBCIsDistTraceEnabled()` is
false — so a stale or ill-formed header would refuse the connection instead of leaving it untraced.
Setting the GUC directly lets me pass WARNING.

src/yb/util/dist_trace.cc
ellabaron-code: `GetActiveSpanContext()` returns a `SpanContext` object, which is what the in-process hops
(thread pool, rpc, operation driver) need. The startup packet carries the serialized W3C string, so
this goes through the global propagator's `Inject` — the exact inverse of
`GetTraceparentSpanContext`, which the receiving side uses to parse it back. Building the string
from a `SpanContext` would mean hand-formatting the hex myself.

src/postgres/src/backend/postmaster/postmaster.c
ellabaron-code: Good catch. Under auth passthrough the parse loop allocates in the auth transaction's context by
design, and the control backend's own Port fields are saved and restored around the call
(postgres.c:7722 and :7824). `yb_dist_traceparent` was not in that set, so it would have been left
pointing into freed memory once that transaction ended. Nothing reads it afterwards today — the only
reader is `InitPostgres`, which runs long before a backend ever serves an 'A' packet — but it is a
trap.

Rather than adding it to the save/restore list, the value is now not stored at all when
`YbIsAuthPassthroughInProgress(port)`. The traceparent belongs to a logical client whose queries run
on a different backend, so it means nothing on the control backend, and the guard keys off the same
predicate that decides the memory context a few lines up.

src/postgres/src/backend/utils/init/postinit.c
ellabaron-code: Right. `PGC_S_CLIENT` is below `PGC_S_OVERRIDE`, so `makeDefault` comes out true (guc.c:11642) and
`reset_val` takes the injected value, leaving `RESET yb_dist_tracecontext` to restore the traceparent
rather than clear it. Switched the source to `PGC_S_SESSION`, which is above the threshold, so
`reset_val` stays empty and `RESET` clears the trace context.

Largely moot in practice — this path only fires on internal connections, which never set or reset the
GUC — but the GUC itself is user-facing and shared with the other entry points, so it should behave
consistently.

src/yb/yql/pgwrapper/dist_trace-test.cc
ellabaron-code: Done. Extracted `WaitForParentedSpan`, which waits for a span matching a child predicate whose
parent in the same trace matches a parent predicate. Both helpers now supply their two matchers plus
the wait message, and the `rpc.system` check folds into `WaitForRemoteChildSpan`'s child matcher.
`WaitForLocalHopToRemoteSpan` is a three-level chain, so I left it as is.

src/postgres/src/backend/utils/init/postinit.c
ellabaron-code: Done. The buffer is now sized as `YB_TRACEPARENT_KEY_PREFIX_LEN + 2 * YB_TRACEPARENT_QUOTE_LEN +
YB_TRACEPARENT_VALUE_LEN + 1` instead of a magic 128, and the string is built from
`YB_TRACEPARENT_KEY_PREFIX` rather than a second copy of the literal.

ellabaron-code:
Updated. The Upgrade/Rollback section had been copied over from the RPC-boundary PR and described
the `rpc_header.proto` field, which this PR does not touch. It now covers the actual wire-visible
change here — the new `yb_dist_traceparent` startup parameter, and why a mixed-build client and
backend cannot occur for the internal connections that set it.

src/postgres/src/backend/postmaster/postmaster.c
IshanChhangani: Not storing it fixes the freed-memory read, but the param still reaches the GUC 'else' case, so `yb_dist_traceparent` and its value get added to port->guc_options. 
What does `process_startup_options` do with it in YbAuthPassthroughSetupGUCAndReport? 
Is it a good idea to consume and discard the value?

ellabaron-code: Good catch — through conn mgr the connection really did abort with "unrecognized configuration parameter". Fixed by always consuming the param and only skipping the stash under auth passthrough, and added a conn-mgr regression test that fails on the old code.

src/yb/util/dist_trace.cc
```diff
@@ -334,6 +340,17 @@ trace::SpanContext GetTraceparentSpanContext(const char* traceparent) {
...
+  if (!HasActiveContext()) {
+    return {};
+  }
+  TraceparentCarrier carrier;
```
IshanChhangani, line 350: This should call `std::string(nullptr, 0)` which seems to be undefined behaviour.

src/yb/yql/pgwrapper/libpq_utils.cc
```diff
@@ -1030,7 +1034,8 @@ PGConnBuilder CreateInternalPGConnBuilder(
        .password = UInt64ToString(postgres_auth_key),
        .connect_timeout = connect_timeout,
        .yb_internal_conn_kind = std::string(yb_internal_conn_kind),
-       .should_stop = std::move(should_stop)});
+       .should_stop = std::move(should_stop),
+       .traceparent = dist_trace::GetActiveTraceparent()});
```
IshanChhangani: `PGConnBuilder` builds conn_str_ once in its constructor, so the traceparent gets frozen there and the backend keeps it for the whole session. This works for short-lived connection, but if we want to hold builder longer (`ysql_upgrade.cc:377`) or a connection reused across polls (`XClusterDdlQueueHandler::pg_conn_`)? Those would keep adding spans to a trace that finished ages ago. Should we grab the traceparent per Connect() instead, or have the backend drop it once init is done?

src/postgres/src/backend/utils/init/postinit.c
```diff
@@ -1130,6 +1130,26 @@ InitPostgresImpl(const char *in_dbname, Oid dboid,
...
 
+	if (!bootstrap && MyProcPort != NULL &&
+		MyProcPort->yb_dist_traceparent != NULL &&
+		MyProcPort->yb_dist_traceparent[0] != '\0')
+	{
+		char		tracecontext[YB_TRACEPARENT_KEY_PREFIX_LEN +
+								 2 * YB_TRACEPARENT_QUOTE_LEN +
+								 YB_TRACEPARENT_VALUE_LEN + 1];
+
+		snprintf(tracecontext, sizeof(tracecontext),
```
IshanChhangani, line 1141: NIT:
An attacker without having password can run something like:
```
ysqlsh "host=a.b.c.d port=5433 user=yugabyte yb_dist_traceparent=$(printf '\nLOG:  statement: DROP TABLE customers;')"
```

which can add misleading lines (at max 55 bytes) to logs like:
```
WARNING:  invalid value for parameter "yb_dist_tracecontext": "traceparent='
LOG:  statement: DROP TABLE customers;'"
```

This isnt a problem for other options because of 
`pg_clean_ascii(tmp_app_name);` 

Although this issue already exists with user param too
`ysqlsh -h a.b.c.d -p 5433 -U $'\nFATAL:  fake line here'` 
But good to avoid it.

src/yb/util/dist_trace.cc
ellabaron-code: Fixed — the constructor now only assigns when the view is non-empty.

src/yb/yql/pgwrapper/libpq_utils.cc
ellabaron-code: Moved the capture into `Connect()`, so a stored builder picks up whatever trace is active at each connect. For a reused session the GUC is `USERSET`, so a holder can `SET`/`RESET` it if it ever needs to re-parent.

src/postgres/src/backend/utils/init/postinit.c
ellabaron-code: Fixed — the stashed value goes through `pg_clean_ascii()`, same as `application_name`.

IshanChhangani (approved):

PR #33116, opened 2026-08-04
[#16670] DocDB: dist-trace: span/scope API and its direct callers

ellabaron-code:
## Summary

Rework the dist_trace library into the span-and-scope API the rest of the stack builds
on: a span bundled with an activated thread-local scope, entry points for starting one
and for re-establishing a parent context, and accessors for reading the current span's
identity. Every in-tree caller of the changed API is updated in lockstep so the tree
keeps building, and the tests follow the renamed entry points along with the span names
and status text that result.

## Test plan

      commit.

### Stack

### DIFF
--- a/src/postgres/src/backend/utils/misc/pg_yb_utils.c
+++ b/src/postgres/src/backend/utils/misc/pg_yb_utils.c
@@ -1098,7 +1098,7 @@ YBInitPostgresBackend(const char *program_name, const YbcPgInitPostgresInfo *ini
 			hex_encode((const char *) YbGetLocalTServerUuid(), UUID_LEN, hex_uuid);
 			hex_uuid[2 * UUID_LEN] = '\0';
 
-			YBCInitDistTrace(MyProcPid, hex_uuid);
+			YBCInitDistTrace(hex_uuid);
 
 			/* Hooks that close node spans left open by a query abort. */
 			YbDistTraceInstallExecutorHooks();
@@ -1110,7 +1110,7 @@ void
 YBOnPostgresBackendShutdown()
 {
 	if (YBCIsDistTraceEnabled())
-		YBCCleanupDistTrace();
+		YBCShutdownDistTrace();
 
 	YBCDestroyPgGate();
 }
--- a/src/yb/rpc/outbound_call.cc
+++ b/src/yb/rpc/outbound_call.cc
@@ -151,13 +151,11 @@ bool FinishedState(RpcCallState state) {
   return false;
 }
 
-void SetSpanStatus(opentelemetry::trace::Span& span, RpcCallState state) {
+void SetSpanStatus(opentelemetry::trace::Span& span, RpcCallState state, const Status& status) {
   switch (state) {
     case TIMED_OUT:
-      span.SetStatus(opentelemetry::trace::StatusCode::kError, "Call TimedOut");
-      return;
     case FINISHED_ERROR:
-      span.SetStatus(opentelemetry::trace::StatusCode::kError, "Call ErroredOut");
+      span.SetStatus(opentelemetry::trace::StatusCode::kError, status.ToUserMessage());
       return;
     case FINISHED_SUCCESS:
       span.SetStatus(opentelemetry::trace::StatusCode::kOk);
@@ -279,9 +277,16 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
   IncrementCounter(rpc_metrics_->outbound_calls_created);
   IncrementGauge(rpc_metrics_->outbound_calls_alive);
 
-  if (dist_trace::HasActiveContext()) {
-    otel_span_ = dist_trace::StartSpan(
-        Format("rpc $0", remote_method_.ToString()), dist_trace::GetPendingRpcAttrPairs());
+  // Capture this call's parent before StartClientSpanWithScope makes otel_span_ current;
+  // InvokeCallbackSync restores it around the callback so its follow-on work nests as a sibling.
+  trace_parent_ = dist_trace::GetActiveSpanContext();
+
+  otel_span_ = dist_trace::StartClientSpanWithScope(Format("rpc $0", remote_method_.ToString()));
+  if (otel_span_) {
+    otel_span_->SetAttribute("rpc.system", "outbound_rpc");
+    otel_span_->SetAttribute("rpc.service", remote_method_.service_name());
+    otel_span_->SetAttribute("rpc.method", remote_method_.method_name());
+    otel_span_->SetAttribute("rpc.call_id", call_id_);
   }
 }
 
@@ -472,7 +477,7 @@ OutboundCall::State OutboundCall::state() const {
   return state_.load(std::memory_order_acquire);
 }
 
-bool OutboundCall::SetState(State new_state) {
+bool OutboundCall::SetState(State new_state, const Status& status) {
   auto old_state = state();
   // Sanity check state transitions.
   DVLOG(3) << "OutboundCall " << this << " (" << ToString() << ") switching from "
@@ -487,7 +492,8 @@ bool OutboundCall::SetState(State new_state) {
     }
     if (state_.compare_exchange_weak(old_state, new_state, std::memory_order_acq_rel)) {
       if (otel_span_ && FinishedState(new_state)) {
-        SetSpanStatus(*otel_span_, new_state);
+        DCHECK(otel_span_->span);
+        SetSpanStatus(*otel_span_->span, new_state, status);
         otel_span_->End();
       }
       return true;
@@ -557,7 +563,13 @@ void OutboundCall::InvokeCallbackSync(std::optional<CoarseTimePoint> now_optiona
   // TODO: consider removing the cycle-based mechanism of reporting slow callbacks below.
 
   int64_t start_cycles = CycleClock::Now();
-  callback_();
+  // Re-activate the call's parent context so RPCs the callback issues nest as siblings, not
+  // parentless roots. No-op when trace_parent_ is invalid; parent_scope drops at block end so it
+  // can't leak.
+  {
+    auto parent_scope = dist_trace::ActivateParentScope(trace_parent_);
+    callback_();
+  }
   // Clear the callback, since it may be holding onto reference counts
   // via bound parameters. We do this inside the timer because it's possible
   // the user has naughty destructors that block, and we want to account for that
@@ -692,7 +704,7 @@ void OutboundCall::SetFailed(const Status &status, std::unique_ptr<ErrorStatusPB
   bool invoke_callback;
   {
     std::lock_guard l(mtx_);
-    invoke_callback = SetState(RpcCallState::FINISHED_ERROR);
+    invoke_callback = SetState(RpcCallState::FINISHED_ERROR, status);
     if (invoke_callback) {
       status_ = status;
       if (status_.IsRemoteError()) {
@@ -731,7 +743,7 @@ void OutboundCall::SetTimedOut() {
         call_id_);
     std::lock_guard l(mtx_);
     status_ = std::move(status);
-    invoke_callback = SetState(RpcCallState::TIMED_OUT);
+    invoke_callback = SetState(RpcCallState::TIMED_OUT, status_);
   }
   if (invoke_callback) {
     InvokeCallback();
--- a/src/yb/rpc/outbound_call.cc
+++ b/src/yb/rpc/outbound_call.cc
@@ -287,6 +287,7 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
     otel_span_->SetAttribute("rpc.service", remote_method_.service_name());
     otel_span_->SetAttribute("rpc.method", remote_method_.method_name());
     otel_span_->SetAttribute("rpc.call_id", call_id_);
+    otel_span_->DropScope();
   }
 }
 
--- a/src/yb/rpc/outbound_call.h
+++ b/src/yb/rpc/outbound_call.h
@@ -72,12 +72,11 @@
 #include "yb/util/memory/memory_usage.h"
 #include "yb/util/monotime.h"
 #include "yb/util/net/sockaddr.h"
-#include "yb/util/dist_trace_fwd.h"
+#include "yb/util/dist_trace.h"
 #include "yb/util/object_pool.h"
 #include "yb/util/ref_cnt_buffer.h"
 #include "yb/util/shared_lock.h"
 #include "yb/util/slice.h"
-#include "yb/util/status_fwd.h"
 #include "yb/util/trace.h"
 
 using namespace std::literals;
@@ -349,6 +348,16 @@ class OutboundCall : public RpcCall {
   void SetConnectionId(const ConnectionId& value, const std::string* hostname) {
     conn_id_ = value;
     hostname_ = hostname;
+    if (otel_span_) {
+      if (hostname) {
+        otel_span_->SetAttribute("network.peer.name", *hostname);
+      }
+      otel_span_->SetAttribute("network.peer.address", yb::ToString(value.remote()));
+      // Drop the OTEL span's scope here, on the calling thread, before the call is queued to the
+      // reactor. The span ends later, but the scope must be released on the thread that installed
+      // it.
+      otel_span_->DropScope();
+    }
   }
 
   void SetThreadPoolFailure(const Status& status) EXCLUDES(mtx_) {
@@ -453,6 +462,10 @@ class OutboundCall : public RpcCall {
   virtual size_t GetSidecarsCount() const;
   virtual size_t TransferSidecars(Sidecars* dest);
 
+  // Distributed-trace span for this call; LocalOutboundCall reads its context to parent the local
+  // inbound span.
+  const dist_trace::SpanWithScopePtr& otel_span() const { return otel_span_; }
+
   // ----------------------------------------------------------------------------------------------
   // Protected fields set in constructor or during initialization
   // ----------------------------------------------------------------------------------------------
@@ -481,7 +494,7 @@ class OutboundCall : public RpcCall {
 
   void NotifyTransferred(const Status& status, const ConnectionPtr& conn) override;
 
-  MUST_USE_RESULT bool SetState(State new_state);
+  MUST_USE_RESULT bool SetState(State new_state, const Status& status = Status::OK());
   State state() const;
 
   // return current status
@@ -595,9 +608,13 @@ class OutboundCall : public RpcCall {
 
   std::unique_ptr<MetadataSerializer> metadata_serializer_;
 
-  // OpenTelemetry span for distributed tracing. Created when the call starts if there is an
-  // active trace context, ended when the call completes (success, failure, or timeout).
-  opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> otel_span_;
+  // OpenTelemetry span for this call, created at start (if a trace context is active) and ended at
+  // completion.
+  dist_trace::SpanWithScopePtr otel_span_;
+
+  // The trace context active when this call was constructed -- its PARENT, re-activated around the
+  // completion callback so follow-on RPCs nest as SIBLINGS of this call.
+  dist_trace::trace::SpanContext trace_parent_ = dist_trace::trace::SpanContext::GetInvalid();
 
   // InvokeCallbackTask should be able to call InvokeCallbackSync and we don't want other that
   // method to be public.
--- a/src/yb/rpc/outbound_call.h
+++ b/src/yb/rpc/outbound_call.h
@@ -77,6 +77,7 @@
 #include "yb/util/ref_cnt_buffer.h"
 #include "yb/util/shared_lock.h"
 #include "yb/util/slice.h"
+#include "yb/util/status.h"
 #include "yb/util/trace.h"
 
 using namespace std::literals;
--- a/src/yb/rpc/outbound_call.h
+++ b/src/yb/rpc/outbound_call.h
@@ -353,10 +353,6 @@ class OutboundCall : public RpcCall {
         otel_span_->SetAttribute("network.peer.name", *hostname);
       }
       otel_span_->SetAttribute("network.peer.address", yb::ToString(value.remote()));
-      // Drop the OTEL span's scope here, on the calling thread, before the call is queued to the
-      // reactor. The span ends later, but the scope must be released on the thread that installed
-      // it.
-      otel_span_->DropScope();
     }
   }
 
--- a/src/yb/util/dist_trace.cc
+++ b/src/yb/util/dist_trace.cc
@@ -11,31 +11,18 @@
 // under the License.
 //
 
-#include "yb/util/dist_trace.h"
-
-#include <deque>
-#include <memory>
-
 #include "opentelemetry/context/propagation/global_propagator.h"
-#include "opentelemetry/context/propagation/text_map_propagator.h"
-#include "opentelemetry/context/runtime_context.h"
 #include "opentelemetry/exporters/otlp/otlp_http_exporter_factory.h"
 #include "opentelemetry/exporters/otlp/otlp_http_exporter_options.h"
 #include "opentelemetry/sdk/common/global_log_handler.h"
 #include "opentelemetry/sdk/trace/batch_span_processor_factory.h"
-#include "opentelemetry/sdk/trace/batch_span_processor_options.h"
-#include "opentelemetry/sdk/trace/tracer_provider.h"
 #include "opentelemetry/sdk/trace/tracer_provider_factory.h"
 #include "opentelemetry/sdk/trace/provider.h"
-#include "opentelemetry/trace/context.h"
 #include "opentelemetry/trace/propagation/http_trace_context.h"
 #include "opentelemetry/trace/provider.h"
-#include "opentelemetry/trace/span.h"
-#include "opentelemetry/trace/span_metadata.h"
-#include "opentelemetry/trace/tracer.h"
 
+#include "yb/util/dist_trace.h"
 #include "yb/util/flag_validators.h"
-#include "yb/util/flags.h"
 #include "yb/util/signal_util.h"
 
 DEFINE_NON_RUNTIME_PREVIEW_string(otel_collector_traces_endpoint, "",
@@ -50,12 +37,17 @@ DEFINE_NON_RUNTIME_string(otel_ssl_ca_cert_string, "",
     "CA certificate bundle contents for OTLP HTTPS trace export. Used only when no CA certificate "
     "bundle path is configured.");
 
-DEFINE_NON_RUNTIME_uint32(otel_batch_max_queue_size, 2048,
+DEFINE_NON_RUNTIME_uint32(otel_batch_max_queue_size, 16384,
     "Maximum number of spans that can be buffered in the batch span processor queue. "
     "Spans arriving after this limit are dropped. Must be greater than 0 and at least as "
     "large as otel_batch_max_export_batch_size.");
 
-DEFINE_NON_RUNTIME_uint32(otel_batch_schedule_delay_ms, 5000,
+DEFINE_NON_RUNTIME_uint32(otel_ysql_batch_max_queue_size, 2048,
+    "Like otel_batch_max_queue_size, but used only by the ysql (postgres backend) process, which "
+    "runs one batch span processor per connection. Must be greater than 0 and at least as large as "
+    "otel_batch_max_export_batch_size.");
+
+DEFINE_NON_RUNTIME_uint32(otel_batch_schedule_delay_ms, 500,
     "Time interval in milliseconds between two consecutive batch exports of spans to the "
     "OpenTelemetry collector. Lower values reduce latency but increase export frequency.");
 
@@ -68,6 +60,9 @@ DEFINE_NON_RUNTIME_string(otel_internal_log_level, "info",
     "values are debug, info, warning, error, and none.");
 
 DEFINE_validator(otel_batch_max_queue_size,
+    FLAG_GE_FLAG_VALIDATOR(otel_batch_max_export_batch_size));
+
+DEFINE_validator(otel_ysql_batch_max_queue_size,
     FLAG_GE_FLAG_VALIDATOR(otel_batch_max_export_batch_size));
 
 DEFINE_validator(otel_internal_log_level,
@@ -83,38 +78,39 @@ namespace context = opentelemetry::context;
 
 namespace {
 
-const nostd::string_view ysql_resource_name = "ysql";
+// Service name for the tracing resource and tracer (e.g. "ysql", "Master", "TabletServer").
+static std::string g_service_name;
 
-// Owns string attribute data and maintains a parallel vector of string_view/AttributeValue pairs
-// that can be passed directly to the OTel Tracer::StartSpan API. Uses std::deque for pointer
-// stability -- unlike std::vector, deque does not relocate existing elements on insertion, so
-// string_views into earlier entries remain valid.
-class RpcSpanAttrs {
- public:
-  void AddStringAttr(std::string key, std::string value) {
-    auto& owned_key = owned_keys_.emplace_back(std::move(key));
-    auto& owned_val = owned_values_.emplace_back(std::move(value));
-    attrs_.emplace_back(owned_key, owned_val);
+// The ysql process gets its own queue-size flag; tserver/master share otel_batch_max_queue_size.
+static uint32_t EffectiveBatchMaxQueueSize() {
+  return g_service_name == kYsqlServiceName ? FLAGS_otel_ysql_batch_max_queue_size
+                                            : FLAGS_otel_batch_max_queue_size;
+}
+
+// A batch of pending RPC span attributes, owned as plain (key, value) strings.
+using PendingRpcSpanAttrs = std::vector<std::pair<std::string, std::string>>;
+
+thread_local PendingRpcSpanAttrs pending_rpc_attrs;
+
+// Moves the pending attributes out of the thread-local buffer, leaving it empty. The returned batch
+// owns the strings.
+static PendingRpcSpanAttrs ConsumePendingRpcAttrs() {
+  PendingRpcSpanAttrs consumed = std::move(pending_rpc_attrs);
+  // std::move leaves the source unspecified; force it empty.
+  pending_rpc_attrs.clear();
+  return consumed;
+}
+
+// Builds the OTel-API attribute vector (string_view/AttributeValue pairs) viewing into `attrs`.
+std::vector<std::pair<nostd::string_view, opentelemetry::common::AttributeValue>> SpanAttrsView(
+    const PendingRpcSpanAttrs& attrs) {
+  std::vector<std::pair<nostd::string_view, opentelemetry::common::AttributeValue>> view;
+  view.reserve(attrs.size());
+  for (const auto& [key, value] : attrs) {
+    view.emplace_back(key, value);
   }
-
-  const std::vector<std::pair<nostd::string_view, opentelemetry::common::AttributeValue>>& attrs()
-      const {
-    return attrs_;
-  }
-
-  void clear() {
-    attrs_.clear();
-    owned_keys_.clear();
-    owned_values_.clear();
-  }
-
- private:
-  std::deque<std::string> owned_keys_;
-  std::deque<std::string> owned_values_;
-  std::vector<std::pair<nostd::string_view, opentelemetry::common::AttributeValue>> attrs_;
-};
-
-thread_local RpcSpanAttrs pending_rpc_attrs;
+  return view;
+}
 
 internal_log::LogLevel GetOtelInternalLogLevel() {
   const auto& flag_value = FLAGS_otel_internal_log_level;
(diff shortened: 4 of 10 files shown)


src/yb/util/dist_trace.cc
```diff
@@ -50,12 +37,17 @@ DEFINE_NON_RUNTIME_string(otel_ssl_ca_cert_string, "",
     "CA certificate bundle contents for OTLP HTTPS trace export. Used only when no CA certificate "
     "bundle path is configured.");
 
-DEFINE_NON_RUNTIME_uint32(otel_batch_max_queue_size, 2048,
+DEFINE_NON_RUNTIME_uint32(otel_batch_max_queue_size, 16384,
```
IshanChhangani: Why were the batching flags changed? And what is the reasoning for these constants? The earlier values were otel defaults.

Updating these values right now seems a little ahead of sequence since tserver tracing isn't initialized yet.

src/yb/util/dist_trace.cc
```diff
@@ -70,6 +62,9 @@ DEFINE_NON_RUNTIME_string(otel_internal_log_level, "info",
 DEFINE_validator(otel_batch_max_queue_size,
     FLAG_GE_FLAG_VALIDATOR(otel_batch_max_export_batch_size));
 
+DEFINE_validator(otel_ysql_batch_max_queue_size,
```
IshanChhangani: Should there be a validator between ysql and tserver's max_queue_size params too?

Also why do we only need a copy of max_queue_size and not the other params for tservers?

src/postgres/src/backend/utils/misc/pg_yb_utils.c
```diff
@@ -1110,7 +1110,7 @@ void
 YBOnPostgresBackendShutdown()
 {
 	if (YBCIsDistTraceEnabled())
-		YBCCleanupDistTrace();
+		YBCShutdownDistTrace();
```
IshanChhangani, line 1113: Yes, for this function, Shutdown seems to be more accurate name.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -283,9 +283,23 @@ struct ResponseReadyTraits;
 std::string_view GetSharedMemSpanName(tserver::PgSharedExchangeReqType req_type) {
   switch (req_type) {
     case tserver::PgSharedExchangeReqType::PERFORM:
-      return "shmem req yb.tserver.PgClientService.Perform";
```
IshanChhangani, line 284: Why do we need this naming change?

src/yb/rpc/outbound_call.cc
```diff
@@ -151,13 +151,11 @@ bool FinishedState(RpcCallState state) {
   return false;
 }
 
-void SetSpanStatus(opentelemetry::trace::Span& span, RpcCallState state) {
+void SetSpanStatus(opentelemetry::trace::Span& span, RpcCallState state, const Status& status) {
   switch (state) {
     case TIMED_OUT:
-      span.SetStatus(opentelemetry::trace::StatusCode::kError, "Call TimedOut");
```
IshanChhangani, line 157:
```
  void EndSharedMemorySpan(const Status& status) {
    if (!otel_span) {
      return;
    }
    if (status.ok()) {
      otel_span->SetStatus(opentelemetry::trace::StatusCode::kOk);
    } else if (status.IsTimedOut()) {
      otel_span->SetStatus(opentelemetry::trace::StatusCode::kError, "Call TimedOut");
    } else {
      otel_span->SetStatus(opentelemetry::trace::StatusCode::kError, "Call ErroredOut");
    }
    otel_span->End();
    otel_span = nullptr;
  }
```
Still maps failures to fixed "Call TimedOut" etc. So RPC timeouts and shared memory timeouts will show different messages.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -283,9 +283,23 @@ struct ResponseReadyTraits;
...
-      return "shmem req yb.tserver.PgClientService.AcquireObjectLock";
+      return "shmem yb.tserver.PgClientService.AcquireObjectLock";
+    case tserver::PgSharedExchangeReqType_INT_MIN_SENTINEL_DO_NOT_USE_: [[fallthrough]];
+    case tserver::PgSharedExchangeReqType_INT_MAX_SENTINEL_DO_NOT_USE_: break;
+  }
+  FATAL_INVALID_ENUM_VALUE(tserver::PgSharedExchangeReqType, req_type);
+}
+
+// Method name attribute for the outbound shared-memory span, mirroring rpc.method on RPC spans (the
+// service is always PgClientService). Paired with the tserver's SharedMemMethodName.
```
IshanChhangani: Function does not exist.

src/yb/yql/pggate/ybc_dist_trace.cc
```diff
@@ -23,6 +23,7 @@
 #include "opentelemetry/trace/tracer.h"
 
 #include "yb/util/dist_trace.h"
+#include "yb/util/flags.h"
```
IshanChhangani: is this used?

src/yb/util/dist_trace.h
```diff
@@ -13,28 +13,85 @@
...
+inline const std::string kYsqlServiceName = "ysql";
+
+void InitDistTrace(
+    opentelemetry::nostd::string_view service_name, opentelemetry::nostd::string_view node_uuid);
+void ShutdownDistTrace();
 nostd::shared_ptr<opentelemetry::trace::Tracer> GetDistTracer();
 bool IsDistTraceEnabled();
 trace::SpanContext GetTraceparentSpanContext(const char* traceparent);
+
+// Get SpanContext of the active span
```
IshanChhangani: Comment does not convey anything that function name does not. Lets add only the necessary comments.

src/yb/util/dist_trace.h
```diff
@@ -13,28 +13,85 @@
 
 #pragma once
 
-#include <cstdint>
```
IshanChhangani, line 16: Let us not rely on transitive inclusions.

src/yb/util/dist_trace.h
```diff
@@ -13,28 +13,85 @@
...
+
+  nostd::shared_ptr<trace::Span> span;
+  std::optional<trace::Scope> scope;
+  std::thread::id owner_thread = std::this_thread::get_id();
+};
+
+using SpanWithScopePtr = std::shared_ptr<SpanWithScope>;
+
+// OTel service.name for the ysql (postgres backend) process, passed to InitDistTrace at startup.
+inline const std::string kYsqlServiceName = "ysql";
```
IshanChhangani: How about `inline constexpr char kYsqlServiceName[] = "ysql"; `
for dynamic initialization and an exit-time destructor.

src/yb/util/dist_trace.cc
```diff
@@ -11,31 +11,18 @@
...
 #include "opentelemetry/sdk/trace/tracer_provider_factory.h"
 #include "opentelemetry/sdk/trace/provider.h"
-#include "opentelemetry/trace/context.h"
 #include "opentelemetry/trace/propagation/http_trace_context.h"
 #include "opentelemetry/trace/provider.h"
-#include "opentelemetry/trace/span.h"
-#include "opentelemetry/trace/span_metadata.h"
-#include "opentelemetry/trace/tracer.h"
 
+#include "yb/util/dist_trace.h"
```
IshanChhangani: Keeping the header at the top seems to be consistent with the codebase.

IshanChhangani (changes requested):
There are some functions which are defined but never used in the codebase (might be used later on), it seems better to delay them until they are used. This should help in justifying the usecase of these functions.

src/yb/util/dist_trace.cc
```diff
@@ -83,38 +78,39 @@ namespace context = opentelemetry::context;
 
 namespace {
 
-const nostd::string_view ysql_resource_name = "ysql";
+// Service name for the tracing resource and tracer (e.g. "ysql", "Master", "TabletServer").
+static std::string g_service_name;
```
IshanChhangani: Better to avoid static constant variables.

src/postgres/src/backend/utils/misc/pg_yb_utils.c
ellabaron-code: Thank you.

src/yb/rpc/outbound_call.cc
ellabaron-code: Fixed — EndSharedMemorySpan now reports status.ToUserMessage(), so a failure reads the same on both transports instead of collapsing to a fixed string.

src/yb/util/dist_trace.h
ellabaron-code: Fixed.

src/yb/yql/pggate/pg_client.cc
ellabaron-code: The name is picked so that both ends of the shared-memory call will share one operation name once the tserver's inbound span lands in a follow-up PR, with span kind distinguishing them — kClient here, kServer there. That mirrors the RPC path, where both ends of a call are named rpc yb.tserver.<Service>.<Method>. Keeping "req" would make the two ends of one call read as two different operations.

src/yb/util/dist_trace.cc
ellabaron-code: Agreed — reverted here, both defaults are back to the OTel values, and the changes move to the next PR, which is where tracing is first initialized in a tserver or master. On the reasoning: a tserver and master produce far more spans than a single postgres backend does, and a full queue drops spans rather than blocking, so the queue is deepened (16384) and drained more often (500 ms).

src/yb/util/dist_trace.cc
ellabaron-code: There's no ordering relation between the two queue-size flags — they're independent per-process knobs, so a validator between them wouldn't assert anything meaningful. Each is validated against otel_batch_max_export_batch_size instead, which the OTel SDK documents as a bound but doesn't enforce; both validators land in the next PR along with the flags themselves.

On the second question: queue depth is the only one of these knobs that scales with per-process span volume. max_export_batch_size is bounded by what the collector accepts per request, and the schedule delay is a latency choice that should stay uniform across processes, so neither needs a per-process copy.

src/yb/yql/pggate/pg_client.cc
ellabaron-code: Removed that reference from the comment.

src/yb/util/dist_trace.h
ellabaron-code: Removed.

src/yb/util/dist_trace.h
ellabaron-code: Fixed.

src/yb/util/dist_trace.cc
ellabaron-code: Fixed.

src/yb/util/dist_trace.cc
ellabaron-code: Dropped the static — it was redundant, since the anonymous namespace already gives internal linkage.

src/yb/yql/pggate/ybc_dist_trace.cc
ellabaron-code: No — the file has no FLAGS_ reference, so the include is removed.

ellabaron-code:
Done. `StartSpanWithScope` and `StartServerSpanWithScope` are both out of this PR.

`StartClientSpanWithScope` was the only caller of `StartSpanWithScope`, and both functions checked the same thing, so the two are now one function.

`StartServerSpanWithScope` moves to the next PR, where its first caller is. It had two forms, one of which took an attribute list; no caller passes attributes, so only the two-argument form is kept.

The batching flags moved for the same reason — nothing uses them here yet.

IshanChhangani:
> Part 1 of 5. This stack replaces the earlier 9-PR split (https://github.com/yugabyte/yugabyte-db/pull/32895–https://github.com/yugabyte/yugabyte-db/pull/32903) with five
themed PRs, each of which carries the tests for what it adds and passes on its own
commit. The content is otherwise the same change, re-partitioned.

I think this is not relevant with the commit itself, can be dropped from the description.

> Start{Client,Server}SpanWithScope

StartServerSpanWithScope is no longer in this diff

> `CleanupDistTrace` → `ShutdownDistTrace`

It seems the description goes into too much implementation-level detail. Can we reduce that.

>  ### Batching flag changes and why they diverge from the OTel SDK defaults

Can be dropped.

> Also drops a comment on `StartClientSpanWithScope` claiming the span may be "an  
optional root gated by `otel_rpc_sampling_ratio`": no such flag exists, and the  
function returns `nullptr` when there is no active context.

Can be dropped.

> Nothing here writes a trace context onto the wire: that needs `TraceContextPB`,  
which arrives with the RPC-boundary PR, so the header write and the matching reader  
land together there.

This information is irrelevant with this PR. Can be dropped.

IshanChhangani:
Can we also trigger jenkins early on to ensure the API change does not break the existing tests. 

edit: seems like agentk does a substring match instead of a string match. So jenkins is already triggered now. 
Can you trigger it on the rest of the PRs. 
And a single run after all PRs with dist-trace enabled by default would help in understanding if any existing (non dist-trace) test fails.

src/yb/rpc/outbound_call.cc
```diff
@@ -279,9 +277,16 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
...
   IncrementGauge(rpc_metrics_->outbound_calls_alive);
 
-  if (dist_trace::HasActiveContext()) {
-    otel_span_ = dist_trace::StartSpan(
-        Format("rpc $0", remote_method_.ToString()), dist_trace::GetPendingRpcAttrPairs());
+  // Capture this call's parent before StartClientSpanWithScope makes otel_span_ current;
+  // InvokeCallbackSync restores it around the callback so its follow-on work nests as a sibling.
+  trace_parent_ = dist_trace::GetActiveSpanContext();
+
+  otel_span_ = dist_trace::StartClientSpanWithScope(Format("rpc $0", remote_method_.ToString()));
```
IshanChhangani: Object no longer represents a span. Can't think of a better name though.

src/yb/util/dist_trace.h
```diff
@@ -16,25 +16,88 @@
...
+      span->End();
+    }
+  }
+
+  nostd::shared_ptr<trace::Span> span;
+  std::optional<trace::Scope> scope;
+  std::thread::id owner_thread = std::this_thread::get_id();
+};
+
+using SpanWithScopePtr = std::shared_ptr<SpanWithScope>;
```
IshanChhangani: Every owner seems to take an exclusive access and pointer is never copied. It might be better to use unique pointer instead of shared pointer if we do not need shared ownership.

src/yb/rpc/outbound_call.h
```diff
@@ -349,6 +348,16 @@ class OutboundCall : public RpcCall {
   void SetConnectionId(const ConnectionId& value, const std::string* hostname) {
```
IshanChhangani, line 349: The function says `SetConnectionId` but it now does more than that. We are dealing with the tracing scope lifecycle here.

src/yb/rpc/outbound_call.h
```diff
@@ -481,7 +494,7 @@ class OutboundCall : public RpcCall {
 
   void NotifyTransferred(const Status& status, const ConnectionPtr& conn) override;
 
-  MUST_USE_RESULT bool SetState(State new_state);
+  MUST_USE_RESULT bool SetState(State new_state, const Status& status = Status::OK());
```
IshanChhangani, line 493: We have removed `#include "yb/util/status_fwd.h"` from the file, is this a transitive import?

IshanChhangani (changes requested):

ellabaron-code:
On the enabled-by-default run: I don't have a way to run that. The on-switch is otel_collector_traces_endpoint, a non-runtime preview flag that's empty by default, so enabling it globally points every daemon at an OTLP endpoint that has to actually be listening — the tests in this stack each start their own collector for that reason. How would you suggest setting it up?

src/yb/rpc/outbound_call.cc
ellabaron-code: It still is the call's span — SpanWithScope owns the span and forwards the span operations (SetAttribute, SetStatus, End); the activated scope is just how its lifetime is tied to the call. A scope-flavoured name would make the common uses read worse, so leaving it as otel_span_.

src/yb/util/dist_trace.h
ellabaron-code: This PR's call sites are single-owner, agreed. But this alias threads through the whole stack, and I can't tell from here what the later PRs need — these scopes get handed to whichever thread responds, and are ended from a callback rather than by their creator. Reworking every call site to unique_ptr and then hitting one place that needs sharing puts us back where we started, so I'd keep shared_ptr.

src/yb/rpc/outbound_call.h
ellabaron-code: It still sets the connection id — the tracing here doesn't change what the function does, it uses it: the peer attributes are the connection id and hostname it just set, and the scope drop has to happen on the thread that installed the scope, which is this last step before the call is queued to the reactor.

src/yb/rpc/outbound_call.h
ellabaron-code: Yes, transitively. status_fwd.h only forward-declares Status, and this header needs the complete type — Status status_ and Status thread_pool_failure_ are by-value members, and SetState now defaults to Status::OK(). The complete type comes from yb/util/flags.h and yb/rpc/lightweight_message.h, which include yb/util/status.h directly and aren't touched here. So the header we dropped was never what satisfied that need; putting it back would only cost compile time.

src/yb/util/dist_trace.h
IshanChhangani: I went through the stack, still couldn't figure out where is the shared ownership established for this. Can you point me to that code.

src/yb/util/dist_trace.cc
IshanChhangani: Problem is not just with the static here. We are having a global variable that is not constant.
- `a global variable satisfies these requirements if its declaration, considered in isolation, could be constexpr.` - Google C++ Style Guide, [Static and Global Variables](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables)
- [I.2: Avoid non-const global variables](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-global)

How about we wrap this into a class to make the lifecycle explicit? In that case we may have have to keep 2 instances one in pg(like in PgApiImpl), one in tserver?

IshanChhangani:
> On the enabled-by-default run: I don't have a way to run that. The on-switch is otel_collector_traces_endpoint, a non-runtime preview flag that's empty by default, so enabling it globally points every daemon at an OTLP endpoint that has to actually be listening — the tests in this stack each start their own collector for that reason. How would you suggest setting it up?

Agreed. Let me check if there is a way in which you can do it.

src/yb/rpc/outbound_call.h
IshanChhangani: > It still sets the connection id
Agreed. Should this code live in this function, or does it make more sense to move it into a separate function.
Something like:
```
controller->call_->SetConnectionId(conn_id, &remote_.host());
controller->call_->OurChange();
```

src/yb/rpc/outbound_call.h
IshanChhangani: Agree that status_fwd.h doesn't help much here. How about we directly include `yb/util/status.h` instead of depending on `yb/util/flags.h` and `yb/rpc/lightweight_message.h`

src/yb/util/dist_trace.h
IshanChhangani: Also on a side note, how about explicitly deleting moves?
```
SpanWithScope(SpanWithScope&&) = delete;
SpanWithScope& operator=(SpanWithScope&&) = delete;
```

IshanChhangani (changes requested):

src/yb/rpc/outbound_call.h
IshanChhangani: Apart from this, implicity our scope push (constructor) and pop (SetConnectionId) should be from the same thread. Is it not the case (generally it is) that `DnsResolver::AsyncResolve` completes in its own IO thread?

src/yb/util/dist_trace.h
ellabaron-code: Changed to unique pointer.

src/yb/rpc/outbound_call.h
ellabaron-code: I'd keep it here. The attributes mirror exactly what this function does — setting the connection id is setting the peer address, so recording it isn't separable work. Extracting it would mean a second function with the same signature called from the same spot, which doesn't buy anything.

You're right. `DnsResolver::AsyncResolve` only completes inline when the host is already in the resolver cache; on a miss the callback runs on the resolver's `io_service` thread and drains the whole waiter queue there, so the detach could pop a token attached by a different thread — and `Detach` pops everything above it as well. Moved the `DropScope()` into the `OutboundCall` constructor, right after the attributes are set: the scope is still top of stack at that point and it's the thread that attached it. Nothing between the two lines hops threads, and nothing in between needs the outbound span to be current — the wire header takes it from `GetContext()` and the local inbound call from `otel_span_context()`.

src/yb/rpc/outbound_call.h
ellabaron-code: Done — added #include `"yb/util/status.h"`.

For the record I don't agree with the general principle here. Including every header a translation unit's declarations happen to name doesn't buy correctness in this codebase — Status arrives one way or another — and the cost is real: more headers parsed per TU, more edges in the dependency graph, longer builds and more rebuilds on unrelated touches. In this specific case the include is already reachable through two paths, so making it explicit changes nothing but the file's length. Taking it because it's one line, but I don't think it should be applied as a blanket rule.

src/yb/util/dist_trace.cc
ellabaron-code: The existing dist_trace surface is free functions, not a class — I`nitDistTrace / ShutdownDistTrace / GetDistTracer / StartClientSpanWithScope / AddPendingRpcStringAttr`. Had the tracing functionality been consolidated into one object up front, the service name would have an obvious home and I'd put it there. Adding a class now solely to hold a single string written once at startup isn't worth it: none of the call sites can reach an instance — `StartClientSpanWithScope` runs from the `OutboundCall` constructor, `AddPendingRpcStringAttr` from PgSession, and `ybc_dist_trace.cc` is entered from C in `pg_yb_utils.c `— so it needs either threading the object through all of them or a global accessor to find it, which is the same global with more code. And the OTel provider is a process-wide singleton anyway, so there's never more than one live instance per process.

The variable is going to be mutable global state either way — a class member reached through a global accessor is no less global than a namespace-scope string. If what concerns you is the static-init and exit-time destruction, I can move it into a function-local static so it's constructed on first use and never destroyed:

`std::string& ServiceName() {
  static auto* s = new std::string();
  return *s;
}`

No static init, no destruction ordering, three call sites. I'll take that if it settles the concern.

ellabaron-code:
Done — made the changes I found acceptable:

- `SpanWithScope` held through `std::unique_ptr`, with copy and move deleted.
- `DropScope()` moved from `SetConnectionId` into the `OutboundCall` constructor, on the thread that attached the scope.
- `outbound_call.h` includes `yb/util/status.h` directly.

Details are on the individual threads.

src/yb/util/dist_trace.cc
IshanChhangani: Since we are extending the framework to tserver now, having a separate class for the surface seems like a good idea (Those functions can be part of the class too).

> The variable is going to be mutable global state either way
My understand of classes here was to have separate child classes with immutable member variable.

ellabaron-code: I'd rather not restructure the dist-trace surface as part of this stack.
`InitDistTrace`, `ShutdownDistTrace`, `GetDistTracer`, `IsDistTraceEnabled`,
`StartSpan` and `AddPendingRpcStringAttr` are already free functions on master, so
reshaping them into a class is a refactor of existing code — it isn't part of what
this effort sets out to do, and it would churn five PRs that are mid-review. Once
they land I'll put up a separate PR for it, and the service name gets its home
there.

On the variable itself: it's assigned once in `InitDistTrace` and read three
times, all inside `dist_trace.cc`, and it isn't visible outside the translation
unit. There are 94 namespace-scope `std::string` globals in non-test `src/yb` code
today, including mutable ones with initialization-order dependencies between them
— see src/yb/consensus/log.cc lines 278 and 279, where one placeholder-prefix
string is built from another at static initialization time. Changing this one
variable wouldn't make the codebase any more consistent, and it's the least
exposed instance of the pattern.

Wrapping it in a class also wouldn't address what you linked: that section covers
static data members of classes as well as namespace-scope variables, so a private
static member is the same kind of variable. And with no instances to hold, it
would be a class existing only to group statics, which the guide's [Nonmember,
Static Member, and Global Functions](https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions) section advises against.

IshanChhangani: @amitanandaiyer Can you offer suggestion about this. I am not sure if this change is required in this PR or can be delayed.

src/yb/rpc/outbound_call.h
```diff
@@ -349,6 +349,12 @@ class OutboundCall : public RpcCall {
   void SetConnectionId(const ConnectionId& value, const std::string* hostname) {
     conn_id_ = value;
     hostname_ = hostname;
+    if (otel_span_) {
+      if (hostname) {
+        otel_span_->SetAttribute("network.peer.name", *hostname);
```
IshanChhangani: Isn't network.peer.address (IP + port) already having the name (IP) within it?

src/yb/yql/pggate/pg_session.cc
```diff
@@ -124,7 +124,17 @@ void PublishPendingRpcTableInfo(const PgsqlOps& ops, const PgSession::TableCache
...
   }
-  dist_trace::ClearPendingRpcAttrs();
+
+  // Publish the details of the Perform RPC.
+  size_t reads = 0;
+  size_t writes = 0;
+  for (const auto& op : ops) {
+    (op->is_read() ? reads : writes)++;
+  }
+  dist_trace::AddPendingRpcStringAttr("rpc.perform.op_count", std::to_string(ops.size()));
```
IshanChhangani: This seems redundant. op_count = reads + writes?

src/yb/yql/pggate/pg_session.cc
```diff
@@ -124,7 +124,17 @@ void PublishPendingRpcTableInfo(const PgsqlOps& ops, const PgSession::TableCache
...
+  dist_trace::AddPendingRpcStringAttr("rpc.perform.reads", std::to_string(reads));
```
IshanChhangani: Addition of read and write op counts is a great addition, as without them it was difficult to understand what the RPC is going to do.

src/yb/yql/pggate/pg_session.cc
```diff
@@ -124,7 +124,17 @@ void PublishPendingRpcTableInfo(const PgsqlOps& ops, const PgSession::TableCache
...
+  dist_trace::AddPendingRpcStringAttr("rpc.perform.writes", std::to_string(writes));
```
IshanChhangani: To avoid the confusion with explain plans lets keep the attributes as `rpc.read_ops` and `rpc.write_ops`.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -429,14 +443,19 @@ struct PgClientData : public FetchBigDataCallback {
...
 
   void StartSharedMemorySpan() {
-    if (dist_trace::HasActiveContext()) {
-      otel_span = dist_trace::StartSpan(
-          GetSharedMemSpanName(kSharedExchangeRequestType), dist_trace::GetPendingRpcAttrPairs());
+    otel_span = dist_trace::StartClientSpanWithScope(
+        GetSharedMemSpanName(kSharedExchangeRequestType));
+    if (otel_span) {
+      // Mirror the attributes the RPC outbound span carries (outbound_call.cc).
+      otel_span->SetAttribute("rpc.system", "outbound_shmem");
```
IshanChhangani: Setting this attribute to outbound/inbound_shmem/rpc seems improper for two reasons:
    1.  [Otel semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/) requires `rpc.system.name` to identify the RPC system, such as grpc or connectrpc.
      2. We already have `span.kind` to distinguish between outbound and inbound spans.

src/yb/rpc/outbound_call.h
```diff
@@ -349,6 +349,12 @@ class OutboundCall : public RpcCall {
...
+      }
+      otel_span_->SetAttribute("network.peer.address", yb::ToString(value.remote()));
```
IshanChhangani: [Otel semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/) seems to have the following rules:
- [server.address](https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/) : string identifying a RPC server instance request is sent to. (Probably the network.peer.name in our case)
- [server.port](https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/)
- [network.peer.address](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/) : Peer address of the network connection - IP address or UNIX domain socket name (The definition used in this code includes the port too while the convention keeps a separate param for it)
- [network.peer.port](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/)

src/yb/yql/pggate/pg_session.cc
ellabaron-code: Thank you.

src/yb/yql/pggate/pg_session.cc
ellabaron-code: Fixed.

src/yb/yql/pggate/pg_session.cc
ellabaron-code: Done.

src/yb/yql/pggate/pg_client.cc
ellabaron-code: Changed across all four span-creation sites.

`rpc.system` now carries the transport only: `yb_rpc` for the wire protocol and `yb_shmem` for the PG↔tserver shared memory exchange. The attribute is `rpc.system` rather than `rpc.system.name`, and the convention allows system-specific values beyond the registered ones — the two transports need to be distinguishable, since they differ in framing and size limits and shmem falls back to RPC.

Direction now comes from `span.kind`. The test collector wasn't recording the kind, so it had been using `rpc.system` as its direction discriminator; it now records the OTLP kind and matches on that, with one explicit assertion per transport so the attribute value itself stays covered.

src/yb/rpc/outbound_call.h
ellabaron-code: Split, as described.

The hostname the proxy resolved is now `server.address`. The other three come straight off the connection endpoint, which exposes the address and port separately, so no parsing is needed: `network.peer.address` is now the bare IP with no brackets on IPv6, and `network.peer.port` holds the port on its own.

`server.port` and `network.peer.port` hold the same number here because the endpoint comes from resolving that same `HostPort`; only the address differs between the pairs — name versus resolved IP.

src/yb/rpc/outbound_call.cc
```diff
@@ -279,9 +277,15 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
   IncrementCounter(rpc_metrics_->outbound_calls_created);
   IncrementGauge(rpc_metrics_->outbound_calls_alive);
 
-  if (dist_trace::HasActiveContext()) {
-    otel_span_ = dist_trace::StartSpan(
-        Format("rpc $0", remote_method_.ToString()), dist_trace::GetPendingRpcAttrPairs());
+  // Capture this call's parent before StartClientSpanWithScope makes otel_span_ current;
```
IshanChhangani: nit: traceparent seems clearer.

IshanChhangani: This is missing the `  if (dist_trace::HasActiveContext()) {` check. Is that intentional?

IshanChhangani: I see that we have added this on the next PR. Given that this PR has not been merged yet lets fix it right here. And remove the same changes from the next PR.

ellabaron-code: Fixed.

ellabaron-code: Fixed.

src/yb/yql/pggate/pg_client.cc
IshanChhangani: Why do we need `rpc.system` when the op_name already conveys wheather the request went through rpc or shmem.

src/yb/rpc/outbound_call.h
```diff
@@ -349,6 +349,15 @@ class OutboundCall : public RpcCall {
   void SetConnectionId(const ConnectionId& value, const std::string* hostname) {
     conn_id_ = value;
     hostname_ = hostname;
+    if (otel_span_) {
+      if (hostname) {
```
IshanChhangani, line 353: Isn't this the only call
```
  controller->call_->SetConnectionId(conn_id, &remote_.host());
```
In that case this `if (hostname)` statement seems always true.

IshanChhangani:
Is the stack of PRs required in the commit messages? Given that Github already shows the stack.

src/yb/rpc/outbound_call.h
```diff
@@ -595,9 +608,13 @@ class OutboundCall : public RpcCall {
 
   std::unique_ptr<MetadataSerializer> metadata_serializer_;
 
-  // OpenTelemetry span for distributed tracing. Created when the call starts if there is an
-  // active trace context, ended when the call completes (success, failure, or timeout).
-  opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> otel_span_;
+  // OpenTelemetry span for this call, created at start (if a trace context is active) and ended at
```
IshanChhangani, line 607: This gets moved and updated in PR2. The change doesn't seem functionally specific to PR2, if thats the case we can make it here itself. Also can you search for other some codepaths where a change made by a PR gets overwritten in an upcoming PR? This should simplify the changes per PR. 
- SpanWithScope internals too seems to have changed between PR1 and PR2.

src/yb/yql/pggate/pg_client.cc
ellabaron-code: The same call shows up under two names depending on transport — `rpc
yb.tserver.PgClientService.Perform` and `shmem yb.tserver.PgClientService.Perform`. If you only
have the name, those look like two unrelated operations, so you can't ask "what's the p99 of
Perform, split by transport" in one query, and you can't pull every shared-memory call in a trace
without regexing over names. With the attribute you just filter on it.

src/yb/rpc/outbound_call.h
ellabaron-code: While `proxy.cc:338` is the only caller and it can't pass null, the parameter is still a pointer,
so the check follows the signature — with a reference there'd be nothing to check.

But changing the signature touches pre-existing code and its one caller, so it belongs in a
separate PR, and the `if` goes away with it.

ellabaron-code:
Not required — it's there for after the merge. GitHub shows the base branch, so from a PR you see
the immediate parent but not the rest of the series, and once these land on master none of that
context is in `git log`. Someone running `git blame` on a change that looks incomplete on its own
then has a way to see it was one layer of a series.

src/yb/rpc/outbound_call.h
ellabaron-code: That's a consequence of splitting the stack this way. Nothing in PR1 reads that member from
outside the class, so PR1 keeps it private with an accessor. PR2 adds `LocalOutboundCall`, which
reads the span to parent the local inbound span — a subclass, so PR2 drops the accessor and makes
the member protected. PR1 has no reason to be protected; PR2 does.

Aug 5 was "PR1 shouldn't carry what PR1 doesn't use"
([r3720652117](https://github.com/yugabyte/yugabyte-db/pull/33116#discussion_r3720652117), which
is why the batching flags and `StartServerSpanWithScope` moved to the PRs that use them). Aug 24
is "PR1 should carry the final form even though PR1 doesn't use it." Those two can't both be
satisfied — the member declaration you're pointing at is only in its PR2 place because PR2's
local-call path reads it.

src/yb/rpc/outbound_call.cc
```diff
@@ -279,9 +277,18 @@ OutboundCall::OutboundCall(const RemoteMethod& remote_method,
   IncrementCounter(rpc_metrics_->outbound_calls_created);
   IncrementGauge(rpc_metrics_->outbound_calls_alive);
 
+  // Guarded so that with tracing off the span name below is never formatted.
```
IshanChhangani: Can be avoided.

ellabaron-code: Sorry, I can't tell what "can be avoided" refers to here — the comment line it's anchored on, or the code under it? Can you clarify?

IshanChhangani: Sorry for the confusion. I meant that the code seems self explanatory, comment is redundant.

ellabaron-code: Fixed.

src/yb/yql/pggate/pg_client.cc
```diff
@@ -1204,8 +1203,30 @@ class PgClient::Impl : public BigDataFetcher {
     if (tablespace_oid) {
       lock_oid.set_tablespace_oid(*tablespace_oid);
     }
-    req.set_lock_type(static_cast<tserver::ObjectLockMode>(mode));
+    const auto lock_type = static_cast<tserver::ObjectLockMode>(mode);
+    req.set_lock_type(lock_type);
     req.set_is_session_lock(is_session_lock);
+
+    // Publish the details of AcquireObjectLock.
+    if (dist_trace::HasActiveContext()) {
```
IshanChhangani, line 1209: Is this required given that `AddPendingRpcStringAttr` already  has a `HasActiveContext` check?

IshanChhangani (approved):

src/yb/yql/pggate/pg_client.cc
ellabaron-code: The outer check short-circuits the whole block: without it every acquire with tracing off would still format all seven attribute strings and pay seven individual `HasActiveContext` checks inside `AddPendingRpcStringAttr` that would each conclude the same thing. One check up front instead of seven.

PR #33118, opened 2026-08-04
[#31461] YSQL: Honor deferred read point for DDLs in regular transaction blocks

### LINKED ISSUE #31461: [YSQL] Allow deferred mode for DDL
Jira Link: [DB-21280](https://yugabyte.atlassian.net/browse/DB-21280)
### Description

Remove ddl_mode check here

```
    if (!options.ddl_mode() && !options.use_catalog_session() && options.defer_read_point()) {
      // For DMLs, only fast path writes cannot be deferred.
      RETURN_NOT_OK(session.read_point()->TrySetDeferredCurrentReadTime());
    }
```

and add a test case for it to work.

### Issue Type

kind/enhancement

[DB-21280]: https://yugabyte.atlassian.net/browse/DB-21280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bmcnama-shopify:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/33118/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

## Summary

Fixes #31461 (Jira: DB-21280).

When `yb_read_after_commit_visibility` is set to `'deferred'` (or a `READ ONLY DEFERRABLE` transaction is used), pggate requests a deferred read point via the `defer_read_point` option. `PgClientSessionImpl::UpdateReadTime` dropped that request for any perform RPC with `ddl_mode: true`, so DDL statements silently ran with a regular, restart-prone read time.

With `ysql_yb_ddl_transaction_block_enabled` (transactional DDL), DDLs run in the regular transaction block and pggate does request deferral for them; the tserver-side drop then causes `Restart read required` errors that the query layer cannot retry (DDL retries are unsupported in READ COMMITTED with transactional DDL), and violates pggate's "Deferred reads do not face read restart errors" invariant (`PgTxnManager::CheckConflictsAcrossReadTimeOptions`). With the flag disabled, pggate never requests deferral for DDLs (`PgTxnManager::CalculateIsolation` returns early before computing `need_defer_read_point_`), so this change is behavior-neutral in that mode.

This PR drops the `ddl_mode` conjunct from the deferral guard so that a requested deferral is honored for DDLs too. Reads on the legacy catalog session still cannot be deferred.

The new `PgReadAfterCommitVisibilityDdlTest.DeferredModeDdl` test reproduces the dropped deferral and verifies the fix. It runs with table-level object locks disabled because the lock acquisition request that otherwise precedes every DDL is not in ddl mode and picks (and defers) the statement read point itself, masking the bug. Without the fix the test fails deterministically:

```
ERROR:  Restart read required (query layer retry isn't possible because retrying DDL
statements are not supported in READ COMMITTED isolation level when transactional DDL
is enabled. If object locking is enabled, kConflict and kReadRestart errors won't occur.)
```

Note: the issue carries "2025.2 Backport Required" and "2026.1 Backport Required" labels; backport PRs are to follow separately (the guard exists in both branches with older field names).

## Upgrade/Rollback safety

No wire-format, catalog, or on-disk changes; no flag defaults change. The behavior change is confined to perform RPCs that carry both `ddl_mode: true` and `defer_read_point: true`, which pggate only sends when `ysql_yb_ddl_transaction_block_enabled` is set and the session opted into deferred visibility (`yb_read_after_commit_visibility = 'deferred'` or a deferrable read-only transaction). During a mixed-version window an un-upgraded tserver simply keeps the old behavior (deferral dropped for DDLs), which is safe; rollback restores it uniformly.

## Test plan

- [x] `./yb_build.sh debug --cxx-test pg_read_after_commit_visibility-test --gtest_filter PgReadAfterCommitVisibilityDdlTest.DeferredModeDdl` (new test; fails with `Restart read required` before the fix, passes after; the tserver log shows the deferral applied to the DDL's requests after the fix and never before)
- [x] `./yb_build.sh debug --cxx-test pg_read_after_commit_visibility-test --gtest_filter PgReadAfterCommitVisibilityTest.DifferentNodeDeferredRead` (existing deferred-mode DML regression control, passes)
- [x] `./build-support/lint.sh --rev origin/master` (0 errors, 0 warnings)

### DIFF
--- a/src/yb/tserver/pg_client_session.cc
+++ b/src/yb/tserver/pg_client_session.cc
@@ -3908,9 +3908,10 @@ class PgClientSession::Impl {
     RETURN_NOT_OK(
         UpdateReadPointForXClusterConsistentReads(options, deadline, session.read_point()));
 
-    if (!options.ddl_mode() && !options.use_legacy_catalog_session() &&
-        read_time_options.defer_read_point()) {
-      // For DMLs, only fast path writes cannot be deferred.
+    if (!options.use_legacy_catalog_session() && read_time_options.defer_read_point()) {
+      // Among DMLs, only fast path writes cannot be deferred. DDLs can also request
+      // deferral, but only when they run in the regular transaction block (see
+      // PgTxnManager::CalculateIsolation).
       RETURN_NOT_OK(session.read_point()->TrySetDeferredCurrentReadTime());
       VLOG_WITH_PREFIX(3) << "Set current read time for deferred mode "
           << session.read_point()->GetReadTime();
--- a/src/yb/yql/pgwrapper/pg_read_after_commit_visibility-test.cc
+++ b/src/yb/yql/pgwrapper/pg_read_after_commit_visibility-test.cc
@@ -91,10 +91,11 @@ class PgReadAfterCommitVisibilityTest : public PgMiniTestBase {
     ANNOTATE_UNPROTECTED_WRITE(FLAGS_yb_enable_read_committed_isolation) = true;
     ANNOTATE_UNPROTECTED_WRITE(FLAGS_time_source) = server::SkewedClock::kName;
     ANNOTATE_UNPROTECTED_WRITE(FLAGS_replication_factor) = 1;
-    // Support DDL concurrency with object locks.
+    // Run DDLs in the regular transaction block, optionally with DDL
+    // concurrency via object locks.
     ANNOTATE_UNPROTECTED_WRITE(FLAGS_ysql_yb_ddl_transaction_block_enabled) = true;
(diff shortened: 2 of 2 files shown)


bmcnama-shopify:
trigger jenkins

rahulb-yb:
Fixed by YB dev in another commit - https://github.com/yugabyte/yugabyte-db/commit/62cc8330aa80dade03f1aad02c2a45e15af5a51b

PR #33121, opened 2026-08-04
[#33120] Docs: Fix sizes and empty cells in the YSQL date-time types table

### LINKED ISSUE #33120: [Docs] YSQL date-time types table: timestamp/timestamptz sizes are wrong, and several cells are empty
## Description

The overview table on the YSQL **Date and time data types** page has several errors, the most significant being wrong storage sizes for `timestamp` and `timestamptz`.

Page: https://docs.yugabyte.com/stable/api/ysql/datatypes/type_datetime/
Source: `docs/content/<version>/api/ysql/datatypes/type_datetime/_index.md` — present in `stable`, `v2025.2`, `v2025.1`, `v2024.2`, and `v2.20`.

### 1. `timestamp` and `timestamptz` are listed as 12 bytes; both are 8 bytes

This is the error a user reported. The table's "Internal format" column says `12-bytes` for both types. Verification:

- `pg_type.dat`: `typname => 'timestamp'` / `'timestamptz'` both have `typlen => '8', typbyval => 'FLOAT8PASSBYVAL'`.
- `src/postgres/src/include/datatype/timestamp.h`: `typedef int64 Timestamp; typedef int64 TimestampTz;`
- `src/postgres/src/backend/catalog/yb_catalog/yb_type.c`: both OIDs map to `YB_YQL_DATA_TYPE_INT64` with `sizeof(int64)`, so it's 8 bytes on the DocDB side too.
- PostgreSQL's own [Table 8.9](https://www.postgresql.org/docs/15/datatype-datetime.html#DATATYPE-DATETIME-TABLE) documents 8 bytes.
- `pg_column_size()` returns 8 for both.

The number appears to have been displaced from the `timetz` row: commit d5f986aa1bcd502fa90f07fe3404fb3442eb1b11 rewrote the table, blanked the `timetz` row's cells, and moved its `12-bytes` onto the two timestamp rows. The table was correct before that commit.

### 2. The `timetz` row has empty Size / Min / Max / Resolution cells

Should be 12 bytes, `00:00:00+15:59` to `24:00:00-15:59`, 1 microsecond. (`TimeTzADT` is `{int64 time; int32 zone}`, declared `typlen => '12'`. The offset limit follows from `MAX_TZDISP_HOUR = 15` in `timestamp.h`; `'00:00:00+16:00'::timetz` raises "time zone displacement out of range".)

### 3. `interval` has no Min / Max

PostgreSQL documents -178000000 years to 178000000 years, and both values are accepted.

### 4. The `(p)` qualifier is described incorrectly

The page says `(p)` "specifies the precision, in microseconds, with which values will be recorded." `p` is the *number of fractional digits retained in the seconds field* (0..6), not a microsecond count — `timestamp(3)` gives milliseconds, not 3-microsecond precision. The page also omits PostgreSQL's "by default, there is no explicit bound on precision."

### 5. The special date-time constants table is under-inclusive and self-inconsistent

`'infinity'` is listed as valid with `timestamptz` but `'-infinity'` is not, which can't be right — they take the same code path. Measured behavior: `epoch`, `infinity`, `-infinity`, `today`, `tomorrow`, and `yesterday` all accept `timestamptz`, and `now` accepts `timestamptz` as well.

### 6. Malformed HTML anchor

`<a name="maximum-and-minimum-supported-values"></br></a>` — `</br>` is not a tag, and it's nested inside the anchor.

### 7. Presentation issues in the same table

- "Internal format" is a poor name for a column that gives a size; PostgreSQL calls it "Storage Size". For YugabyteDB it's also worth stating that this is the size of the value's representation, not the on-disk footprint in DocDB.
- Sizes are hyphenated (`12-bytes`) rather than `12 bytes`.
- The `interval` size cell carries a trailing "3-field struct" that no other row has.

### Measured sizes

Against a local build:

```
    type     | pg_column_size
-------------+----------------
 date        |              4
 time        |              8
 timetz      |             12
 timestamp   |              8
 timestamptz |              8
 interval    |             16
```

kai-franz:
## Summary

Fixes #33120.

The overview table on the YSQL [Date and time data types](https://docs.yugabyte.com/stable/api/ysql/datatypes/type_datetime/) page listed `timestamp` and `timestamptz` as `12-bytes`. Both are 8 bytes:

- `pg_type.dat` declares `typlen => '8', typbyval => 'FLOAT8PASSBYVAL'` for both.
- `Timestamp` and `TimestampTz` are typedefs of `int64` (`datatype/timestamp.h`).
- `yb_type.c` maps both OIDs to `YB_YQL_DATA_TYPE_INT64` with `sizeof(int64)`, so it's 8 bytes on the DocDB side too.
- PostgreSQL's [Table 8.9](https://www.postgresql.org/docs/15/datatype-datetime.html#DATATYPE-DATETIME-TABLE) documents 8 bytes, and `pg_column_size()` returns 8.

The `12` was displaced from the `timetz` row by d5f986aa1bcd502fa90f07fe3404fb3442eb1b11, which blanked that row's cells when it rewrote the table. The table was correct before that commit, so the error has been live since 2021.

While in the table, this also fixes the surrounding problems:

| Problem | Fix |
| --- | --- |
| `timestamp` / `timestamptz` sizes | 12-bytes → 8 bytes |
| `timetz` row had empty Size / Min / Max / Resolution | 12 bytes, `00:00:00+15:59` to `24:00:00-15:59`, 1 microsecond |
| `timetz` "Purpose" held only a recommendation | real purpose, with the "avoid this" pointer kept alongside it |
| `interval` had no Min / Max | -178000000 years to 178000000 years, with a footnote on those cells noting the true limits are per-field |
| "Internal format" column name | renamed to "Size" (PostgreSQL calls it "Storage Size") |
| `12-bytes` | `12 bytes` |
| `interval` size cell carried "3-field struct" | moved into the footnote |
| `(p)` described as "the precision, in microseconds" | it's the number of fractional digits retained in the seconds field; `timestamp(3)` gives milliseconds |
| `'-infinity'` omitted `timestamptz` while `'infinity'` listed it | added `timestamptz` to every special-constant row that accepts it |
| `<a name="..."></br></a>` | `</br>` isn't a tag and was nested inside the anchor |

One addition to the prose under the table: a note that `(p)` has no effect on a value's size.

The `Min`, `Max`, and `Resolution` values are unchanged except where they were blank, and are identical in the PostgreSQL 11 and 15 versions of Table 8.9, so the same values apply to every docs version touched here. Applied identically to `stable`, `v2025.2`, `v2025.1`, `v2024.2`, and `v2.20`; the only per-version difference is that `v2024.2` and `v2.20` cite the PostgreSQL 11 table, matching the link those pages already used.

One judgment call worth a reviewer's eye: the special-constants table now lists `timestamptz` where it applies, but still doesn't mention `timetz` (which also accepts `'now'` and `'allballs'`), because the page states "No further reference will be made to _timetz_." Happy to add it if the docs team would rather the table be exhaustive.

## Test plan

Documentation only — no code or runtime behavior changes.

- [x] Every size, Min, Max, and Resolution value in the new table verified against a local YugabyteDB build: `pg_column_size()` returns 4 / 8 / 12 / 8 / 8 / 16 for `date` / `time` / `timetz` / `timestamp` / `timestamptz` / `interval`.
- [x] `timetz` UTC-offset limits verified: `'00:00:00+15:59'::timetz` and `'24:00:00-15:59'::timetz` are accepted; `'00:00:00+16:00'::timetz` raises `time zone displacement out of range`.
- [x] `interval` limits verified: `'178000000 years'::interval` and `'-178000000 years'::interval` are both accepted.
- [x] `(p)` semantics verified: `timestamp(3) '2025-06-15 15:06:40.123456'` renders `...15:06:40.123`, confirming `p` is a fractional-digit count.
- [x] Special-constants table verified by casting each of the 8 constants to each of the 6 date-time types and recording which succeed.
- [x] Cross-checked against `pg_type.dat`, `datatype/timestamp.h`, `utils/date.h`, and `yb_catalog/yb_type.c`.
- [x] All 5 version copies confirmed to receive an identical diff (modulo the PostgreSQL 11 vs 15 doc link).
- [x] Markdown tables checked for consistent row length and column count; `git diff --check` reports no whitespace errors.
- [ ] Hugo render of the page reviewed on the Netlify deploy preview.


ddhodge (approved):
LGTM

PR #33132, opened 2026-08-05
[#33125] DocDB: Dump thread stacks to the log when RPC service queues stay full

### LINKED ISSUE #33125: [DocDB]  When rpc queues are full, dump worker thread stacks to log file
Jira Link: [DB-22832](https://yugabyte.atlassian.net/browse/DB-22832)
### Description

When master rpc queues are full for a long time, it would be helpful to have a threadz like dump of the stacks of workers in the rpc worker pool. The feature could be default false and work as follows
1. Poll queue size every N secs and look for queue size more than some min value like 100
2. If rpc queue is >= 100 for some N * M secs consecutively, dump stacks to regular log file
3. As long as the condition remains true, keep dumping stacks but the frequency should exponentially back off to once every 5 mins or so.

Flag should be default off but it shd be possible to enable on both tserver and master. We could start off enabling on master for specific scenarios. 

### Issue Type

kind/enhancement

[DB-22832]: https://yugabyte.atlassian.net/browse/DB-22832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bmcnama-shopify:
DO NOT MERGE!! [CSI]() ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

## Summary

Fixes #33125 (Jira: DB-22832)

Currently, when an RPC service queue overflows, the server only logs a rate-limited backpressure warning in `ServicePoolImpl::Overflow` and increments the `rpcs_queue_overflow` counter. Nothing captures what the RPC worker threads are doing, so diagnosing why the workers are not draining the queue requires manually hitting the `/threadz` endpoint while the incident is in progress.

This change adds an opt-in queue depth watchdog to the shared `src/yb/rpc/` code, so it works for both `yb-master` and `yb-tserver`:

- Lightweight per-`Messenger` registrations share one process-wide `ServiceQueueMonitor` worker and one dump backoff sequence. Services expose a best-effort queue snapshot through `RpcService::QueueSize()`, and registrations are removed before Messenger shutdown starts.
- Queue persistence is tracked independently per service. A dump becomes eligible after any queue remains at or above `rpc_queue_stack_dump_threshold` for `rpc_queue_stack_dump_min_interval_ms`; the warning summarizes up to ten affected services.
- Automatic captures are limited to the normal, tagged-normal, or high-priority RPC worker categories associated with affected services. If no expected worker category exists, collection falls back to all managed threads rather than producing an empty diagnostic.
- Threads sharing an identical stack are grouped and logged once. Every record carries a process-wide dump ID, and a completion record reports elapsed time, thread count, and group count.
- Repeated dumps use `CoarseBackoffWaiter` exponential backoff with jitter, starting at `rpc_queue_stack_dump_min_interval_ms` and capped by `rpc_queue_stack_dump_max_interval_ms` (default 5 minutes). The backoff resets after the sustained overload clears.

The single process-wide monitor thread starts when the first server Messenger registers a service. While `rpc_queue_stack_dump_threshold` is 0, it sleeps on a condition variable and continues to support enabling the feature on a live server.

## Upgrade/Rollback safety

Safe in both directions:

- The four new runtime flags (`rpc_queue_stack_dump_threshold`, `rpc_queue_stack_dump_poll_interval_ms`, `rpc_queue_stack_dump_min_interval_ms`, `rpc_queue_stack_dump_max_interval_ms`) default to the feature being disabled (`rpc_queue_stack_dump_threshold = 0`). Support can enable the watchdog on a live cluster without a restart.
- No wire-format, RPC-versioning, catalog, or on-disk format changes; the change is process-local diagnostics consisting of one background monitor thread per server process and opt-in log output.
- Rollback removes the flags and leaves no persistent state.
- Existing `NON_RUNTIME` queue-length flags such as `master_svc_queue_length` are unchanged.

## Test plan

- [x] `./yb_build.sh release --cxx-test mt-rpc-test --gtest_filter MultiThreadedRpcTest.QueueStackDumpOnSustainedQueueBuildup`: blocks two real RPC workers, fills the service queue, verifies both worker names appear in one grouped stack, and verifies jittered 200 ms then 400 ms backoff ranges.
- [x] `./yb_build.sh release --cxx-test mt-rpc-test --gtest_filter MultiThreadedRpcTest.NoQueueStackDumpWhenDisabled`: applies the same worker and queue pressure with the feature disabled and verifies no dump.
- [x] `./yb_build.sh release --cxx-test mt-rpc-test --gtest_filter MultiThreadedRpcTest.QueueStackDumpCanBeEnabledAtRuntime`: registers services while disabled, enables the runtime threshold after the queue fills, and verifies capture.
- [x] `./yb_build.sh release --cxx-test mt-rpc-test --gtest_filter MultiThreadedRpcTest.QueueStackDumpUsesOneProcessGlobalMonitor`: registers services with two Messengers and verifies one process-wide monitor thread.
- [x] `./yb_build.sh release --cxx-test mt-rpc-test --gtest_filter MultiThreadedRpcTest.TestBlowOutServiceQueue`: existing queue-overflow coverage still passes.
- [x] `./yb_build.sh release daemons --sj --skip-pg-parquet --no-odyssey --no-ybc`: clean compile of `yb-master`, `yb-tserver`, PostgreSQL, AutoFlags generation, and `yb-admin`.
- [x] `./build-support/lint.sh --rev origin/master`: 0 errors, 0 warnings.

### DIFF
--- a/src/yb/rpc/messenger.cc
+++ b/src/yb/rpc/messenger.cc
@@ -57,6 +57,7 @@
 #include "yb/rpc/rpc_metrics.h"
 #include "yb/rpc/rpc_service.h"
 #include "yb/rpc/rpc_util.h"
+#include "yb/rpc/service_queue_monitor.h"
 #include "yb/rpc/tcp_stream.h"
 #include "yb/rpc/yb_rpc.h"
 
@@ -206,12 +207,15 @@ void Messenger::Shutdown() {
   std::vector<Reactor*> reactors;
   std::unique_ptr<Acceptor> acceptor;
   std::unique_ptr<ReactorMonitor> reactor_monitor;
+  std::unique_ptr<ServiceQueueMonitor> service_queue_monitor;
   {
     std::lock_guard guard(lock_);
 
     acceptor.swap(acceptor_);
 
     reactor_monitor.swap(reactor_monitor_);
+
+    service_queue_monitor.swap(service_queue_monitor_);
 
     for (const auto& reactor : reactors_) {
       reactors.push_back(reactor.get());
@@ -225,6 +229,11 @@ void Messenger::Shutdown() {
   if (reactor_monitor) {
     reactor_monitor->Shutdown();
     reactor_monitor.reset();
+  }
+
+  if (service_queue_monitor) {
+    service_queue_monitor->Shutdown();
+    service_queue_monitor.reset();
   }
 
   for (auto* reactor : reactors) {
@@ -448,6 +457,13 @@ Status Messenger::RegisterService(
     const std::string& service_name, const scoped_refptr<RpcService>& service) {
   DCHECK(service);
   rpc_services_.emplace(service_name, service);
+  {
+    std::lock_guard guard(lock_);
+    if (!service_queue_monitor_) {
+      service_queue_monitor_ = std::make_unique<ServiceQueueMonitor>(name_);
+    }
+    service_queue_monitor_->Track(service_name, service);
+  }
   return Status::OK();
 }
 
--- a/src/yb/rpc/messenger.cc
+++ b/src/yb/rpc/messenger.cc
@@ -197,25 +197,32 @@ void Messenger::Shutdown() {
   }
   VLOG_WITH_PREFIX(1) << "shutting down messenger";
 
+  // Since we're shutting down, it's OK to block.
+  ThreadRestrictions::ScopedAllowWait allow_wait;
+
+  std::unique_ptr<ServiceQueueMonitor> service_queue_monitor;
+  {
+    std::lock_guard guard(lock_);
+    service_queue_monitor.swap(service_queue_monitor_);
+  }
+  if (service_queue_monitor) {
+    service_queue_monitor->Shutdown();
+    service_queue_monitor.reset();
+  }
+
   ShutdownThreadPools();
   ShutdownAcceptor();
   UnregisterAllServices();
 
-  // Since we're shutting down, it's OK to block.
-  ThreadRestrictions::ScopedAllowWait allow_wait;
-
   std::vector<Reactor*> reactors;
   std::unique_ptr<Acceptor> acceptor;
   std::unique_ptr<ReactorMonitor> reactor_monitor;
-  std::unique_ptr<ServiceQueueMonitor> service_queue_monitor;
   {
     std::lock_guard guard(lock_);
 
     acceptor.swap(acceptor_);
 
     reactor_monitor.swap(reactor_monitor_);
-
-    service_queue_monitor.swap(service_queue_monitor_);
 
     for (const auto& reactor : reactors_) {
       reactors.push_back(reactor.get());
@@ -229,11 +236,6 @@ void Messenger::Shutdown() {
   if (reactor_monitor) {
     reactor_monitor->Shutdown();
     reactor_monitor.reset();
-  }
-
-  if (service_queue_monitor) {
-    service_queue_monitor->Shutdown();
-    service_queue_monitor.reset();
   }
 
   for (auto* reactor : reactors) {
@@ -455,6 +457,12 @@ Result<ThreadPoolPtr> Messenger::TaggedThreadPool(TaggedThreadPools::Tag tag) {
 // Register a new RpcService to handle inbound requests.
 Status Messenger::RegisterService(
     const std::string& service_name, const scoped_refptr<RpcService>& service) {
+  return RegisterService(service_name, service, ServicePriority::kNormal);
+}
+
+Status Messenger::RegisterService(
+    const std::string& service_name, const scoped_refptr<RpcService>& service,
+    ServicePriority priority) {
   DCHECK(service);
   rpc_services_.emplace(service_name, service);
   {
@@ -462,7 +470,7 @@ Status Messenger::RegisterService(
     if (!service_queue_monitor_) {
       service_queue_monitor_ = std::make_unique<ServiceQueueMonitor>(name_);
     }
-    service_queue_monitor_->Track(service_name, service);
+    service_queue_monitor_->Track(service_name, service, priority);
   }
   return Status::OK();
 }
--- a/src/yb/rpc/mt-rpc-test.cc
+++ b/src/yb/rpc/mt-rpc-test.cc
@@ -43,12 +43,17 @@
 #include "yb/rpc/yb_rpc.h"
 
 #include "yb/util/countdown_latch.h"
+#include "yb/util/logging_test_util.h"
 #include "yb/util/metrics.h"
 #include "yb/util/net/net_util.h"
 #include "yb/util/status_log.h"
 #include "yb/util/test_macros.h"
 #include "yb/util/test_util.h"
 #include "yb/util/thread.h"
+
+DECLARE_int32(rpc_queue_stack_dump_min_polls);
+DECLARE_int32(rpc_queue_stack_dump_poll_interval_ms);
+DECLARE_int64(rpc_queue_stack_dump_threshold);
 
 METRIC_DECLARE_counter(rpc_connections_accepted);
 METRIC_DECLARE_counter(rpcs_queue_overflow);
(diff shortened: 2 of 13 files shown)


bmcnama-shopify:
trigger jenkins

src/yb/rpc/mt-rpc-test.cc
```diff
@@ -96,8 +101,107 @@ class MultiThreadedRpcTest : public RpcTestBase {
...
+
+  // Starts a server whose only service has a tiny queue served by a worker pool with no workers,
+  // then saturates it so that the service queue depth stays at its maximum until shutdown. When
+  // enabled is true, verifies that the RPC queue depth monitor dumps thread stacks to the log
+  // and that repeated dumps are suppressed with an exponentially growing backoff. When enabled
+  // is false, verifies that no dump is produced while the queue stays full.
+  void VerifyQueueStackDump(bool enabled);
 };
 
+void MultiThreadedRpcTest::VerifyQueueStackDump(bool enabled) {
```
iSignal: Can we enhance the test case slightly to actually have "sleeping" workers maybe by adding a TEST flag to a harmless call like Ping and check that we see that many stacks in the output.

src/yb/rpc/service_pool.cc
```diff
@@ -205,6 +206,11 @@ class ServicePoolImpl final : public InboundCallHandler {
     thread_pool->Enqueue(task);
   }
 
+  size_t QueuedCalls() const {
```
iSignal: minor: QueueSize is a better name for this and similar fns

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+namespace yb::rpc {
+
+namespace {
+
+constexpr int32_t kMinPollIntervalMs = 10;
+
+// Maximum number of thread names listed per group of threads sharing the same stack.
+constexpr size_t kMaxThreadNamesPerGroup = 10;
+
+// Guards against multiple messengers in the same process (e.g. a tablet server and its CQL
```
iSignal: We don't seem to use anything specific in messenger, could we make this process global instead to avoid this problem? That way we also only have one monitor thread per process

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+        break;
+      }
+      auto threshold = FLAGS_rpc_queue_stack_dump_threshold;
+      if (threshold <= 0) {
+        ResetPollState();
+        continue;
+      }
+      size_t max_depth = 0;
+      const std::string* worst_service = nullptr;
+      for (const auto& [service_name, service] : services_) {
```
iSignal: This means that if different services hit queue depth in sequence, we will still consider it a global trigger. Instead could we maintain a consecutive fail count per service to make this more accurate?

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+      if (worst_service == nullptr || max_depth < static_cast<size_t>(threshold)) {
+        ResetPollState();
+        continue;
+      }
+      if (++consecutive_polls_over_threshold_ <
+              std::max(FLAGS_rpc_queue_stack_dump_min_polls, 1)) {
+        continue;
+      }
+      auto now = CoarseMonoClock::Now();
+      if (next_dump_time_ != CoarseTimePoint() && now < next_dump_time_) {
```
iSignal: ideally, we can write this code more cleanly if we split into two parts
1. checktriggersfordump - which only cares if conditions are met from service depth pov
2. maybedump - which takes the expoential backoff into account. if we decide not to dump, this dump is skipped, we still evaluate triggers as before.
Also, I think we have an exponential backoff utility somewhere, would be good to use it. It adds some jitter etc so all nodes don't dump in lockstep.

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+                std::max(FLAGS_rpc_queue_stack_dump_min_polls, 1)
+          : std::min(suppress_ms_ * 2, max_backoff_ms);
+      next_dump_time_ = now + suppress_ms_ * 1ms;
+      // Copy what the dump needs, then release the lock so that Track and Shutdown are not
+      // blocked while stacks are being collected.
+      auto service_name = *worst_service;
+      auto consecutive_polls = consecutive_polls_over_threshold_;
+      auto suppress_ms = suppress_ms_;
+      lock.unlock();
+      DumpThreadStacks(service_name, max_depth, threshold, consecutive_polls, suppress_ms);
```
iSignal: we should pass stop_ to the thread dump utility so that we can interrupt it on shutdown, otherwise we will be stuck here for a while.

src/yb/util/thread.cc
```diff
@@ -1090,6 +1107,10 @@ void RenderAllThreadStacks(std::ostream& output) {
   thread_manager->RenderThreadGroup(kAllGroups, output);
 }
 
+std::vector<ThreadIdAndName> ListThreadsForStackTrace() {
+  return thread_manager->ListThreads();
```
iSignal, line 1112: this returns all threads. can we filter to the specific thread pools involved? it is a bit of a hassle but it could be useful to avoid all the consensus threads for example when overflowing read/write service queues

src/yb/util/thread.cc
```diff
@@ -626,6 +627,22 @@ void ThreadMgr::RenderThreadGroup(const std::string& group, std::ostream& output
...
+  std::vector<ThreadIdAndName> result;
+  std::lock_guard lock(mutex_);
+  for (const auto& [_, category] : thread_categories_) {
+    for (const auto& descriptor : category) {
+#if defined(__linux__)
+      auto tid_for_stack = descriptor.thread_id;
+#else
+      auto tid_for_stack = descriptor.pthread_id;
+#endif
+      result.push_back(ThreadIdAndName{tid_for_stack, descriptor.name});
```
iSignal: minor: result.reserve up front

src/yb/util/thread.cc
```diff
@@ -626,6 +627,22 @@ void ThreadMgr::RenderThreadGroup(const std::string& group, std::ostream& output
   RenderThreadGroupUnlocked(group, output);
 }
 
+std::vector<ThreadIdAndName> ThreadMgr::ListThreads() {
+  std::vector<ThreadIdAndName> result;
+  std::lock_guard lock(mutex_);
+  for (const auto& [_, category] : thread_categories_) {
```
iSignal: Can we use the categories to only dump threads relevant to the service pools that are full

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+      std::vector<std::string> names;
+      names.reserve(std::min(group.thread_indexes.size(), kMaxThreadNamesPerGroup + 1));
+      for (auto index : group.thread_indexes) {
+        if (names.size() == kMaxThreadNamesPerGroup) {
+          names.push_back("...");
+          break;
+        }
+        names.push_back(threads[index].name);
+      }
+      LOG(WARNING) << name_ << ": " << group.thread_indexes.size() << " thread(s) with stack ["
```
iSignal: 1. would be good to log time taken for dump itself at the end.
2. this is one giant log, which makes it easier to grep but not sure if it will cause glog to stall. It may be better to log each stack as we get it with some common identifier and we can grep that later? But no strong opinions.

iSignal (changes requested):

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+    thread->Join();
+  }
+
+  void Track(const std::string& service_name, const RpcServicePtr& service) {
+    std::lock_guard lock(mutex_);
+    if (stop_.load(std::memory_order_acquire)) {
+      return;
+    }
+    services_.emplace_back(service_name, service);
+    if (!thread_) {
```
iSignal, line 102: can we only create the thread if the feature is on

src/yb/rpc/messenger.cc
```diff
@@ -227,6 +231,11 @@ void Messenger::Shutdown() {
     reactor_monitor.reset();
   }
 
+  if (service_queue_monitor) {
```
iSignal: Let's move this to the 1st thing in shutdown, it is not essential and we may see some queues etc naturally here.

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+    "polls. Helps diagnosing why RPC worker threads are not draining the service queues. "
+    "0 disables the RPC queue depth monitor.");
+TAG_FLAG(rpc_queue_stack_dump_threshold, advanced);
+
+DEFINE_RUNTIME_int32(rpc_queue_stack_dump_poll_interval_ms, 1000,
+    "Interval (in ms) between RPC service queue depth polls, when the RPC queue depth monitor "
+    "is enabled via rpc_queue_stack_dump_threshold.");
+TAG_FLAG(rpc_queue_stack_dump_poll_interval_ms, advanced);
+
+DEFINE_RUNTIME_int32(rpc_queue_stack_dump_min_polls, 3,
```
iSignal: this could be a time flag as well, so it is easier to set. We can poll every 1s but this can be a min dump interval like 10s, so that this and the max backoff are two related flags.

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+  void ResetPollState() {
+    consecutive_polls_over_threshold_ = 0;
+    suppress_ms_ = 0;
+    next_dump_time_ = CoarseTimePoint();
+  }
+
+  void DumpThreadStacks(
+      const std::string& service_name, size_t depth, int64_t threshold, int32_t consecutive_polls,
+      int64_t suppress_ms) {
+    LOG(WARNING) << name_ << ": RPC service queue for " << service_name << " has " << depth
```
iSignal: we can log all affected services here as multiple services will likely hit queues. if we keep a count per service above in the detection phase, we can summarize the situation in one log that covers max 10 services.

src/yb/rpc/service_queue_monitor.cc
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  void Track(const std::string& service_name, const RpcServicePtr& service) {
+    std::lock_guard lock(mutex_);
+    if (stop_.load(std::memory_order_acquire)) {
+      return;
+    }
+    services_.emplace_back(service_name, service);
+    if (!thread_) {
+      thread_ = CHECK_RESULT(Thread::Make(
```
iSignal: minor: technically we can ignore this failure and move on without monitoring but failure to create a thread is uncommon.

iSignal (changes requested):

src/yb/rpc/mt-rpc-test.cc
bmcnama-shopify: Updated the test to block two real service workers, queue two more calls, and verify both workers appeared in one stack group. Used a test-local service instead of adding a global `Ping` test flag.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_pool.cc
bmcnama-shopify: Renamed the accessor and its call sites to `QueueSize()`, and clarified that actively executing calls were excluded.

---
_automated · OpenCode (gpt-5.6)_

src/yb/util/thread.cc
bmcnama-shopify: Added the category to `ThreadIdAndName` and filtered threads before stack collection.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/messenger.cc
bmcnama-shopify: Removed the Messenger's monitor registration before shutting down pools, services, or the acceptor.

---
_automated · OpenCode (gpt-5.6)_

src/yb/util/thread.cc
bmcnama-shopify: Reserved `descriptors_by_id_.size()` before populating the thread list.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Tracked the above-threshold duration independently per service, preventing alternating services from sharing trigger state.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Logged up to ten affected services in deterministic order and included an omitted count. The complete affected set still selected the relevant worker categories.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Kept one sleeping process-wide monitor thread so every runtime flag update path remained reliable. Starting it only when enabled would have made activation callback-dependent and missed flags-file refreshes; making it process-global removed the per-Messenger thread cost.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Added a shared dump ID and completion timing. Kept one log per unique stack group rather than combining them into a potentially truncated message.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Kept `ThreadStacks()` unchanged because cancellation after sending signals could leave late responses that interfere with the next process-wide capture. Moved monitor teardown to the start of shutdown and added cancellation checks before and after collection and between symbolizations; the collector wait remained bounded.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Replaced `CHECK_RESULT` with a rate-limited warning and retried thread creation on later registrations, avoiding both a server crash and silent failure.

---
_automated · OpenCode (gpt-5.6)_

src/yb/util/thread.cc
bmcnama-shopify: Filtered captures to worker categories associated with affected services, with an all-thread fallback when no expected category matched.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Split the logic into `CheckTriggersForDump()` and `MaybeDump()`. Used nonblocking `CoarseBackoffWaiter` with jitter, so trigger evaluation continued while dumps were suppressed.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Replaced the poll count with `rpc_queue_stack_dump_min_interval_ms` and paired it with `rpc_queue_stack_dump_max_interval_ms`. Used the minimum for both sustained-overload detection and the initial repeat interval.

---
_automated · OpenCode (gpt-5.6)_

src/yb/rpc/service_queue_monitor.cc
bmcnama-shopify: Made the monitor and backoff process-global while retaining per-Messenger registrations for safe teardown. Added coverage that registered services with two Messengers and verified one monitor thread.

---
_automated · OpenCode (gpt-5.6)_

bmcnama-shopify:
trigger jenkins

Issue #33201, opened 2026-08-10
[DocDB] Commit hangs after transaction promotion when a participant tablet has no completed batches


ellabaron-code:
Jira Link: [DB-22907](https://yugabyte.atlassian.net/browse/DB-22907)
## Description

**Jira Link:** n/a

### Summary

When a region-local YSQL transaction auto-promotes to global, `COMMIT` can block until the pggate
RPC deadline expires. It happens when one of the involved tablets never completes a write batch —
for example, a write that fails inside a savepoint without aborting the transaction.

`auto_promote_nonlocal_transactions_to_global` defaults to `true`, so no non-default flags are
required; the cluster only needs region-local transaction status tablets (geo-partitioned /
regional tablespaces).

### Root cause

`StartPromotionToGlobal` (`src/yb/client/transaction.cc`) registers **every** involved tablet for a
status-location move, with no filter on whether that tablet has any completed batches:

```c++
for (const auto& tablet_state : tablets_) {
  const auto& participant_tablet = tablet_state.first;
  transaction_status_move_tablets_.emplace(participant_tablet);
  transaction_status_move_handles_.emplace(participant_tablet,
                                           manager_->rpcs().InvalidHandle());
}
```

`DoSendUpdateTransactionPromotingRpcs` then skips tablets with no completed batches, but does not
erase their handle:

```c++
if (tablet_state->second.num_completed_batches == 0) {
  VLOG_WITH_PREFIX(1) << "Tablet " << tablet_id
                      << " has no completed batches, skipping"
                      << " UpdateTransaction(PROMOTING) rpc for now";
  continue;
}
```

The skip is deliberate — sending `UpdateTransaction(PROMOTING)` to a participant that has not
registered the transaction makes `ApplyUpdateTransactionPromoting` return an expired status
(`src/yb/tablet/transaction_participant.cc`), which surfaces to the client as a 40001. The
assumption is that a later successful batch on that tablet will re-trigger the send.

`Commit()` parks whenever the handle map is non-empty:

```c++
if (!transaction_status_move_handles_.empty()) {
  DCHECK(!commit_waiter_);
  VLOG_WITH_PREFIX(1) << "Waiting for transaction move RPCs to finish";
  commit_waiter_ = Waiter(std::bind(
      &Impl::DoCommit, this, deadline, seal_only, _1, transaction));
  return;
}
```

Handles are erased and `commit_waiter_` is consumed only in `UpdateTransactionPromotingDone`, which
[...]

PR #33202, opened 2026-08-10
[#33201] DocDB: Drop pending status moves for participants with no completed batches

ellabaron-code:
## Summary

`StartPromotionToGlobal` registers every tablet in `tablets_` for a status-location move.
`DoSendUpdateTransactionPromotingRpcs` then skips any tablet with no completed batches, but leaves
its entry in `transaction_status_move_handles_`. That entry is only ever erased by
`UpdateTransactionPromotingDone`, which runs when the rpc completes — so for a tablet whose rpc is
never sent, it stays forever, and `Commit()` parks on `commit_waiter_` until the pggate rpc deadline
expires (600 s by default).

A write that fails inside a savepoint reaches commit in exactly this state: the tablet is registered
in `tablets_` with `num_completed_batches == 0`, and `CanAbortTransaction` leaves the transaction
running because the failure carries a non-minimum `subtransaction_id`.

`Commit()` now prunes those tablets from `transaction_status_move_tablets_` and
`transaction_status_move_handles_` before deciding whether to wait.

Pruning rather than sending the rpc anyway: `ApplyUpdateTransactionPromoting` returns an expired
status for a transaction the participant has not registered, which would surface as a spurious
abort. The participant is not part of the commit either — `DoCommit()` drops tablets without
metadata, which is recorded in the same branch that increments `num_completed_batches`. This is safe
at commit time only, since `CheckCouldCommitUnlocked` rejects a commit with running requests, so no
further batch can complete.

The new test uses a fixture that lowers `ysql_client_read_write_timeout_ms`, so that on a regression
the commit fails with an rpc timeout inside the test rather than being killed by the harness timeout
with no assertion output.

## Test plan

- [x] New test `GeoTransactionsPromotionTest.TestCommitWithZeroBatchParticipant` in
  `src/yb/yql/pgwrapper/geo_transactions_promotion-test.cc` — verified it hangs on the unmodified
  tree and passes with the fix:
  ```
  ybd --cxx-test pgwrapper_geo_transactions_promotion-test \
    --gtest_filter GeoTransactionsPromotionTest.TestCommitWithZeroBatchParticipant
  ```
- [ ] Full `pgwrapper_geo_transactions_promotion-test` binary (CI)

### DIFF
--- a/src/yb/client/transaction.cc
+++ b/src/yb/client/transaction.cc
@@ -677,6 +677,7 @@ class YBTransaction::Impl final : public internal::TxnBatcherIf {
         RequestStatusTablet(deadline);
         return;
       }
+      DropPendingStatusMovesWithoutCompletedBatches();
       if (!transaction_status_move_handles_.empty()) {
         DCHECK(!commit_waiter_);
         VLOG_WITH_PREFIX(1) << "Waiting for transaction move RPCs to finish";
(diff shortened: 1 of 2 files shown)


es1024:
Trigger Jenkins

src/yb/client/transaction.cc
```diff
@@ -2328,6 +2329,31 @@ class YBTransaction::Impl final : public internal::TxnBatcherIf {
...
+  // Only safe at commit time: CheckCouldCommitUnlocked() rejects a commit with running requests, so
+  // no further batch can complete. Such a participant is not part of the commit either, since
+  // DoCommit() drops tablets without metadata, which is recorded in the same place as a completed
+  // batch.
+  void DropPendingStatusMovesWithoutCompletedBatches() REQUIRES(mutex_) {
+    for (auto it = transaction_status_move_tablets_.begin();
+         it != transaction_status_move_tablets_.end();) {
+      const auto tablet_state = tablets_.find(*it);
+      CHECK(tablet_state != tablets_.end());
+      if (tablet_state->second.num_completed_batches != 0) {
```
es1024: Would prefer has_metadata check instead to match the check in DoCommit.

ellabaron-code: Fixed.

ellabaron-code:
Fixed your review comment, please rereview.

es1024:
Trigger Jenkins

es1024 (approved):

Issue #33068, opened 2026-08-03
[DocDB] ThreadSafeObjectPool sizes pools_ by CPU count but indexes it by sched_getcpu(), causing an out-of-bounds read


acudovs:
Jira Link: [DB-22787](https://yugabyte.atlassian.net/browse/DB-22787)
### Description

`ThreadSafeObjectPool` allocates one pool per CPU using `base::RawNumCPUs()`, then indexes that array with the raw return value of `sched_getcpu()`. These two numbers come from different spaces: the first is a count of online CPUs, the second is an absolute CPU id. Whenever the number of online CPUs is smaller than the highest CPU id a thread can be scheduled on, `Take()` and `Release()` read past the end of `pools_` and the process dies with SIGSEGV.

## Affected code

All references are against master at `ab23d4f1cb811946e1ecd7077a22f3cd4cde3290`.

Sizing, in `src/yb/util/object_pool.h:206`:

```cpp
// Need the actual number of CPUs, so we do not use the Gflag value
size_t num_cpus = base::RawNumCPUs();
pools_.reserve(num_cpus);
```

Indexing, in `src/yb/util/object_pool.h:245`:

```cpp
size_t cpu = sched_getcpu();
DCHECK_LT(cpu, pools_.size());
return cpu;
```

`DCHECK_LT` is compiled out in release builds, so nothing catches the bad index.

## Why this is a bug

`base::RawNumCPUs()` returns `cpuinfo_num_cpus`, which on Linux is set from `std::thread::hardware_concurrency()` in `src/yb/gutil/sysinfo.cc:408`. Both standard library implementations resolve that to the count of online CPUs: libc++ calls `sysconf(_SC_NPROCESSORS_ONLN)` and libstdc++ calls `get_nprocs()`. In glibc both read `/sys/devices/system/cpu/online`. So `RawNumCPUs()` is a count, and it says nothing about which CPU ids exist.

`src/yb/gutil/sysinfo.h:56` already documents this:

```
// Note that, if not all CPUs are online, this may return a value lower than the maximum
// value of sched_getcpu().
```

and the header goes on to provide the correct primitive at line 64:

```
// Return the maximum CPU index that may be returned by sched_getcpu(). For example, on
// an 8-core machine, this will return '7' even if some of the CPUs have been disabled.
extern int MaxCPUIndex();
```

[...]

Issue #9619, opened 2021-08-06
percpu_rwlock assert sched_getcpu() within the range of visible CPUs is broken on OpenVZ


FranckPachot:
Jira Link: [DB-1453](https://yugabyte.atlassian.net/browse/DB-1453)
In https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/util/locks.h#L201 we have the following assert: `CHECK_LT(cpu, n_cpus_);` that checks that the processor we are scheduled on is within the range visible from `/sys/devices/system/cpu/present`

This, on some hypervisors like OpenVZ, fails because sched_getcpu() maps all host processors, but /sys exposes only the visible range. All details here: https://dev.to/yugabyte/visible-number-of-cpu-on-openvz-2h7d

bmatican:
I believe we've seen something like this in robin.io as well, but I think the checks are in place because the locks implementation rely on being able to explicitly use a range of cpus.

I think some of this functionality is ported over from chromium, not sure if there's something upstream we could bring in to cover for this here.

Issue #5057, opened 2020-07-13
Reuse CQL parser between processors


spolitov:
When CQLProcessor is created it consumes ~16kb of memory.
Most of this memory ~14kb is consumed by Parser.
The parser is required for a very short period of time in just a single function.
So we could share parsers between CQLProcessors.

PR #33203, opened 2026-08-10
[#33068] DocDB: Fix random tserver SIGSEGV from out-of-bounds per-CPU pool indexing

acudovs:
## Summary

Issue: https://github.com/yugabyte/yugabyte-db/issues/33068

`ThreadSafeObjectPool` keeps one lock-free stack per CPU. It sizes that array from `base::RawNumCPUs()`, a count of online CPUs, and then indexes it with `sched_getcpu()`, an absolute CPU id. Those are different number spaces. Whenever a thread can be scheduled on a CPU whose id is greater than or equal to the number of online CPUs, `Take()` and `Release()` index past the end of `pools_` and the process dies with SIGSEGV. `DCHECK_LT` is compiled out in release builds, so nothing catches it.

The two numbers agree only when the online CPUs form a contiguous range starting at 0, which is the usual case and why this has gone unnoticed. They diverge in two situations:

- Some CPUs are offline. `/sys/devices/system/cpu/online` loses entries, so a thread can run on a CPU id above the online count. This is the case `MaxCPUIndex()` already documents in `src/yb/gutil/sysinfo.h:64`: "on an 8-core machine, this will return '7' even if some of the CPUs have been disabled".
- Something makes `/sys/devices/system/cpu/online` container aware. `lxcfs` has done this since 3.1.2. With a cpuset of `32-36` it reports 5 online CPUs while threads keep running on host CPU ids 32 through 36, and under `--enable-cfs` it instead renumbers the range to `0-4` from the CFS quota. Either way the count is 5 and `sched_getcpu()` returns up to 36.

Every user of the pool is exposed:

- `src/yb/util/trace.cc:293`, the arena pool behind tracing
- `src/yb/yql/cql/ql/ql_processor.cc:157` and `src/yb/yql/cql/cqlserver/cql_service.h:289`, the CQL parser pools

### Fix

Size `pools_` by `base::MaxCPUIndex() + 1`. `MaxCPUIndex()` reads `/sys/devices/system/cpu/present` and is the documented bound for anything indexed by `sched_getcpu()`. This matches the two per-CPU structures that already use it: `PerCpuRwMutex` (`src/yb/util/locks.h:179`) and `ClockboundClock` (`src/yb/server/clockbound_clock.cc:189`).

`GetCPU()` also converted a failed `sched_getcpu()` (it returns `int` and can return `-1`) into `SIZE_MAX` through `size_t cpu = sched_getcpu()`, an out of bounds index regardless of how `pools_` is sized. The negative case now falls back to the thread-id hash that macOS already used, extracted as `PoolIndexForThread()`.

`MaxCPUIndex()` is cached under `call_once`, so a CPU that comes online later would still fall outside `pools_`, guarded only by a `DCHECK` that release builds drop. `GetCPU()` therefore wraps with `% pools_.size()` and never indexes raw, following `ClockboundClock::Now()`, which does the same.

This is the same class of bug as #9619, where `PerCpuRwMutex` assumed `sched_getcpu()` stayed inside the online CPU range. `ThreadSafeObjectPool` was the last per-CPU structure in the tree sizing by a count while indexing by CPU id.

The remaining caller of `RawNumCPUs()` is `spinlock.cc`, so its doc comment in `sysinfo.h` no longer points readers at `object_pool.h` for "the actual number of CPUs".

## Test plan

- [x] `build-support/lint.sh --rev origin/master` reports 0 errors, 0 warnings
- [x] `object_pool-test`, covering the existing cases plus the new `TestThreadSafeObjectPool.TakeAndReleaseOnEveryCpu`

The new test pins the calling thread to each CPU in its affinity mask in turn, takes and releases an object from each, and restores the original mask on exit. It also asserts that pool destruction frees every pooled object.

Scope worth stating plainly: on a machine where `/sys/devices/system/cpu/online` matches `/sys/devices/system/cpu/present`, every CPU id is a valid index either way, so the test passes with and without the fix. Reproducing the SIGSEGV needs an environment where the online CPU count is lower than the highest CPU id a thread can be scheduled on, which is what LXC containers with a CPU quota and a wide cpuset produce.

### DIFF
--- a/src/yb/gutil/sysinfo.h
+++ b/src/yb/gutil/sysinfo.h
@@ -57,8 +57,8 @@ namespace base {
 // value of sched_getcpu().
 extern int NumCPUs();
 
-// Use the default value for when gflags are not initialized yet (i.e. spinlock.cc)
-// or if we need the actual number of CPUs (i.e. object_pool.h)
+// Return the number of online CPUs. Like NumCPUs(), but ignoring the --num_cpus flag override.
+// Use when gflags are not initialized yet (i.e. spinlock.cc).
 extern int RawNumCPUs();
 
 // Return the maximum CPU index that may be returned by sched_getcpu(). For example, on
--- a/src/yb/util/object_pool-test.cc
+++ b/src/yb/util/object_pool-test.cc
@@ -30,9 +30,16 @@
 // under the License.
 //
 
+#ifdef __linux__
+#include <sched.h>
+#endif
+
 #include "yb/util/logging.h"
 #include <gtest/gtest.h>
 
+#include "yb/gutil/sysinfo.h"
+
+#include "yb/util/errno.h"
 #include "yb/util/object_pool.h"
 
 namespace yb {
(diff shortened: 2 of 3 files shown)


acudovs:
@spolitov you added `ThreadSafeObjectPool` in #5057, so you are likely the right person for this one. Apologies for the direct ping, the PR has had no reviewer assigned since it was opened.

The short version: the pool sizes `pools_` from `base::RawNumCPUs()`, a count of online CPUs, then indexes it with `sched_getcpu()`, an absolute CPU id. When those two diverge, `Take()` and `Release()` read past the end of the array, and `DCHECK_LT` is compiled out in release builds, so it surfaces as a random SIGSEGV rather than a check failure. This shows up on tservers running in containers where a CPU quota shrinks the online count while the cpuset still spans high host CPU ids. Every user of the pool is exposed.

One design question I would love to get your opinion on. Sizing by `MaxCPUIndex() + 1` closes the gap in practice, but the value is cached under `call_once`, so a CPU that comes online later still falls outside `pools_`, and the only thing guarding that is a `DCHECK`. The tree already answers this twice, in opposite ways:

- `PerCpuRwMutex::get_lock()` is strict. `LOG_IF(FATAL, cpu < 0)` followed by `CHECK_LT(cpu, n_cpus_)`, so a bad index is a loud crash instead of a silent read.
- `ClockboundClock::Now()` is permissive. It falls back to a thread id hash when `sched_getcpu()` fails, and it never indexes raw, applying `% num_ctxs` on every access.

This PR currently sits between the two. It follows the `ClockboundClock` fallback for the negative case and keeps the existing `DCHECK` for the range. I am happy to move it either way, strict like `PerCpuRwMutex` or permissive like `ClockboundClock`, as a follow-up commit. Which do you prefer for this class?

Is there something I should do on my side to move this forward, or does that need to happen internally?

spolitov:
I suppose we could be permissive as `ClockboundClock` here.

acudovs:
Thanks, pushed as `91f0e39`. `GetCPU()` wraps with `% pools_.size()` and no longer asserts the range, so it never indexes raw, same as `ClockboundClock::Now()`. The negative `sched_getcpu()` case keeps the thread id hash fallback. Summary updated to describe it.

The branch is rebased on current master, lint is clean, and `object_pool-test` passes.

PR #33205, opened 2026-08-10
[DOC-1438] Update YBA Python requirement

ddhodge:
YBA v2026.1 supports Python 3.10-3.13


nkhogen (approved):
Thanks for the quick turnaround. LGTM.

Issue #33207, opened 2026-08-10
[DocDB] Index backfill is never resumed if a tablet report clears the ALTERING state mid-backfill


ellabaron-code:
Jira Link: [DB-22909](https://yugabyte.atlassian.net/browse/DB-22909)
## Description

An index backfill can be permanently stranded if a tablet reports its schema version to the master while the backfill is still running and the master leader subsequently changes. The index never becomes usable, and the indexed table has already been marked as if the alter completed.

### Mechanism

`MultiStageAlterTable::StartBackfillingData` deliberately leaves the indexed table persisted as `ALTERING` for the duration of the backfill, records the work in the table's `backfill_jobs`, and tracks progress only through the in-memory `is_backfilling_` flag.

If a tablet of the indexed table sends a schema-version report during that window, `CatalogManager::HandleTabletSchemaVersionReport` re-enters `MultiStageAlterTable::LaunchNextTableInfoVersionIfNecessary`. For a YSQL table at `INDEX_PERM_DO_BACKFILL`, `indexes_to_update` and `indexes_to_delete` are both empty, and `is_backfilling` short-circuits the remaining check, so this guard is taken:

```cpp
  if (indexes_to_update.empty() &&
      indexes_to_delete.empty() &&
      (is_backfilling || indexes_to_backfill.empty())) {
    return ClearFullyAppliedAndUpdateState(
        catalog_manager, indexed_table, current_version, /* change state to RUNNING */ true, epoch);
  }
```

Declining to launch another backfill is correct — one is already running — but the branch also persists `RUNNING`, which is not.

Nothing goes visibly wrong until the master leader changes. The new leader has `is_backfilling_ == false` and reads a table whose state is `RUNNING`, so the recovery path that rebuilds the index list from `backfill_jobs` (guarded on `!is_backfilling`) never applies. The backfill is never resumed and the index stays at `INDEX_PERM_DO_BACKFILL`.

### Steps to reproduce

1. Start a cluster and create a table with a single tablet.
2. Issue `CREATE INDEX` and let it reach the backfill stage.
3. While the backfill is running, cause the table's tablet to send a schema-version report — a tablet leader election is the simplest trigger, since `RaftConsensus::SetLeaderUuidUnlocked` marks the tablet dirty and the new leader includes its schema version in the next heartbeat's tablet report. A tserver restart or a periodic full tablet report does the same.
[...]

PR #33208, opened 2026-08-10
[#33207] DocDB: Resume a lost index backfill after the indexed table leaves ALTERING

ellabaron-code:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/33208/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

## Summary

`MultiStageAlterTable::StartBackfillingData` leaves the indexed table persisted as `ALTERING` for the duration of an index backfill, records the work in the table's `backfill_jobs`, and tracks progress only through the in-memory `is_backfilling_` flag.

A tablet schema-version report arriving during that window re-enters `LaunchNextTableInfoVersionIfNecessary`. For a YSQL table at `INDEX_PERM_DO_BACKFILL` both `indexes_to_update` and `indexes_to_delete` are empty and `is_backfilling` short-circuits the remaining check, so the "not necessary to launch next version" guard is taken and `ClearFullyAppliedAndUpdateState(..., /* change state to RUNNING */ true, ...)` persists `RUNNING` while the backfill is still running.

Nothing goes visibly wrong until the master leader changes. The new leader has `is_backfilling_ == false` and reads a table in state `RUNNING`, so the recovery path that rebuilds `indexes_to_backfill` from `backfill_jobs` — guarded on `!is_backfilling` — never applies. The backfill is never resumed and the index is left at `INDEX_PERM_DO_BACKFILL`. The window lasts as long as the backfill itself, so it is wide on a large table, and any tablet leader election, tserver restart, or periodic full tablet report will produce the report that triggers it.

Persisting `RUNNING` mid-backfill is not itself the defect: concurrent DDL on the indexed table clears `ALTERING` the same way, so table state cannot be relied on to mean "a backfill is outstanding". What is missing is a resume path that does not depend on it. `HandleTabletSchemaVersionReport` early-returns unless the table is `ALTERING`, and it is the only entry point to the resume logic, so once the state is cleared the backfill becomes unreachable.

The fix relaxes that gate on durable evidence instead: the function also proceeds when the table has a recorded backfill job and no backfill running (`backfill_jobs_size() > 0 && !IsBackfilling()`), which is exactly the condition the recovery path in `LaunchNextTableInfoVersionIfNecessary` already looks for. The backfill is then relaunched from the persisted job regardless of whether the table is `ALTERING` or `RUNNING`. Keeping the resume behind this one function means PITR restores and post-failover resumption continue to share a single code path, including the `IsTableUndergoingPitrRestore` check and the DDL-verification cleanup.

Resumption otherwise depends on a tablet report arriving after the election, which is not guaranteed. `SysCatalogLoaded` therefore queues every table in that state with `CatalogManagerBgTasks` (`EnqueuePendingBackfillsAfterLoad`), reusing the `pending_backfill_tables_` queue that PITR already feeds, so a newly elected leader picks the job up on its own.

## Upgrade/Rollback safety

No wire-format, catalog-schema, on-disk-format, or gflag-default change. The fix only adds a resume path over state that is already persisted in `backfill_jobs`.

Mixed-version clusters are safe in both directions: masters do not coordinate on this path, and only the leader acts on it. During a rolling upgrade an unfixed master leader still fails to resume a backfill whose table has left `ALTERING`; a fixed leader resumes it, including for indexes stranded before the upgrade, since the decision is made from the persisted job.

Rollback is safe: reverting restores the previous behavior with no state to migrate. An index stranded while the unfixed code is in place is resumed as soon as a fixed master holds leadership.

## Test plan

New regression test `PgIndexBackfillMultiMaster.BackfillResumesAfterMasterFailover`: it holds the backfill with `TEST_block_do_backfill`, steps down the indexed table's tablet leader so the new leader reports its schema version to the master, waits for the indexed table to leave `ALTERING`, then fails the master leader over and unblocks the backfill. Without the fix the index never becomes valid.

- [x] `./yb_build.sh release --cxx-test pg_index_backfill-test --gtest_filter 'PgIndexBackfillMultiMaster.BackfillResumesAfterMasterFailover/*'` — fails on `/0` and `/1` before the change, passes on both after
- [x] `./yb_build.sh release --cxx-test pg_index_backfill-test --gtest_filter 'PgIndexBackfillMultiMaster.MasterLeaderStepdown/*'` — the existing master-failover coverage, passes on `/0` and `/1`

### DIFF
--- a/src/yb/master/backfill_index.cc
+++ b/src/yb/master/backfill_index.cc
@@ -335,7 +335,7 @@ Status MultiStageAlterTable::StartBackfillingData(
     const std::vector<IndexInfoPB>& idx_infos,
     std::optional<uint32_t> current_version, const LeaderEpoch& epoch,
     std::optional<TransactionMetadata> requester_transaction) {
-  // We leave the table state as ALTERING so that a master failover can resume the backfill.
+  // Stay in ALTERING: IsAlterTableDone must not report done while the backfill is running.
   RETURN_NOT_OK(ClearFullyAppliedAndUpdateState(
       catalog_manager, indexed_table, current_version, /* change_state to RUNNING */ false, epoch));
 
--- a/src/yb/master/catalog_manager.cc
+++ b/src/yb/master/catalog_manager.cc
@@ -12054,7 +12054,9 @@ Status CatalogManager::HandleTabletSchemaVersionReport(
   // Verify if it's the last tablet report, and the alter completed.
   {
     auto l = table->LockForRead();
-    if (l->pb.state() != SysTablesEntryPB::ALTERING) {
+    // A recorded backfill job with no backfill running means the backfill was lost.
+    const bool resume_backfill = l->pb.backfill_jobs_size() > 0 && !table->IsBackfilling();
+    if (l->pb.state() != SysTablesEntryPB::ALTERING && !resume_backfill) {
       VLOG_WITH_PREFIX_AND_FUNC(2) << "Table " << table->ToString() << " is not altering";
       return Status::OK();
     }
@@ -14052,6 +14054,7 @@ void CatalogManager::SysCatalogLoaded(SysCatalogLoadingState&& state) {
       "Failed to backfill plugin name for notifications CDC streams");
 
   SchedulePostTabletCreationTasksForPendingTables(state.epoch);
+  EnqueuePendingBackfillsAfterLoad();
   restoring_sys_catalog_ = false;
 
   if (FLAGS_enable_ysql) {
@@ -14275,6 +14278,25 @@ void CatalogManager::SchedulePostTabletCreationTasksForPendingTables(const Leade
       continue;
     }
     SchedulePostTabletCreationTasks(*table_info_result, epoch);
+  }
+}
+
+void CatalogManager::EnqueuePendingBackfillsAfterLoad() {
+  std::vector<TableInfoPtr> tables;
+  {
+    SharedLock lock(mutex_);
+    tables.reserve(tables_->Size());
+    for (const auto& table_info : tables_->GetAllTables()) {
+      tables.push_back(table_info);
+    }
+  }
+
+  for (const auto& table_info : tables) {
+    if (table_info->LockForRead()->pb.backfill_jobs_size() == 0 || table_info->IsBackfilling()) {
+      continue;
+    }
+    LOG(INFO) << "Queueing " << table_info->ToString() << " for backfill resumption";
+    AddPendingBackFill(table_info->id());
   }
 }
 
--- a/src/yb/master/catalog_manager.h
+++ b/src/yb/master/catalog_manager.h
@@ -3286,6 +3286,10 @@ class CatalogManager : public CatalogManagerIf, public SnapshotCoordinatorContex
 
   void SchedulePostTabletCreationTasksForPendingTables(const LeaderEpoch& epoch) EXCLUDES(mutex_);
 
+  // Queues every table that has a backfill job recorded but no backfill running for resumption by
+  // CatalogManagerBgTasks.
+  void EnqueuePendingBackfillsAfterLoad() EXCLUDES(mutex_);
+
   Status BumpVersionAndStoreClusterConfig(
       ClusterConfigInfo* cluster_config, ClusterConfigInfo::WriteLock* l);
 
--- a/src/yb/yql/pgwrapper/pg_index_backfill-test.cc
+++ b/src/yb/yql/pgwrapper/pg_index_backfill-test.cc
@@ -2261,6 +2261,81 @@ TEST_P(PgIndexBackfillMultiMaster, MasterLeaderStepdown) {
   thread_holder_.JoinAll();
 }
 
+// Make sure that a tablet schema version report arriving during backfill does not strand the
+// backfill across a master leader change.  The report moves the indexed table out of ALTERING, and
+// the new master leader has to resume the backfill anyway.  Simulate the following:
+//   Session A                                    Session B
+//   --------------------------                   ----------------------
+//   CREATE INDEX
+//   - indislive
+//   - indisready
+//   - backfill
+//     - get safe time for read
+//                                                tablet leader stepdown
+//                                                - new leader reports schema version
+//                                                master leader stepdown
+TEST_P(PgIndexBackfillMultiMaster, BackfillResumesAfterMasterFailover) {
+  auto client = ASSERT_RESULT(cluster_->CreateClient());
+
+  // 1. Create a single-tablet table and start CREATE INDEX on a separate thread.
+  ASSERT_OK(conn_->ExecuteFormat(
+      "CREATE TABLE $0 (i int, PRIMARY KEY (i)) SPLIT INTO 1 TABLETS", kTableName));
+  ASSERT_OK(conn_->ExecuteFormat("INSERT INTO $0 VALUES (1)", kTableName));
+
+  const auto table_id = ASSERT_RESULT(
+      GetTableIdByTableName(client.get(), kDatabaseName, kTableName));
+
+  // conn_ should be used by at most one thread for thread safety.
+  thread_holder_.AddThreadFunctor([this] {
+    LOG(INFO) << "Begin create thread";
+    PGConn create_conn = ASSERT_RESULT(ConnectToDB(kDatabaseName));
+    ASSERT_OK(create_conn.ExecuteFormat("CREATE INDEX $0 ON $1 (i ASC)", kIndexName, kTableName));
+    LOG(INFO) << "Done create thread";
+  });
+
+  // 2. Wait for backfill safe time, at which point the backfill is blocked by
+  //    TEST_block_do_backfill.
+  ASSERT_OK(WaitForBackfillSafeTime(kYBTableName));
+
+  // 3. Step down the tablet leader so the new leader reports its schema version to the master
+  //    while the backfill is in progress.
+  google::protobuf::RepeatedPtrField<master::TabletLocationsPB> tablets;
+  ASSERT_OK(client->GetTabletsFromTableId(table_id, 0, &tablets));
+  ASSERT_EQ(tablets.size(), 1);
+  const auto tablet_id = tablets[0].tablet_id();
+  const auto old_leader_idx = ASSERT_RESULT(cluster_->GetTabletLeaderIndex(tablet_id));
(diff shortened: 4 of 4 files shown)


amitanandaiyer:
Trigger Jenkins

amitanandaiyer (commented):
Thanks Ella. The fix makes sense.

Testing wise, would be good to also add a test to invoke the master-failover code path; that the GH talks about. Thanks

amitanandaiyer:
> **AI-generated review.** This comment was produced by Claude Opus 5 (Anthropic), running in Claude Code, and posted by the account owner. All code references were checked against `master` at the time of writing, but the analysis should be verified independently before acting on it.

The change was reviewed against master. The diagnosis in #33207 holds up — every link in the chain checks out:

- `StartBackfillingData` deliberately leaves the table `ALTERING` (`backfill_index.cc:330`) with `backfill_jobs` as the durable record.
- The resume branch is guarded on `!is_backfilling && backfill_jobs_size() > 0` (`backfill_index.cc:458`).
- The only way to reach it after a failover is `HandleTabletSchemaVersionReport`, which bails out unless the state is `ALTERING` (`catalog_manager.cc:12001`).

So persisting `RUNNING` mid-backfill doesn't just look wrong, it permanently disables the only entry point to recovery. Good find, and worth fixing.

That said, **the current patch appears to be the wrong lever, and the CSI run shows it.**

## The CSI failures are caused by this change

The whole `PgDdlAtomicityStressTest` family regressed on `alma8-clang21-tsan`, not just the five flagged as new consistent failures:

| Variant | This PR | Baseline |
|---|---|---|
| `Main/colocated` | 11 fail / 0 pass | 2/35 |
| `Main/colocated_ddl_rollback_failure_probability` | 11 / 0 | 0/36 |
| `Main/colocated_ddl_transaction_verification_failure_probability` | 11 / 0 | 1/36 |
| `Main/colocated_ddl_verification_failure_probability` | 11 / 0 | 0/35 |
| `Main/colocated_fail_probability_of_catalog_writes_by_ddl_verification` | 11 / 0 | 1/36 |
| `Main/ddl_rollback_failure_probability` | 10 / 1 | 0/36 |
| `Main/ddl_verification_failure_probability` | 9 / 2 | 0/36 |
| `Main/fail_probability_of_catalog_writes_by_ddl_verification` | 10 / 1 | 0/35 |
| `Main/none` | 9 / 2 | 2/34 |

Nine independent parameterizations going from ~0/36 to ~10-of-11 failing, tagged `timeout` / `SIGQUIT`. The colocated ones only stand out because they hit zero passes. And it is exactly the test that runs concurrent `CREATE INDEX` alongside `ALTER TABLE` on the same table — `TestConcurrentIndex` (`pg_ddl_atomicity_stress-test.cc:258`, driven from `:369`).

To reproduce locally without CSI access:

```
# the family that regressed (TSAN, matches the failing CSI build)
./yb_build.sh tsan --cxx-test pg_ddl_atomicity_stress-test \
    --gtest_filter '*PgDdlAtomicityStressTest.Main/colocated*'

# the non-colocated variants regressed too
./yb_build.sh tsan --cxx-test pg_ddl_atomicity_stress-test \
    --gtest_filter '*PgDdlAtomicityStressTest.Main/none*'

# whole binary
./yb_build.sh tsan --cxx-test pg_ddl_atomicity_stress-test
```

These are slow and were flagged as timeouts, so each needs several runs — baseline is ~0 failures in 36 runs, so even 2-3 failures in 10 is signal. The mechanism below is not TSAN-specific; TSAN just widens the backfill window enough to make the overlap near-certain.

## Why it breaks

The guard being short-circuited is not only the backfill's exit. It is **the only place a YSQL table's `ALTERING` -> `RUNNING` transition happens at the end of _any_ alter.**

Trace a plain `ALTER TABLE ADD COLUMN`: state set to `ALTERING` + version bump (`catalog_manager.cc:8376`) -> `SendAlterTableRequest` -> tablets apply -> `HandleTabletSchemaVersionReport` -> `LaunchNextTableInfoVersionIfNecessary`. For a YSQL table on this path `indexes_to_update` is empty (the loop only fills it when `!is_ysql_table` or `update_ysql_to_backfill`, and reports pass neither) and `indexes_to_delete` is empty. So the guard is taken and `RUNNING` is persisted. That *is* how every YSQL `ALTER TABLE` finishes.

Now add an index backfill in flight on the same table: `is_backfilling` is true, the new early return fires, and that unrelated alter's completion is discarded. The table never leaves `ALTERING`, so the client's alterer keeps polling `IsAlterTableDone` — which is literally `done = state != ALTERING` (`catalog_manager.cc:8526`, waited on at `client/table_alterer.cc:123`) — until the backfill finishes or the deadline expires. A workload issuing DDL against a table that is being indexed now stalls every one of those DDLs for the full length of the backfill.

Note that the backfill's *own* completion is safe: `ClearIsBackfilling()` (`backfill_index.cc:1280`) runs before `SendAlterTableRequest` (`:1284`), so the post-backfill reports arrive with `is_backfilling == false` and still reach the `RUNNING` transition. It is the *other* alters landing inside the window that get swallowed.

## Suggested direction

The root problem is that `ALTERING` is doing double duty: "an alter is in flight, clients keep waiting" and "a backfill is owed, a new leader must resume it". This patch resolves the conflict in favor of the backfill and breaks the client contract.

The durable marker for the second meaning already exists — it is `backfill_jobs`. The bug is that the *reader* is gated on table state (`catalog_manager.cc:12001`), so a table in `RUNNING` with a live `backfill_jobs` entry is unreachable. The fix belongs on that side, leaving the state machine alone.

There is already scaffolding for it: `AddPendingBackFill()` (`catalog_manager.h:1818`) feeds `pending_backfill_tables_`, drained every bg-task iteration by `CatalogManagerBgTasks::TryResumeBackfillForTables` (`catalog_manager_bg_tasks.cc:172`, wired at `:273`). Today it is only fed by PITR restore. Two changes make it a general recovery path:

1. On leader startup (after `RunLoaders`), enqueue every table with `backfill_jobs_size() > 0 && !IsBackfilling()`.
2. Have that path call `MultiStageAlterTable::LaunchNextTableInfoVersionIfNecessary` **directly** rather than going through `HandleTabletSchemaVersionReport`, which otherwise re-imposes the same `ALTERING` gate. The recovery branch at `backfill_index.cc:458` already rebuilds `indexes_to_backfill` from the job.

That recovers stranded backfills both in the reported scenario and in any case where the table legitimately reached `RUNNING`, without changing when alters complete.

If the current shape is preferred, the early return needs narrowing to reports for the backfill's *own* alter — e.g. skip only while the table's version still matches the version recorded when the backfill job was created, and fall through to the normal `RUNNING` transition once a later alter has bumped past it. Fiddlier, and `ALTERING` stays overloaded.

## A related gap worth closing either way

`is_backfilling_` is set in one place (`backfill_index.cc:342`) and cleared in exactly two (`:366`, `:1280`). Between `:1271` and `:1280` there are four early returns that leave it set: the `backfill_jobs_size() == 0` return at `:1244`, and failures of `AllowCompactionsToGCDeleteMarkers` (`:1267`), `UpdateIndexPermission` (`:1271`), and `ClearCheckpointStateInTablets` (`:1279`). `TEST_skip_index_backfill` (`:355`) leaks it by design.

The leak is bounded — `RunLoaders` clears the table map (`catalog_manager.cc:1592`) and rebuilds fresh `TableInfo` objects via `NewTableInfo` (`catalog_loaders.cc:85`), so the flag resets on every leadership acquisition — but the current patch makes correctness depend on that flag much more heavily than it does today. A scope guard that clears it on every exit of `UpdateIndexPermissionsForIndexes` would be worth having regardless of which fix lands.

## On the new test

Two mechanical notes:

- The `SleepFor(5s)` allows a false pass: if the report never reaches the master the test still passes, and would keep passing if someone later broke the trigger. A deterministic signal is available — the master learns the new tablet leader from `ProcessCommittedConsensusState` in the *same* report whose third pass calls `HandleTabletSchemaVersionReport` (`master_heartbeat_service.cc:1058-1069`). Polling `GetTabletsFromTableId` until the master reports the new leader proves the report was processed, and drops 5s of wall clock.
- `leader_stepdown` without a `dest_ts_uuid` does not guarantee a different node wins; the old leader can be re-elected and the 30s `WaitFor` then hard-fails. Passing an explicit destination would make it deterministic.

More substantively: the test asserts the *invariant* (still `ALTERING`), not the *bug* (backfill never resumes). It would be better written as the failover test — hold the backfill, trigger the tablet report, step down the master leader, clear `TEST_block_do_backfill` on all masters, then assert `CREATE INDEX` completes and the index is valid. That fails on the timeout without a fix, and it is the version that survives either fix design; asserting "stays ALTERING" would be wrong under the decoupled approach above. There is currently no coverage of resume-after-failover at all, so that test is worth having on its own.

## Unrelated in the same run

`PgWaitQueuesProbeForwardingTest.ForwardProbeDuringDetectorShutdown` is 11/0 with a `tsan race` tag in a different subsystem, and the single-failure entries (`BackupTxnTest`, `MasterPathHandlersItest.TestClusterBalancerWarnings`, `MasterTxnStatusCheck.NoActionOnScaleDown`, `TestPgRegressAggregates`) look independent of this change — worth confirming against master before attributing them here.

ellabaron-code:
@amitanandaiyer 
Thank you for detailed write up. I will investigate.

ellabaron-code:
@amitanandaiyer I think I've fixed it. Please, rereview.

ellabaron-code:
I added a new test as part of this commit: `PgIndexBackfillMultiMaster.BackfillResumesAfterMasterFailover`. It blocks the backfill, gets the indexed table out of `ALTERING` via a tablet schema-version report, then fails the master leader over before unblocking it, so `CREATE INDEX` returns only if the new leader resumes the job from `backfill_jobs`. It times out without the fix and passes on both parameter instances with it. The existing `MasterLeaderStepdown` case still passes as well.

ellabaron-code:
trigger jenkins

amitanandaiyer:
Trigger Jenkins

Schitiz:
Trigger Jenkins

amitanandaiyer:
the tests look good.

amitanandaiyer (approved):
The tests look good. 

Given that we have a cleaner way to resume a backfill after a failover, we could perhaps get rid of the comment/mechanism that leaves it in an ALTERING state. Could be done as a later follow up/improvement.

ellabaron-code:
Fixed a comment.

amitanandaiyer:
Screenshot 2026-08-28 at 3 13 46 PM
CSI didn't clear automatically. But the tests look good. Merging.

PR #33244, opened 2026-08-11
[#33243] DocDB: Stripe the retryable request tracker to reduce request ID lock contention

### LINKED ISSUE #33243: [DocDB] YBClient retryable-request ID generation serializes on a single tracker lock at high connection counts
Jira Link: [DB-22939](https://yugabyte.atlassian.net/browse/DB-22939)
## Description

On hosts with high core counts, `YBClient` write throughput is limited by lock contention on the retryable-request ID path. Every operation acquires a single per-client lock twice — once in `YBClient::NextRequestIdAndMinRunningRequestId` to allocate a request ID and register it in the running-requests set, and once in `RequestsFinished` to retire it. At high connection counts all client threads funnel through this one lock.

Measured on a 3-node cluster of large (hundreds of vCPUs) nodes running a single-row INSERT workload at ~3,500-5,000 client connections:

- Lock-contention profiling (`/pprof/contention`) attributed ~90% of total lock wait time in the tserver's embedded YSQL client to this single spinlock pair.
- Reducing the lock hold time and then striping the tracker (independent lock + ID space per stripe, lock-free monotone lower bound per stripe for the `min_running_request_id` watermark) removed the register/unregister path from the contention profile (~2% residual) and improved saturation throughput measurably, as part of a series of changes that raised single-row write throughput ~15% on the same hardware.

The watermark remains conservative under striping: each stripe publishes a monotone lower bound (front of its running list when non-empty, otherwise its next unassigned ID), and the client-wide minimum across stripes may lag but can never overshoot, so server-side dedup-state GC based on `min_running_request_id` stays correct.

Issue Type: kind/enhancement

- [x] Please confirm this issue does not contain any sensitive information

[DB-22939]: https://yugabyte.atlassian.net/browse/DB-22939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

craigsoules:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/33244/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

## Summary

At high connection counts on many-core hosts, every YBClient operation crosses one per-client lock twice: `YBClient::NextRequestIdAndMinRunningRequestId` takes it to allocate a request ID and register it in the running-requests set, and `RequestsFinished` takes it again to retire the ID. Lock-contention profiling on a 3-node cluster of large (hundreds of vCPUs) nodes under a single-row INSERT workload at 3,500-5,000 connections attributed ~90% of the embedded YSQL client's total lock wait time to this pair.

Two commits:

1. **Reduce work under the lock** on the ID-generation/retirement path (narrows the critical section without changing the data structure).
2. **Stripe the RetryableRequestTracker.** The tracker is partitioned into independently locked stripes; stripe `s` of `S` assigns request IDs `s, s+S, s+2S, ...` under its own lock, keeping ID assignment and active-list insertion atomic per stripe. Each stripe publishes a monotone lower bound (the front of its running list when non-empty, otherwise its next unassigned ID) in a lock-free atomic; `min_running_request_id` is the minimum across stripes. The watermark may lag but can never overshoot, regardless of read interleaving, so server-side dedup-state GC driven by the watermark remains correct. Stripes are chosen by a per-thread round-robin counter so every stripe keeps advancing on an active client. The stripe count comes from a new non-runtime flag `--client_retryable_request_tracker_stripes` (default 16, rounded up to a power of two; default preserves enough striping to remove the hot lock while keeping per-stripe ID-space fragmentation negligible).

With both changes (measured with 64 stripes), the register/unregister path dropped from ~90% of client lock wait to ~2%, contributing to a ~15% single-row write throughput improvement on the same hardware in combination with reactor/connection tuning.

Fixes #33243

## Test plan

- [x] `client-unittest` passes (12/12), including five new/extended tracker tests:
  - `RetryableRequestTrackerMaintainsMinimumActiveRequestId`
  - `RetryableRequestTrackerTransfersMovedRegistration`
  - `RetryableRequestTrackerStripeCountRoundsUpToPowerOfTwo`
  - `RetryableRequestTrackerStripedIdsAreUniqueAndWatermarkConservative`
  - `RetryableRequestTrackerWatermarkAdvancesAfterRetirement`
- [x] Built and tested on both x86_64/clang19 (release, full LTO) and aarch64/clang21 (release).
- [x] Backport of the same changes ran a multi-week write-scaling benchmark campaign on a 3-node cluster (RF3, single-row INSERT, up to 10,000 connections) with no consensus/dedup errors; contention profiles confirm the lock is no longer hot.

### DIFF
--- a/src/yb/client/CMakeLists.txt
+++ b/src/yb/client/CMakeLists.txt
@@ -48,6 +48,7 @@ set(CLIENT_SRCS
   meta_data_cache.cc
   namespace_alterer.cc
   permissions.cc
+  retryable_request_tracker.cc
   session.cc
   schema.cc
   stateful_services/stateful_service_client_base.cc
--- a/src/yb/client/async_rpc.cc
+++ b/src/yb/client/async_rpc.cc
@@ -697,13 +697,14 @@ WriteRpc::WriteRpc(const AsyncRpcData& data, rpc::ThreadPoolTag pool_tag)
     // request ID and details (see https://github.com/yugabyte/yugabyte-db/issues/14005).
     if (first_yb_op->request_id()) {
       const auto& request_detail = batcher_->GetRequestDetails(*first_yb_op->request_id());
-      req_.set_request_id(*first_yb_op->request_id());
-      req_.set_min_running_request_id(request_detail.min_running_request_id);
+      req_.set_request_id(request_detail.registration.request_id());
+      req_.set_min_running_request_id(
+          request_detail.registration.min_running_request_id());
     } else {
-      const auto request_pair = batcher_->NextRequestIdAndMinRunningRequestId();
-      req_.set_request_id(request_pair.first);
-      req_.set_min_running_request_id(request_pair.second);
-      batcher_->RegisterRequest(request_pair.first, request_pair.second);
+      const auto& request_detail = batcher_->RegisterRequest();
+      req_.set_request_id(request_detail.registration.request_id());
+      req_.set_min_running_request_id(
+          request_detail.registration.min_running_request_id());
     }
     FillRequestIds(req_.request_id(), &ops_);
   }
--- a/src/yb/client/batcher.cc
+++ b/src/yb/client/batcher.cc
@@ -688,12 +688,22 @@ server::Clock* Batcher::Clock() const {
   return client_->Clock();
 }
 
-std::pair<RetryableRequestId, RetryableRequestId> Batcher::NextRequestIdAndMinRunningRequestId() {
-  return client_->NextRequestIdAndMinRunningRequestId();
+const RequestDetails& Batcher::RegisterRequest() {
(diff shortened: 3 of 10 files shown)


bmcnama-shopify:
trigger jenkins

spolitov (changes requested):
I would prefer a lock-free approach.

The minimum running ID is used for cleanup, so it is OK if it is a little conservative.

We could do the following:

1. Make `request_id` atomic.
2. Make `min_running_id` atomic.

`NextRequestIdAndMinRunningRequestId` increments the first and reads the second.

`RequestsFinished` pushes finished requests onto a lock-free stack and then tries to run the background processor that handles them (see `PreparerImpl::Submit` for reference).

Issue #32600, opened 2026-07-10
[YSQL] Client retryable-request-ID allocation serializes all writes on a single spinlock


hkimura:
Jira Link: [DB-22322](https://yugabyte.atlassian.net/browse/DB-22322)
### Problem

Every write RPC issued by a YBClient allocates a retryable request ID at
construction and retires it at completion. Both operations take a single
process-wide spinlock (`YBClient::Data::tablet_requests_mutex_`) and mutate a
`std::set` inside the critical section (`YBClient::NextRequestIdAndMinRunningRequestId`
and `YBClient::RequestsFinished` in `src/yb/client/client.cc`).

With `pg_client_use_shared_memory=true`, all PostgreSQL backends on a node funnel
their writes through the one shared pggate YBClient inside the tserver process, so
this one lock is acquired twice per row written by the entire node, from many
threads, with tree-node allocation/free under the lock. The cost scales with
per-node write rate and core count, and it cannot be mitigated by adding tablets,
tables, or connections, because the lock is per-process and sits above tablet
sharding.

Lock-contention profiling of a single-row insert workload on a large many-core
host shows the node plateauing well below CPU saturation, with prominent
`base::SpinLock` contention stacks under `WriteRpc` construction and `Batcher`
destruction. Those are the two code paths that take this lock, which makes it the
prime suspect; a fully symbolized profile to confirm the exact frames is still
pending.

### Why it matters

This is a vertical-scalability cap on the write path: a node with a large core
count cannot convert those cores into small-row OLTP ingest throughput regardless
of sharding or client parallelism, forcing horizontal scaling even when abundant
CPU is available.

### Suggested direction

The global lock is only needed to (a) hand out unique, roughly-dense IDs and
(b) maintain the `min_running_request_id` lower bound advertised to servers for
retryable-request cleanup. Both permit a mostly-lock-free design: reserve blocks
of consecutive IDs per allocating thread (one shared operation per block instead
of two per request), and maintain the min at block granularity. The advertised
min may lag (which only delays server-side cleanup) but must never exceed any
running or still-issuable ID, since servers reject smaller IDs as Expired
(`src/yb/consensus/retryable_requests.cc`).

Constraint for any alternative: server-side dedup state is stored as ID intervals
merged by adjacency, so IDs should remain dense per allocating stream (interleaved
ID sharding would fragment the interval map).

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

[...]

PR #33246, opened 2026-08-11
[#32600] DocDB: Allocate retryable request ids in per-thread blocks

hkimura:
## Summary

Every write RPC allocates a retryable request id at construction and retires it
at completion. Both operations took a single process-wide spinlock
(`YBClient::Data::tablet_requests_mutex_`) and mutated a `std::set` inside the
critical section. With `pg_client_use_shared_memory=true`, all PostgreSQL
backends on a node funnel their writes through the one shared pggate YBClient
inside the tserver process, so this lock was acquired twice per written row by
the entire node. Under single-row insert load on many-core hosts it shows up as
the dominant contended spinlock in the tserver process and caps per-node write
throughput while most cores stay idle.

Replace the lock with a block allocator (`RequestIdAllocator`):

- Each allocating thread reserves a block of consecutive ids
  (`client_request_id_block_size`, default 256) and hands them out with
  thread-local operations only. The shared registry spinlock is taken once per
  block lifecycle instead of twice per request.
- `min_running_request_id` is maintained at block granularity: the minimum
  floor across active blocks (or the sequence head when none are active),
  ratcheting forward only. It may lag the exact minimum of running requests,
  which merely delays server-side cleanup of retryable-request state; it never
  exceeds a running or still-issuable id, which is the invariant servers rely
  on when rejecting stale ids as Expired.
- A block is retired once it is sealed and every allocated id has finished. An
  idle sweep (`client_request_id_block_idle_sec`, default 30) seals blocks
  abandoned by idle threads so they cannot pin `min_running_request_id`
  forever. The seal-vs-allocate race is resolved with a sequentially consistent
  protocol plus discard compensation.
- Ids stay consecutive within a block, so the server-side dedup intervals
  (merged by adjacency in `retryable_requests.cc`) do not fragment.
- The Batcher's `RequestDetails` carries a handle to the id's block, so request
  completion is a counter bump instead of a global-lock `std::set` erase.

## Test plan

- [x] New unit test `request_id_allocator-test`, covering id uniqueness and min
      monotonicity, block-exhaustion retirement, idle-block sealing,
      sealed-block-with-running-request tracking, and a 16-thread hammer
      asserting no id is ever issued below a previously advertised
      `min_running_request_id`:

      ./yb_build.sh --cxx-test request_id_allocator-test

- [x] Sanity of the retryable-request machinery end to end:

      ./yb_build.sh --cxx-test ql-stress-test --gtest_filter QLStressTest.RetryWrites
      ./yb_build.sh --cxx-test ql-stress-test --gtest_filter QLStressTest.RetryWritesWithRestarts
      ./yb_build.sh --cxx-test retryable_request-test

### DIFF
--- a/src/yb/client/CMakeLists.txt
+++ b/src/yb/client/CMakeLists.txt
@@ -48,6 +48,7 @@ set(CLIENT_SRCS
   meta_data_cache.cc
   namespace_alterer.cc
   permissions.cc
+  request_id_allocator.cc
   session.cc
   schema.cc
   stateful_services/stateful_service_client_base.cc
@@ -142,6 +143,7 @@ ADD_YB_TEST(ql-map-test)
 ADD_YB_TEST(ql-tablet-test)
 ADD_YB_TEST(ql-transaction-test)
 ADD_YB_TEST(ql-stress-test)
+ADD_YB_TEST(request_id_allocator-test)
 ADD_YB_TEST(seal-txn-test)
 ADD_YB_TEST(snapshot-txn-test)
 ADD_YB_TEST(snapshot_schedule-test)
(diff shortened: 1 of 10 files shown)


spolitov (changes requested):
What idea of doing this ineffective and complex design instead of lock-free approach?

Issue #33271, opened 2026-08-12
[ClaudeCode] Support Phorge-based backports in the /backport-commit command


kai-franz:
## Description

**Is your feature request related to a problem or an enhancement?**

The `/backport-commit` Claude Code command only knows one way to backport: `.agents/scripts/backport-commit.sh`, which cherry-picks onto release branches and opens GitHub PRs.

Changes that originally landed through Phorge/arc (their commit body carries a `Differential Revision:` line) are backported with a different tool, which publishes arc revisions instead of GitHub PRs. A backport should go through the same review system the original change went through, so the command currently covers only half the cases — and when it's pointed at a Phorge-landed commit it silently produces a backport in the wrong system.

**Provide a brief solution of what you want to happen.**

Extend the command with a second workflow and a mechanism-selection step:

- Add an optional `--phorge` / `--github` argument to force one mechanism.
- Default by detecting how the original landed: a `Dxxxxx` revision ID argument, or a commit whose body has a `Differential Revision:` line, selects Phorge; a commit with a `(#NNNNN)` subject and no such line selects GitHub. Ask when neither pattern matches cleanly.
- Document the Phorge path end to end: non-interactive invocation, exit-code handling, conflict resolution and the `-x <branch>` rerun, where merge-conflict notes belong in the commit message so they reach the published revision summary, and the fact that a new revision auto-fires CI (so an extra trigger comment is a no-op).

Issue type: `kind/enhancement` (tooling change under `.claude/`; no product code affected).

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

PR #33272, opened 2026-08-12
[#33271] ClaudeCode: Add /backport-diff for Phorge backports

kai-franz:
## Summary

The `/backport-commit` command only documented the GitHub-PR backport path (`.agents/scripts/backport-commit.sh`). Changes that landed through Phorge/arc are backported with `backport.sh`, which publishes arc revisions instead — and a backport should go through the same review system the original went through. Pointing `/backport-commit` at a Phorge-landed commit silently produced a backport in the wrong system.

Per review feedback, the Phorge path is a **separate command** rather than a second workflow inside `/backport-commit`: the two files share no content, so once YugabyteDB finishes moving to GitHub PRs, the cleanup is deleting `backport-diff.md` as a whole.

**`.claude/commands/backport-commit.md`** stays GitHub-only. The only addition is a stop-and-redirect: if the argument is a `Dxxxxx` revision ID or the commit body carries a `Differential Revision:` line, use `/backport-diff`; if neither pattern matches cleanly, ask the user rather than guessing.

**`.claude/commands/backport-diff.md`** (new) documents the arc path end to end:

- The `EDITOR=true GIT_EDITOR=true … </dev/null` invocation, with each part explained — without it `arc diff` opens an editor and the run hangs, and the script's `read -p` fallbacks hang instead of failing fast.
- Exit `2` disambiguated: the script uses it for both a dirty workspace and a merge conflict.
- The `-x <branch>` rerun, and the fact that task branches already carrying a `Differential Revision:` line are skipped, so reruns don't create duplicate revisions.
- Merge-conflict notes go into the commit message's Summary section rather than a post-hoc comment: the script's `git commit --amend` carries them into the published revision summary, and there's no safe Phorge equivalent of the GitHub body patch. They have to sit above the `Reviewers:`/`Subscribers:` trailers — free text after `Subscribers:` makes `arc diff` prompt.
- CI: creating a revision auto-fires Harbormaster, and a `trigger jenkins` comment dedups per diff-version, so posting one is a no-op unless the revision has no build at all.

Conflict classification, the lint gate, and reporting are deliberately duplicated in the new file rather than cross-referenced, to keep it independently deletable. In the shared conflict-resolution step, `lint.sh` is called with `--rev origin/<release-branch>`: a backport task branch has no `@{upstream}`, so the base would otherwise be inferred rather than pinned.

`backport.sh` is an internal tool and is not shipped in this repo; the new command's prerequisites say to stop rather than silently fall back to the GitHub path if it isn't on `PATH`.

## Test plan

- [x] Documentation-only change under `.claude/`; no product code or runtime behavior is affected.
- [x] Verified every documented interface against the actual tools rather than from memory: `backport.sh` flags (`-x`, `-r`, `-n`), both exit-`2` paths, the `.git/MERGE_MSG` rewrite that strips the stale `Differential Revision:` line, and the skip of task branches that already carry one; and `build-support/lint.py`'s `--rev` handling for branches with no `@{upstream}`.
- [ ] Not exercised end to end — running it would publish real backport revisions. The next real Phorge backport is the first live run.

### DIFF
--- a/.claude/commands/backport-commit.md
+++ b/.claude/commands/backport-commit.md
@@ -1,11 +1,13 @@
 ---
-description: Backport a commit to one or more YugabyteDB release branches
+description: Backport a commit to one or more YugabyteDB release branches as GitHub PRs
 argument-hint: <commit-sha> [<branch> ...]
 allowed-tools: Bash(.agents/scripts/backport-commit.sh *)
 model: sonnet
 ---
 
 Backport a merged commit to one or more YugabyteDB release branches using the bundled `backport-commit.sh` at `.agents/scripts/backport-commit.sh`. The script handles cherry-picking, message rewriting, fork detection, push, and PR creation. Wrap it with conflict-resolution logic: trivial whitespace conflicts are fixed automatically; anything more complex is escalated to the user.
+
+**This command produces GitHub PRs.** A backport should go through the same review system the original went through, so if the original landed via Phorge — the argument is a `Dxxxxx` revision ID, or `git log -n1 --format=%b <commit>` shows a `Differential Revision:` line — stop and use `/backport-diff` instead. If neither pattern matches cleanly, ask the user rather than guessing.
 
 **Always pass the full branch list in a single invocation.** After each successful branch, the script chains the next branch's cherry-pick from the previous branch's task branch — so a conflict you resolve on `b2` carries forward, and `b3+` typically apply cleanly without re-resolving the same conflict.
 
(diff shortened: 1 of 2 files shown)


hari90 (changes requested):
Lets fork this into a separate command backport-diff.md
Because once we move to PRs we will need to cleanup the Phorge parts and it will be easy to just delete the file as a whole

.claude/commands/backport-commit.md
```diff
@@ -17,7 +19,7 @@ Parse `$ARGUMENTS` (whitespace-separated):
 
 If `<commit>` is missing, stop and ask the user.
 
-If no `<branches>` are provided, ask the user which release branches to target before invoking the script. Do **not** rely on the script's interactive prompt mode — Claude Code's Bash tool cannot answer interactive `read -p` prompts and the run will hang.
+If no `<branches>` are provided, ask the user which release branches to target before invoking the script. Do **not** rely on the script's interactive prompt mode — Claude Code's Bash tool cannot answer interactive `read -p` prompts and the run will hang. Active stable branches: `curl -s https://release.dev.yugabyte.com/version/active/text`. Backport to the branches that actually carry the affected code.
```
iSignal: if this should be friendly to external users, it might be better to suggest querying stable branches from https://docs.yugabyte.com/stable/releases/ybdb-releases/

iSignal (approved):

hari90 (approved):

PR #33273, opened 2026-08-13
[#33252] DocDB: Reduce lock contention in LongOperationTracker

### LINKED ISSUE #33252: [DocDB] Reduce lock contention in LongOperationTracker registration
Jira Link: [DB-22947](https://yugabyte.atlassian.net/browse/DB-22947)
### Description

`LongOperationTracker` registration funnels every tracked operation through a single process-wide mutex in `LongOperationTrackerHelper` (`src/yb/util/debug/long_operation_tracker.cc`):

1. Every `Register()` call acquires the global `std::mutex`, pushes into a shared `std::priority_queue` (O(log n) sift-up under the lock), and then unconditionally calls `cond_.notify_one()` — a futex wake syscall on nearly every registration, since the watcher thread is almost always waiting.
2. The woken watcher immediately re-acquires the same mutex, contending with concurrent registrations. The watcher holds the mutex for all of its work except stack dumping, and pops expired entries one at a time under the lock.
3. Because the queue retains a `shared_ptr` to each operation until its deadline (typically 1s), both the push rate and the pop rate equal the tracked-operation rate — all serialized on one mutex.

This sits on several hot paths: every read query (`read_query.cc`), every Raft `UpdateReplica` (`raft_consensus.cc`), log append (`log.cc`), master heartbeat processing (`master_heartbeat_service.cc`), and — in debug builds — every `ScopedRWOperation` (`operation_counter.cc`). At high throughput this becomes a measurable point of cross-thread contention for a purely diagnostic facility.

There is also a latent bug: `bool stop_;` in `LongOperationTrackerHelper` is never initialized, so the watcher thread's first reads of it are of an indeterminate value.

### Proposed fix

- Replace the mutex-protected shared priority queue with a lock-free `MPSCQueue` (`src/yb/util/lockfree.h`) intake: registration becomes one allocation plus one CAS, with no mutex and no condvar notify.
- Make `TrackedOperation` intrusively ref-counted (`RefCountedThreadSafe` + `scoped_refptr`) so the queue's raw pointer carries an owned reference; the `shared_ptr::use_count() > 1` completion check maps to `!HasOneRef()`.
- The watcher thread drains the intake into a private (uncontended) priority queue and sleeps until `min(next deadline, ~100ms)` instead of being woken per registration. Since the minimum tracked duration in the tree is 1s, the bounded emission delay is not observable in practice. Reporting fidelity is unchanged: each overdue operation still logs its own message, exact elapsed time, and stack trace.
- Initialize `stop_` and drain queue-held references on shutdown.

No wire, disk, or flag changes; purely in-process.

[DB-22947]: https://yugabyte.atlassian.net/browse/DB-22947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

craigsoules:
## Summary

Fixes #33252. Initial review in Shopify/yugabyte-db#23.

Every `LongOperationTracker` registration went through a single process-wide mutex in `LongOperationTrackerHelper`: `Register` locked the mutex, pushed into a shared `std::priority_queue`, and issued an unconditional `notify_one` (a futex wake on nearly every registration, since the checker thread is almost always waiting). The woken checker thread immediately re-acquired the same mutex, contending with concurrent registrations. This sits on hot paths: every read query, every Raft `UpdateReplica`, log append, master heartbeat processing, and (in debug builds) every `ScopedRWOperation`.

Make registration lock-free instead:

- Operations are handed to the checker thread through the lock-free `MPSCQueue` from `yb/util/lockfree.h`. Registration is now one allocation plus one CAS, with no mutex and no condvar notify.
- `TrackedOperation` ownership switches from `std::shared_ptr` to intrusive ref-counting (`RefCountedThreadSafe` + `scoped_refptr`) so the queue's raw pointer can carry an owned reference; the `use_count() > 1` completion check becomes `!HasOneRef()`.
- The checker thread drains the intake queue (bounded to 1000 per iteration so deadline processing and shutdown are never starved) into a private, uncontended priority queue and sleeps until the next deadline, capped at 100ms. Reporting is unchanged: each overdue operation still logs its own message, exact elapsed time, and stack trace.
- Deadlines shorter than 200ms (possible via runtime flags such as `master_ts_heartbeat_long_operation_warning_ms`) additionally wake the checker thread through a mutex-protected predicate, so short-deadline warnings cannot be lost; typical (1s+) registrations never touch the mutex.
- Fix `stop_` being uninitialized.

Registration throughput on macOS arm64 (8 threads x 200k registrations): 4.2M/sec before vs 3.7-4.4M/sec after, i.e. parity within run-to-run noise on that platform. The expected production win is eliminating the per-registration futex wake and the mutex convoy between registering threads and the checker thread on Linux; a Linux perf/futex comparison is still to be collected.

## Test plan

- [x] `./yb_build.sh release --cxx-test debug-util-test --gtest_filter DebugUtilTest.LongOperationTracker` (also debug)
- [x] `./yb_build.sh release --cxx-test debug-util-test --gtest_filter DebugUtilTest.LongOperationTrackerStress` (new: concurrent registration/move churn; also debug)
- [x] `./yb_build.sh release --cxx-test debug-util-test --gtest_filter DebugUtilTest.LongOperationTrackerUnderLoad` (new: overdue operations detected under sustained registration traffic and a 50k burst exercising the bounded drain; exact warning counts validate move construction/assignment clearing; also debug, and 10x repeated in release)
- [x] `./yb_build.sh release daemons` (all call sites compile)
- [ ] Linux sanitizer (ASAN/TSAN) runs via CI (`Register` is a no-op under sanitizers, pre-existing; macOS prebuilt thirdparty lacks instrumented libc++ locally)

### DIFF
--- a/src/yb/util/debug/long_operation_tracker.cc
+++ b/src/yb/util/debug/long_operation_tracker.cc
@@ -63,7 +63,7 @@ struct TrackedOperationComparer {
 // registered operation is noticed within this interval, so a warning could be logged at most
 // this much later than the operation deadline. Since tracked durations are typically much
 // larger than this bound, the delay is not observable in practice.
-constexpr std::chrono::milliseconds kMaxWaitTime(100);
+constexpr auto kMaxWaitTime = std::chrono::milliseconds(100);
 
 // Deadlines below this threshold cannot rely on the periodic scan alone: the operation could
 // expire and complete entirely within one kMaxWaitTime sleep, losing the stack trace warning.
--- a/src/yb/util/debug/long_operation_tracker.cc
+++ b/src/yb/util/debug/long_operation_tracker.cc
@@ -13,25 +13,32 @@
 
 #include "yb/util/debug/long_operation_tracker.h"
 
+#include <algorithm>
 #include <condition_variable>
 #include <mutex>
 #include <queue>
-#include <thread>
+#include <vector>
 
 #include "yb/util/debug-util.h"
+#include "yb/util/lockfree.h"
 #include "yb/util/status_log.h"
 #include "yb/util/thread.h"
 #include "yb/util/tsan_util.h"
 
 namespace yb {
 
-struct LongOperationTracker::TrackedOperation {
+// While an operation is running, one reference is held by the LongOperationTracker returned to
+// the caller and one is transferred through LongOperationTrackerHelper's lock-free intake queue
+// to the checker thread. So HasOneRef() held by the checker thread means the operation has
+// completed.
+struct LongOperationTracker::TrackedOperation :
+    public RefCountedThreadSafe<LongOperationTracker::TrackedOperation>,
+    public MPSCQueueEntry<LongOperationTracker::TrackedOperation> {
   ThreadIdForStack thread_id;
   const char* message;
   CoarseTimePoint start;
   // time when we should log warning
   CoarseTimePoint time;
-  bool complete = false;
 
   TrackedOperation(
       ThreadIdForStack thread_id_, const char* message_, CoarseTimePoint start_,
@@ -42,7 +49,7 @@ struct LongOperationTracker::TrackedOperation {
 
 namespace {
 
-typedef std::shared_ptr<LongOperationTracker::TrackedOperation> TrackedOperationPtr;
+typedef scoped_refptr<LongOperationTracker::TrackedOperation> TrackedOperationPtr;
 
 struct TrackedOperationComparer {
   // Order is reversed, because priority_queue keeps track of the "largest" element.
@@ -51,8 +58,20 @@ struct TrackedOperationComparer {
   }
 };
 
+// Upper bound on how long the checker thread sleeps between scans of the intake queue. A newly
+// registered operation is noticed within this interval, so a warning could be logged at most
+// this much later than the operation deadline. Since tracked durations are much larger than
+// this bound, the delay is not observable in practice.
+constexpr std::chrono::milliseconds kMaxWaitTime(100);
+
 // Singleton that maintains queue of tracked operation and runs thread that checks for expired
 // operations.
+//
+// Synchronization strategy: operations are registered through the lock-free intake_ queue, so
+// registration does not contend on any mutex. The checker thread is the only consumer of
+// intake_ and exclusively owns the priority queue of pending operations (local to Execute).
+// mutex_ and cond_ protect only stop_ and are used only by the checker thread and the
+// destructor.
 class LongOperationTrackerHelper {
  public:
   LongOperationTrackerHelper() {
@@ -72,6 +91,9 @@ class LongOperationTrackerHelper {
     if (thread_) {
       thread_->Join();
     }
+    // Release references of operations that were registered after the checker thread performed
+    // its final drain.
+    ReleaseIntakeRefs();
   }
 
   static LongOperationTrackerHelper& Instance() {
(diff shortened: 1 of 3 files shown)


bmcnama-shopify:
trigger jenkins

src/yb/util/debug/long_operation_tracker.cc
```diff
@@ -51,8 +59,31 @@ struct TrackedOperationComparer {
   }
 };
 
+// Upper bound on how long the checker thread sleeps between scans of the intake queue. A newly
+// registered operation is noticed within this interval, so a warning could be logged at most
+// this much later than the operation deadline. Since tracked durations are typically much
+// larger than this bound, the delay is not observable in practice.
+constexpr std::chrono::milliseconds kMaxWaitTime(100);
```
spolitov: This format seems more readable:
`constexpr kMaxWaitTime = std::chrono::milliseconds(100);`

cuivienor:
Updated review of `825f3a93f8550380f2738505d348e525635f3dd4`: queue ordering is confirmed, but I have not reproduced a missed live stack trace under realistic load. This corrects my earlier medium-severity bug classification.

The details and a test-coverage question are now [in the inline discussion](https://github.com/yugabyte/yugabyte-db/pull/33273#discussion_r3988997816).

src/yb/util/debug/long_operation_tracker.cc
```diff
@@ -84,65 +120,122 @@ class LongOperationTrackerHelper {
...
+    while (!stop_.load(std::memory_order_acquire)) {
+      // Adopt a bounded number of registrations, so that this loop terminates even when
+      // producers are registering operations faster than we drain them.
+      size_t drained = 0;
+      while (drained < kMaxDrainPerIteration) {
+        auto* operation = intake_.Pop();
+        if (!operation) {
+          break;
+        }
+        queue.push(AdoptIntakeRef(operation));
```
cuivienor, line 172: Test-coverage question: could we also exercise a short-deadline operation registered after an older intake batch is already being drained?

`MPSCQueue::Pop()` consumes its detached `pop_head_` batch before checking newer pushes. The 1,000-entry cap here bounds one drain iteration; it does not move a newly registered short-deadline operation ahead of the remaining batch. A standalone extraction of the unchanged queue confirmed that, after queuing 5,000 entries and draining 1,000, a new entry was returned after all 4,000 older entries.

That establishes queue ordering, not a missed stack trace under a realistic workload. I have not reproduced the latter. `LongOperationTrackerUnderLoad` registers its overdue operations before the 50,000-entry burst and keeps them alive until warnings appear, so it does not directly exercise this later-arrival ordering.

Would a test with a short-deadline operation arriving during an existing backlog help establish the intended behavior, or is reporting for that case explicitly best effort? The destructor's elapsed-time warning is separate; the question is about capturing the live stack trace.

This is a test-coverage question, correcting the stronger bug classification in my earlier top-level comment.

cuivienor (commented):
Follow-up to my earlier review: the queue ordering is confirmed, but I have not reproduced a missed live stack trace under realistic load. I have reframed the concern as the inline test-coverage question below.

---
_Review assisted by pair-review_

src/yb/util/debug/long_operation_tracker.cc
cuivienor, line 172: 🐛 **Bug**: The checker only orders operations by deadline after it pops them from the FIFO MPSC intake queue. A burst of older long-deadline registrations can therefore leave a later short-deadline operation invisible to the priority queue while the checker drains 1000-entry chunks. The short-deadline notify wakes the thread, but it does not move that operation ahead of the backlog; if the operation exceeds its deadline and then completes before this loop reaches its intake entry, the later `HasOneRef()` check skips the stack-trace warning. This regresses the intended guarantee for short runtime-configured deadlines under high registration load.

cuivienor (commented):
Testing review

---
_Review assisted by [pair-review](https://github.com/in-the-loop-labs/pair-review)_

src/yb/util/debug/long_operation_tracker.cc
```diff
@@ -42,7 +50,7 @@ struct LongOperationTracker::TrackedOperation {
 
 namespace {
 
-typedef std::shared_ptr<LongOperationTracker::TrackedOperation> TrackedOperationPtr;
+typedef scoped_refptr<LongOperationTracker::TrackedOperation> TrackedOperationPtr;
```
spolitov: We use "using TrackedOperationPtr = ...;" in new code.
Would be nice to switch here also.

spolitov (approved):

src/yb/util/debug/long_operation_tracker.cc
craigsoules: Done in b5cf39ccd84, switched to `using TrackedOperationPtr = scoped_refptr<...>;`.

src/yb/util/debug/long_operation_tracker.cc
craigsoules: Thanks, the ordering observation is correct: `MPSCQueue::Pop()` consumes the detached FIFO batch before looking at newer pushes, so a later short-deadline operation is deadline-ordered only after the checker adopts it into its priority queue. The old code ordered by deadline at registration time, so this is a behavioral difference.

I don't think it rises to a bug, though, because there was never a hard guarantee to regress:

- The stack trace warning has always been best effort. The old path had condvar wakeup latency plus mutex contention with registering threads before the checker could look at the queue, and the checker re-checks `HasOneRef()` after `DumpThreadStack` precisely because the operation may complete mid-dump. The guaranteed signal for an overdue operation is the `took a long time` warning from `~LongOperationTracker`, which is unaffected.
- The window is small. The checker does not sleep while a backlog exists (`drained == kMaxDrainPerIteration` → `continue`), and the sleep is capped at 100ms, so the backlog is bounded by 100ms of registrations. Draining is a pointer chase plus an `O(log n)` `priority_queue::push`; in the new test a 200k-entry backlog registered from 8 threads delays the stack trace by ~5-11ms past the deadline (release) and ~6ms (debug). Missing the trace requires a sub-200ms deadline, an overrun, and completion inside that few-ms drain window, at a registration rate well beyond the production call sites.
- Restoring deadline ordering at registration time would mean touching shared state on every `Register`, which is what this PR removes from the hot paths.

I've documented this explicitly in the `LongOperationTrackerHelper` class comment ("Reporting semantics") in b5cf39ccd84 and added a test for the later-arrival case, see the other thread. If we later want a stronger guarantee, a separate lock-free intake queue for sub-threshold deadlines that the checker drains fully before the bounded main drain would close the window in ~10 lines; I'd rather keep the single queue for now.

craigsoules: Agreed, that case wasn't covered. Added `DebugUtilTest.LongOperationTrackerShortDeadlineBehindBacklog` in b5cf39ccd84:

- 8 threads each register 25k long-deadline (60s) operations that are kept alive, so the intake queue holds a real backlog (200k entries, well past the 1000-per-iteration drain bound).
- A single 20ms-deadline operation is then registered behind it (this also exercises the `short_deadline_pending_` wakeup path).
- Asserts the `running for` stack trace warning is logged while the operation is still alive and within a bounded delay of its deadline (500ms × time multiplier, generous for loaded CI hosts; observed ~5-11ms in release and ~6ms in debug), exactly one checker warning and one destructor warning, `in thread` present, and zero warnings for the backlog operations.

To answer the question directly: reporting for that case is best effort, and the class comment now says so. The destructor's elapsed-time warning is the guaranteed signal; the stack trace can only be captured while the operation is running.

jasonyb:
This appears to fail `./yb_build.sh asan --cxx-test util_debug-util-test --gtest_filter 'DebugUtilTest.LongOperationTrackerShortDeadlineBehindBacklog' --clang21` 11/11 times on ASAN with

```
I0923 22:07:33.098811 418893 thread_holder.cc:30] JoinAll done
../../src/yb/util/debug-util-test.cc:718: Failure
Expected equality of these values:
  log_sink.MessagesSize()
    Which is: 4
  0
```

@craigsoules, can you fix it?

hkimura:
> This appears to fail `./yb_build.sh asan --cxx-test util_debug-util-test --gtest_filter 'DebugUtilTest.LongOperationTrackerShortDeadlineBehindBacklog' --clang21` 11/11 times on ASAN with
> 
> ```
> I0923 22:07:33.098811 418893 thread_holder.cc:30] JoinAll done
> ../../src/yb/util/debug-util-test.cc:718: Failure
> Expected equality of these values:
>   log_sink.MessagesSize()
>     Which is: 4
>   0
> ```
> 
> TSAN is also reported as affected.
> 
> @craigsoules, can you fix it?

FYI, https://github.com/yugabyte/yugabyte-db/pull/34205 on it. Reviewing...

Issue #33169, opened 2026-08-07
[Build] pr-title workflow executes the PR body as shell code


kai-franz:
## Description

The `pr-title` workflow (`.github/workflows/pr-title.yml`) interpolates the PR
title and body directly into the `run:` script with `${{ ... }}`. GitHub Actions
substitutes those expressions **textually, before bash parses the script**, so
the values are pasted raw into the shell source.

In the *Check Backport references* step this is done inside double quotes:

```yaml
body="${{ github.event.pull_request.body }}"
```

Inside double quotes, backticks are command substitution. So any backport PR
whose description contains a backtick — i.e. almost any PR that mentions a
function, option, or file name in code formatting, or includes a fenced code
block — has its description **executed as shell code**. Because `run:` steps use
`bash -e`, the first failing word aborts the step and the check reports failure
with exit code 127.

### Impact

- **False CI failures.** The step only runs for `[BACKPORT ...]` titles, so every
  backport PR with a backtick in its body fails `pr-title` regardless of whether
  the title and the original-commit reference are actually valid. Observed on
  PR #32113, whose title and `Original commit:` line are both well-formed:

  ```
  line 105: read_replica_placement: command not found
  line 105: ./yb_build.sh: No such file or directory
  command substitution: line 117: syntax error near unexpected token `('
  ##[error]Process completed with exit code 127
  ```

- **Command injection.** PR-author-controlled text is executed on the runner.
  Severity is limited because the workflow triggers on `pull_request` rather than
  `pull_request_target`, so it runs with a read-only token and no access to
  secrets — but arbitrary code execution on CI infrastructure is still not
  intended behavior.

### Steps to reproduce

1. Open a PR against a release branch with a title beginning `[BACKPORT <x.y>]`.
2. Include a backtick-quoted word or a fenced code block anywhere in the body.
3. The `pr-title` check fails with exit code 127.

### Suggested fix

Pass the values through the step's `env:` block so they reach the script as
ordinary environment variables and are never parsed as shell source:

```yaml
- name: Check Backport references
  env:
    PR_TITLE: ${{ github.event.pull_request.title }}
    PR_BODY: ${{ github.event.pull_request.body }}
  run: |
    body="$PR_BODY"
```

The same applies to the four `github.event.pull_request.title` interpolations and
to `github.base_ref` in the other two steps.

## Issue Type

kind/bug

## Warning: Please confirm that this issue does not contain any sensitive information

I confirm this issue does not contain any sensitive information.

PR #31948, opened 2026-05-28
[#6640] YSQL: Support ALTER TABLESPACE SET/RESET for placement options

austenLacy:
This patch makes `ALTER TABLESPACE ... SET/RESET` work for the YB placement options (`replica_placement` and `read_replica_placement`), so an existing tablespace can be updated in place. This avoids the awkward evolution of tablespace definitions which require creating new ts, altering all relations that use it, and then deleting the old ts.

tracking issue: #6640 

Some details:

 - `ALTER TABLESPACE` validates the final `spcoptions` the same way we validate `ALTER TABLE/INDEX/MATERIALIZED VIEW ... SET TABLESPACE`.
- If the tablespace is already in use, the new placement has to be satisfiable
- Existing objects are not moved synchronously. We keep using the current master tablespace refresh + load balancer path, so movement happens asynchronously after the new tablespace options are picked up
- `read_replica_placement` still requires `replica_placement`, so `RESET (replica_placement)` fails if `read_replica_placement` is still set.

Test coverage added for:

- `ALTER TABLESPACE ... SET/RESET` regression cases
- movement for tables, indexes, materialized views, and tablegroups
- reset behavior
- invalid placement / invalid JSON cases
- read replica validation, including the `RESET (replica_placement)` case above

Also includes YSQL docs for `ALTER TABLESPACE`.

Tested successfully with:

- `./yb_build.sh release --target postgres --skip-java --no-odyssey`
- `JAVA_TOOL_OPTIONS='--add-opens=java.base/java.lang=ALL-UNNAMED' ./yb_build.sh release --java-test org.yb.pgsql.TestPgRegressTablespaces`
- `JAVA_TOOL_OPTIONS='--add-opens=java.base/java.lang=ALL-UNNAMED' ./yb_build.sh release --java-test 'org.yb.pgsql.TestTablespaceProperties#testAlterTablespacePlacementMovesExistingObjects'`
- `JAVA_TOOL_OPTIONS='--add-opens=java.base/java.lang=ALL-UNNAMED' ./yb_build.sh release --java-test 'org.yb.pgsql.TestTablespaceProperties#negativeTest'`
- `JAVA_TOOL_OPTIONS='--add-opens=java.base/java.lang=ALL-UNNAMED' ./yb_build.sh release --java-test 'org.yb.pgsql.TestTablespaceProperties#readReplicaWithTablespaces'`

Practical testing with `ysqlsh`:

```sql
[...]

PR #33288, opened 2026-08-13
[BACKPORT 2024.2][#33169] Build: Quote PR title and body in the pr-title workflow (#33171)

kai-franz:
DO NOT MERGE!! [CSI](<https://csiweb.dev.yugabyte.com/pull/33288/>) ❌. Once CSI passes, comment `trigger jenkins` to clear this warning.

The `pr-title` workflow interpolated the PR title and body directly into
its
`run:` scripts with `${{ ... }}`. GitHub Actions substitutes those
expressions
**textually, before bash parses the script**, so the values were pasted
raw into
the shell source. In the *Check Backport references* step that happened
inside
double quotes:

```yaml
body="${{ github.event.pull_request.body }}"
```

Inside double quotes, backticks are command substitution — so a backport
PR whose
description contained a backtick had its description **executed as shell
code**,
and `bash -e` aborted the step on the first failing word. The check then
reported
failure with exit code 127 no matter how well-formed the title and the
original-commit reference actually were.

Since the step only runs for `[BACKPORT ...]` titles, every backport PR
that
mentioned a flag, function or file name in code formatting hit this.
#32113 is a
live example: its title and `Original commit:` line are both valid, but
the check
fails with

```
line 105: read_replica_placement: command not found
line 105: ./yb_build.sh: No such file or directory
command substitution: line 117: syntax error near unexpected token `('
##[error]Process completed with exit code 127
```

This change passes the title, body and base ref through each step's
`env:` block
so they arrive as ordinary environment variables and are never parsed as
shell
source. **No validation logic changes** — the regexes, the branch
comparison and
every error message are untouched.

It also removes an unintended code-execution path on the runner.
Exposure was
limited, since the workflow triggers on `pull_request` rather than
`pull_request_target` and so runs with a read-only token and no access
to
secrets, but PR-author-controlled text should not reach a shell parser.

tracking issue: #33169

Original commit: 886bdeca5979b0663167aa9205524c3e34b2c691 / #33171

## Test plan

The three `run:` scripts were extracted from the workflow before and
after the
change and executed under `bash -e` — the shell Actions uses, and the
reason this
fails at all; plain `bash` treats the command-not-found errors as
non-fatal.

Driven with the real title and body of #32113, a backport whose body
contains 22
lines with backticks plus a fenced `sql` block:

| | result |
|---|---|
| before | step 3 exits **127**, reproducing CI (`ALTER: command not
found`, `syntax error near unexpected token '('`) |
| after | step 3 exits **0** — `Found original commit info:
0af82c65c1a6e5487e50e6fa80059c58436646d2 / #31948` |

Confirmed the validation still rejects what it should, each at the
correct step:

- a master PR carrying a `[BACKPORT]` prefix → step 2
- a backport whose title branch disagrees with the base ref → step 2
- a backport body with no `Original commit` line → step 3
- a title with no issue ID → step 1

Also confirmed a well-formed master title passes all three steps, and
that the
only remaining `${{ }}` occurrences in the file are `env:` passthroughs.

## Backports

2026.1, 2025.2 and 2024.2 carry the same line and need this. 2025.1 does
not —
its copy of the workflow has no such interpolation.

Note that each backport PR will need a backtick-free description to get
past the
very check it is fixing, until this lands on that branch.

Co-authored-by: Kai Franz <kai-franz@users.noreply.github.com>

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -14,13 +14,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check PR title
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
           TITLE_RE='^(\[(BACKPORT )?([0-9]+(\.[0-9]+)+)\])?'
           TITLE_RE+='\[((#[0-9]+(,#[0-9]+)*)|([A-Z]+-[0-9]+(,[A-Z]+-[0-9]+)*))\]'
           TITLE_RE+=' [a-zA-Z0-9_-]+(,[[a-zA-Z0-9_]-]+)*:'
           TITLE_RE+=' \S.{8,98}\S$'
 
-          if [[ ! "${{ github.event.pull_request.title }}" =~ $TITLE_RE ]]; then
+          if [[ ! "$PR_TITLE" =~ $TITLE_RE ]]; then
             echo "::error ::Pull Request title must include issue ID, code area,"
             echo "::error ::  and a short description, with no extra whitespace."
             echo "::error ::Issue referenced must be a github issue or YB-internal Jira ID."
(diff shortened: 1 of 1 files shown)


kai-franz:
Trigger Jenkins

svarnau (approved):

Issue #32502, opened 2026-07-06
[Tools] No guard against agent tooling pushing directly to master/main


pao214:
## Problem
AI coding agents used in this repo -- Claude Code and Cursor -- run shell commands, including `git`, directly. A `git push` targeting a protected branch can be executed straight from an agent session, e.g.:

```
git push origin master
```

Nothing in the repository intercepts or blocks such a command before it reaches the remote. Any protection that exists today is per-developer and untracked, so it is applied inconsistently, or not at all.

## User impact
A contributor running these agents can accidentally push directly to `master`/`main`, bypassing the feature-branch + pull-request workflow.

PR #33290, opened 2026-08-13
[#32502] Tools: Tell Claude to use git-push.sh when a push is denied

kai-franz:
## Summary

`permissions.deny` blocks raw `git push` in Claude Code, but a deny rule carries no message. The agent sees only that its command was refused, with no hint that `.agents/scripts/git-push.sh` is the sanctioned way to push. Cursor has had that hint since `block-git-push.sh` landed:

> Blocked: raw git push is disabled. Use .agents/scripts/git-push.sh to push to your fork instead.

This gives Claude Code the same message. **The deny rule is unchanged and still does the blocking** — the `.claude/settings.json` diff is purely additive.

## One script, three registrations

The obvious reading is that there are two harnesses and so two registrations. There are three: Cursor also *imports* the `PreToolUse` hook out of `.claude/settings.json`, so the script has to answer in two dialects across three event names. It reads `hook_event_name` off stdin and decides for itself, rather than trusting either harness's matcher:

| | Claude Code | Cursor, own entry | Cursor, imported entry |
|---|---|---|---|
| registered in | `.claude/settings.json` | `.cursor/hooks.json` | `.claude/settings.json`, rewritten on import |
| `hook_event_name` | `PreToolUse` | `beforeShellExecution` | `preToolUse` |
| scoped by | `matcher` = tool name (`Bash`) | `matcher` regex `\bpush\b` over the command | `matcher` = `Shell`; the `if` is dropped on import |
| command arrives as | `tool_input.command` | top-level `command` | `tool_input.command` |
| deny reply | `hookSpecificOutput.permissionDecision` | `permission` + `user_message` + `agent_message` | same as Cursor's own |
| neutral reply | silence | explicit `{"permission":"ask"}` | silence |
| what enforces | `permissions.deny` | the hook (`failClosed: true`) | nothing — `failClosed` is forced false |

In every case the script's own command check is what finally decides, so a matcher that over-includes costs a process spawn, not a false refusal.

### Cursor imports the Claude hook

Cursor reads `.claude/settings.json`, maps `PreToolUse`/`Bash` to `preToolUse`/`Shell`, and rebuilds the entry keeping only `type`, `command`, `matcher` and `timeout` — the `if` condition is dropped and `failClosed` is forced false. Its dedupe is keyed on the exact command string per event name, so the `.cursor/hooks.json` entry never cancelled the import: the two differed only by the flag they passed. The result is a third registration firing on every Shell call.

That was harmless here only because the substring pre-reject kept it quiet. Any revision that leaned on `if` to scope the *refusal* — as one earlier in this branch did — rejects every shell command in Cursor, because the `if` that scoped it does not survive the import.

### Three defects that fell out of the same investigation

- **The Cursor branch refused unconditionally**, on the premise that Cursor "passes no command in". It does: `beforeShellExecution` carries it as a top-level `command`. Any matcher over-match was refused with a misleading message. The script now reads the command from `tool_input.command` or the top-level `command`, whichever is present.
- **Silence is not neutral under `failClosed`.** Cursor blocks a hook that exits 0 with empty stdout, reporting "no output, policy: fail closed". So the `failClosed` registration gets an explicit `{"permission":"ask"}`, which defers to Cursor's normal approval flow rather than asserting an unrecognised command is safe. Claude and the imported hook still take silence as "no opinion".
- **The command regex over-matched.** Its middle group `([[:space:]][^;&|]*)?` was meant for git's global options but accepted any words, so `git log --author push` and `git commit -m "fix push"` matched. Global options start with `-` and subcommands do not, so the group now requires a leading `-` plus at most one value token.

An earlier revision routed the dialect through a `--format=cursor|claude` flag set per registration. Imported hooks are rebuilt from a fixed field list, so a third registration inherits whatever flag the Claude entry passes and there is no place to correct it — the dialect has to come from the payload. That flag is gone.

### Why the Cursor matcher moves the other way

`.cursor/hooks.json` narrows nothing: its matcher is a bare `\bpush\b`. That side has no equivalent of `permissions.deny`, so the hook *is* the enforcement, and a push spelling its matcher missed would run unblocked. Over-including in the matcher and narrowing in the script is the safe direction there. On the Claude side `permissions.deny` still enforces and the script only supplies the message.

### Why a hook, given hooks are the weaker primitive

Claude Code hooks fail *open*: if one crashes, or the running session hasn't reloaded a changed settings file, the command proceeds. That makes a hook a poor choice for enforcement but a fine messenger. The split is explicit — `permissions.deny` refuses the push, the hook supplies the reason. If the hook fails, the push is still refused, just tersely.

An earlier revision of this PR replaced the deny rule with an owner-aware hook that resolved the remote and allowed fork pushes. That's reverted in the first commit: the fail-open behavior bit during development, and direct fork pushes mostly just create a way to bypass the lint/rebase work `git-push.sh` does.

### Why the script filters, rather than an `if` condition

Two things can scope a Claude hook, and neither can do this job alone:

- `matcher` is matched against the **tool name**, never the command. For `PreToolUse` the dispatcher sets its match query to `tool_name`, and the schema describes `matcher` as a "String pattern to match (e.g. tool names like `Write`)". So `matcher` can say "every Bash call" and nothing narrower.
- `if` takes permission-rule syntax and does see the command, but it **over-matches**: `diff <(echo a) <(echo b)`, which contains no push at all, trips `Bash(git push *)`. A `PreToolUse` deny cannot be approved by the user, so each false positive is a hard stop on an unrelated command, refused with a misleading push message. And as above, `if` does not survive Cursor's import, so anything that depends on it is unscoped on that side.

So `matcher` stays `Bash`, the hook runs on every Bash call, and the script decides. It rejects any payload without `push` in it before spawning `python3`, then extracts the command and requires `git` at command position with `push` as its own word, allowing git's global options in between. `.agents/scripts/git-push.sh` does not qualify — its "git" follows a `/` and its "push" a `-`, so neither sits at a word boundary — while `git -C <dir> push` and `git -c k=v push` carry the message, which the narrow `if` never gave them.

Anything unparseable or unexpected on stdin gets the neutral answer for its event. `permissions.deny` is what refuses a real push, so a miss in the script costs the message, never the enforcement.

### CI routing for `.agents/`

135f47bc37a created `.agents/` without updating either build-path list. `.claude/**` and `.cursor/**` were already in both `bld-db.yml` `paths-ignore` and `bld-none.yml` `paths`; `.agents/**` was in neither, so every file under it fell through to a full DB build — 12 build flavors and ~13k tests for a directory holding five dev scripts and one doc, none of which the build reads.

`bld-db.yml`'s header states the invariant: every file must be covered by one of the `bld-*` workflows. That is why the entry goes in both files — adding it only to `bld-db` `paths-ignore` would leave an `.agents/`-only PR matching no `bld-*` workflow at all, so nothing would report a build status. Raised by @svarnau in review.

## Test plan

- [x] 21-case offline matrix over the three payload shapes, each captured live from the harness that emits it: 8 Claude `PreToolUse`, 4 Cursor `preToolUse` (the imported hook), 4 Cursor `beforeShellExecution`, 5 malformed-stdin cases. All pass.
- [x] Regex spot-checks: the global-option (`git -C /tmp/repo push`, `git -c user.name=x push`) and subshell (`x=$(git push)`, `foo | git push`) spellings are still refused; `git -C /tmp/r log --author push`, `git commit -m "fix push"` and `git stash push` no longer are.
- [x] Earlier 33-case matrix over the Claude shape: 13 push spellings denied with the documented `PreToolUse` shape, 14 non-push commands silent (`.agents/scripts/git-push.sh` in five spellings, the `diff <(echo a) <(echo b)` over-match, `git pushx`, `npm run push-docs`), 6 malformed payloads silent with exit 0
- [x] Live through cursor-agent (`agent -p --trust`) against a scratch repo holding these exact configs: `ls` runs; the plain and `git -C <dir>` spellings are both rejected with the message; `git log --author push --oneline` falls through to Cursor's own permission gate instead of a false refusal
- [x] Live through Claude Code (`claude -p`) against the same configs: `echo HELLO-FROM-CLAUDE` runs, and the dry-run spelling is blocked quoting the message verbatim
- [x] Live in this session through the registered Claude hook: `git push origin --dry-run …` refused **with** the message
- [x] Live: `.agents/scripts/git-push.sh` runs — it is what pushed every commit on this branch
- [x] Both hook configs parse; `bld-db.yml` and `bld-none.yml` parse as YAML and resolve `.agents/**` as the first entry of their respective lists (23 entries for `bld-db`, 20 for `bld-none`)
- [x] `lint.sh --rev origin/master` clean — 0 errors; 3 long-line warnings, all pre-existing (run by `git-push.sh` before every push above)

This PR is its own end-to-end case for the CI change: it touches `.agents/`, `.claude/`, `.cursor/` and `.github/` only. All four are now ignored by `bld-db` and matched by `bld-none`, so `bld-db` should skip on the next sync of this branch, where it previously ran a full build.

Two notes on hook load conditions found the hard way: Cursor hooks only load when the workspace is a git repo, and `-f`/`--yolo` skips them entirely — which is why an earlier attempt saw nothing fire.

## Known limitation, left as is

`Bash(git * push *)` over-matches by the same mechanism the script regex just stopped over-matching by: it is a hard stop on any git command with the word `push` somewhere in it. `git log --author push` is refused outright, and a `git commit -m` whose message quotes the word is too — that happened twice while writing these commits (`git commit -F` is the workaround). The rule is unchanged deliberately: for the thing that actually enforces, over-matching is the safe direction, and narrowing it is a separate decision from this PR.

## Cost note

Dropping `if` means the hook process starts on every Bash tool call — `if` existed partly to avoid that ("Avoids spawning hooks for non-matching commands"). The substring pre-reject keeps the common case to one `bash` and no `python3`.


.agents/scripts/block-git-push.sh
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env bash
-printf '{"permission":"deny","user_message":"%s%s"}\n' \
-  'Blocked: raw git push is disabled. ' \
-  'Use .agents/scripts/git-push.sh to push to your fork instead.'
```
hari90, line 4: Why?

hari90 (changes requested):

kai-franz:
Closing without merging — the blanket `git push` deny is staying.

This replaced the deny rule with a hook that resolves the remote and blocks only the upstream owner, so agents could push to a fork directly. Two things argue against it:

1. **`permissions.deny` is the more robust primitive.** It's evaluated by the harness and always applies. Claude Code hooks fail *open* — there is no `failClosed` equivalent to Cursor's. That bit during development: a rebase momentarily reverted `.claude/settings.json`, the session latched the pre-change copy, and a push to origin that the hook denies when invoked directly went straight through. (Dry run; `git ls-remote` confirmed no ref was created.) A guard with a silent-failure mode is worse than a blunt one without.

2. **Direct fork pushes aren't needed.** `.agents/scripts/git-push.sh` is the sanctioned path and does more than push — integrates commits already on the fork branch, rebases onto fresh upstream, runs lint, force-pushes with lease. Allowing raw `git push` to a fork mostly creates a way to bypass that.

So the existing arrangement stands: blanket deny in `.claude/settings.json`, `block-git-push.sh` for Cursor, and `src/AGENTS.md` telling agents to push via `git-push.sh`.

One observation from the investigation, recorded here rather than fixed: that instruction lives at `src/AGENTS.md:63`, which is only loaded when the agent is working under `src/`. Root `AGENTS.md` has no push guidance and only defers to `src/AGENTS.md` "when working on DB code". For repo-tooling work outside `src/`, an agent gets a bare permission denial with no pointer to `git-push.sh` — Cursor at least gets a message from `block-git-push.sh`. Worth a look if that trips anyone up later.

svarnau:
Perhaps you could add .agents/** to the ignore list in .github/workflows/bld-db and bld-none

hari90 (approved):

PR #33292, opened 2026-08-14
[DOC-1440] Docs: Clarify catalog version mismatch troubleshooting

iSignal:
Explain that the error is specific to breaking DDL, correct which statements are actually breaking (plain CREATE TABLE/ALTER TABLE/DROP are not), add an example log line for locating the responsible DDL, and link to the existing SQLSTATE 40001 retry documentation.


pgyogesh (commented):
Looks good to me.

ddhodge (approved):

PR #33300, opened 2026-08-16
[#33298] YSQL: Make pg_localtime/pg_gmtime buffer thread-local

### LINKED ISSUE #33298: [CDCSDK] Data race on shared static struct pg_tm in pg_localtime() corrupts timestamptz values decoded by concurrent GetChanges calls
Jira Link: [DB-22976](https://yugabyte.atlassian.net/browse/DB-22976)
### Summary

`pg_localtime()` / `pg_gmtime()` (src/postgres/src/timezone/localtime.c) return a
pointer to a single process-wide `static struct pg_tm tm`. Fine in vanilla PostgreSQL,
but the tserver calls it from many RPC threads:
CDCSDK decodes every timestamptz column via `timestamptz_out` through ybgate
(`cdcsdk_producer.cc` -> `docdb_pgapi.cc SetValueFromQLBinary` -> `ybgate_api.c
DecodeTZDatum` -> `timestamp2tm` -> `pg_localtime`). Concurrent GetChanges calls
overwrite each other's fields in that shared buffer, so consumers receive corrupted
timestamps. (Related GetChanges concurrency: #23394.)

Observed corruption classes (all reproduced under concurrent load and by the unit test
below):

1. Wrong seconds/minutes, milliseconds intact. `2026-06-28 22:13:20.1+00` comes out as
   `2026-06-28 22:13:34.1+00`. `timestamp2tm` reads the tm fields back from the shared
   buffer after another thread's `timesub` already overwrote them. `fsec` is computed
   separately from the raw value, so the milliseconds survive.
2. Year off by +1900, e.g. `3926-...` for a 2026 date. `timesub` writes the full year
   into the shared `tm_year`, then subtracts `TM_YEAR_BASE` in a second step. A reader
   that hits the `tm_year + 1900` line in that window gets 2026 + 1900 = 3926.
3. Year off by -1900, e.g. `0126-...`. Two threads run the read-modify-write
   subtraction on the same shared field at the same time.
4. Crash. A reader can catch `tm_mon` mid-update, out of range. In cassert builds that
   aborts: `TRAP: FailedAssertion("tm->tm_mon >= 1 && tm->tm_mon <= MONTHS_PER_YEAR",
   datetime.c:4258)` -> SIGABRT. In release builds it is an out-of-bounds read of the
   month-name array in `EncodeDateTime`.

### Reproduction

The unit test `YbGateTest.TimestamptzDecodeIsThreadSafe` in
`src/yb/docdb/docdb_pgapi-test.cc` reproduces it reliably. It runs 8 threads, each
decoding its own fixed timestamptz value in a loop via
`docdb::SetValueFromQLBinary(..., 1184, ...)`, and checks every result against the
value decoded single-threaded up front. The same test on plain `timestamp` (oid 1114)
is a control - that path never touches the shared buffer and stays clean.

Results (debug build, 2026.1, 8 threads x 2M = 16M decodes):

- before fix: `mismatches=101697 year_mismatches=847 year_3926_hits=825
  year_0126_hits=22` (~0.64% of decodes corrupted); one run aborted with the
  EncodeDateTime TRAP above; the control test is always clean;
- after fix: all counters are zero on the same workload; wall time unchanged
  (6.46s vs 6.57s for 16M decodes).

### Fix

Make the buffer thread-local. Covers both `pg_localtime` and `pg_gmtime` (same buffer),
no call-site changes, no locking, throughput unaffected (see Results).

```diff
diff --git a/src/postgres/src/timezone/localtime.c b/src/postgres/src/timezone/localtime.c
index fa3c059038..964085840e 100644
--- a/src/postgres/src/timezone/localtime.c
+++ b/src/postgres/src/timezone/localtime.c
@@ -99,7 +99,7 @@ static bool typesequiv(struct state const *, int, int);
  * Thanks to Paul Eggert for noting this.
  */
 
-static struct pg_tm tm;
+static __thread struct pg_tm tm;
 
 /* Initialize *S to a value based on UTOFF, ISDST, and DESIGIDX.  */
 static void
```

### Source connector version

server-side bug

### Connector configuration

server-side bug

### YugabyteDB version

_No response_

### Issue Type

kind/bug

[DB-22976]: https://yugabyte.atlassian.net/browse/DB-22976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

grigoriizubov:
## Summary

Fixes #33298.

`pg_localtime()` / `pg_gmtime()` return a pointer to a single process-wide `static struct pg_tm`. Under ybgate's multi-threaded timestamptz decoding (CDCSDK GetChanges), concurrent calls overwrite each other's fields in that shared buffer - corrupting timestamps and occasionally crashing on an out-of-range `tm_mon`. Making the buffer thread-local fixes the race, with no call-site changes and no measurable overhead.

See #33298 for the corruption classes, reproduction numbers, and throughput comparison.

## Test plan

- [x] Added `YbGateTest.TimestamptzDecodeIsThreadSafe` (8 threads decoding a timestamptz value) and control `TimestampDecodeIsThreadSafe` (plain timestamp, does not touch the shared buffer) in `docdb_pgapi-test.cc`.
- [x] `docdb_pgapi-test` passes with the fix (fastdebug, clang21, arm64).
- [x] Reverting the `__thread` change makes `TimestamptzDecodeIsThreadSafe` fail while the control still passes, confirming the test catches the race.

### DIFF
--- a/src/postgres/src/timezone/localtime.c
+++ b/src/postgres/src/timezone/localtime.c
@@ -99,7 +99,7 @@ static bool typesequiv(struct state const *, int, int);
  * Thanks to Paul Eggert for noting this.
  */
 
-static struct pg_tm tm;
+static __thread struct pg_tm tm;
 
 /* Initialize *S to a value based on UTOFF, ISDST, and DESIGIDX.  */
 static void
--- a/src/yb/docdb/docdb_pgapi-test.cc
+++ b/src/yb/docdb/docdb_pgapi-test.cc
@@ -10,12 +10,24 @@
 // or implied.  See the License for the specific language governing permissions and limitations
 // under the License.
 
+#include <atomic>
+#include <string>
+#include <vector>
+
 #include "yb/util/logging.h"
 #include <gtest/gtest.h>
 
 #include "ybgate/ybgate_api.h"
 #include "ybgate/ybgate_status.h"
 #include "ybgate/ybgate_api-test.h"
+
+#include "yb/common/common.pb.h"
+#include "yb/common/value.pb.h"
+#include "yb/docdb/docdb_pgapi.h"
+#include "yb/util/env.h"
+#include "yb/util/path_util.h"
+#include "yb/util/result.h"
+#include "yb/util/test_thread_holder.h"
 #include "yb/util/test_util.h"
 #include "yb/util/yb_pg_errcodes.h"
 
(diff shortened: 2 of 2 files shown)


src/postgres/src/timezone/localtime.c
```diff
@@ -99,7 +99,7 @@ static bool typesequiv(struct state const *, int, int);
  * Thanks to Paul Eggert for noting this.
  */
 
-static struct pg_tm tm;
+static __thread struct pg_tm tm;
```
kai-franz: We use `YB_THREAD_LOCAL` instead of `__thread`

PR #33322, opened 2026-08-17
[#33268] YSQL: Fix hang in PollTransactionStatusBase::Shutdown racing VerifyTransaction

### LINKED ISSUE #33268: [YSQL] PollTransactionStatusBase::Shutdown can hang forever when it races VerifyTransaction
Jira Link: [DB-22956](https://yugabyte.atlassian.net/browse/DB-22956)
### Description

`PollTransactionStatusBase::Shutdown` waits on `sync_` with an untimed
`Wait()` to drain an in-flight `GetTransactionStatus` RPC. If it races a
concurrent `VerifyTransaction`, that wait never completes and the calling
thread is stuck permanently.

`VerifyTransaction` consumes the signal before it decides whether to send
an RPC:

  1. `RETURN_NOT_OK(sync_.Wait()); sync_.Reset();`
  2. take `rpc_mutex_`
  3. `SCHECK(!shutdown_, ...)` -- returns without sending anything
  4. `rpcs_.Prepare()` -- also returns without sending if the handle is invalid

`Reset()` clears `assigned_`, so the latch is re-armed and now expects a
callback. On either early exit no RPC is sent, so no callback ever runs and
nothing signals it again. A later `Shutdown()` calls `sync_.Wait()` on that
re-armed latch and blocks forever. The wait has no deadline, and it is held
under `rpc_mutex_`, so the task object can no longer be torn down.

The window is small but reachable in the master: an abort arriving while a
DDL verification task is polling runs `Shutdown()` from the abort path while
the task's own step is inside `VerifyTransaction`.

### Issue Type

kind/bug

[DB-22956]: https://yugabyte.atlassian.net/browse/DB-22956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

egladysh:
## Summary

Fixes #33268.

`PollTransactionStatusBase` uses `sync_` as a one-token gate: signaled means no
`GetTransactionStatus` callback is outstanding. The constructor primes it, a send consumes the
token and hands the duty of re-signaling to the rpc callback, and `Shutdown()` waits for the
token to prove no callback is in flight before setting `shutdown_`.

`VerifyTransaction` consumed that token before it knew whether an rpc would actually be sent:

```
  RETURN_NOT_OK(sync_.Wait());
  sync_.Reset();                 // token consumed here

  std::lock_guard l(rpc_mutex_);
  SCHECK(!shutdown_, ...);       // returns without sending
  auto rpc_handle = rpcs_.Prepare();
  if (rpc_handle == rpcs_.InvalidHandle()) { return ...; }   // returns without sending
```

`Reset()` clears `assigned_`, so the latch is re-armed and now expects a callback. On either
early exit no rpc is sent, so no callback ever runs and nothing signals it again. A later
`Shutdown()` calls the untimed `sync_.Wait()` on that re-armed latch and blocks forever. The
wait is held under `rpc_mutex_`, so the task can no longer be torn down either, and the
destructor's `DCHECK(shutdown_)` can never be satisfied.

In the master this is reachable whenever an abort arrives while a DDL verification task is
polling: the abort path runs `PerformAbort()` -> `Shutdown()` while the task's own step is
inside `VerifyTransaction`. The aborting thread then hangs with no deadline and no log output.

The fix moves `sync_.Reset()` down to just before the callback is created, so the token is
consumed only on the path that actually sends. `Reset()` through `SendRpc()` now happens
entirely under `rpc_mutex_`, which `Shutdown()` also takes, so a concurrent `Shutdown()` can
only land on one side or the other: before, and `VerifyTransaction` sees `shutdown_` and
returns with the signal intact; after, and the rpc is already registered in `rpcs_`, so
`rpcs_.Shutdown()` aborts it and the callback fires. There is no longer a window where the
token is gone and no callback is pending.

`Reset()` is safe at its new position: `must_wait_` was cleared by the `Wait()` above, so
`EnsureWaitDone()` is a no-op.

Two alternatives were considered and rejected. Restoring the token on the early-exit paths
with a scope guard works but re-signals a latch that was never consumed for anything, and the
guard has to be cancelled before `AsStdStatusCallback` hands the token out or it double-assigns.
Moving the `Wait()` itself under `rpc_mutex_` is unsafe: the callback path can reach
`Shutdown()` via `FinishPollTransaction` -> `Complete()` -> `TaskCompleted()`, which takes
`rpc_mutex_` -- the deadlock already documented in `DdlRequesterLivenessTask`.

## Test plan

New regression test `CatalogEntityTaskTest.TestAbortWithInFlightTransactionPoll` in
`master_multi_step_monitored_task-test`. It uses sync points to park a task's step inside
`VerifyTransaction` past the point where the token is consumed, lets the abort path run
`Shutdown()` once, then releases the step so it returns via the `shutdown_` check. The abort
is run on a separate thread and the test waits on it with a 30 second bound, so a
reproduction fails with a named assertion instead of wedging the binary until the harness
kills it.

The test also asserts that the step exited at the shutdown `SCHECK` specifically, by matching
`"Task has been shut down"`. Three exits in `VerifyTransaction` return `IllegalState`, and
only that one strands the signal, so without the message check a reordering of the function
could make the test pass without exercising the bug.

- [x] `CatalogEntityTaskTest.TestAbortWithInFlightTransactionPoll` fails on the parent commit:
      30 second hang, then `Abort deadlocked on the transaction status poll Synchronizer`.
- [x] Same test passes with this fix, in 11ms.
- [x] `CatalogEntityTaskTest.TestMultipleSteps` passes (4005ms).
- [x] `CatalogEntityTaskTest.TestEpochChanges` passes (4009ms).
- [ ] CI.

Run with:

```
ybd --cxx-test master_multi_step_monitored_task-test \
  --gtest_filter CatalogEntityTaskTest.TestAbortWithInFlightTransactionPoll
```

### DIFF
--- a/src/yb/master/multi_step_monitored_task-test.cc
+++ b/src/yb/master/multi_step_monitored_task-test.cc
@@ -279,6 +279,10 @@ TEST(CatalogEntityTaskTest, TestAbortWithInFlightTransactionPoll) {
   auto abort_returned = std::make_shared<std::atomic<bool>>(false);
   std::thread abort_thread(
       [task, abort_returned, catalog_entity, thread_pool, messenger, client]() {
+        (void)catalog_entity;
+        (void)thread_pool;
+        (void)messenger;
+        (void)client;
         task->AbortAndReturnPrevState(STATUS(Aborted, "Test abort"));
         abort_returned->store(true);
       });
--- a/src/yb/master/multi_step_monitored_task-test.cc
+++ b/src/yb/master/multi_step_monitored_task-test.cc
@@ -11,12 +11,19 @@
 // under the License.
 //
 
+#include <atomic>
+#include <future>
+#include <thread>
+
 #include <gtest/gtest.h>
 
+#include "yb/client/client.h"
 #include "yb/master/catalog_entity_base.h"
 #include "yb/master/multi_step_monitored_task.h"
 #include "yb/master/tasks_tracker.h"
+#include "yb/master/ysql_ddl_verification_task.h"
 #include "yb/rpc/messenger.h"
+#include "yb/util/backoff_waiter.h"
 #include "yb/util/scope_exit.h"
 #include "yb/util/sync_point.h"
 #include "yb/util/test_macros.h"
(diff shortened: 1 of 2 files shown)


bmcnama-shopify:
trigger jenkins

myang2021 (approved):
LGTM!

PR #33332, opened 2026-08-17
[#16670] DocDB: dist-trace: DO NOT MERGE - enable dist-trace by default for a full CI run

ellabaron-code:
## Summary

Sets the `otel_collector_traces_endpoint` default to `http://127.0.0.1:14318`, a
local port with no listener, so every process in a CI run initializes
distributed tracing and exercises the span code paths. Nothing collects the
spans — the exporter gets connection refused and fails fast.

Changing the default rather than overriding the flag also avoids the preview
flag gate, which only rejects non-default values.

Not for merge — this branch exists only to get a full Jenkins run with tracing
on, to confirm no test outside `dist_trace-test.cc` breaks when it is enabled.

## Test plan

- [ ] Full Jenkins run on this branch; compare failures against the same run on `feature-stack/dt2-05-pg-backend`.


ellabaron-code:
trigger jenkins

Issue #33376, opened 2026-08-19
[DocDB] RPC response callbacks share the inbound service thread pool, allowing self-starvation deadlock


ellabaron-code:
Jira Link: [DB-23067](https://yugabyte.atlassian.net/browse/DB-23067)
## Description

`Messenger::CallbackThreadPool` hands out the same thread pool that serves inbound calls:

```c++
rpc::ThreadPool& CallbackThreadPool(ServicePriority priority) override {
  return ThreadPool(priority);
}
```

`Proxy::GetCallbackThreadPool` routes an outbound call's response callback to that pool whenever
the call's `InvokeCallbackMode` is `kThreadPoolNormal` or `kThreadPoolHigh`.
`kThreadPoolNormal` is the default in `rpc_controller.h`, and only a handful of call sites in the
tree override it, so in practice nearly every outbound response callback on a server is enqueued
onto the pool that is also executing that server's inbound RPC handlers.

That makes the dependency graph cyclic. An inbound handler that blocks waiting for an outbound
call to complete occupies a worker in the very pool that has to run that call's completion. With
enough concurrent handlers doing this, every worker is blocked, the response callbacks sit in the
pool queue with no worker to pick them up, and the server never makes progress again.

The wait is unbounded, so RPC deadlines cannot break the cycle. In the case below the handler
waits on a `Synchronizer`, and `Synchronizer::Wait()` is
`WaitUntil(std::chrono::steady_clock::time_point::max())`. The deadline belongs to the outbound
call, which already succeeded — what never happens is the delivery of its response.

### Reproduction

`PgTxnTest.ManyCommitsWithSmallRpcWorkers` hangs on macOS with
`--rpc_workers_limit=32`. macOS makes it easy to hit because
`pg_client_use_shared_memory` defaults to false there, so all `Perform` and `AcquireObjectLock`
traffic goes through the RPC pool instead of shared memory; the cycle itself is
platform-independent.

Observed state once wedged:

- All 32 workers of the tserver's normal pool are parked with zero CPU, waiting for transaction
  metadata — before any lock request has been issued:

```
    @  _sigtramp
    @  _pthread_cond_wait
    @  std::__1::condition_variable::wait()
    @  yb::Synchronizer::WaitUntil()
    @  yb::tserver::PgClientSession::Impl::NextObjectLockingTxnMeta()
    @  yb::tserver::PgClientSession::Impl::DoAcquireObjectLock()
[...]

PR #33377, opened 2026-08-19
[#33376] DocDB: Use separate thread pools for outbound RPC response callbacks

ellabaron-code:
## Summary

`Messenger::CallbackThreadPool` returned the same pool that serves inbound calls, so an inbound
handler that waits for an outbound completion could occupy every worker in the pool that has to
invoke that completion. The response arrives, its callback is enqueued, and no worker is ever free
to run it.

Give each messenger dedicated normal- and high-priority callback pools, sized by the new
`rpc_callback_workers_limit` flag, which defaults to `rpc_workers_limit`. Inbound handlers now
depend on outbound completions, and outbound completions no longer need an inbound handler slot.

## Test plan

- [x] `pg_txn-test` — hung before the change (`PgTxnTest.ManyCommitsWithSmallRpcWorkers`),
      passes after it.
- [x] `rpc-test`, `rpc_stub-test` — messenger init/shutdown with the added pools.

### DIFF
--- a/src/yb/rpc/messenger.cc
+++ b/src/yb/rpc/messenger.cc
@@ -98,6 +98,10 @@ DEPRECATE_FLAG(int32, rpc_queue_limit, "03_2024");
 
 DEFINE_NON_RUNTIME_int32(rpc_workers_limit, 1024, "Workers limit for rpc server");
 
+DEFINE_NON_RUNTIME_int32(rpc_callback_workers_limit, 0,
+    "Workers limit for the outbound RPC response callback thread pools. 0 means use the value of "
+    "rpc_workers_limit.");
+
 DEFINE_NON_RUNTIME_int32(rpc_reactor_task_timeout_ms, 0,
     "Report if reactor task task takes longer that specified amount of milliseconds. "
     "0 to disable monitoring.");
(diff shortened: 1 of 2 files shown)


src/yb/rpc/messenger.cc
```diff
@@ -653,6 +673,21 @@ Reactor* Messenger::RemoteToReactor(const Endpoint& remote, uint32_t idx) {
...
 
+  auto callback_workers_limit = FLAGS_rpc_callback_workers_limit > 0
+      ? FLAGS_rpc_callback_workers_limit : thread_pool_workers_limit_;
+  normal_callback_thread_pool_ = std::make_shared<rpc::ThreadPool>(
+      rpc::ThreadPoolOptions {
+        .name = name_ + "-cb",
+        .max_workers = callback_workers_limit,
+        .cgroup = system_med_cgroup_,
+      });
+  high_priority_callback_thread_pool_ = std::make_shared<rpc::ThreadPool>(
```
es1024: We don't create normal high priority thread pool unless needed (because many messengers may not ever use it); should do something similar for this.

ellabaron-code: Done — both callback pools are now built on the first `CallbackThreadPool()` call for that
priority, using the same double-checked-locking shape as `high_priority_thread_pool_`, and
`Messenger::Init` no longer creates them up front.

Issue #33380, opened 2026-08-19
[DocDB] Leaked pending Raft operation state when NotifyAddedToLeader fails


jmeehan16:
Jira Link: [DB-23069](https://yugabyte.atlassian.net/browse/DB-23069)
## Description

`RaftConsensus::DoAppendNewRoundsToQueueUnlocked` allocates a Raft OpId via `ReplicaState::NewIdUnlocked` (which advances `next_index_`) and then calls `round->NotifyAddedToLeader(op_id, committed_op_id)`:

```cpp
OpId op_id = VERIFY_RESULT(state_->NewIdUnlocked(round->replicate_msg()->op_type()));
RETURN_NOT_OK(round->NotifyAddedToLeader(op_id, committed_op_id));

auto s = state_->AddPendingOperation(round, OperationMode::kLeader);
if (!s.ok()) {
  RollbackIdAndDeleteOpId(round->replicate_msg(), /*should_exists=*/false);
  return s;
}
```

The `AddPendingOperation` failure path rolls back the allocated index, but the `NotifyAddedToLeader` failure path is a bare `RETURN_NOT_OK` with no cleanup. `NotifyAddedToLeader` can fail: it calls `Operation::AddedToLeader`, whose first step is `VERIFY_RESULT(tablet_safe())`, which fails while the tablet is shutting down.

Consequences of returning without cleanup:

1. `ReplicaState::next_index_` has already advanced, so the allocated Raft index is silently skipped: the leader's log ends at `N - 1` while the next replicate is assigned `N + 1`. `ReplicaState::CancelPendingOperation` exists precisely to roll this back (`CHECK_EQ(next_index_, id.index + 1); next_index_ = id.index;`) and is invoked for the sibling failure path.
2. The existing `RollbackIdAndDeleteOpId(round->replicate_msg(), ...)` helper cannot be reused as-is for this path, because `Operation::AddedToLeader` fails *before* it copies the OpId into the replicate message (`op_id.ToPB(replicate_msg->mutable_id())`), so the message's id field is still unset. The cleanup must use the locally allocated `op_id`.

Proposed fix: on `NotifyAddedToLeader` failure, call `state_->CancelPendingOperation(op_id, /* should_exist= */ false)` before returning the status, mirroring the `AddPendingOperation` failure handling.

[...]

PR #33381, opened 2026-08-19
[#33380] DocDB: Roll back allocated Raft index when NotifyAddedToLeader fails

jmeehan16:
## Summary

`RaftConsensus::DoAppendNewRoundsToQueueUnlocked` allocates the operation's Raft index via `NewIdUnlocked` (which advances `ReplicaState::next_index_`) and then calls `round->NotifyAddedToLeader(op_id, committed_op_id)`. The `AddPendingOperation` failure path directly below rolls the allocation back via `RollbackIdAndDeleteOpId`, but a `NotifyAddedToLeader` failure returned with no cleanup, silently skipping the allocated index: the leader's log ends at `N - 1` while the next replicate is assigned `N + 1`.

`NotifyAddedToLeader` can fail: `Operation::AddedToLeader`'s first step is `tablet_safe()`, which fails while the tablet shuts down.

The fix mirrors the `AddPendingOperation` failure handling, with one difference: `Operation::AddedToLeader` fails before it copies the OpId into the replicate message (`op_id.ToPB(replicate_msg->mutable_id())`), so the cleanup must use the locally allocated `op_id` via `ReplicaState::CancelPendingOperation(op_id, /* should_exist= */ false)` rather than `RollbackIdAndDeleteOpId(round->replicate_msg(), ...)`, whose message id field is still unset on this path.

Since the natural trigger is a shutdown race that cannot be reproduced deterministically, this also adds `TEST_fail_operation_added_to_leader` to force the failure in tests.

A related but separate symptom of the same abort window (an operation aborted after OpId assignment but before it was added as pending unregisters a never-registered operation filter) is tracked separately and will be fixed in a follow-up PR.

## Test plan

- [x] New test: `./yb_build.sh release --cxx-test tablet_peer-test --gtest_filter TabletPeerTest.PendingStateCleanedUpWhenAddedToLeaderFails` — with the failure injected, the write is rejected with `IllegalState`, the committed and logged OpIds are unchanged, and with the injection removed the next write succeeds reusing the rejected index.
- [x] Regression for the `ExecuteWrite` test-helper refactor: `./yb_build.sh release --cxx-test tablet_peer-test --gtest_filter TabletPeerTest.TestLogAnchorsAndGC`

### DIFF
--- a/src/yb/consensus/raft_consensus.cc
+++ b/src/yb/consensus/raft_consensus.cc
@@ -1398,7 +1398,16 @@ Status RaftConsensus::DoAppendNewRoundsToQueueUnlocked(
     // the write batch inside the write operation.
     //
     // TODO: we could allocate multiple HybridTimes in batch, only reading system clock once.
-    RETURN_NOT_OK(round->NotifyAddedToLeader(op_id, committed_op_id));
+    auto notify_status = round->NotifyAddedToLeader(op_id, committed_op_id);
+    if (!notify_status.ok()) {
+      // Roll back the index allocated by NewIdUnlocked above, mirroring the AddPendingOperation
+      // failure handling below: otherwise the index is silently skipped (the log ends before it
+      // while the next replicate is assigned a later one). Operation::AddedToLeader fails before
+      // it copies the OpId into the replicate message, so the cleanup must use the local op_id
+      // rather than RollbackIdAndDeleteOpId(round->replicate_msg(), ...).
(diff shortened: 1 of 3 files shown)


jmeehan16:
Resubmitting from an in-repo feature-stack/Shopify/ branch so the PR stack (#33383 and successors) follows the upstream feature-stack convention. Replacement PR link to follow.

---
_automated · opencode (claude-fable-5)_

jmeehan16:
Replacement PR: https://github.com/yugabyte/yugabyte-db/pull/33403

Issue #33382, opened 2026-08-19
[DocDB] Operation aborted before AddPendingOperation unregisters a never-registered operation filter


jmeehan16:
Jira Link: [DB-23070](https://yugabyte.atlassian.net/browse/DB-23070)
## Description

An operation that is aborted after its Raft OpId was assigned but before it was added as pending takes the was-pending abort path and unregisters an operation filter it never registered, corrupting the tablet's operation-filter list.

The window: `OperationDriver::AddedToLeader` stores `op_id_copy_` *before* invoking `Operation::AddedToLeader`:

```cpp
Status OperationDriver::AddedToLeader(const OpId& op_id, const OpId& committed_op_id) {
  ...
  op_id_copy_.store(op_id, boost::memory_order_release);
  RETURN_NOT_OK(operation_->AddedToLeader(op_id, committed_op_id));
  ...
```

`Operation::AddedToLeader` can fail at its first step (`tablet_safe()` fails while the tablet shuts down), which is *before* it calls `AddedAsPending(tablet)` at the end. On the resulting abort, `OperationDriver::HandleFailure` infers `was_pending` from OpId validity:

```cpp
operation_->Aborted(status, op_id_copy_.load().valid());
```

which is now true, so `Operation::Aborted` calls `RemovedFromPending(tablet)` even though `AddedAsPending` never ran.

Two operation types register an operation filter in `AddedAsPending` and unregister it unconditionally in `RemovedFromPending`:

- `SplitOperation` (always), and
- `SnapshotOperation` (for `RESTORE_ON_TABLET` / `RESTORE_SYS_CATALOG`).

`Tablet::UnregisterOperationFilter` erases the filter from an intrusive list via `iterator_to`, which is undefined behavior for an unlinked node; in practice this crashes with SIGSEGV.

Proposed fix: track filter registration in these operations (register sets a flag; `RemovedFromPending` unregisters only when the flag is set). A deeper follow-up option is to make the driver/operation track added-as-pending explicitly instead of inferring it from OpId validity; the targeted guards keep the fix minimal.

This is the second symptom of the abort window fixed in #33380 / #33381 (leaked pending Raft index on `NotifyAddedToLeader` failure); the `TEST_fail_operation_added_to_leader` hook added there makes this deterministically testable.

## Issue Type

kind/bug

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[...]

PR #33383, opened 2026-08-19
[#33382] DocDB: Do not unregister a never-registered operation filter on abort

jmeehan16:
## Summary

> **Stacked PR**: based on #33381 (`[#33380] DocDB: Roll back allocated Raft index when NotifyAddedToLeader fails`), whose commit this branch includes and whose `TEST_fail_operation_added_to_leader` hook the tests reuse. Only the top commit is new here. Once #33381 merges, this PR will be retargeted to `master` and rebased.

An operation aborted after its Raft OpId was assigned but before it was added as pending takes the was-pending abort path: `OperationDriver::AddedToLeader` stores `op_id_copy_` *before* invoking `Operation::AddedToLeader`, whose first step (`tablet_safe()`) can fail while the tablet shuts down -- before `AddedAsPending` runs -- while `OperationDriver::HandleFailure` infers `was_pending` from OpId validity (`operation_->Aborted(status, op_id_copy_.load().valid())`). `Operation::Aborted` then calls `RemovedFromPending` for an operation that was never added as pending.

`SplitOperation` (always) and `SnapshotOperation` (RESTORE operations) register an operation filter in `AddedAsPending` and unregistered it unconditionally in `RemovedFromPending`. Unregistering the never-registered filter erases an unlinked node from the tablet's intrusive operation-filter list (`operation_filters_.erase(operation_filters_.iterator_to(*filter))`), which is undefined behavior and crashes in practice.

This change tracks filter registration in both operations and unregisters only when the filter was actually registered. A deeper follow-up option is to track added-as-pending explicitly in the driver/operation instead of inferring it from OpId validity; these targeted guards keep the fix minimal -- happy to take that on if reviewers prefer the root fix.

## Test plan

- [x] `./yb_build.sh release --cxx-test tablet_peer-test --gtest_filter TabletPeerTest.SplitAbortBeforePendingDoesNotUnregisterFilter` -- a `SplitOperation` submitted with `TEST_fail_operation_added_to_leader` aborts cleanly (no SIGSEGV), and the committed and logged OpIds are unchanged.
- [x] `./yb_build.sh release --cxx-test tablet_peer-test --gtest_filter TabletPeerTest.SnapshotAbortBeforePendingDoesNotUnregisterFilter` -- same for a `RESTORE_ON_TABLET` `SnapshotOperation`.

### DIFF
--- a/src/yb/tablet/operations/snapshot_operation.cc
+++ b/src/yb/tablet/operations/snapshot_operation.cc
@@ -150,11 +150,17 @@ bool SnapshotOperation::NeedOperationFilter() const {
 void SnapshotOperation::AddedAsPending(const TabletPtr& tablet) {
   if (NeedOperationFilter()) {
     tablet->RegisterOperationFilter(this);
+    filter_registered_.store(true, std::memory_order_release);
   }
 }
 
 void SnapshotOperation::RemovedFromPending(const TabletPtr& tablet) {
-  if (NeedOperationFilter()) {
+  // The operation can be aborted without ever having been added as pending: AddedAsPending runs
+  // at the end of Operation::AddedToLeader, whose earlier steps can fail (e.g. tablet_safe()
+  // while the tablet shuts down), while the abort path infers was_pending from OpId validity,
+  // which the operation driver sets before invoking the operation. Unregistering a
+  // never-registered filter would corrupt the tablet's operation-filter list.
+  if (filter_registered_.exchange(false, std::memory_order_acq_rel)) {
     tablet->UnregisterOperationFilter(this);
   }
 }
--- a/src/yb/tablet/operations/snapshot_operation.h
+++ b/src/yb/tablet/operations/snapshot_operation.h
@@ -13,6 +13,7 @@
 
 #pragma once
 
+#include <atomic>
 #include <mutex>
 #include <string>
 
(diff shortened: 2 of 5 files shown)


jmeehan16:
Resubmitting: this PR was opened directly against the stack-base branch, which makes the pr-title Backport check evaluate a non-master base and fail. The replacement follows the upstream feature-stack recipe (open against master, then retarget to the stack base). Replacement PR link to follow.

---
_automated · opencode (claude-fable-5)_

jmeehan16:
Replacement PR: https://github.com/yugabyte/yugabyte-db/pull/33404

src/yb/tablet/operations/snapshot_operation.cc
```diff
@@ -150,11 +150,17 @@ bool SnapshotOperation::NeedOperationFilter() const {
...
 }
 
 void SnapshotOperation::RemovedFromPending(const TabletPtr& tablet) {
-  if (NeedOperationFilter()) {
+  // The operation can be aborted without ever having been added as pending: AddedAsPending runs
+  // at the end of Operation::AddedToLeader, whose earlier steps can fail (e.g. tablet_safe()
+  // while the tablet shuts down), while the abort path infers was_pending from OpId validity,
+  // which the operation driver sets before invoking the operation. Unregistering a
+  // never-registered filter would corrupt the tablet's operation-filter list.
+  if (filter_registered_.exchange(false, std::memory_order_acq_rel)) {
```
spolitov: Do we care about races here?

PR #33404, opened 2026-08-20
[#33382] DocDB: Do not unregister a never-registered operation filter on abort

jmeehan16:
## Summary

> **Stacked PR**: builds on #33403 (`[#33380] DocDB: Roll back allocated Raft index when NotifyAddedToLeader fails`), whose commit this branch includes and whose `TEST_fail_operation_added_to_leader` hook the tests reuse. **Only the top commit is new to this PR**; review #33403 first. Once #33403 merges, this branch will be rebased and the diff here collapses to the single new commit.

An operation aborted after its Raft OpId was assigned but before it was added as pending takes the was-pending abort path: `OperationDriver::AddedToLeader` stores `op_id_copy_` *before* invoking `Operation::AddedToLeader`, whose first step (`tablet_safe()`) can fail while the tablet shuts down -- before `AddedAsPending` runs -- while `OperationDriver::HandleFailure` infers `was_pending` from OpId validity (`operation_->Aborted(status, op_id_copy_.load().valid())`). `Operation::Aborted` then calls `RemovedFromPending` for an operation that was never added as pending.

`SplitOperation` (always) and `SnapshotOperation` (RESTORE operations) register an operation filter in `AddedAsPending` and unregistered it unconditionally in `RemovedFromPending`. Unregistering the never-registered filter erases an unlinked node from the tablet's intrusive operation-filter list (`operation_filters_.erase(operation_filters_.iterator_to(*filter))`), which is undefined behavior and crashes in practice.

This change tracks filter registration in both operations and unregisters only when the filter was actually registered. A deeper follow-up option is to track added-as-pending explicitly in the driver/operation instead of inferring it from OpId validity; these targeted guards keep the fix minimal -- happy to take that on if reviewers prefer the root fix.

## Test plan

- [x] `./yb_build.sh release --cxx-test tablet_peer-test --gtest_filter TabletPeerTest.SplitAbortBeforePendingDoesNotUnregisterFilter` -- a `SplitOperation` submitted with `TEST_fail_operation_added_to_leader` aborts cleanly (no SIGSEGV), and the committed and logged OpIds are unchanged.
- [x] `./yb_build.sh release --cxx-test tablet_peer-test --gtest_filter TabletPeerTest.SnapshotAbortBeforePendingDoesNotUnregisterFilter` -- same for a `RESTORE_ON_TABLET` `SnapshotOperation`.

### DIFF
--- a/src/yb/tablet/operations/snapshot_operation.cc
+++ b/src/yb/tablet/operations/snapshot_operation.cc
@@ -150,11 +150,17 @@ bool SnapshotOperation::NeedOperationFilter() const {
 void SnapshotOperation::AddedAsPending(const TabletPtr& tablet) {
   if (NeedOperationFilter()) {
     tablet->RegisterOperationFilter(this);
+    filter_registered_.store(true, std::memory_order_release);
   }
 }
 
 void SnapshotOperation::RemovedFromPending(const TabletPtr& tablet) {
-  if (NeedOperationFilter()) {
+  // The operation can be aborted without ever having been added as pending: AddedAsPending runs
+  // at the end of Operation::AddedToLeader, whose earlier steps can fail (e.g. tablet_safe()
+  // while the tablet shuts down), while the abort path infers was_pending from OpId validity,
+  // which the operation driver sets before invoking the operation. Unregistering a
+  // never-registered filter would corrupt the tablet's operation-filter list.
+  if (filter_registered_.exchange(false, std::memory_order_acq_rel)) {
     tablet->UnregisterOperationFilter(this);
   }
 }
--- a/src/yb/tablet/operations/snapshot_operation.cc
+++ b/src/yb/tablet/operations/snapshot_operation.cc
@@ -150,7 +150,7 @@ bool SnapshotOperation::NeedOperationFilter() const {
 void SnapshotOperation::AddedAsPending(const TabletPtr& tablet) {
   if (NeedOperationFilter()) {
     tablet->RegisterOperationFilter(this);
-    filter_registered_.store(true, std::memory_order_release);
+    filter_registered_ = true;
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/yb/tablet/operations/snapshot_operation.cc
```diff
@@ -150,11 +150,17 @@ bool SnapshotOperation::NeedOperationFilter() const {
...
 }
 
 void SnapshotOperation::RemovedFromPending(const TabletPtr& tablet) {
-  if (NeedOperationFilter()) {
+  // The operation can be aborted without ever having been added as pending: AddedAsPending runs
+  // at the end of Operation::AddedToLeader, whose earlier steps can fail (e.g. tablet_safe()
+  // while the tablet shuts down), while the abort path infers was_pending from OpId validity,
+  // which the operation driver sets before invoking the operation. Unregistering a
+  // never-registered filter would corrupt the tablet's operation-filter list.
+  if (filter_registered_.exchange(false, std::memory_order_acq_rel)) {
```
spolitov: Do we care about race here?

spolitov (changes requested):

src/yb/tablet/operations/snapshot_operation.cc
jmeehan16: Good question. The `atomic` was misleading here, so it's been removed.

In reality there is no concurrent access possible here. There are two touchpoints that are serialized:
- A pre-replication abort runs on the submitting thread strictly after Operation::AddedToLeader returns: the failure propagates up the same call stack to OperationDriver::HandleFailure, which calls Aborted → RemovedFromPending.
- Post-replication completion (Replicated/Aborted via ReplicationFinished) happens-after the append that ran AddedAsPending, ordered by the ReplicaState lock and the driver's lock_; replication_state_ makes the two terminal paths mutually exclusive, so RemovedFromPending runs at most once.

PR #33447, opened 2026-08-21
[#33434] DocDB: Stop framing every yb-admin remote error as a version mismatch

### LINKED ISSUE #33434: [DocDB] yb-admin reports "The cluster doesn't support <operation>" for every remote error
Jira Link: [DB-23118](https://yugabyte.atlassian.net/browse/DB-23118)

## Summary

Every remote RPC failure in `yb-admin` is reported as `The cluster doesn't support <operation>`, whatever the real error was. A permission denial, a not-ready leader, an unreachable peer — all of them get framed as a version-compatibility problem.

## The bug

`ClusterAdminCli::RunCommand()` on `master` (`src/yb/tools/yb-admin_cli.cc:502`):

```cpp
if (s.IsRemoteError() && s.ToString().find("rpc error 2")) {
```

The comparison against `std::string::npos` is missing. `find()` returns `npos` — a very large, non-zero value — when the substring is absent, so the condition is **true when the text is not found**. When the text *is* found, `find()` returns a positive index, which is also true. The branch is therefore taken for every `RemoteError`.

Present since `10ef220697d` (2022-07-13).

## Impact

```
The cluster doesn't support <operation>: <the real error>
```

The real status is still printed, so no information is lost. The framing is what is wrong, and it is wrong on an error path where the reader is least able to second-guess it.

## Fix

Detect the condition by RPC error code instead of by scraping rendered text:

```cpp
bool IsUnsupportedRpcError(const Status& s) {
  const auto rpc_error = rpc::RpcError(s);
  return rpc_error == rpc::ErrorStatusPB::ERROR_NO_SUCH_METHOD ||
         rpc_error == rpc::ErrorStatusPB::ERROR_NO_SUCH_SERVICE;
}
```

`ERROR_NO_SUCH_METHOD` (2) and `ERROR_NO_SUCH_SERVICE` (3) are the two codes `Messenger::QueueInboundCall()` produces for a call it cannot route — the method missing, or the whole service missing. A cluster that predates the operation returns one or the other. `client.cc` and `client_master_rpc.cc` already test the same condition this way. A status carrying no RPC code decodes to 0, so it never matches.

Adding just the missing `!= std::string::npos` would also fix the reported bug, but leaves the intent expressed as a string match. The code check is the same size and cannot drift if the rendered text changes.

## Reproduce

Point `--master_addresses` at a master that is up but has no elected leader, then run an operation that requires the leader. The output is prefixed with `The cluster doesn't support <operation>` instead of `Error running <operation>`.

## Status

Fixed in [#33470](https://github.com/yugabyte/yugabyte-db/pull/33470) (first two commits).

Found during the [#33431](https://github.com/yugabyte/yugabyte-db/issues/33431) design review, but independent of that work.

## Issue Type

`kind/bug`

[DB-23118]: https://yugabyte.atlassian.net/browse/DB-23118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ag26jan:
## Summary

Fixes [#33434](https://github.com/yugabyte/yugabyte-db/issues/33434). `RunCommand()` decides whether to print `The cluster doesn't support <operation>` by looking for `rpc error 2` (`ERROR_NO_SUCH_METHOD`) in the status text — but the `find()` result was used as a bool, and `npos` is truthy, so the branch fired for **every** `RemoteError`. A leaderless master or an unreachable peer was reported as a version-compatibility problem, on the error path where a user is least able to second-guess the framing. Present since `10ef220697d` (2022).

The predicate moves to `yb-admin_util` as `IsNoSuchMethodError()` with the missing `!= std::string::npos`. The extraction is what makes it testable: `yb-admin_cli.cc` compiles only into the binary, while `yb-admin_util.cc` is in `yb-admin_lib`, which `yb-admin-test` links — so the regression is pinned by a unit test with synthetic statuses instead of arranging a real non-`NO_SUCH_METHOD` remote error in a cluster. Matching the rendered text rather than the structural RPC error code is unchanged; that follow-up is noted in the issue.

**Stacked on [#33442](https://github.com/yugabyte/yugabyte-db/pull/33442)** (which is stacked on #33395): the fix touches the same `RunCommand()` the 1b-b restructure reshaped, so only the top commit (`[#33434] DocDB: Stop framing every yb-admin remote error as a version mismatch`) is new here.

## Test plan

- [x] New unit test `AdminCliTest.NoSuchMethodErrorDetection` (no cluster needed): `RemoteError` containing `rpc error 2` → true; generic `RemoteError` → false (the regression); non-remote status with the same text → false.
- [x] `yb-admin-test` help/regression tests pass with the change: 9/9 (`Help*`, `InvalidOperationSuggestsClosestCommands`, `MalformedInitMasterAddrs`, `PrintArgumentExpressions`, `NoSuchMethodErrorDetection`), release build, macOS arm64.
- [ ] Full CI on this PR.

### DIFF
--- a/bin/yugabyted
+++ b/bin/yugabyted
@@ -3124,9 +3124,12 @@ class ControlScript(object):
         status_details = [(Output.make_yellow("Admin operation"),
                            "Find the response below."),]
         Output.print_out(self.get_status_string_common(status_details))
-        if 0 == ret_code:
+        # yb-admin writes usage and help text to stdout even when it exits non-zero, so show
+        # stdout unconditionally; discarding it on failure rendered usage errors as an empty red
+        # block (#33433).
+        if out:
           Output.print_out(out)
-        else:
+        if 0 != ret_code:
           Output.log(err)
           Output.print_out(Output.make_red(err))
 
--- a/docs/content/stable/admin/yb-admin.md
+++ b/docs/content/stable/admin/yb-admin.md
@@ -59,10 +59,22 @@ yb-admin \
 
 ### Online help
 
-To display the online help, run `yb-admin --help` from the YugabyteDB home directory.
+To list all operations, run `yb-admin help` from the YugabyteDB home directory.
 
 ```sh
-./bin/yb-admin --help
+./bin/yb-admin help
+```
+
+To display the usage of a single operation, pass its name to `help`:
+
+```sh
+./bin/yb-admin help delete_table
+```
+
+You can also pass any text to list the operations whose name contains it. For example, to list all snapshot operations:
+
+```sh
+./bin/yb-admin help snapshot
 ```
 
 ## Commands
--- a/src/yb/tools/yb-admin-test.cc
+++ b/src/yb/tools/yb-admin-test.cc
@@ -32,7 +32,9 @@
 // Tests for the yb-admin command-line tool.
 
 #include <algorithm>
+#include <fstream>
 #include <regex>
+#include <sstream>
 #include <thread>
 
 #include <boost/algorithm/string.hpp>
@@ -66,6 +68,7 @@
 #include "yb/util/backoff_waiter.h"
 #include "yb/util/date_time.h"
 #include "yb/util/format.h"
+#include "yb/util/monotime.h"
 #include "yb/util/json_document.h"
 #include "yb/util/net/net_util.h"
 #include "yb/util/scope_exit.h"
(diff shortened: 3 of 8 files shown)


ag26jan:
Reopened as #33469 — same 30-line fix, now based directly on `master` instead of on a three-PR stack, so the diff shows only the change itself rather than 801 lines of stacked context.

The unit test moved with it. One thing that surfaced when splitting this out: Phase 1c's test depended on the `#include "yb/tools/yb-admin_util.h"` that this commit introduced, so separating them broke that build until the include was added on the other side.
---

Issue #32640, opened 2026-07-14
[DocDB] Improve yb-admin help UX with progressive disclosure


ag26jan:
Jira Link: [DB-22370](https://yugabyte.atlassian.net/browse/DB-22370)

## Summary

`yb-admin` has ~133 operations, and its help treats every question the same way: dump everything. This umbrella issue makes help **progressive** — short by default, more detail only when asked for — without changing any command's name, syntax, or behavior, and without ever auto-running a suggested command.

The everyday pain, concretely:

- `yb-admin` with no operation prints the full 133-operation catalog plus placeholder definitions.
- `yb-admin --help` — the exact command [our own docs tell users to run](https://docs.yugabyte.com/stable/admin/yb-admin/) — prints `Warning: SetUsageMessage() never called`, then 1,564 lines of raw gflags output covering 118 source files.
- A near-miss like `list_server` gets no suggestion at all.

## Roadmap

| Phase | What it delivers | Design | Status |
|---|---|---|---|
| 1a | Misspelled operation → closest-match suggestions instead of the full dump | [#9624](https://github.com/yugabyte/yugabyte-db/issues/9624) | ✅ Done ([#32513](https://github.com/yugabyte/yugabyte-db/pull/32513)) |
| 1b-a | Four usage-text defect fixes (bracketed placeholders, numbering gaps, gflags leak, redundant footer) | [#33017](https://github.com/yugabyte/yugabyte-db/issues/33017) | In review ([#33470](https://github.com/yugabyte/yugabyte-db/pull/33470)) |
| 1b-b | Repair all twelve help surfaces; add the `help` operation; short overview by default | [#33431](https://github.com/yugabyte/yugabyte-db/issues/33431) | In review ([#33470](https://github.com/yugabyte/yugabyte-db/pull/33470)) |
| 1c | Token suggestions for abbreviations (`list_server` → tablet-server commands) | this issue, sketch below | In review ([#33470](https://github.com/yugabyte/yugabyte-db/pull/33470)) |
| 2b | Command metadata registry: grouped help, per-operation descriptions, per-command flags | [#33449](https://github.com/yugabyte/yugabyte-db/issues/33449) | Prototyped; PR follows after [#33470](https://github.com/yugabyte/yugabyte-db/pull/33470) lands |
| 2c | Global `--output json` flag: JSON output for data operations (3 list operations so far), so scripts stop parsing text | this issue; sub-issue filed at PR time | Prototyped |
[...]

PR #33448, opened 2026-08-21
[#32640] DocDB: yb-admin help UX Phase 1c: token-match suggestions for abbreviated operations

ag26jan:
## Summary

Phase 1c of [#32640](https://github.com/yugabyte/yugabyte-db/issues/32640). An abbreviation like `list_server` gets no suggestion today: it is not a prefix of any operation, and at edit distance 7 from `list_tablet_servers` it is far beyond the fuzzy tier's tolerance. Widening that tolerance is not an option — it makes arbitrary garbage match random commands.

This adds a third suggestion tier, tried only when the prefix and fuzzy tiers both return nothing: split the typed operation and each command name on `_` and suggest the commands whose tokens cover every typed token, where a token covers another when either is a prefix of the other (so `server` finds `servers` and vice versa). Ranked by fewest uncovered name tokens, then alphabetically; capped at 5. Requiring every typed token to be covered is the precision guard.

```
$ yb-admin list_server
Invalid operation: list_server
Did you mean one of these?
  list_tablet_servers
  list_all_tablet_servers

Run 'yb-admin help' to list all operations.
```

The matcher lives in `yb-admin_util` as `SuggestByNameTokens()` so its semantics are pinned by a unit test with synthetic name lists — `yb-admin_cli.cc` compiles only into the binary and is not unit-testable. Existing suggestion output is unchanged: the new tier never runs when the earlier tiers produce candidates, and garbage input still gets no suggestions.

**Stacked on [#33447](https://github.com/yugabyte/yugabyte-db/pull/33447)** (→ #33442 → #33395): only the top commit (`[#32640] DocDB: yb-admin help UX Phase 1c: token-match suggestions for abbreviated operations`) is new here.

## Test plan

- [x] New unit test `AdminCliTest.TokenMatchSuggestions` (no cluster needed): ranking, bidirectional prefix coverage (`server` ↔ `servers`), cap, disqualification when any typed token is uncovered, empty/underscore-only input.
- [x] New stanza in `AdminCliTest.InvalidOperationSuggestsClosestCommands` runs the real binary with `list_server` and asserts the tablet-server listings are suggested.
- [x] `yb-admin-test` help/regression tests pass with the change: 10/10 (`Help*`, `InvalidOperationSuggestsClosestCommands`, `MalformedInitMasterAddrs`, `PrintArgumentExpressions`, `NoSuchMethodErrorDetection`, `TokenMatchSuggestions`), release build, macOS arm64.
- [ ] Full CI on this PR.

### DIFF
--- a/bin/yugabyted
+++ b/bin/yugabyted
@@ -3124,9 +3124,12 @@ class ControlScript(object):
         status_details = [(Output.make_yellow("Admin operation"),
                            "Find the response below."),]
         Output.print_out(self.get_status_string_common(status_details))
-        if 0 == ret_code:
+        # yb-admin writes usage and help text to stdout even when it exits non-zero, so show
+        # stdout unconditionally; discarding it on failure rendered usage errors as an empty red
+        # block (#33433).
+        if out:
           Output.print_out(out)
-        else:
+        if 0 != ret_code:
           Output.log(err)
           Output.print_out(Output.make_red(err))
 
--- a/docs/content/stable/admin/yb-admin.md
+++ b/docs/content/stable/admin/yb-admin.md
@@ -59,10 +59,22 @@ yb-admin \
 
 ### Online help
 
-To display the online help, run `yb-admin --help` from the YugabyteDB home directory.
+To list all operations, run `yb-admin help` from the YugabyteDB home directory.
 
 ```sh
-./bin/yb-admin --help
+./bin/yb-admin help
+```
+
+To display the usage of a single operation, pass its name to `help`:
+
+```sh
+./bin/yb-admin help delete_table
+```
+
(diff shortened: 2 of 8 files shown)


ag26jan:
Superseded by #33470, which now carries Phase 1c along with Phase 1b.

Folded because the four-PR stack all targeted `master`: this PR rendered 937 lines when only ~136 were new, and the diff repeated everything below it. Phase 1c genuinely depends on the `help` operation for its "Run 'yb-admin help'" footer, so it is the one part that could not be published independently — it is the last of the seven commits in #33470.

The token matcher and its unit test are unchanged. The #33434 fix that sat below this one is now separate as #33469.
---

PR #33457, opened 2026-08-23
[#33454] DocDB: don't double-count rows scanned when a response is size-truncated

### LINKED ISSUE #33454: [YSQL] Rows scanned metrics double count the row rescanned after response size truncation
Jira Link: [DB-23133](https://yugabyte.atlassian.net/browse/DB-23133)
### Description

The rows scanned metrics (`scanned_table_rows_` / `scanned_index_rows_` in `src/yb/docdb/pgsql_operation.cc`, surfaced as `Storage Table Rows Scanned` / `Storage Index Rows Scanned` in `EXPLAIN (ANALYZE, DIST)` and as the session statistics behind pg_stat_statements storage metrics) count a row twice when a page ends by exceeding the response size limit.

Mechanism: in `PgsqlReadOperation::ExecuteScalar`, `++scanned_table_rows_` happens before the size check. When appending the row pushes the response buffer over the limit, the row is truncated from the response and the paging state is set to `ReadKey::kCurrent`, so the next request scans the same row again and counts it a second time. The same pattern exists in `PgsqlReadOperation::ExecuteBatchKeys`, where the truncated key is not counted in `batch_arg_count` and is resent by the client, and for the index row that produced the truncated table row when the scan is driven by a colocated index (the index iterator also resumes at the current row).

The overcount is one row per size-truncated page. A page is only guaranteed to return one row, so with rows wide enough that every page truncates after a single result, an N row table reports up to ~2N rows scanned. This is not hypothetical: the expected output of the `yb.orig.fetch_limits` regress test records `Storage Table Rows Scanned: 9613` for a 5000 row table (batch keys path under `yb_fetch_size_limit = '1kB'`), and that test's header comment documents the rescan counting.

Impact:
- `EXPLAIN (ANALYZE, DIST)` rows scanned values are inflated for any scan that pages by response size, which misleads the rows-scanned-to-returned efficiency reading.
- #30372 derives the base table `pg_class.reltuples` and `pg_stat_progress_create_index.tuples_done` from this metric when a partial index predicate is pushed down to the base table scan of an index backfill. Those consumers need the count to equal the rows of the table exactly for the values to be right and for their tests to assert exact values.

Proposed fix: decrement the counters when the row is truncated, in both fetch loops and for the accompanying colocated index row, so the metrics report each row once. The number of read requests still reflects the rescans, so paging behavior remains observable. The `yb.orig.fetch_limits` and `yb.orig.fetch_limits_joins` expected output changes accordingly (rows scanned values drop to the actual row counts, read request values unchanged).

Note this redefines what the rows scanned metrics mean for size-truncated scans: rows examined, counted once, instead of row read operations including rescans. `yb.orig.fetch_limits` deliberately asserted the latter.

### Environment

- Observed on master commit de6443e8ef30d5b8ec7718eef924998f5b616f6f (also present in all releases with the size-limit truncation logic).

[DB-23133]: https://yugabyte.atlassian.net/browse/DB-23133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

waterWang:
In `PgsqlReadOperation::ExecuteScalar` and `PgsqlReadOperation::ExecuteBatchKeys`, `scanned_table_rows_` is incremented before the response size check. When appending a row pushes the response buffer over the size limit, the row is truncated from the response and the paging state is set to `ReadKey::kCurrent` (or `batch_arg_count` is set to `processed_keys`), so the next request scans the same row again and counts it a second time.

Decrement the counters when the row is truncated, so the `Rows Scanned` metrics report each row exactly once. The number of read requests still reflects the rescans, so paging behavior remains observable.

**For the colocated index path:** the index iterator (`index_state->scanned_rows`) also resumes at the current row, so decrement that counter too.

**Test expected outputs:** `yb.orig.fetch_limits` and `yb.orig.fetch_limits_joins` expected output will change — rows scanned values drop to the actual row counts, while read request values are unchanged. The test header comment has been updated to reflect the new semantics.

Fixes #33454.

### DIFF
--- a/src/postgres/src/test/regress/expected/yb.orig.fetch_limits.out
+++ b/src/postgres/src/test/regress/expected/yb.orig.fetch_limits.out
@@ -1,4 +1,7 @@
--- If size limit is exceeded, last row has to be rescanned.
+-- If size limit is exceeded, the last row is truncated from the response and
+-- rescanned on the next request. The rows scanned metric counts each row
+-- once (the rescanned row is not double-counted). The number of read
+-- requests still reflects the rescans, so paging behavior remains observable.
 -- with too large row and too low size limit it means rescan of the each row,
 -- except the last on the tablet.
 -- Use explicit number of tablets to achieve predictable number of rows scanned
--- a/src/postgres/src/test/regress/sql/yb.orig.fetch_limits.sql
+++ b/src/postgres/src/test/regress/sql/yb.orig.fetch_limits.sql
@@ -1,4 +1,7 @@
--- If size limit is exceeded, last row has to be rescanned.
+-- If size limit is exceeded, the last row is truncated from the response and
+-- rescanned on the next request. The rows scanned metric counts each row
+-- once (the rescanned row is not double-counted). The number of read
+-- requests still reflects the rescans, so paging behavior remains observable.
 -- with too large row and too low size limit it means rescan of the each row,
 -- except the last on the tablet.
 -- Use explicit number of tablets to achieve predictable number of rows scanned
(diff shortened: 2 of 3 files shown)


jasonyb (changes requested):
fetch_limits_joins needs updating. also, this is being discussed internally, which may mean abandoning the approach.

PR #33470, opened 2026-08-24
[#33431,#32640,#33434] DocDB: Make yb-admin help progressive and fix remote-error framing

### LINKED ISSUE #33017: [DocDB] yb-admin help UX Phase 1b-a: fix SetUsage defects (bracketed placeholders, numbering, gflags leak, footer)
Jira Link: [DB-22736](https://yugabyte.atlassian.net/browse/DB-22736)

## Summary

Four defects in yb-admin's usage text. Each is a real bug on its own, independent of the help restructure, and separately backportable.

Phase **1b-a** of [#32640](https://github.com/yugabyte/yugabyte-db/issues/32640).

## The four defects

**1. Bracketed placeholders lose their definitions.** `GetArgumentExpressions()` matches placeholders by exact whitespace-split token equality, so a placeholder inside optional-argument brackets never matches and its definition is silently omitted. Affects `import_snapshot` (`[<namespace> <table_name> [<table_name>]...]`) and `list_change_data_streams` (`[<namespace>]`).

**2. The operation list numbers by the wrong index.** The list numbers each command by its index in the full command table, hidden commands included, so a hidden command's slot leaves a permanent gap in the visible numbering — `85.` followed directly by `87.`.

**3. The help path leaks gflags internals.** `main()` calls `google::ShowUsageWithFlagsRestrict(argv[0], __FILE__)`, which appends every gflag defined in the file with its build-relative source path, C++ type, and default:

```
  Flags from ../../src/yb/tools/yb-admin_cli.cc:
    -exclude_dead (Exclude dead tservers from output) type: bool default: false
    ...
```

**4. A redundant global footer.** After the ~133-operation list, a footer prints the `<namespace>` / `<table>` / `<index>` definitions unconditionally, though only a minority of operations use one. `RunCommand()` already prints the relevant definition next to the specific operation when its arguments are invalid.

For that per-operation mechanism to fully replace the footer it must be self-contained: the `<table>` and `<index>` definitions are themselves written in terms of `<namespace>`, so the `<namespace>` definition has to be emitted transitively whenever they are.

Riding along on the same lines: reformatting the three definition strings, and adding `--init_master_addrs` to the flags section.

## Where the fixes go

| Area | Location | Change |
|---|---|---|
| `GetArgumentExpressions()` | `yb-admin_cli.cc` | Strip leading `[` and trailing `]` / `.` before comparing; emit each definition at most once; emit `<namespace>` whenever `<table>` / `<index>` is emitted |
| Operation numbering | `ClusterAdminCli::SetUsage` | Number only the operations actually printed |
| Global footer | `ClusterAdminCli::SetUsage` | Remove |
| Gflags dump | `main()` | Print `google::ProgramUsage()` instead of `ShowUsageWithFlagsRestrict` |
| Definition strings | `yb-admin_cli.cc` | Reformat; update matching literals in the test |

## Acceptance criteria

- [x] `import_snapshot` and `list_change_data_streams` with bad arguments print their `<namespace>` definition, where they previously printed none. A repeated placeholder (`create_snapshot`'s `<table> [<table>]...`) prints its definition once, not twice.
- [x] A malformed `--init_master_addrs` (`host:99999`, `host:abc`, `,`) prints an error naming the flag and exits 1 — not the gflags unset-usage warning, and not a SIGSEGV ([#33435](https://github.com/yugabyte/yugabyte-db/issues/33435)).
- [x] The printed operation list is numbered `1..N` with no gap or duplicate, asserted generically rather than against current hidden-command positions.
- [x] No operator-facing path prints `Flags from` or any build-relative source path.
- [x] The global definitions footer is gone, and a `<table>` / `<index>` operation's bad-argument output defines `<namespace>` too.
- [x] No change to command names, argument syntax, or auto-execution of suggested commands.

## Out of scope

The header restructure, the `help` operation, the short overview, and the repair of the other gflags `--help*` surfaces are [#33431](https://github.com/yugabyte/yugabyte-db/issues/33431) (Phase 1b-b). Abbreviation matching is Phase 1c. Machine-readable catalog is [#33432](https://github.com/yugabyte/yugabyte-db/issues/33432). Grouping and per-operation descriptions are [#33449](https://github.com/yugabyte/yugabyte-db/issues/33449).

## Status

Fixed in [#33470](https://github.com/yugabyte/yugabyte-db/pull/33470), as its own commit.

Prior work: [#9624](https://github.com/yugabyte/yugabyte-db/issues/9624) / [#32513](https://github.com/yugabyte/yugabyte-db/pull/32513) (closest-match suggestions), [#32408](https://github.com/yugabyte/yugabyte-db/pull/32408) (`--flagfile` tip).

## Issue Type

`kind/improve-ux`

[DB-22736]: https://yugabyte.atlassian.net/browse/DB-22736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### LINKED ISSUE #33431: [DocDB] yb-admin help UX Phase 1b: repair all twelve help surfaces, add a help operation
Jira Link: [DB-23114](https://yugabyte.atlassian.net/browse/DB-23114)

**This issue body is the design doc for phase 1b of [#32640](https://github.com/yugabyte/yugabyte-db/issues/32640) — review it here.** Implemented by [#33470](https://github.com/yugabyte/yugabyte-db/pull/33470).

**Scope:** every help surface yb-admin exposes — the `help` operation, `help <operation>`, `--help`, and the seven other gflags `--help*` flags reachable today.
**Out of scope:** token matching for abbreviations (phase 1c), descriptions and grouping ([#33449](https://github.com/yugabyte/yugabyte-db/issues/33449)), JSON catalog ([#33432](https://github.com/yugabyte/yugabyte-db/issues/33432)), completion and man page (phase 3).

All measurements are from a `release-clang21-dynamic-arm64-ninja` build containing the phase 1b-a fixes ([#33017](https://github.com/yugabyte/yugabyte-db/issues/33017)).

---

## 1. The problem, measured

yb-admin has **twelve** reachable help surfaces. Nine are broken, and the one the documentation tells users to run is the worst.

| Invocation | Lines | Exit | What you actually get |
|---|--:|--:|---|
| `yb-admin` | 151 | 1 | Overview + all 133 operations, on stdout |
| `yb-admin --help` / `-help` | **1564** | 1 | `Warning: SetUsageMessage() never called`, then every flag of 118 source files |
| `yb-admin <operation> --help` | 1564 | 1 | The same dump — gflags parses flags after positionals |
| `yb-admin -h` | 1 | 1 | `ERROR: unknown command line flag 'h'` |
| `yb-admin --helpfull` | 1564 | 1 | Warning, then the same dump |
| `yb-admin --helpshort` | 3 | 1 | Warning, then `No modules matched: use -help` |
| `yb-admin --helppackage` | 1 | 1 | `Unable to find a package for file=yb-admin` |
| `yb-admin --helpon=<module>` | varies | 1 | Warning, then that file's flags |
| `yb-admin --helpmatch=<substr>` | varies | 1 | Warning, then matching files' flags |
| `yb-admin --helpxml` | 446 | 0 | Correct |
| `yb-admin --help_auto_flag_json` | 65 | 0 | Correct |
| `yb-admin --version` | 1 | 0 | Correct |

`docs/content/*/admin/yb-admin.md` tells the user to run `yb-admin --help`. That is row 2.

### 1.1 One cause, six symptoms

`ClusterAdminCli::Run()` calls `ParseCommandLineFlags()` but reaches `SetUsage()` only seventeen lines later. That call is **`yb::ParseCommandLineFlags`**, not `google::`; its body ends in `google::HandleCommandLineHelpFlags()`. So gflags renders help before a usage message exists.

That single ordering bug produces the `Warning:` header on all six of `--help`, `-help`, `--helpfull`, `--helpshort`, `--helpon` and `--helpmatch`. **Moving one call fixes all six.** `--helpxml`, `--help_auto_flag_json` and `--version` are unaffected because the wrapper handles them earlier and they never print the usage message.

### 1.2 `--helpshort` is not useless, it is misaddressed

gflags' `--helpshort` means "flags of the main module for this program", and it locates that module by matching the *program* name against the defining file — it looks for `yb-admin.cc`, but `main()` lives in `yb-admin_cli.cc`. Hence `No modules matched`. Proof: `--helpon=yb-admin_cli` prints exactly the flags you would want.

The flags yb-admin itself defines are **ten**, and they are the ten a user needs:

```
certs_dir_name   client_node_name   disable_graceful_transition   exclude_dead
init_master_addrs   master_addresses   read_time_wait_ms   timeout_ms
wait_if_no_leader_master   TEST_metadata_file_format_version
```

So the missing tier — *"what flags does this tool take?"* — is already computable. §6.2 uses it.

### 1.3 A second defect on the same path

`Run()` reads `FLAGS_init_master_addrs` and calls `HostPort::ParseStrings()`, which returns `InvalidArgument` **before** `SetUsage()`. `main()` then prints `google::ProgramUsage()` with nothing set, so the user sees the `Warning:` line as the entire error message. Reproduce with `--init_master_addrs=host:99999` or `=host:abc`.

Not with a value lacking a colon: `ParseString()` takes its `pos == npos` branch and returns OK with the default port, so `=not_a_host_port` fails later at connect time — a different path that never touches the usage message.

This is the **only** `InvalidArgument` that can escape `Run()` before `SetUsage()`: operation failures and the unknown-operation path both return `RuntimeError`. That is what makes the §7.3 reorder a complete fix rather than a partial one.

A third defect sits on the same two lines: `ParseStrings()` splits with `SkipEmpty()`, so `--init_master_addrs=,` yields zero addresses, returns OK, and indexes an empty vector — a verified SIGSEGV. Tracked separately as [#33435](https://github.com/yugabyte/yugabyte-db/issues/33435).

### 1.4 The 151-line catalog has no machine consumer

`docs/content/*/admin/yb-admin.md` is hand-written prose with its own grouping. No script, test, or completion script reads the catalog. This is not an argument that no consumer *should* exist — see §9 — only that today's format is answerable to no one.

---

## 2. Goals

1. Every one of the twelve surfaces gives a correct, useful answer. No surface prints a gflags warning.
2. No help surface needs a master connection. The `--help` flag spellings additionally need no valid `--flagfile` and no valid flag values, because they are answered before the parse.
3. The operation list, the `help <operation>` page, and the bad-argument error each come from one code path, so they cannot drift.
4. Finding one operation among 133 takes one command, not a scroll, and searching by *topic* works even though the names are verb-first.
5. No regressions: no removed capability, no new crash path, no changed exit code that a consumer depends on.

**Non-goals:** per-operation descriptions, grouping, aliases, a pager, and any new `--help-*` flag.

> **Known gap against goal 2.** The pre-parse scan covers the `--help` *flags*, not the `help` *operation*. `yb-admin help` still runs through the flag parse, so it fails on a missing `--flagfile`, an unparsable typed flag value, or an unknown flag. Follow-up, not addressed by this phase: intercepting a positional operation before the parse means deciding what to do with the flags that follow it.

---

## 3. The surface map after this phase

| Invocation | Lines | Exit | Stream | Change |
|---|--:|--:|---|---|
| `yb-admin` | 22 | 1 | stdout | Overview only; the operation list moves to `help` |
| `yb-admin --help` `-help` `--h` `-h` | 22 | **0** | stdout | Was 1564 lines / exit 1 |
| `yb-admin <operation> --help` | ~5 | **0** | stdout | Was 1564 lines / exit 1 |
| `yb-admin help` | 138 | 0 | stdout | New. Numbered, alphabetical, with argument syntax |
| `yb-admin help <text>` | 1–20 | 0 | stdout | New. Substring filter |
| `yb-admin help <operation>` | ~5 | 0 | stdout | New |
| `yb-admin help <unknown>` | ~5 | 1 | stderr | New. Suggestions |
| `yb-admin --helpshort` | ~35 | **0** | stdout | Overview + the ten own flags. Was 3 garbage lines |
| `yb-admin --helpfull` | 1583 | 1 | stdout | Unchanged in kind; gains a real header |
| `--helpon` `--helpmatch` `--helppackage` | varies | 1 | stdout | Gain a real header |
| `--helpxml` `--help_auto_flag_json` `--version` | — | 0 | stdout | Untouched |
| `yb-admin <unknown_operation>` | ~5 | 1 | stderr | Hint text changes |
| `yb-admin <operation>` with bad args | ~6 | 1 | stderr | Unchanged, plus §7.5 |

Two exit codes change, both 1 to 0, both on paths where the user asked for help and got it. Nothing in the tree consumes them: I grepped `bin/`, `managed/`, `python/`, `build-support/` and `java/` for every `--help*` spelling and found only YBA's `gflags_metadata/*.xml`, which are master/tserver dumps and do not run yb-admin.

Bare `yb-admin` keeps exit 1 and keeps stdout, matching `git` and the current binary.

### 3.1 One string, every surface

`SetUsage()` builds the overview once, hands it to `google::SetUsageMessage()`, and the interception path prints `prog_name + ": " + google::ProgramUsage()`. gflags formats its header identically — `progname: <usage>`, confirmed against `yb-ts-cli` — so the intercepted `--help` and the gflags-rendered `--helpfull` header are byte-identical by construction, not by discipline.

Because gflags prefixes `progname: `, the usage message must **open with a one-line summary of the tool**, not with `Usage:`. That is why line 1 of §4.1 exists; it is not decoration.

---

### 4.1 `yb-admin`, `--help`, `-help`, `--h`, `-h`

```
yb-admin: administer a YugabyteDB cluster from the command line.

Usage:
  yb-admin [global flags] <operation> [args]

Get help:
  yb-admin help                     List all 134 operations
  yb-admin help <operation>         Usage of one operation
  yb-admin help <text>              Operations whose name contains <text>
  yb-admin --helpshort              yb-admin's own global flags
  yb-admin --helpmatch=<substring>  Inherited flags matching a substring
  yb-admin --helpfull               All flags (long)

Common global flags:
  --master_addresses host:port[,host:port,...]  (default: localhost:7100)
  --init_master_addrs host:port                 (alternative to --master_addresses)
  --timeout_ms <millisec>                       (default: 60000)
  --certs_dir_name <dir>
  --flagfile <path>                             (use the master's server.conf)

Example:
  yb-admin --flagfile /path/to/master/conf/server.conf list_all_masters
```

Exit 1 with no operation (still an incorrect invocation); exit 0 for the flags. **22 lines**, which is what test 23 guards.

Three details are load-bearing:

- **The `Get help:` block** is what makes it acceptable to drop the operation list from the bare invocation, and it is the only place the flag tiers are discoverable.
- **It comes before `Common global flags:`.** A user with muscle memory for the bare catalog must hit the pointer to `help` in the first lines they read, not after a flag table.
- **The count is computed** at `SetUsage()` time (`std::count_if` over `commands_` for `!hidden_`). Do not hard-code it. It does the job a transitional "the list moved" notice would: a user learns in one glance that the list still exists *and* that it is complete, with no line to delete two releases from now.

The two `(default: …)` values are read from `google::GetCommandLineFlagInfo(...).default_value`, so they cannot drift from the `DEFINE_`.

### 4.2 `yb-admin help`

```
Operations (134):
    1. abort_snapshot_restore [<restoration_id>]
    2. add_namespace_to_xcluster_checkpoint <replication_group_id> <namespace_name>
  ...
  134. ysql_catalog_version [ysql.][<database_name>]

Run 'yb-admin help <operation>' for the usage of one operation.
Run 'yb-admin help <text>' to filter, e.g. 'yb-admin help snapshot'.
```

Numbered, alphabetical, argument syntax retained but truncated to one terminal line (§6.1). Hidden operations are excluded.

### 4.3 `yb-admin help <text>`

```
$ yb-admin help snapshot
Operations matching 'snapshot' (16):
  * abort_snapshot_restore [<restoration_id>]
  * create_database_snapshot [ysql.]<database_name> [retention_duration_hours] ...
  ...
  * restore_snapshot_schedule <schedule_id> (<timestamp> | minus <interval>)

Run 'yb-admin help <operation>' for the usage of one operation.
```

Not numbered — see §6.1.

### 4.4 `yb-admin help <operation>` and `yb-admin <operation> --help`

Byte-identical to the block `RunCommand()` prints on a bad-argument error, because it is the same function (§7.4).

```
$ yb-admin help delete_table
Usage: yb-admin delete_table <table>
Definitions: <table>
  <namespace> <table_name> | tableid.<table_id>
<namespace>
  [(ycql|ysql).]<namespace_name> (default: ycql.)
```

`Usage:` on one line is `RunCommand()`'s existing format; `Definitions: ` is `GetArgumentExpressions()`'s existing label. The `<namespace>` block is new — see §7.5.

### 4.5 `yb-admin --helpshort`

The overview of §4.1, then:

```
yb-admin's own flags:
  --certs_dir_name <string>                     (default: "")
  --disable_graceful_transition <bool>          (default: false)
  ...
  --timeout_ms <int64>                          (default: 60000)
```

Name, type and default for each, derived from `google::GetAllFlags()` — see §6.2.

### 4.6 Unknown operation, with or without `help`

```
$ yb-admin help delete_tabel
Invalid operation: delete_tabel
Did you mean one of these?
  delete_table

Run 'yb-admin help' to list all operations.
```

Suggestions come from the existing `GetSuggestedCommands()`. `yb-admin delete_tabel` prints the same thing.

---

## 5. Control flow

```mermaid
flowchart TD
    A["main(): ClusterAdminCli().Run(argc, argv)"] --> B["FLAGS_logtostderr, FLAGS_minloglevel"]
    B --> C["RegisterCommandHandlers()<br/>(no client, no flags)"]
    C --> D["SetUsage(prog_name)<br/>google::SetUsageMessage()"]
    D --> E{"ScanForHelpRequest(raw argv)"}
    E -- "--help / -h / --helpshort<br/>+ operation token" --> F["PrintCommandUsage()<br/>exit 0"]
    E -- "--help / -h only" --> G["PrintOverview()<br/>exit 0"]
    E -- "--helpshort only" --> H["PrintOverview() +<br/>own flags<br/>exit 0"]
    E -- "none" --> I["yb::ParseCommandLineFlags()<br/>(handles --helpfull/--helpxml/...)"]
    I --> J["InitGoogleLoggingSafe()"]
    J --> K{"argc >= 2 ?"}
    K -- no --> L["PrintOverview()<br/>exit 1"]
    K -- yes --> M{"operation registered ?"}
    M -- no --> N["Suggestions to stderr<br/>exit 1"]
    M -- yes --> O{"command.needs_client_ ?"}
    O -- no --> Q["RunCommand()"]
    O -- yes --> P["Read master flags<br/>Make client, client_->Init()"]
    P --> Q
```

Three properties follow from this order:

- The scan precedes the flag wrapper, so the `--help` spellings work with a malformed `--flagfile` or a garbage flag value. A help flag *inside* a flagfile is not intercepted (§7.4).
- Client construction moved **below** the operation lookup, so `help` never touches the network and `--init_master_addrs` is validated only when an operation needs it. This is the §1.3 fix.
- `help` reaches `RunCommand()` like every other operation, so its error formatting cannot diverge.

---

### 6.1 Keep the numbering. Keep the argument syntax, truncated.

- **Numbers** give an instant count and a scroll anchor, and keeping them is the zero-regression option.
- **Numbers appear in the full list only.** Global numbers in a filtered view (`12`, `41`, `87`) read as a rendering bug, and renumbering per view gives one operation two numbers depending on how you got there. Filtered entries get a `*` marker instead.
- **Numbers are display-only, forever.** `yb-admin help 42` must be an unknown operation, not operation #42. Written down here because otherwise someone will request it and we will have committed to numbering stable across releases.
- **Argument syntax is kept, truncated to one terminal line.** Measured over today's 133 operations: median entry 45 characters, p90 96, max **338** (`alter_universe_replication`); 12 entries exceed 100. Left alone those 12 wrap to 2–4 terminal lines each, so the list renders as ~160 unevenly-wrapped lines. Truncating at 100 columns with a trailing `...` gives 134 uniform lines and loses nothing, because the full syntax is one `help <operation>` away.
- The numbering-gap fix from #33017 and its regex test **must stay**: hidden commands used to leave permanent gaps (`85.` → `87.`).

The list changes from registration order to alphabetical. Registration order is near-arbitrary today, and `command_indexes_` is already a sorted `std::map`, so alphabetical is free. This renumbers every operation relative to the shipped binary — see §10.

### 6.2 Do not remove `--helpshort`. Repair it.

It **cannot** be removed: it is a `DEFINE_bool` inside third-party gflags. Tagging it hidden would only hide it from YB's own flag listings — passing it would still work. Real removal means patching thirdparty gflags.

It **should not** be removed either. §1.2 shows the tier it is supposed to fill is real, is missing, and is ten flags wide.

**Format the flags from `google::GetAllFlags()`, filtered on the basename of `flag_info.filename`; do not call `google::ShowUsageWithFlagsRestrict()`.** The two-line version is tempting, but it emits gflags' own `Flags from ../../src/yb/tools/…` headers — reintroducing exactly the build-relative path leak #33017 removed. Filtering `GetAllFlags()` is still fully auto-derived, so a hand-curated list cannot go stale, and it costs about fifteen lines instead of two. Match on basename so the filter does not depend on build layout.

Additionally filter on `FlagTag::kHidden`: the defining file alone is not sufficient, since `DEFINE_test_flag` tags a flag hidden but still defines it in `yb-admin_client.cc`, and a help surface that advertises a TEST-only knob invites its use.

`--helppackage`, `--helpon` and `--helpmatch` need no interception; §1.1's one-call move repairs their headers.

### 6.3 `--helpfull` stays the all-flags dump

Do **not** change its `hidden` tag. `util/flags/flags.cc` is shared with master and tserver, where `--help` still *is* `--helpfull`; advertise the flag from yb-admin's own overview instead and the blast radius stays inside yb-admin. Fix the now-false comment there that says helpfull is the same as help.

### 6.4 No pager, in any form

Not `less`, not an internal pager, not `isatty()`-gated auto-paging. yb-admin runs in break-glass and minimal-container contexts where `less` may be absent; its stdout is captured by subprocess in both `bin/yugabyted` and YBA; and fork/exec plus SIGPIPE plus terminal-state handling is new failure surface for something users already get from `| less`.

### 6.5 Filtering must be **substring**, not prefix

yb-admin operation names are verb-first, so topic words sit in the middle. Measured over the 133 visible names:

| `help <text>` | Prefix matches | Substring matches |
|---|--:|--:|
| `snapshot` | **0** | **16** |
| `xcluster` | 1 | 16 |
| `cdc` | **0** | 4 |
| `list` | 19 | 22 |
| `delete` | 12 | 12 |

`yb-admin help snapshot` finding zero operations while sixteen exist is worse than no filter at all. Use a case-insensitive substring match.

Prefix-then-fuzzy stays correct where it already lives: `GetSuggestedCommands()` handles *typos*, and a typo usually shares a prefix. Two different jobs, both needed.

Dispatch rule for `help <arg>`:

| `<arg>` | Output |
|---|---|
| a registered operation, exactly | Full usage + definitions (§4.4) |
| a substring of one or more names | Those entries with argument syntax (§4.3) |
| neither | `GetSuggestedCommands()` suggestions, exit 1 (§4.6) |

This ordering also makes typos degrade gracefully: `help snapsho` substring-matches 16 operations, while `help delete_tabel` matches nothing and falls through to the fuzzy suggestion of `delete_table`.

### 6.6 No grouped list, no columns, no descriptions here

Grouping needs per-operation topic metadata ([#33449](https://github.com/yugabyte/yugabyte-db/issues/33449)). Deriving groups from name prefixes would be wrong (`list_*` is a verb, not a topic). Columns leave no room for later descriptions, so adding them now guarantees a second rewrite.

### 6.7 No new flag, and no `help --all`

A third full-output surface recreates the problem this issue exists to remove, and a flag hides it from most users anyway. The machine-readable consumer that *is* missing is [#33432](https://github.com/yugabyte/yugabyte-db/issues/33432).

### 6.8 Bare `yb-admin` shows the overview only

This is the one deliberate discoverability trade, and the one users are most likely to notice. A user who runs `yb-admin`, needs `list_tables`, and does not know it exists now types two commands instead of one. Accepted because the 151-line wall was never a good answer and the `Get help:` block makes the replacement unmissable. Curating a "common operations" shortlist was rejected: a permanent bikeshed and a permanent staleness risk. Needs a release note (§10).

---

## 7. Implementation

**Files:** `src/yb/tools/yb-admin_cli.cc`, `yb-admin_cli.h`, `yb-admin-test.cc`, `docs/content/stable/admin/yb-admin.md`.

### 7.1 Two prerequisites

**(a) Delete the `DCHECK_ONLY_NOTNULL(client_)` at the top of `RegisterCommandHandlers()`.** Registration must run before the client exists, and that macro is a live `DCHECK_NOTNULL` in debug builds — leaving it aborts **every** debug-build invocation. The assertion is already vacuous: it is the only reference to `client_` in that function, since all 133 actions receive `ClusterAdminClient*` as a parameter.

**(b) Move the client-construction block below the operation lookup.** Otherwise `yb-admin help --init_master_addrs=garbage` still returns `InvalidArgument` before dispatch. This is also the §1.3 fix.

### 7.2 All of it in `Run()`, none in `main()`

`SetUsage()` and `RegisterCommandHandlers()` are `protected`, and `main()` is neither member nor friend. `Run()` also calls `RegisterCommandHandlers()`, so a second call would append to `commands_` and every operation would appear twice.

### 7.3 The new head of `Run()`

Keep `yb::ParseCommandLineFlags()`. Do **not** substitute `google::ParseCommandLineNonHelpFlags()` + `google::HandleCommandLineHelpFlags()`: that drops `--helpxml`, `--dump_flags_xml`, `--help_auto_flag_json`, `--dump_metrics_json`, the flag validators, the callbacks, the allowlist, and the preview-flag gate. Other tooling reads that metadata. Moving `SetUsage()` earlier is sufficient.

```cpp
Status ClusterAdminCli::Run(int argc, char** argv) {
  const string prog_name = BaseName(argv[0]);
  FLAGS_logtostderr = true;
  FLAGS_minloglevel = 2;

  RegisterCommandHandlers();   // No flag and no client dependency (see 7.1a).
  SetUsage(prog_name);         // Builds the overview.

  if (auto request = ScanForHelpRequest(argc, argv)) {
    PrintHelpRequest(*request, prog_name, std::cout);
    return Status::OK();       // Exit 0.
  }

  ParseCommandLineFlags(&argc, &argv, true);
  InitGoogleLoggingSafe(prog_name.c_str());
  ...
}
```

No `LOG()` on the help path — `InitGoogleLoggingSafe()` runs after the early return.

`prog_name` changes from the full `argv[0]` to `BaseName(argv[0])`. **It must change in both places.** `RunCommand()` is currently invoked with `args[0]`; miss it and `help <op>` prints `Usage: yb-admin …` while the bad-argument path prints `Usage: build/latest/bin/yb-admin …` — exactly the drift §7.4 exists to prevent, and the existing test only asserts the literal `"Usage:"`.

### 7.4 `ScanForHelpRequest()` and the shared printers

`ScanForHelpRequest()` reads raw `argv` before the parse. It recognizes:

- `--help`, `-help`, `--h`, `-h`, `--helpshort`, `-helpshort`, and each with `=<bool>`. gflags accepts `-` and `--` interchangeably, so both spellings are required; gflags has no `-h` at all, so that spelling is new and desirable.
- `=<bool>` is **parsed, not merely matched**. `--help=false`, `--help=0` and `--help=no` are not requests, and must keep falling through to the bare-invocation path.
- A stop at a bare `--`, matching gflags.
- **A token gflags would consume as a flag *value* is not a request.** gflags accepts the two-token form for every non-boolean flag, so `yb-admin --master_addresses --help` sets `master_addresses` to the string `--help`. A scan that misses this turns a malformed command line into help-and-exit-0. Rule: for a token `--foo` carrying no `=`, call `gflags::GetCommandLineFlagInfo("foo", &info)`, and if `info.type != "bool"`, skip the following token. The registry is populated by static initializers, so this works before the parse.

To find the operation, **scan every token; do not check `argv[1]`.** Real commands lead with a flag — `yb-admin --flagfile master/conf/server.conf delete_table --help` has `--flagfile` at `argv[1]`. Take the first token that is a key of `command_indexes_`. Operation arguments are free text, so a table named `list_tables` is legal; the scan therefore selects only *which help page to show* and never which operation to run. The worst outcome is the wrong help page, and no action can be triggered.

**One surface the scan does not reach.** A help flag supplied inside a `--flagfile` is invisible to an argv scan. After the reorder it still reaches gflags' handler — so it still dumps every flag and still exits 1 — but the `Warning:` line is gone, because `SetUsage()` has run. That is an improvement, not an interception. Do not chase it: reading the flagfile before the parse would duplicate gflags, and nobody writes `--help` into a `server.conf`.

Two functions move out of `RunCommand()` so both callers share them:

- `PrintCommandUsage(const Command&, const string& prog_name, std::ostream&)` — the `Usage:` line plus `GetArgumentExpressions()`.
- `PrintOverview(const string& prog_name, std::ostream&)` — emits `prog_name + ": " + google::ProgramUsage()` (§3.1).

`GetArgumentExpressions()` is not `const` today, so either make it `static` or leave `PrintCommandUsage()` non-`const`.

### 7.5 Transitive argument definitions

`GetArgumentExpressions()` reads only top-level tokens of `usage_arguments_`. `delete_table_args` is `"<table>"`, so the tool prints the `<table>` definition — which itself names `<namespace>` — and stops, leaving `<namespace>` undefined.

Collect referenced placeholders into a set, expand `<table>` → `<namespace>` and `<index>` → `<namespace>`, and emit each once in a stable order (referencing definition first, as in §4.4). Not two ad-hoc `if`s.

### 7.6 `help` as a real operation, routed through `RunCommand()`

Register `help` with `Register()`. Two problems disappear: `GetSuggestedCommands()` reads only registered operations, so without registration `yb-admin hepl` suggests nothing; and the operation list is built from registered operations, so `help` would be missing from it.

`help` must not build a client, but it should still get the standard error formatting. Add a field rather than a special case:

```cpp
struct Command {
  std::string name_;
  std::string usage_arguments_;
  Action action_;
  bool hidden_;
  bool needs_client_ = true;   // false for `help`
};
```

```cpp
auto& command = commands_[cmd->second];
if (command.needs_client_) {
  // the relocated FLAGS_init_master_addrs / client_.reset() / client_->Init() block
}
return RunCommand(command, command_args, prog_name);
```

`RunCommand()` only does `command.action_(command_args, client_.get())`, so a null client is fine for an action that ignores it. Six lines, and a general mechanism for any future client-free operation.

The action then needs almost no error handling of its own:

| Arguments | Behavior |
|---|---|
| none | Print the operation list, `Status::OK()` |
| one, exact operation | `PrintCommandUsage()`, `Status::OK()` |
| one, a substring of some names | Print the filtered list, `Status::OK()` |
| one, no match | Print suggestions to stderr, return `RuntimeError` |
| more than one | Return `kInvalidArguments` |

The last row is why the routing matters: `RunCommand()` already prints `Usage: yb-admin help [<operation>]` for `InvalidArgument` and converts it to `RuntimeError`, so `main()` prints no overview. Identical treatment to the other 133 operations. The no-match row returns `RuntimeError` directly so the suggestions are not preceded by an `Error running help:` line.

`help <hidden_operation>` prints that operation's usage; the list still omits it. Support engineers can then read the syntax of an operation whose name they already know.

### 7.7 `PrintOperationNames(std::ostream&, const string& filter = "") const`

Iterate `command_indexes_` (a sorted `std::map`, so alphabetical is free). Skip hidden commands. Keep an entry when `filter` is empty or is a case-insensitive substring of the name. Number only when `filter` is empty (§6.1). Print `name <usage_arguments_>`, truncated to 100 columns with a trailing `...`.

### 7.8 `main()` and a bad `--init_master_addrs`

`main()` prints `google::ProgramUsage()` when the status is `InvalidArgument`. Guard it with `IsUsageMessageSet()` so no path can ever print `Warning: SetUsageMessage() never called` again — belt and braces alongside §1.1, since a future reorder could reintroduce it.

For `--init_master_addrs`, prepend a targeted message and return `RuntimeError` instead of `InvalidArgument`, so a one-line problem does not get a 22-line answer.

---

## 8. Test plan

`src/yb/tools/yb-admin-test.cc`.

**Change:** `InvalidOperationSuggestsClosestCommands` — the no-operation path no longer contains `Operations:`, numbers, or the operation names; assert the overview markers instead. `PrintArgumentExpressions` — extend for §7.5.

**Add:**

| # | Assertion |
|--:|---|
| 1 | `help` prints every non-hidden operation, alphabetically, numbered 1..N with no gap or duplicate |
| 2 | `help <op>` and the bad-argument path of the same operation print an **identical** block from `Usage:` to the end, compared both directions with real strings |
| 3 | `help <unknown>` prints suggestions, exit 1, on stderr |
| 4 | `help <far-off garbage>` prints no suggestions |
| 5 | `help <hidden_operation>` prints its usage |
| 6 | `help extra1 extra2` prints `Usage: yb-admin help [<operation>]` to stderr, exit 1, no overview |
| 7 | `yb-admin hepl` suggests `help` |
| 8 | `--help`, `-help`, `--h`, `-h`, `--help=true` each print the overview, exit 0 |
| 9 | `yb-admin <op> --help` matches `yb-admin help <op>` exactly |
| 10 | `yb-admin --flagfile <master.conf> <op> --help` prints that operation's help — the test an `argv[1]` check fails |
| 11 | `yb-admin help --init_master_addrs=garbage` prints the list, exit 0 (covers §7.1b) |
| 12 | `yb-admin` with no operation prints the overview, exit 1 |
| 13 | `--helpshort` prints the overview, exit 0, and does **not** contain `No modules matched` |
| 14 | `--helpfull` **still** contains `Flags from` — the guard proving test 16 did not break the one surface that legitimately dumps flags |
| 15 | No output on any path contains `SetUsageMessage() never called` |
| 16 | No output on the overview, `help`, or `help <op>` paths contains `Flags from` |
| 17 | `help snapshot` returns the 16 substring matches, exit 0 — the test a prefix filter fails (§6.5) |
| 18 | `help 42` is an unknown operation, not operation #42 (§6.1) |
| 19 | `help <op> extra` and the `<op>` bad-argument path print the same `prog_name` spelling (§7.3) |
| 20 | Help text on stdout, error text on stderr, asserted per path |
| 21 | No line of `help` output exceeds 100 columns, and `alter_universe_replication`'s entry ends in `...` while `help alter_universe_replication` shows the full 338-character syntax |
| 22 | `yb-admin help --master_addresses 127.0.0.1:1` — a well-formed address on a closed port — prints the list, exit 0, in **under two seconds**. This is the one a missing `needs_client_` fails, by hanging for the full `--timeout_ms` |
| 23 | The bare-invocation overview is **at most 25 lines**. It is 22 today; the margin is for later phases, and the ceiling is what stops it regrowing into the 151-line wall |
| 24 | `yb-admin --master_addresses --help` behaves as it does today — stdout, exit 1, **not** help-and-exit-0 — because `--help` is that flag's value, not a request (§7.4) |

Rule for test 1: assert presence, ordering, and numbering — not the full text of a line, since a later phase appends a description to the same entry.

Tests 2 and 19 are the highest-value pair; they are the only defense against the divergence this design exists to prevent.

---

## 9. Documentation

Repoint `docs/content/stable/admin/yb-admin.md` at `yb-admin help` and add `yb-admin help <operation>`.

Five trees carry that pointer: `stable`, `v2025.2`, `v2025.1`, `v2024.2`, `v2.20`. **Change only `stable`**, and gate the entry on the release that will carry it — the others document released binaries where `--help` still dumps flags and no `help` operation exists.

---

## 10. Behavior changes needing a release note

1. **Bare `yb-admin` no longer lists the operations.** Write this as *the pipe breaking*, not the display changing: `yb-admin 2>/dev/null | grep snapshot` returns 16 lines today and will return nothing — exit 1, no error text — for anyone who has it in a script or a shell alias. `yb-admin help | grep snapshot` still works; `yb-admin help snapshot` is better. The notes are the only place this reaches someone whose script discards the output.
2. **`yb-admin --help` prints the overview** instead of every gflag, and exits 0 instead of 1.
3. **The operation list is alphabetical**, so every operation's number differs from earlier releases. Numbers were never stable and are not accepted as input.
4. **`yb-admin -h` works**, where it was previously an unknown-flag error.
5. **`--helpshort` prints the overview and yb-admin's own ten flags**, and exits 0.
6. **Remote failures that are not a missing method or service** now print `Error running <operation>` instead of `The cluster doesn't support <operation>` ([#33434](https://github.com/yugabyte/yugabyte-db/issues/33434)).
7. **A malformed `--init_master_addrs`** prints `Invalid --init_master_addrs '<value>': <reason>` and exits 1, where it previously printed the unset-usage warning or crashed ([#33435](https://github.com/yugabyte/yugabyte-db/issues/33435)).
8. **The operation-list marker changed** from `<operation> must be one of:` to `Operations:`, and the global definitions footer after the list is gone ([#33017](https://github.com/yugabyte/yugabyte-db/issues/33017)). No in-tree consumer (§1.4), but out-of-tree scripts are not covered by that grep.

---

## 11. Rejected alternatives

Recorded so they are not re-proposed.

- **A transitional `Note: the full operation list moved` line.** The computed count on the `help` line carries the same signal permanently. A self-expiring line does not expire; it would still be there, with a stale `TODO`, in three years.
- **An `isatty` split that keeps the legacy dump when stdout is a pipe.** Output that differs between terminal and pipe is the defect nobody can reproduce, and it forks all 24 tests. Break the pipe once, in the release notes.
- **`-h`-only interception, restricted to tokens before the first non-flag.** The restriction breaks `yb-admin delete_table --help`, the most natural thing a user types, which is why §7.4 scans every token.
- **`yb-admin ?` as a help alias.** `?` is a single-character glob. Under zsh's default `nomatch` it errors when the working directory holds no one-character file, and silently becomes `yb-admin a` when it does. Broken either way.
- **A `commands` synonym for `help`.** A second name for one surface, against §3.1.
- **A curated "Common operations" block, git-style.** The strongest known pattern for this migration, and genuinely tempting — but it is opinionated, it overlaps the later grouping work, and every line erodes the 25-line ceiling. Kept on the roadmap, not here.
- **Widening the fuzzy tier's edit-distance tolerance** instead of adding a token tier. Makes arbitrary garbage match random commands.

---

## Issue Type

`kind/improve-ux`

[DB-23114]: https://yugabyte.atlassian.net/browse/DB-23114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### LINKED ISSUE #33433: [yugabyted] configure admin_operation discards yb-admin stdout on non-zero exit, showing an empty red block
Jira Link: [DB-23116](https://yugabyte.atlassian.net/browse/DB-23116)

## Summary

`yugabyted configure admin_operation` shows an empty red block instead of yb-admin's output whenever yb-admin exits non-zero.

## The bug

`bin/yugabyted` prints stdout only on a zero exit code:

```python
out, err, ret_code = run_process(cmd, timeout=timeout, log_cmd=True)
...
if 0 == ret_code:
  Output.print_out(out)
else:
  Output.log(err)
  Output.print_out(Output.make_red(err))
```

yb-admin writes help and usage text to **stdout** while exiting 1. That is deliberate — it matches `git`. So for every yb-admin usage error surfaced through yugabyted, `out` holds the actual explanation and is discarded, while `err` is empty and gets printed in red.

## Reproduce

Run `yugabyted configure admin_operation` with no operation, or with an operation whose arguments are invalid. yugabyted renders an empty red block and the user sees nothing.

## Fix

Print `out` whenever it is non-empty, and keep the red block for the case where `err` actually has content:

```python
if out:
  Output.print_out(out)
if 0 != ret_code:
  Output.log(err)
  Output.print_out(Output.make_red(err))
```

## Status

Reproduces on `master` today — the current binary already prints its operation catalog to stdout with exit 1 — so this is **not** a regression from the help UX work ([#32640](https://github.com/yugabyte/yugabyte-db/issues/32640)). That work makes the gap easier to notice, because the stdout-only content becomes the short overview a user most needs to see.

Fixed in [#33470](https://github.com/yugabyte/yugabyte-db/pull/33470), as its own commit.

## Issue Type

`kind/bug`

[DB-23116]: https://yugabyte.atlassian.net/browse/DB-23116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### LINKED ISSUE #33434: [DocDB] yb-admin reports "The cluster doesn't support <operation>" for every remote error

### LINKED ISSUE #33435: [DocDB] yb-admin segfaults on --init_master_addrs=,
Jira Link: [DB-23119](https://yugabyte.atlassian.net/browse/DB-23119)

## Summary

`yb-admin --init_master_addrs=,` crashes with SIGSEGV before doing any work.

```
$ yb-admin --init_master_addrs=, list_tables
*** SIGSEGV (@0x17) received by PID 9758 ***
$ echo $?
139
```

`--init_master_addrs=,,` crashes the same way. A bare space does not — it parses as a single host.

## Root cause

`ClusterAdminCli::Run()` indexes the parsed vector without checking that anything was parsed (`src/yb/tools/yb-admin_cli.cc:532` on `master`):

```cpp
std::vector<HostPort> init_master_addrs;
RETURN_NOT_OK(HostPort::ParseStrings(
    FLAGS_init_master_addrs, master::kMasterDefaultPort, &init_master_addrs));
client_.reset(new ClusterAdminClient(
    init_master_addrs[0], MonoDelta::FromMilliseconds(FLAGS_timeout_ms)));
```

`HostPort::ParseStrings()` splits with `strings::SkipEmpty()`, so `","` yields **zero** fields. The loop body never runs, the function returns `Status::OK()`, and `init_master_addrs[0]` indexes an empty vector.

The usual guard does not help: `Run()` only reaches this block when `!FLAGS_init_master_addrs.empty()`, and `","` is not empty.

## Fix

Check the result at the call site, not inside `ParseStrings()`. That function has 39 callers outside `net_util`, and some legitimately accept an empty result (for example `tablet_server_main_impl.cc` parses possibly-unset bind-address flags), so rejecting empty there would change behavior for all of them.

The call site now reports a bad value on its own terms and returns `RuntimeError`:

```
$ yb-admin --init_master_addrs=, list_tables
Invalid --init_master_addrs ',': no addresses found
$ echo $?
1
```

`RuntimeError` rather than `InvalidArgument` because `main()` answers `InvalidArgument` by printing the whole usage overview — a 22-line reply to a one-line problem.

`--init_master_addrs` is defined only in `src/yb/tools/yb-admin_cli.cc`, so the blast radius is `yb-admin` alone.

## Related

A second defect lives on the same two lines: a bad *port* (`--init_master_addrs=host:99999`) returns `InvalidArgument` before `SetUsage()` has run, so the user sees `Warning: SetUsageMessage() never called` as the entire error message. Different failure, same block; tracked in [#33431](https://github.com/yugabyte/yugabyte-db/issues/33431) and fixed by the same reorder.

## Status

Pre-existing on `master`. Not caused by the help UX work ([#32640](https://github.com/yugabyte/yugabyte-db/issues/32640)), filed separately so it could be fixed on its own.

Fixed in [#33470](https://github.com/yugabyte/yugabyte-db/pull/33470), as its own commit.

## Issue Type

`kind/bug`

[DB-23119]: https://yugabyte.atlassian.net/browse/DB-23119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ag26jan:
## Summary

`yb-admin --help` — the command [the docs point at](https://docs.yugabyte.com/stable/admin/yb-admin/) — printed `Warning: SetUsageMessage() never called` and 1,564 lines of raw gflags output, because `Run()` parsed flags (where gflags renders help) before `SetUsage()`. Bare `yb-admin` dumped every operation. A typo like `list_server` got no suggestion. Separately, every remote RPC failure was reported as `The cluster doesn't support <operation>`.

This PR makes help progressive and fixes the bugs found on the way. Design notes and measurements of all help surfaces: #33431.

## UX changes

| Command | Before | After |
|---|---|---|
| `yb-admin` | Every operation plus a flag list | 22-line overview, exit 1 |
| `yb-admin --help` | Warning + raw flag dump, exit 1 | Overview, exit 0 |
| `yb-admin -h` | Unknown-flag error | Overview, exit 0 |
| `yb-admin --helpshort` | `No modules matched` | Overview + yb-admin's own flags, exit 0 |
| `yb-admin --helpfull` | Warning + all flags | Overview + all flags |
| `yb-admin help` | `Invalid operation: help` | All operations, numbered, alphabetical |
| `yb-admin help <operation>` | — | Usage of that operation |
| `yb-admin help <text>` | — | Operations whose name contains `<text>` |
| `yb-admin <operation> --help` | Warning + raw flag dump | Same output as `help <operation>` |
| `yb-admin <operation> <bad args>` | Error, then `Usage:` and definitions — none for bracketed placeholders (`import_snapshot` showed no `<namespace>`), and `<table>` defined without `<namespace>` | Error, then the same usage block `help <operation>` prints |
| `yb-admin list_server` | `Invalid operation`, no suggestion | Suggests the tablet-server listings |

`help` works when the cluster does not:

- `help` never builds a client, so it answers at once against an unreachable master.
- `--help`, `-h`, `--helpshort`, and a leading `help` are read from argv before the flag parse, so they answer even when the parse would fail — a missing `--flagfile`, a bad flag value, an unknown flag. `--helpfull`, `--helpmatch`, and `--helpon` are rendered by gflags and still need a clean parse.
- `help <operation>`, `<operation> --help`, and the bad-argument error print one shared block, so they cannot drift apart. The operation count in the overview is computed.

<details>
<summary>Examples — output of this branch's binary</summary>

```
$ yb-admin --help
yb-admin: administer a YugabyteDB universe from the command line.

Usage:
  yb-admin [global flags] <operation> [args]

Get help:
  yb-admin help                     List all 136 operations
  yb-admin help <operation>         Usage of one operation
  yb-admin help <text>              Operations whose name contains <text>
  yb-admin --helpshort              yb-admin's own global flags
  yb-admin --helpmatch=<substring>  Inherited flags matching a substring
  yb-admin --helpfull               All flags (long)

Common global flags:
  --master_addresses host:port[,host:port,...]  (default: localhost:7100)
  --init_master_addrs host:port                 (alternative to --master_addresses)
  --timeout_ms <millisec>                       (default: 60000)
  --certs_dir_name <dir>
  --flagfile <path>                             (use the master's server.conf)

Example:
  yb-admin --flagfile /path/to/master/conf/server.conf list_all_masters
```

```
$ yb-admin help
Operations (136):
    1. abort_snapshot_restore [<restoration_id>]
    2. add_namespace_to_xcluster_checkpoint <replication_group_id> <namespace_name>
    ...
  136. ysql_catalog_version [ysql.][<database_name>]

Run 'yb-admin help <operation>' for the usage of one operation.
Run 'yb-admin help <text>' to filter, e.g. 'yb-admin help snapshot'.
```

```
$ yb-admin help snapshot
Operations matching 'snapshot' (16):
  * abort_snapshot_restore [<restoration_id>]
  * create_database_snapshot [ysql.]<database_name> [retention_duration_hours] (set retention_dur...
  ...
  * restore_snapshot_schedule <schedule_id> (<timestamp> | minus <interval>)

Run 'yb-admin help <operation>' for the usage of one operation.
```

```
$ yb-admin help delete_table        # byte-identical to: yb-admin delete_table --help
Usage: yb-admin delete_table <table>
Definitions: <table>
  <namespace> <table_name> | tableid.<table_id>
<namespace>
  [(ycql|ysql).]<namespace_name> (default: ycql.)
```

```
$ yb-admin --helpshort              # overview, then:

yb-admin's own flags:
  --certs_dir_name <string>               (default: "")
  --client_node_name <string>             (default: "")
  --disable_graceful_transition <bool>    (default: false)
  --exclude_dead <bool>                   (default: false)
  --init_master_addrs <string>            (default: "")
  --master_addresses <string>             (default: "localhost:7100")
  --read_time_wait_ms <uint32>            (default: 5000)
  --timeout_ms <int64>                    (default: 60000)
  --wait_if_no_leader_master <bool>       (default: false)
  --yb_admin_force_use_private_ip <bool>  (default: false)
```

```
$ yb-admin list_server
Invalid operation: list_server
Did you mean one of these?
  list_tablet_servers
  list_all_tablet_servers

Run 'yb-admin help' to list all operations.
```

```
$ yb-admin --flagfile=/nope/server.conf help     # previously died in the flag parse
Operations (136):
    1. abort_snapshot_restore [<restoration_id>]
    ...

$ yb-admin --init_master_addrs=, list_all_masters   # previously a segfault
Invalid --init_master_addrs ',': no addresses found
```

</details>

## Bug fixes

- **#33434** — every remote error printed as `The cluster doesn't support <operation>`. The check was `s.ToString().find("rpc error 2")`, and `find()` returns `npos` (truthy) on a miss. Now keyed on the RPC error code, `ERROR_NO_SUCH_METHOD` / `ERROR_NO_SUCH_SERVICE`, the same test `client.cc` uses.
- **#33017** — usage-text defects: the bare-invocation page appended every `yb-admin_cli.cc` gflag with its build-relative source path; bracketed placeholders never matched their definitions, so `import_snapshot` showed none; `<table>` and `<index>` were defined in terms of an undefined `<namespace>`; numbering skipped a slot at every hidden command; a footer defining every placeholder printed after the whole catalog.
- **#33435** — `--init_master_addrs=,` segfaulted: zero addresses parsed as OK, then `[0]` on an empty vector. Now reported as a malformed value.
- **#33433** — `yugabyted configure admin_operation` dropped yb-admin's stdout on a non-zero exit, so a usage error showed as an empty red block. Now prints stdout, and prints the red stderr block only when stderr is non-empty.
- **#32640 phase 1c** — a third suggestion tier for abbreviations, tried only when the prefix and edit-distance tiers find nothing. Every typed token must match a name token, so garbage still gets no suggestion.
- `--helpshort` lists the flags yb-admin defines, filtered by `FlagTag::kHidden`, so `DEFINE_test_flag` knobs stay out of the help.

## Docs

`docs/content/stable/admin/yb-admin.md` now shows `yb-admin help`, gated "v2026.2 and later". Earlier versions are pointed at bare `./bin/yb-admin`, which lists the operations on those binaries. Other versioned trees are unchanged.

## Upgrade/Rollback safety

No wire, catalog, on-disk, or gflag-default change. Only yb-admin's output changes, which matters to anything that scrapes it:

- **Bare `yb-admin` no longer lists operations.** `yb-admin 2>/dev/null | grep <op>` returns nothing; use `yb-admin help | grep <op>`. This is the change most likely to be reported as a regression.
- Remote errors other than a missing method or service print `Error running <op>: <status>` instead of `The cluster doesn't support <op>`. The genuine unsupported case keeps the old text.
- `--help` and `--helpshort` exit 0 (were 1).
- Operation numbers change (the list is now alphabetical). Numbers are display-only and never accepted as input.
- `help` ignores flags on its own command line, since it is answered before the parse. Its output never depended on a flag.

Rollback is a binary swap with no state to undo.

## Test plan

- CI green at head: `alma8-clang21-release`, `alma8-gcc15-fastdebug`, `alma9-clang21-asan`; `yb-required` 0 failed checks.
- New tests in `tools_yb-admin-test`: `HelpFlagsPrintOverview`, `HelpListsAllOperations`, `HelpOperationDispatch`, `HelpNeedsNoCluster` (both `help` and `--help` against a missing `--flagfile`, a bad flag value, and an unknown flag), `HelpMatchesBadArgumentUsage`, `TokenMatchSuggestions`, `MalformedInitMasterAddrs`, `UnsupportedRpcErrorDetection`. Updated: `InvalidOperationSuggests*`, `PrintArgumentExpressions`.
- Local, macOS arm64 release build: the help and suggestion tests on every commit; `build-support/lint.sh --rev origin/master` 0 errors (7 line-length warnings, all on pre-existing `bin/yugabyted` lines).

## Issues

Fixes #33431. Fixes #33434. Fixes #33017. Fixes #33435. Fixes #33433.

Advances #32640 — phases 2a, 2b and 3 remain open.

Supersedes #33395, #33442, #33447, #33448 and #33469 (all closed). Commits are one per fix for review; the repo squash-merges.

### DIFF
--- a/bin/yugabyted
+++ b/bin/yugabyted
@@ -3124,9 +3124,12 @@ class ControlScript(object):
         status_details = [(Output.make_yellow("Admin operation"),
                            "Find the response below."),]
         Output.print_out(self.get_status_string_common(status_details))
-        if 0 == ret_code:
+        # yb-admin writes usage and help text to stdout even when it exits non-zero, so show
+        # stdout unconditionally; discarding it on failure rendered usage errors as an empty red
+        # block (#33433).
+        if out:
           Output.print_out(out)
-        else:
+        if 0 != ret_code:
           Output.log(err)
           Output.print_out(Output.make_red(err))
 
--- a/src/yb/tools/yb-admin_cli.cc
+++ b/src/yb/tools/yb-admin_cli.cc
@@ -107,6 +107,14 @@ const Status ClusterAdminCli::kInvalidArguments =
 
 namespace {
 
+// Read a flag's compiled-in default so the usage text cannot drift from the DEFINE_ that sets
+// it. yb::ParseCommandLineFlags() rewrites a flag's default only when the program set it
+// programmatically before parsing, so this returns the DEFINE_'d value on either side of the parse.
+std::string FlagDefault(const char* flag_name) {
+  google::CommandLineFlagInfo info;
+  return google::GetCommandLineFlagInfo(flag_name, &info) ? info.default_value : "";
+}
+
 constexpr auto kBlacklistAdd = "ADD";
 constexpr auto kBlacklistRemove = "REMOVE";
 constexpr int32 kDefaultRpcPort = 9100;
@@ -538,16 +546,17 @@ Status ClusterAdminCli::Run(int argc, char** argv) {
   const string addrs = FLAGS_master_addresses;
   if (!FLAGS_init_master_addrs.empty()) {
     std::vector<HostPort> init_master_addrs;
-    RETURN_NOT_OK(HostPort::ParseStrings(
-        FLAGS_init_master_addrs, master::kMasterDefaultPort, &init_master_addrs));
-    // ParseStrings() splits with SkipEmpty(), so a value of "," parses to zero addresses and
-    // returns OK; indexing that is out of bounds (#33435). Not InvalidArgument: SetUsage() has
-    // not run yet, so main()'s InvalidArgument branch would print an unset
-    // google::ProgramUsage() -- "Warning: SetUsageMessage() never called" as the entire error
-    // message.
-    if (init_master_addrs.empty()) {
-      cerr << "Invalid --init_master_addrs '" << FLAGS_init_master_addrs
-           << "': no addresses found" << endl;
+    const auto parse_status = HostPort::ParseStrings(
+        FLAGS_init_master_addrs, master::kMasterDefaultPort, &init_master_addrs);
+    // Neither failure may return InvalidArgument: SetUsage() has not run yet, so main()'s
+    // InvalidArgument branch would print an unset google::ProgramUsage() -- "Warning:
+    // SetUsageMessage() never called" as the entire error message. ParseStrings() also splits with
+    // SkipEmpty(), so a value of "," parses to zero addresses and returns OK; indexing that is out
+    // of bounds (#33435).
+    if (!parse_status.ok() || init_master_addrs.empty()) {
+      cerr << "Invalid --init_master_addrs '" << FLAGS_init_master_addrs << "': "
+           << (parse_status.ok() ? "no addresses found" : parse_status.message().ToBuffer())
+           << endl;
       return STATUS(RuntimeError, "Invalid --init_master_addrs");
     }
     client_.reset(new ClusterAdminClient(
@@ -621,10 +630,12 @@ void ClusterAdminCli::SetUsage(const string& prog_name) {
       << "  " << prog_name << " [global flags] <operation> [args]" << endl
       << endl
       << "Common global flags:" << endl
-      << "  --master_addresses host:port[,host:port,...]  (default: localhost:7100)" << endl
+      << "  --master_addresses host:port[,host:port,...]  (default: "
+      << FlagDefault("master_addresses") << ")" << endl
       << "  --init_master_addrs host:port                 (alternative to --master_addresses)"
       << endl
-      << "  --timeout_ms <millisec>                       (default: 60000)" << endl
+      << "  --timeout_ms <millisec>                       (default: "
+      << FlagDefault("timeout_ms") << ")" << endl
       << "  --certs_dir_name <dir>" << endl
       << "  --flagfile <path>" << endl
       << endl
--- a/src/yb/tools/yb-admin_cli.cc
+++ b/src/yb/tools/yb-admin_cli.cc
@@ -499,7 +499,7 @@ Status ClusterAdminCli::RunCommand(
     const Command& command, const CLIArguments& command_args, const std::string& program_name) {
   auto s = command.action_(command_args, client_.get());
   if (!s.ok()) {
-    if (s.IsRemoteError() && s.ToString().find("rpc error 2")) {
+    if (IsNoSuchMethodError(s)) {
       cerr << "The cluster doesn't support " << command.name_ << ": " << s << std::endl;
     } else {
       cerr << "Error running " << command.name_ << ": " << s << endl;
--- a/src/yb/tools/yb-admin_cli.cc
+++ b/src/yb/tools/yb-admin_cli.cc
@@ -499,7 +499,7 @@ Status ClusterAdminCli::RunCommand(
     const Command& command, const CLIArguments& command_args, const std::string& program_name) {
   auto s = command.action_(command_args, client_.get());
   if (!s.ok()) {
-    if (IsNoSuchMethodError(s)) {
+    if (IsUnsupportedRpcError(s)) {
       cerr << "The cluster doesn't support " << command.name_ << ": " << s << std::endl;
     } else {
       cerr << "Error running " << command.name_ << ": " << s << endl;
--- a/src/yb/tools/yb-admin_cli.cc
+++ b/src/yb/tools/yb-admin_cli.cc
@@ -113,9 +113,11 @@ constexpr int32 kDefaultRpcPort = 9100;
 const string kMinus = "minus";
 
 const std::string namespace_expression =
-    "<namespace>:\n [(ycql|ysql).]<namespace_name> (default ycql.)";
-const std::string table_expression = "<table>:\n <namespace> <table_name> | tableid.<table_id>";
-const std::string index_expression = "<index>:\n  <namespace> <index_name> | tableid.<index_id>";
+    "<namespace>\n  [(ycql|ysql).]<namespace_name> (default: ycql.)";
+const std::string table_expression =
+    "<table>\n  <namespace> <table_name> | tableid.<table_id>";
+const std::string index_expression =
+    "<index>\n  <namespace> <index_name> | tableid.<index_id>";
 
 Status GetUniverseConfig(ClusterAdminClient* client, const ClusterAdminCli::CLIArguments&) {
   RETURN_NOT_OK_PREPEND(client->GetUniverseConfig(), "Unable to get universe config");
@@ -430,11 +432,21 @@ std::string ClusterAdminCli::GetArgumentExpressions(const std::string& usage_arg
   std::stringstream ss(usage_arguments);
   std::string next_argument;
   while (ss >> next_argument) {
-    if (next_argument == "<namespace>" || next_argument == "<source_namespace>") {
+    // usage_arguments_ marks optional arguments with surrounding '[' ']' and repeated ones with
+    // a trailing "...", e.g. "[<namespace> <table_name> [<table_name>]...]". Strip those
+    // decorations before comparing, otherwise a placeholder inside brackets never matches and its
+    // definition is silently omitted.
+    const auto begin = next_argument.find_first_not_of('[');
+    const auto end = next_argument.find_last_not_of("].");
+    const std::string token = (begin == std::string::npos || end == std::string::npos ||
+                                begin > end)
+                                   ? std::string()
+                                   : next_argument.substr(begin, end - begin + 1);
+    if (token == "<namespace>" || token == "<source_namespace>") {
       expressions += namespace_expression + '\n';
-    } else if (next_argument == "<table>") {
+    } else if (token == "<table>") {
       expressions += table_expression + '\n';
-    } else if (next_argument == "<index>") {
+    } else if (token == "<index>") {
       expressions += index_expression + '\n';
     }
   }
@@ -595,32 +607,42 @@ void ClusterAdminCli::Register(
 void ClusterAdminCli::SetUsage(const string& prog_name) {
   ostringstream str;
 
-  str << prog_name << " [--master_addresses server1:port,server2:port,server3:port,...] "
-      << " [--timeout_ms <millisec>] [--certs_dir_name <dir_name>]" << endl
-      << "  [--flagfile <path/to/master/conf/server.conf>]" << endl
-      << "  <operation>" << endl
+  str << "Usage:" << endl
+      << "  " << prog_name << " [global flags] <operation> [args]" << endl
       << endl
-      << "Tip: Use --flagfile with the master's server.conf to automatically pick up" << endl
-      << "master_addresses and certs_dir, avoiding manual flag entry." << endl
-      << "Example: " << prog_name
-      << " --flagfile master/conf/server.conf list_all_masters" << endl
+      << "Common global flags:" << endl
+      << "  --master_addresses host:port[,host:port,...]  (default: localhost:7100)" << endl
+      << "  --init_master_addrs host:port                 (alternative to --master_addresses)"
       << endl
-      << "<operation> must be one of:" << endl;
+      << "  --timeout_ms <millisec>                       (default: 60000)" << endl
+      << "  --certs_dir_name <dir>" << endl
+      << "  --flagfile <path>" << endl
+      << endl
+      << "Tip:" << endl
+      << "  Use --flagfile with the master's server.conf to automatically pick up" << endl
+      << "  master_addresses and certs_dir, avoiding manual flag entry." << endl
+      << endl
+      << "Example:" << endl
+      << "  " << prog_name << " --flagfile /path/to/master/conf/server.conf list_all_masters"
+      << endl
+      << endl
+      << "Operations:" << endl;
 
-  for (size_t i = 0; i < commands_.size(); ++i) {
-    const auto& command = commands_[i];
+  // Number only the operations actually printed, so a hidden command's slot in commands_ doesn't
+  // leave a gap in the visible list (e.g. "85. ..." followed by "87. ..." with no "86.").
+  size_t visible_number = 0;
+  for (const auto& command : commands_) {
     if (command.hidden_) {
       continue;
     }
-    str << ' ' << i + 1 << ". " << command.name_ << (command.usage_arguments_.empty() ? "" : " ")
-        << command.usage_arguments_ << endl;
+    str << "  " << ++visible_number << ". " << command.name_
+        << (command.usage_arguments_.empty() ? "" : " ") << command.usage_arguments_ << endl;
   }
 
-  str << endl;
-  str << namespace_expression << endl;
-  str << table_expression << endl;
-  str << index_expression << endl;
-
+  // Argument placeholders like <namespace>/<table>/<index> are defined per-command instead of
+  // in a global footer here: only a minority of operations use them (see GetArgumentExpressions),
+  // and RunCommand() already prints the relevant definition alongside a specific command's usage
+  // when that command's arguments are invalid.
   google::SetUsageMessage(str.str());
 }
 
@@ -3312,7 +3334,11 @@ int main(int argc, char** argv) {
   }
 
   if (s.IsInvalidArgument()) {
-    google::ShowUsageWithFlagsRestrict(argv[0], __FILE__);
+    // Print the usage message set up by ClusterAdminCli::SetUsage directly, rather than
+    // google::ShowUsageWithFlagsRestrict(argv[0], __FILE__), which additionally dumps every gflag
+    // defined in this file with its build-relative source path, type, and default -- noise that
+    // buries the operation catalog the usage message already lists in full.
+    std::cout << google::ProgramUsage();
   }
 
   return 1;
--- a/src/yb/tools/yb-admin_cli.cc
+++ b/src/yb/tools/yb-admin_cli.cc
@@ -118,6 +118,10 @@ std::string FlagDefault(const char* flag_name) {
   return google::GetCommandLineFlagInfo(flag_name, &info) ? info.default_value : "";
 }
 
+// The operation that prints the catalog. Named here because ScanForHelpRequest() must
+// recognize it in raw argv, before the flag parse, and Run() dispatches it by name.
+constexpr auto kHelpOperation = "help";
+
 constexpr auto kBlacklistAdd = "ADD";
 constexpr auto kBlacklistRemove = "REMOVE";
 constexpr int32 kDefaultRpcPort = 9100;
@@ -652,22 +656,18 @@ std::optional<ClusterAdminCli::HelpRequest> ClusterAdminCli::ScanForHelpRequest(
     int argc, char** argv) const {
   bool help_requested = false;
   bool helpshort = false;
-  std::string operation;
+  std::vector<std::string> positionals;
+  bool positional_only = false;
   for (int i = 1; i < argc; ++i) {
     const std::string token = argv[i];
-    if (token == "--") {
-      // gflags treats everything after a bare "--" as positional; so does this scan.
-      break;
+    if (!positional_only && token == "--") {
+      // gflags stops parsing at a bare "--" and drops it, leaving everything after it in the
+      // positional argv; so does this scan.
+      positional_only = true;
+      continue;
     }
-    if (token.empty() || token[0] != '-') {
-      // The first positional token naming a registered operation selects which help page to
-      // show. Every token is scanned, not just argv[1]: real commands lead with flags, e.g.
-      // "yb-admin --flagfile server.conf delete_table --help". The scan only ever selects a help
-      // page and never which operation to run, so a free-text argument that happens to equal an
-      // operation name can at worst pick the wrong page.
-      if (operation.empty() && command_indexes_.count(token) > 0) {
-        operation = token;
-      }
+    if (positional_only || token.empty() || token[0] != '-') {
+      positionals.push_back(token);
       continue;
     }
     const auto name_begin = token.find_first_not_of('-');
@@ -706,13 +706,32 @@ std::optional<ClusterAdminCli::HelpRequest> ClusterAdminCli::ScanForHelpRequest(
       }
     }
   }
-  if (!help_requested) {
-    return std::nullopt;
+  if (help_requested) {
+    HelpRequest request;
+    request.helpshort = helpshort;
+    // The first positional naming a registered operation selects which help page to show. Every
+    // token is scanned, not just argv[1]: real commands lead with flags, e.g. "yb-admin
+    // --flagfile server.conf delete_table --help". The scan only ever selects a help page and
+    // never which operation to run, so a free-text argument that happens to equal an operation
+    // name can at worst pick the wrong page.
+    for (const auto& positional : positionals) {
+      if (command_indexes_.count(positional) > 0) {
+        request.operation = positional;
+        break;
+      }
+    }
+    return request;
   }
-  HelpRequest request;
-  request.operation = operation;
-  request.helpshort = helpshort;
-  return request;
+  // A --help* flag anywhere wins over the operation, so this runs only when none was given.
+  // Only a *leading* positional counts: "yb-admin --flagfile server.conf help delete_table" asks
+  // for help, but "yb-admin list_tables help" passes "help" to list_tables as an argument.
+  if (!positionals.empty() && positionals.front() == kHelpOperation) {
+    HelpRequest request;
+    request.help_operation = true;
+    request.help_args.assign(positionals.begin() + 1, positionals.end());
+    return request;
+  }
+  return std::nullopt;
 }
 
 void ClusterAdminCli::PrintHelpRequest(
@@ -777,8 +796,16 @@ Status ClusterAdminCli::Run(int argc, char** argv) {
   SetUsage(prog_name_);
 
   // Help must work on a broken cluster: the scan reads raw argv, so it answers before the flag
-  // parse can fail on a malformed --flagfile or a garbage flag value.
+  // parse can fail on a malformed --flagfile, an unparsable flag value, or an unknown flag.
   if (const auto request = ScanForHelpRequest(argc, argv)) {
+    if (request->help_operation) {
+      // Dispatched through RunCommand() rather than answered inline so that intercepting `help`
+      // early leaves its output and its error framing -- suggestions for an unknown operation,
+      // help's own usage line for extra arguments -- identical to ordinary dispatch. The action
+      // ignores the client, which is still null here.
+      return RunCommand(
+          commands_[command_indexes_[kHelpOperation]], request->help_args, prog_name_);
+    }
     PrintHelpRequest(*request, prog_name_, std::cout);
     return Status::OK();
   }
@@ -3370,8 +3397,10 @@ void ClusterAdminCli::RegisterCommandHandlers() {
   // help is registered like any other operation so that it appears in the operation list, gets
   // suggested for typos ("yb-admin hepl"), and inherits RunCommand()'s standard error formatting.
   // needs_client is false: help must answer on a broken cluster, without waiting out --timeout_ms.
+  // Run() reaches this action from ScanForHelpRequest() before the flag parse, so the action must
+  // not depend on any flag having been applied.
   Register(
-      "help", "[<operation>]",
+      kHelpOperation, "[<operation>]",
       [this](const CLIArguments& args, ClusterAdminClient*) -> Status {
         if (args.size() > 1) {
           return kInvalidArguments;
--- a/src/yb/tools/yb-admin_cli.cc
+++ b/src/yb/tools/yb-admin_cli.cc
@@ -35,6 +35,7 @@
 #include <algorithm>
 #include <limits>
 #include <memory>
+#include <optional>
 #include <unordered_set>
 #include <utility>
 
@@ -104,6 +105,8 @@ namespace tools {
 
 const Status ClusterAdminCli::kInvalidArguments =
     STATUS(InvalidArgument, "Invalid arguments for operation");
+
+const Status ClusterAdminCli::kErrorReported = STATUS(RuntimeError, "Error already reported");
 
 namespace {
 
@@ -436,7 +439,9 @@ size_t EditDistance(const string& lhs, const string& rhs) {
 }  // namespace
 
 std::string ClusterAdminCli::GetArgumentExpressions(const std::string& usage_arguments) {
-  std::string expressions;
+  bool has_namespace = false;
+  bool has_table = false;
+  bool has_index = false;
   std::stringstream ss(usage_arguments);
   std::string next_argument;
   while (ss >> next_argument) {
@@ -451,12 +456,25 @@ std::string ClusterAdminCli::GetArgumentExpressions(const std::string& usage_arg
                                    ? std::string()
                                    : next_argument.substr(begin, end - begin + 1);
     if (token == "<namespace>" || token == "<source_namespace>") {
-      expressions += namespace_expression + '\n';
+      has_namespace = true;
     } else if (token == "<table>") {
-      expressions += table_expression + '\n';
+      has_table = true;
     } else if (token == "<index>") {
-      expressions += index_expression + '\n';
+      has_index = true;
     }
+  }
+  // The <table> and <index> definitions themselves reference <namespace>, so it must be defined
+  // whenever they are. Each placeholder is emitted once, referencing definitions first.
+  has_namespace |= has_table || has_index;
+  std::string expressions;
+  if (has_table) {
+    expressions += table_expression + '\n';
+  }
+  if (has_index) {
+    expressions += index_expression + '\n';
+  }
+  if (has_namespace) {
+    expressions += namespace_expression + '\n';
   }
   return expressions.empty() ? "" : "Definitions: " + expressions;
 }
@@ -519,14 +537,17 @@ Status ClusterAdminCli::RunCommand(
     const Command& command, const CLIArguments& command_args, const std::string& program_name) {
   auto s = command.action_(command_args, client_.get());
   if (!s.ok()) {
+    if (s.IsRuntimeError() && s.message() == kErrorReported.message()) {
+      // The action already wrote its complete report to stderr (e.g. help's suggestions for an
+      // unknown operation); prefixing it with "Error running ..." would bury it.
+      return s;
+    }
     if (IsUnsupportedRpcError(s)) {
       cerr << "The cluster doesn't support " << command.name_ << ": " << s << std::endl;
     } else {
       cerr << "Error running " << command.name_ << ": " << s << endl;
       if (s.IsInvalidArgument()) {
-        cerr << Format("Usage: $0 $1 $2", program_name, command.name_, command.usage_arguments_)
-             << endl
-             << GetArgumentExpressions(command.usage_arguments_);
+        PrintCommandUsage(command, program_name, cerr);
       }
     }
     return STATUS(RuntimeError, "Error running command");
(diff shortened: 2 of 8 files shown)


src/yb/tools/yb-admin_cli.cc
```diff
@@ -492,50 +530,255 @@ std::vector<std::string> ClusterAdminCli::GetSuggestedCommands(const std::string
...
 Status ClusterAdminCli::RunCommand(
     const Command& command, const CLIArguments& command_args, const std::string& program_name) {
   auto s = command.action_(command_args, client_.get());
   if (!s.ok()) {
+    if (s.IsRuntimeError() && s.message() == kErrorReported.message()) {
+      // The action already wrote its complete report to stderr (e.g. help's suggestions for an
+      // unknown operation); prefixing it with "Error running ..." would bury it.
+      return s;
+    }
     if (s.IsRemoteError() && s.ToString().find("rpc error 2")) {
```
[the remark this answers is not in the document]
ag26jan: The bug is real — `find()` returning `npos` in a boolean context makes this fire for every remote error — but it is not this PR's to fix. This line is untouched here; the diff hunk shows it as context below the `kErrorReported` block that #33470 adds.

It is fixed in **#33469**, which extracts the predicate so a unit test can pin both directions, and which is intended to land first. Both PRs edit this `if`, so whichever lands second needs a one-line rebase, and the resolution has to keep *both* the `kErrorReported` early return from this PR and `IsUnsupportedRpcError(s)` from #33469 — taking either side wholesale drops a fix.

---

src/yb/tools/yb-admin_util.cc
```diff
@@ -35,6 +38,28 @@ using master::ListTabletServersResponsePB;
 
 namespace {
 
+std::vector<string> SplitOnUnderscore(const string& s) {
```
hulien22: this already exists in our libraries:
const auto op_tokens = strings::Split(op, "_", strings::SkipEmpty());

src/yb/tools/yb-admin_util.cc
```diff
@@ -73,6 +98,64 @@ bool CompareListTabletServersEntries(
...
+  // Read the code off the Status rather than matching Status::ToString(): OutboundCall::SetFailed()
+  // attaches it with CloneAndAddErrorCode(RpcError(...)), which is how client.cc and
+  // client_master_rpc.cc test the same condition. A status with no rpc code decodes to 0, so it
+  // never matches.
+  const auto rpc_error = rpc::RpcError(s);
+  return rpc_error == rpc::ErrorStatusPB::ERROR_NO_SUCH_METHOD ||
+         rpc_error == rpc::ErrorStatusPB::ERROR_NO_SUCH_SERVICE;
+}
+
+std::vector<string> SuggestByNameTokens(
```
hulien22: this should be case-insensitive

src/yb/tools/yb-admin_cli.h
```diff
@@ -70,25 +80,64 @@ class ClusterAdminCli {
     std::string usage_arguments_;
     Action action_;
     bool hidden_;
+    // Whether Run() must construct and connect a ClusterAdminClient before dispatching to
+    // action_. False for operations like `help` that must work with no reachable cluster.
+    bool needs_client_ = true;
```
hulien22: is this really needed? from what i can tell, we only ever check for command.needs_client_ == true after help has already been processed, so need_client_ is always true at that point

can remove this and the register arg too

src/yb/tools/yb-admin_cli.cc
```diff
@@ -427,18 +444,43 @@ size_t EditDistance(const string& lhs, const string& rhs) {
 }  // namespace
 
 std::string ClusterAdminCli::GetArgumentExpressions(const std::string& usage_arguments) {
```
hulien22: simpler alternative
```
  std::string ClusterAdminCli::GetArgumentExpressions(const std::string& usage_arguments) {
    auto contains = [&](const char* p) { return usage_arguments.find(p) != std::string::npos; };
    const bool has_table = contains("<table>");
    const bool has_index = contains("<index>");
    // The <table> and <index> definitions reference <namespace>.
    const bool has_namespace =
        has_table || has_index || contains("<namespace>") || contains("<source_namespace>");

```

src/yb/tools/yb-admin_cli.cc
```diff
@@ -549,78 +828,114 @@ Status ClusterAdminCli::Run(int argc, char** argv) {
...
 
 void ClusterAdminCli::Register(
-    string&& cmd_name, const std::string& cmd_args, Action&& action, bool hidden) {
+    string&& cmd_name, const std::string& cmd_args, Action&& action, bool hidden,
+    bool needs_client) {
   command_indexes_[cmd_name] = commands_.size();
-  commands_.push_back({std::move(cmd_name), cmd_args, std::move(action), hidden});
+  // The usage strings are hand-written; trim them once here so a stray space (list_snapshots'
+  // args led with one) cannot render a double separator on any surface that prints them.
+  commands_.push_back({std::move(cmd_name), boost::algorithm::trim_copy(cmd_args),
```
hulien22: this seems overly specific on some edge case with list_snapshots - can't we just fix that case so everything works the same? also list_snapshot_args isn't handwritten, its generated by GetListSnapshotsFlagList

can remove the test for this too

src/yb/tools/yb-admin_cli.cc
```diff
@@ -493,50 +535,287 @@ std::vector<std::string> ClusterAdminCli::GetSuggestedCommands(const std::string
...
+  FLAGS_minloglevel = 2;
+
+  // Registration and SetUsage() run before the flag parse: RegisterCommandHandlers() needs no
+  // flags and no client, and every surface gflags renders inside ParseCommandLineFlags()
+  // (--helpfull, --helpon, --helpmatch, a help flag inside a --flagfile) prints
+  // google::ProgramUsage() as its header -- which is "Warning: SetUsageMessage() never called"
+  // until SetUsage() has run.
   RegisterCommandHandlers();
-  SetUsage(prog_name);
+  SetUsage(prog_name_);
```
hulien22: why are we passing prog_name_ around if it is a member?

src/yb/tools/yb-admin_util.cc
```diff
@@ -35,6 +38,28 @@ using master::ListTabletServersResponsePB;
...
+    }
+    start = end + 1;
+  }
+  return tokens;
+}
+
+bool EitherIsPrefix(const string& a, const string& b) {
+  const auto& shorter = a.size() <= b.size() ? a : b;
+  const auto& longer = a.size() <= b.size() ? b : a;
+  return longer.compare(0, shorter.size(), shorter) == 0;
```
hulien22: HasPrefixString(a, b) || HasPrefixString(b, a)

bin/yugabyted
```diff
@@ -3124,9 +3124,14 @@ class ControlScript(object):
         status_details = [(Output.make_yellow("Admin operation"),
                            "Find the response below."),]
         Output.print_out(self.get_status_string_common(status_details))
-        if 0 == ret_code:
+        # yb-admin writes usage and help text to stdout even when it exits non-zero, so show
```
hulien22: why is this change in this diff?? #33433 is not in the title of this diff and seems unrelated?

src/yb/tools/yb-admin_cli.h
```diff
@@ -62,6 +64,14 @@ class ClusterAdminCli {
 
   static const Status kInvalidArguments;
 
+  // Returned by an action that has already written its complete error report to stderr.
+  // RunCommand() adds nothing on top of it; main() still exits non-zero.
+  static const Status kErrorReported;
```
hulien22: this seems weird to me.. from what I can tell, this only exists because we are routing help through RunCommand. if we run help outside of that in its own smaller handler, then we can simplify all of this
```
  Status ClusterAdminCli::RunHelp(const CLIArguments& args) {
    if (args.size() > 1) {
      cerr << "Error running " << kHelpOperation << ": " << kInvalidArguments << endl;
      PrintCommandUsage(commands_[command_indexes_[kHelpOperation]], cerr);
      return STATUS(RuntimeError, "Error running command");
    }
    if (args.empty()) {
      PrintOperationNames(std::cout);
      ...
      return Status::OK();
    }
    // exact match / substring filter as today
    ...
    ReportUnknownOperation(args[0]);
    return STATUS_FORMAT(RuntimeError, "Invalid operation: $0", args[0]);
  }

  In Run():

  if (const auto request = ScanForHelpRequest(argc, argv)) {
    if (request->help_operation) {
      return RunHelp(request->help_args);
    }
    PrintHelpRequest(*request, std::cout);
    return Status::OK();
  }
```

hulien22 (changes requested):
Please take a pass over all the comments in this change and clean them up. There are a lot of redundant comments that just repeat the code, repeat other existing comments, are too verbose, or are too specific on things that are not relevant for others

Issue #33482, opened 2026-08-24
[DocDB] RobustLentObjectTest.TestCrash hangs on macOS (no PTHREAD_MUTEX_ROBUST)


ellabaron-code:
Jira Link: [DB-23158](https://yugabyte.atlassian.net/browse/DB-23158)
## Description

`RobustLentObjectTest.TestCrash` hangs on macOS until the test harness timeout.

The test simulates a process dying while holding a `RobustMutex` in shared memory: the child takes the lock through `RobustLentObjectReference` and calls `std::_Exit(0)` without releasing it, then the parent calls `RobustLendGuard::reset()` → `RobustLentObject::Revoke()`, which locks the same mutex.

On Linux the mutex is created with `PTHREAD_MUTEX_ROBUST`, so the parent's `pthread_mutex_lock` returns `EOWNERDEAD` and recovery proceeds. Darwin does not implement `PTHREAD_MUTEX_ROBUST`; `RobustMutexImpl` compiles the `setrobust` call out (`src/yb/util/shmem/robust_mutex.cc`) and falls back to a plain `PTHREAD_PROCESS_SHARED` mutex, as the comment in `robust_mutex.h` notes ("we just use a non-robust mutex and hope for the best"). With no owner-death notification, the parent blocks forever in `Revoke()` and the test never completes — it hangs rather than fails, consuming the suite's time budget.

The crash-recovery mechanism under test intentionally does not exist on macOS, so the test should be disabled there with `YB_DISABLE_TEST_ON_MACOS`.

Steps to reproduce: run `RobustLentObjectTest.TestCrash` from `robust_lent_object-test` on macOS.

## Issue Type

kind/failing-test

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23158]: https://yugabyte.atlassian.net/browse/DB-23158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33483, opened 2026-08-24
[#33482] DocDB: Disable RobustLentObjectTest.TestCrash on macOS

ellabaron-code:
## Summary

`RobustLentObjectTest.TestCrash` verifies that a process dying while holding a `RobustMutex` in shared memory does not block the lender: the child exits via `std::_Exit(0)` with the lock held, and the parent recovers it in `Revoke()` through the `EOWNERDEAD` path.

That path requires `PTHREAD_MUTEX_ROBUST`, which Darwin does not implement. `RobustMutexImpl` compiles the robustness attribute out on macOS and uses a plain process-shared mutex ("hope for the best", per the comment in `robust_mutex.h`), so the parent blocks forever on the dead child's lock and the test hangs until the harness timeout instead of failing.

The mechanism under test intentionally does not exist on macOS, so disable the test there with `YB_DISABLE_TEST_ON_MACOS`.

## Test plan

- [x] Verified the hang on macOS: parent blocked in `RobustLentObject::Revoke()` after the child exits with the lock held.
- [x] No change on Linux: `YB_DISABLE_TEST_ON_MACOS` expands to the bare test name there.

### DIFF
--- a/src/yb/util/shmem/robust_lent_object-test.cc
+++ b/src/yb/util/shmem/robust_lent_object-test.cc
@@ -88,7 +88,8 @@ TEST_F(RobustLentObjectTest, TestPointerNotDestroyedEarly) {
   }));
 }
 
-TEST_F(RobustLentObjectTest, TestCrash) {
+// macOS has no PTHREAD_MUTEX_ROBUST, so the parent would block forever on the dead child's lock.
+TEST_F(RobustLentObjectTest, YB_DISABLE_TEST_ON_MACOS(TestCrash)) {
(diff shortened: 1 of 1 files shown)


neera-mital (approved):

neera-mital:
A GTEST_SKIP() with a message under #ifdef __APPLE__ would make the skip visible in test output rather than silently hiding the test from the runner — but the YB_DISABLE_TEST_ON_MACOS macro is the house style, so this change is consistent with current guidelines.

Issue #33486, opened 2026-08-24
[YSQL] PgIndexBackfillBackendsManager.NoAbortTxn deadlocks with table locks enabled


ellabaron-code:
Jira Link: [DB-23162](https://yugabyte.atlassian.net/browse/DB-23162)
## Description

With object locking (table locks) enabled, `CREATE INDEX` waits for pre-existing lockers (`WaitForLockers`, #31758) between the `indisready` and backfill phases, before the backfill safe time can be established.

The `PgIndexBackfillBackendsManager.NoAbortTxn` test (table-locks variant) opens a transaction, updates a row of the indexed table (taking a ROW EXCLUSIVE lock), and only commits *after* `WaitForBackfillSafeTime`. With the lock held, backfill cannot get past `GetSafeTime`, and the test times out waiting for it.

The test does not fail on Linux only because the lock check misses shared-memory locks: with the default `--enable_object_lock_fastpath=true`, the UPDATE's lock is acquired through the shared-memory fastpath and sits in the pending-requests table, which `ObjectLockManagerImpl::WaitForConflictingLockers` never consults — it scans only granted locks in `txn_locks_` — so the conflicting lock is invisible and the wait returns immediately. Running the test with `--enable_object_lock_fastpath=false --pg_client_use_shared_memory=false` reproduces the deadlock on Linux.

Fix: in the table-locks variant, commit the transaction before waiting for the backfill safe time.

## Issue Type

kind/failing-test

- [x] I confirm this issue does not contain any sensitive information.

[DB-23162]: https://yugabyte.atlassian.net/browse/DB-23162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33487, opened 2026-08-24
[#33486] YSQL: Fix PgIndexBackfillBackendsManager.NoAbortTxn under table locks

ellabaron-code:
## Summary

With object locking enabled, `CREATE INDEX` waits for pre-existing lockers (`WaitForLockers`) between the `indisready` and backfill phases, so the backfill safe time cannot be established while the test's transaction still holds its ROW EXCLUSIVE lock. The test committed only after `WaitForBackfillSafeTime`, which deadlocks whenever the lock is visible to the wait. It passes on Linux today only because fastpath (shared-memory) locks are invisible to `WaitForConflictingLockers`' conflict scan.

Reorder the table-locks variant to commit before waiting for the backfill safe time; the non-table-locks variant keeps the original order.

## Test plan

- [x] `pg_index_backfill-test --gtest_filter 'PgIndexBackfillBackendsManager.NoAbortTxn/*'` passes on macOS arm64 (release), both variants.
- [x] On Linux, verified the old order deadlocks once the conflicting lock is visible to `WaitForLockers` (object lock fastpath disabled), and the reordered test passes.

### DIFF
--- a/src/yb/yql/pgwrapper/pg_index_backfill-test.cc
+++ b/src/yb/yql/pgwrapper/pg_index_backfill-test.cc
@@ -2836,20 +2836,21 @@ TEST_P(PgIndexBackfillBackendsManager, YB_DISABLE_TEST_IN_TSAN(NoAbortTxn)) {
   // Reset connection to eliminate cache/heartbeat-delay issues of indislive=t, indisready=t.
   conn_->Reset();
 
+  // With object locking, CREATE INDEX waits for existing lockers before requesting the backfill,
+  // so open the transaction only after that wait, paused just before the read time is picked.
+  ASSERT_OK(cluster_->SetFlagOnMasters("TEST_pause_compute_safe_time_for_backfill_read", "true"));
+  LogWaiter pause_log_waiter(
+      cluster_->GetLeaderMaster(),
+      "Pausing due to flag TEST_pause_compute_safe_time_for_backfill_read");
+  ASSERT_OK(cluster_->SetFlagOnTServers("ysql_yb_test_block_index_phase", "none"));
+  ASSERT_OK(pause_log_waiter.WaitFor(30s * kTimeMultiplier));
+
   LOG(INFO) << "Begin txn";
   ASSERT_OK(conn_->Execute("BEGIN"));
   ASSERT_OK(conn_->ExecuteFormat("UPDATE $0 SET j = 5 WHERE i = 3", kTableName));
-  ASSERT_OK(cluster_->SetFlagOnTServers("ysql_yb_test_block_index_phase", "none"));
-  if (EnableTableLocks()) {
-    // With object locking, CREATE INDEX waits for existing lockers (WaitForLockers) before
-    // choosing the backfill safe time, so the transaction must commit before the safe time
-    // can be established.
-    ASSERT_OK(conn_->Execute("COMMIT"));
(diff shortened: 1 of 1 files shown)


src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2840,8 +2840,16 @@ TEST_P(PgIndexBackfillBackendsManager, YB_DISABLE_TEST_IN_TSAN(NoAbortTxn)) {
...
   ASSERT_OK(cluster_->SetFlagOnTServers("ysql_yb_test_block_index_phase", "none"));
-  ASSERT_OK(WaitForBackfillSafeTime(kYBTableName));
-  ASSERT_OK(conn_->Execute("COMMIT"));
+  if (EnableTableLocks()) {
+    // With object locking, CREATE INDEX waits for existing lockers (WaitForLockers) before
+    // choosing the backfill safe time, so the transaction must commit before the safe time
+    // can be established.
+    ASSERT_OK(conn_->Execute("COMMIT"));
+    ASSERT_OK(WaitForBackfillSafeTime(kYBTableName));
+  } else {
```
basavaraj29: let's not change the test semantic, the test specifically asserts that the update is preserved despite backfill picking a time prior to the update's commit time.

you can use `TEST_pause_compute_safe_time_for_backfill_read` to pause just before backfill read time is picked, and start the update transaction, reset the flag, wait for backfill safetime to be picked, and then commit the update.

basavaraj29 (changes requested):

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
ellabaron-code: Done as suggested — used the pause flag and dropped the mode-specific branching. Both parameter values pass locally.

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
```diff
@@ -2836,10 +2836,19 @@ TEST_P(PgIndexBackfillBackendsManager, YB_DISABLE_TEST_IN_TSAN(NoAbortTxn)) {
   // Reset connection to eliminate cache/heartbeat-delay issues of indislive=t, indisready=t.
   conn_->Reset();
 
+  // With object locking, CREATE INDEX waits for existing lockers before requesting the backfill,
+  // so open the transaction only after that wait, paused just before the read time is picked.
+  ASSERT_OK(cluster_->SetFlagOnMasters("TEST_pause_compute_safe_time_for_backfill_read", "true"));
+  LogWaiter pause_log_waiter(
+      cluster_->GetLeaderMaster(),
+      "Pausing due to flag TEST_pause_compute_safe_time_for_backfill_read");
+  ASSERT_OK(cluster_->SetFlagOnTServers("ysql_yb_test_block_index_phase", "none"));
```
basavaraj29: what's the need for `ysql_yb_test_block_index_phase` now? we can remove it, right? update the comment above the test accordingly.

basavaraj29 (changes requested):

src/yb/yql/pgwrapper/pg_index_backfill-test.cc
ellabaron-code: Fixed.

basavaraj29 (approved):

Issue #33445, opened 2026-08-21
[YSQL] Use-after-free when a status update fails during ROLLBACK TO SAVEPOINT on a promoted transaction


egladysh:
Jira Link: [DB-23126](https://yugabyte.atlassian.net/browse/DB-23126)
### Description

When a transaction is moved from a region-local transaction status tablet to a global one, it is registered at both the old and the new status tablet for a while. Rolling back to a savepoint (PG uses savepoints internally too) in that state sends one status update to each of them.

If the update to the old status tablet fails while the update to the new one is still in flight, the rollback returns immediately with that error. The client-side transaction object is then destroyed while the second request is still outstanding and tries to access the same destroyed transaction object.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23126]: https://yugabyte.atlassian.net/browse/DB-23126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33498, opened 2026-08-25
[#33445] DocDB: Keep the transaction alive until sub-txn rollback heartbeats complete

egladysh:
## Summary

`YBTransaction::Impl::RollbackToSubTransaction` sends a `PENDING` heartbeat carrying the aborted
sub-txn set to every status tablet the transaction is still registered at. A promoted transaction
whose heartbeats to the old status tablet are failing stays registered at both, so two heartbeats
go out.

The response callback bound no reference to the transaction, and the wait loop returned as soon as
one heartbeat failed. The other rpc could therefore still be registered, holding pointers into
`Impl`, after the caller destroyed the transaction - a use-after-free when the response finally
arrived, in `UpdateClock` and `Rpcs::Unregister`.

- `SendHeartBeatOnRollback` binds a `YBTransactionPtr` in the response callback, matching every
  other rpc callback in the file.
- The wait loop drains all futures and returns the first error after the loop, so no rollback
  heartbeat is in flight once the call returns. This also keeps a retried rollback from hitting the
  `RegisterAndStart` CHECK on a still-occupied handle.
- `~Impl` aborts `rollback_heartbeat_handle_` and `old_rollback_heartbeat_handle_` along with the
  other handles.

Reaching this state requires the transaction to stay registered at the old status tablet with no
status-move rpc holding a reference to it - the `last_old_heartbeat_failed_` path in
`UpdateTransactionPromotingDone`. The test drives that through
`TEST_simulate_failing_heartbeats_to_old_status_tablet`; it and
`TEST_delay_rollback_heartbeat_response_ms` are inert by default. `Rpcs::TEST_NumActiveCalls` is
new and test-only.

## Test plan

Debug build, macOS/clang, `--no-ccache`, one test per `yb_build.sh` invocation.

- [x] `client_transaction_promotion-test` - `RollbackWithHeartbeatInFlight` (new),
      `PromotedTransactionIntentsReadableForCDC`, `MidTransactionRestartReplayPromotedOp`
- [x] `pg_mini-test` - `PgRecursiveAbortTest.MockRollbackToSubtransactionFailure`, both
      parameterizations
- [x] `pgwrapper_geo_transactions_promotion-test` - all 31 tests in the file, covering promotion
      with both status tablets live, `EnsureGlobal` promotion, and savepoint rollback under
      promotion


egladysh:
trigger jenkins

Issue #33503, opened 2026-08-25
[DocDB] DiskFullTest fails on hosts with more than 10TB of free disk space


ellabaron-code:
Jira Link: [DB-23180](https://yugabyte.atlassian.net/browse/DB-23180)
## Description

YCqlDiskFullTest.TestDiskFull and YSqlDiskFullTest.TestDiskFull simulate a full disk by setting `reject_writes_min_disk_space_mb` to a hardcoded 10TB. On a host with more than 10TB free on the test filesystem, the write is not rejected and `ASSERT_NOK_STR_CONTAINS(..., "has insufficient disk space")` fails.

Fix: derive the threshold from the filesystem's actual free space.

## Issue Type

kind/failing-test

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23180]: https://yugabyte.atlassian.net/browse/DB-23180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33504, opened 2026-08-25
[#33503] DocDB: Fix DiskFullTest on hosts with more than 10TB of free disk space

ellabaron-code:
## Summary

Both TestDiskFull cases fail on machines with more than 10TB of free disk space: they simulate a full disk by hardcoding `reject_writes_min_disk_space_mb` to 10TB, so on bigger disks the write is not rejected. Derive the threshold from the test filesystem's actual free space (doubled for margin) instead.

## Test plan

- [x] `disk_full-test` passes on a Linux host with ~14TB free on the test filesystem (release build); both TestDiskFull cases failed there before this change.

### DIFF
--- a/src/yb/integration-tests/disk_full-test.cc
+++ b/src/yb/integration-tests/disk_full-test.cc
@@ -18,6 +18,8 @@
 
 #include "yb/server/clock.h"
 
+#include "yb/util/size_literals.h"
+
 DECLARE_uint64(reject_writes_min_disk_space_mb);
 DECLARE_uint32(reject_writes_min_disk_space_pct);
 DECLARE_uint32(reject_writes_min_disk_space_check_interval_sec);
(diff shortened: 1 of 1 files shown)


src/yb/integration-tests/disk_full-test.cc
```diff
@@ -122,8 +126,10 @@ TEST_F(YSqlDiskFullTest, YB_DISABLE_TEST_IN_ASAN(TestDiskFull)) {
   }
   ++i;
 
-  // Set a large limit to simulate disk full.
-  ANNOTATE_UNPROTECTED_WRITE(FLAGS_reject_writes_min_disk_space_mb) = 10 * 1024 * 1024;  // 10TB
+  // Set a limit above the actual free space to simulate disk full.
+  const auto free_space_mb =
+      ASSERT_RESULT(Env::Default()->GetFreeSpaceBytes(GetTestDataDirectory())) / 1_MB;
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_reject_writes_min_disk_space_mb) = free_space_mb * 2;
```
ttyusupov: Can we extract that common code into helper function and also use `std::max(free_space_mb * 2, 1)` to properly handle less than 1MB of disk space (although it would most likely fail in other tests, still worth handle it correctly here).

ttyusupov (changes requested):

src/yb/integration-tests/disk_full-test.cc
ellabaron-code: Fixed.

ttyusupov (approved):

Issue #33175, opened 2026-08-07
[DocDB] Support ignore_after_hybrid_time on the write path to fence writes from an expired lease


hari90:
Jira Link: [DB-22882](https://yugabyte.atlassian.net/browse/DB-22882)
## Description

**Is your feature request related to a problem or an enhancement?**

A client that holds a time-bounded lease over some resource currently has no way to guarantee that its writes stop taking effect once that lease expires. There is no server-side mechanism to reject a write that arrives (or lands) after a caller-supplied point in time.

An RPC deadline cannot substitute for one. Once an operation is handed to Raft nothing cancels it — `Consensus::ReplicateBatch` / `TEST_Replicate` take no deadline (`src/yb/consensus/consensus.h:213-215`). The round commits whenever a majority acknowledges it, and the response is simply dropped if the inbound call has already timed out. So a write that the client observed as `TimedOut` can still commit afterwards, which is precisely the outcome a lease is meant to prevent.

The primitive already exists in the codebase, but only for object locks:

- `optional fixed64 ignore_after_hybrid_time` on `AcquireObjectLockRequestPB` (`src/yb/tserver/tserver.proto:466`), mirrored on the master side (`src/yb/master/master_ddl.proto:835`)
- enforced in `TSLocalLockManager::CheckRequestForDeadline` (`src/yb/tserver/ts_local_lock_manager.cc:329-337`), which calls `server::UpdateClock` and then compares the field against `clock_->Now()`

There is no equivalent on the write path: nothing on `PgPerformOptionsPB` / `PgPerformRequestPB` (`src/yb/tserver/pg_client.proto:833-898`), `PgsqlWriteRequestPB` (`src/yb/common/pgsql_protocol.proto:203`), or `WriteRequestPB`.

**Provide a brief solution (if any) of what you want to happen.**

Add an `ignore_after_hybrid_time` field to the write path and have the tablet leader reject the operation when the fence has passed.

### Where the check belongs

The natural point is `RaftConsensus::DoAppendNewRoundsToQueueUnlocked` (`src/yb/consensus/raft_consensus.cc:1345`), because that is where the operation's hybrid time is minted. `round->NotifyAddedToLeader(op_id, committed_op_id)` (`:1380`) lands in `Operation::AddedToLeader`:

```cpp
hybrid_time = tablet->mvcc_manager()->AddLeaderPending(op_id);
...
replicate_msg->set_hybrid_time(hybrid_time.ToUint64());
...
[...]

PR #33507, opened 2026-08-25
[#33175] DocDB: Fence writes past ignore_after_hybrid_time

hari90:
## Summary

A lease lets one client be the sole writer to some data for a bounded window without
coordinating on every write: while it holds the lease it writes freely, and once the lease expires
another holder may take over. That is only safe if the outgoing holder's writes cannot take effect
after its window ends -- otherwise a delayed write from a former owner lands on top of the new
owner's data.

This change adds a per-write fence. The caller states the hybrid time after which the write must not take
effect, and the tablet leader checks it when admitting the write, rejecting it if that moment has
already passed. The check happens before the operation can have any effect, so a rejected write
definitively did not apply. It is reported as `Expired`
carrying `TabletServerErrorPB::WRITE_FENCE_EXPIRED`; the thin client surfaces it as `YBTHIN_FENCED`.

The fence rides on `PgsqlWriteRequestPB.ignore_after_hybrid_time` per op, and is lifted onto
`WritePB` as the earliest fence over the batch, that being the only message the leader's admission
path sees. 0 means no fence.

Two limits, both noted on the request field: each tablet leader judges the fence independently, so a
batch spanning tablets can be partly applied if it straddles the boundary; and being a hybrid time
comparison, the fence is only as tight as `max_clock_skew_usec`.

## Upgrade/Rollback safety

The feature can only be used after the upgrade is completed.

## Test Plan

`yb_thin_client-itest` -- full suite, 11/11.

`PgThinClientTest.WriteFencedByIgnoreAfterHybridTime` asserts on table contents as well as status: a
past fence is rejected and leaves no row, a far-future fence applies, no fence behaves as before, and
every peer reports zero running retryable requests afterwards.


src/yb/consensus/raft_consensus.cc
```diff
@@ -1397,6 +1416,21 @@ Status RaftConsensus::DoAppendNewRoundsToQueueUnlocked(
...
+    // filter is consulted below rather than after NotifyAddedToLeader: once an op is added as
+    // pending its side effects have happened and rolling back the op id does not undo them.
+    //
+    // The comparison is therefore against the clock rather than the op's own hybrid time, which
+    // AddLeaderPending has not assigned yet. The assigned time is >= this reading, so the check is
+    // permissive only by the interval between the two under the same lock. An op that passes and
+    // then stalls keeps its earlier position, so it cannot clobber a newer writer, and safe time
+    // will not advance past it.
+    if (auto s = CheckWriteFenceUnlocked(round); !s.ok()) {
+      round->NotifyReplicationFinished(s, round->bound_term(), /* applied_op_ids = */ nullptr);
```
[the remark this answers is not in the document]
hari90: Fixed. The fence check now runs *before* `RegisterRetryableRequest` rather than after, so a fenced round is never registered and there is nothing to deregister — it becomes kind (1) in `AppendNewRoundsToQueueUnlocked`'s taxonomy.

I took the reordering rather than routing through `state_->NotifyReplicationFinishedUnlocked` because it keeps the rejection off the retryable-request path entirely instead of relying on the unwind being correct. The comment at the new call site now states both ordering constraints — before registration, and before `NotifyAddedToLeader` — so neither is left to be re-derived.

`WriteFencedByIgnoreAfterHybridTime` now also asserts every peer reports zero running retryable requests after the rejection, which is the direct symptom you described: a leaked entry never drains, so the condition would never hold.

src/yb/consensus/raft_consensus.cc
```diff
@@ -1397,6 +1416,21 @@ Status RaftConsensus::DoAppendNewRoundsToQueueUnlocked(
...
+    // A write fenced by a caller-held lease is rejected here, for the same reason the operation
+    // filter is consulted below rather than after NotifyAddedToLeader: once an op is added as
+    // pending its side effects have happened and rolling back the op id does not undo them.
+    //
+    // The comparison is therefore against the clock rather than the op's own hybrid time, which
+    // AddLeaderPending has not assigned yet. The assigned time is >= this reading, so the check is
+    // permissive only by the interval between the two under the same lock. An op that passes and
+    // then stalls keeps its earlier position, so it cannot clobber a newer writer, and safe time
+    // will not advance past it.
+    if (auto s = CheckWriteFenceUnlocked(round); !s.ok()) {
```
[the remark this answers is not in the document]
hari90: Same issue as the thread just above, fixed the same way: the fence check moved ahead of `RegisterRetryableRequest`, so nothing is registered for a round that will be rejected. Your reading of the three rejection kinds is exactly right, and the new call-site comment records it. Covered by a new assertion in `WriteFencedByIgnoreAfterHybridTime` that every peer reports zero running retryable requests after the rejection.

src/yb/tablet/operations.proto
```diff
@@ -110,6 +110,12 @@ message WritePB {
 
   // See SysCDCStreamEntryPB.xcluster_use_target_applied_filter.
   optional bool xcluster_target_applied = 23;
+
+  // Fence from PgsqlWriteRequestPB.ignore_after_hybrid_time: the earliest fence over the ops of
+  // this batch. Lifted here because the replicated message carries only the DocDB write batch --
+  // the pgsql ops are not copied (WriteQuery's SetupKeyValueBatch) -- and the leader enforces the
+  // fence in RaftConsensus::DoAppendNewRoundsToQueueUnlocked, which sees only this message.
+  optional fixed64 ignore_after_hybrid_time = 24;
```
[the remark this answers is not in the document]
hari90: Fixed. Enforcement is now behind `--enable_write_fence_ignore_after_hybrid_time`, a runtime gFlag defaulting to `false`, and the gate sits at the point where the fence enters the replicated message (`WriteQuery`'s `SetupKeyValueBatch`) as well as at enforcement — so with the flag off field 24 never reaches `WritePB` and neither the new field nor the new behaviour can reach a mixed-version cluster. The PR description now carries a `## Upgrade/Rollback safety` section covering forward/backward behaviour and rollback.

I went with a default-off gFlag rather than an `AutoFlag` because the feature is still in development and an `AutoFlag` would not by itself close the gap for the out-of-tree thin client (your point on `yb_thin_client.h`). Promotion to an `AutoFlag` plus a capability signal is the shipping step; that is stated in the flag description, the proto comments and the PR body. The "No capability or version signal is included" line is gone.

src/yb/thin_client/yb_thin_client.cc
```diff
@@ -198,6 +198,12 @@ ybthin_status_code ClassifyStatus(const Status& status) {
   if (status.IsInvalidArgument() || status.IsNotSupported()) {
     return YBTHIN_INVALID;
   }
+  // The only Expired status this service produces is a fenced write (RaftConsensus rejects the op
+  // when ignore_after_hybrid_time has passed), so the caller can tell it apart from a failure that
+  // might still have taken effect. A fenced write definitively did not.
+  if (status.IsExpired()) {
```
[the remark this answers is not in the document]
hari90: Fixed — the premise was indeed false. The status now carries `WriteFenceExpiredError` (`yb/common/write_fence.h`, category 25) and `ClassifyStatus` matches that instead of `Status::IsExpired()`, so `DeadlineInfo`'s read deadline and `RetryableRequests::Register`'s "less than min running" / "too old" no longer report `YBTHIN_FENCED`.

One note on placement: the tag lives in `common/` rather than `consensus/` deliberately. The thin client does not link consensus (`THIN_CLIENT_LIBS` in its CMakeLists says so explicitly), so a `ConsensusError` code would not be decodable there. `MinRunningRequestIdStatusData` in `common/retryable_request.h` is the precedent for a consensus-produced code defined in `common/`. Verified end to end: the code survives the wire and the test asserts `YBTHIN_FENCED`.

src/yb/thin_client/yb_thin_client.h
```diff
@@ -225,6 +226,15 @@ typedef struct {
...
   size_t n_values;
+  // Fences this row against a lease the caller holds: the tablet leader rejects the write if this
+  // hybrid time has already passed by the time the op is assigned its own, reporting
+  // YBTHIN_FENCED. 0 means no fence.
+  //
+  // This exists because an RPC timeout cannot express it -- nothing cancels an operation once it
+  // reaches Raft, so a write that timed out from the caller's point of view can still commit. Each
+  // tablet leader judges the fence separately, so a batch spanning tablets can be partly applied
+  // when it straddles the boundary.
+  uint64_t ignore_after_hybrid_time;
```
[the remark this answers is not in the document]
hari90: Documented rather than gated on a capability signal, and the header now says so outright instead of leaving it to be inferred. `ignore_after_hybrid_time` is called out as NOT covered by the versioning rule at the top of the file: a tserver that does not enforce the fence — older than the field, or with `--enable_write_fence_ignore_after_hybrid_time` off, which is now the default — applies the write unfenced and reports `YBTHIN_OK`, with nothing in the response distinguishing that from a write inside its fence. So a caller must establish out of band that every tserver it writes to enforces the fence before treating success as proof, and leave the field 0 otherwise.

A capability signal is the right end state and belongs with `AutoFlag` promotion; it is not in this change. Tracked on #33175 point 5.

src/yb/tablet/write_query.cc
```diff
@@ -122,6 +124,20 @@ void SetupKeyValueBatch(const tserver::WriteRequestMsg& client_request, LWWriteP
...
+  // fences the batch -- these ops share a hybrid time, so they cannot be fenced apart.
+  uint64_t ignore_after_hybrid_time = 0;
+  for (const auto& op : client_request.pgsql_write_batch()) {
+    if (op.has_ignore_after_hybrid_time()) {
+      ignore_after_hybrid_time = ignore_after_hybrid_time
+          ? std::min(ignore_after_hybrid_time, op.ignore_after_hybrid_time())
+          : op.ignore_after_hybrid_time();
+    }
+  }
+  if (ignore_after_hybrid_time) {
```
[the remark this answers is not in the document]
hari90: Fixed — presence carries through now instead of the value being sentinel-tested, in both `SetupKeyValueBatch` (a `std::optional`, so an explicit 0 can no longer collapse the batch minimum) and `CheckWriteFenceUnlocked`. A present-and-zero fence therefore rejects.

Worth flagging that this is the opposite of gitar-bot's suggestion on the same lines; I went your way, for your reason: `PgsqlWriteRequestPB` is a public wire format that does not reserve 0, and 0 is `HybridTime::kMin`, so treating it as "no fence" would fail open on exactly the input a caller is most likely to get wrong. The 0-means-no-fence convention now lives only in the thin client ABI, which omits the field rather than sending a zero, and both protos state that presence is the signal.

src/yb/tablet/write_query.cc
```diff
@@ -122,6 +124,20 @@ void SetupKeyValueBatch(const tserver::WriteRequestMsg& client_request, LWWriteP
...
+    out_request->set_ignore_after_hybrid_time(ignore_after_hybrid_time);
+  }
```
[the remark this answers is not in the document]
hari90: Not taking this one, though the underlying concern is real — I fixed it the other way round.

`PgsqlWriteRequestPB` is a public wire format that does not reserve 0 as "no fence", and 0 is `HybridTime::kMin`: a fence that has unconditionally already passed. Treating a present-and-zero fence as "no fence" would fail open on exactly the input a caller is most likely to get wrong. So presence carries through end to end instead — a `std::optional` in `SetupKeyValueBatch` and a presence-only check in `CheckWriteFenceUnlocked` — which also removes the collapse-the-minimum problem you flagged, since an explicit 0 now rejects the batch rather than silently disabling the fence for its neighbours.

`yb-agentk-dev` raised the same lines from that angle. The 0-means-no-fence convention stays only in the thin client ABI, which omits the field rather than sending a zero.

src/yb/tablet/operations.proto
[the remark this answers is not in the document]
hari90: Correction to my reply above: the gFlag is gone again, and the change ships ungated.

Setting `ignore_after_hybrid_time` is itself the opt-in, and the only producer is `yb_thin_client`, which ships against tservers that carry the field. Nothing that exists today sets it, so no workload changes behaviour by upgrading — with the field unset the write path is byte-for-byte what it is today, and there is no new behaviour for a flag to keep out of a mixed-version cluster. A flag would gate something nobody can reach without asking for it.

The `## Upgrade/Rollback safety` section stays and now argues from that rather than from a flag: forward behaviour, per-leader enforcement in a mixed cluster (and the fact that a caller cannot infer from success that the fence held), and rollback — existing log entries carrying field 24 replay unfenced, which is correct because the fence is an admission decision taken at append time, so replay has nothing to re-decide. The proto comments say the same.

src/yb/thin_client/yb_thin_client.cc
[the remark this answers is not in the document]
hari90: Correction to my reply above: this is now a new code in an existing category rather than a new category.

`TabletServerErrorPB::Code` gains `WRITE_FENCE_EXPIRED = 34` (appended, nothing renumbered), and consensus attaches it through the same `tserver::TabletServerError` it already uses for `NOT_THE_LEADER` on this path. `src/yb/common/write_fence.h` and category 25 are gone.

The thin client can still decode it without linking consensus: `tserver_error.h` reaches it through `thin_client_proto` -> `pg_client_proto` -> `tserver_service_proto` -> `tserver_proto`. Verified — it links, and the test asserts `YBTHIN_FENCED` end to end.

src/yb/consensus/raft_consensus.h
```diff
@@ -406,6 +406,12 @@ class RaftConsensus : public std::enable_shared_from_this<RaftConsensus>,
         const ConsensusRounds& rounds, size_t* processed_rounds,
         std::vector<ReplicateMsgPtr>* replicate_msgs);
 
+  // Rejects a write whose caller-supplied fence (WritePB::ignore_after_hybrid_time) has already
+  // passed, so that a client holding a time-bounded lease can stop its writes taking effect once
+  // the lease is gone. Must be called before the round is registered with retryable requests and
+  // before it is added as pending -- see the call site for both.
```
hari90: remove

src/yb/tablet/write_query.cc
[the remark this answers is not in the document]
hari90: Reversing my reply above: 0 now means "no fence" everywhere, not just in the thin client ABI, so this is a plain value check rather than a presence check.

Making 0 a real fence meant the ABI's 0-means-no-fence convention had to be translated at the boundary, and left two layers disagreeing about what an explicit 0 means. Defining 0 as "no fence" at every layer removes the disagreement instead of documenting it — the `std::optional` and the `has_` tests are gone, an op with 0 is skipped when taking the batch minimum, and a `WritePB` fence of 0 is not enforced. Both protos state it.

So the fail-open case you identified is now the defined behaviour rather than a bug: a caller that wants a fence sets a non-zero one.

src/yb/tablet/write_query.cc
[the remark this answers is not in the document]
hari90: Taking this after all — 0 is now treated as "no fence", which is what you suggested.

The fix is a bit broader than the guard you proposed: rather than only skipping present-and-zero fences when taking the batch minimum, 0 means "no fence" at every layer, so the code is a plain value check with no `has_` tests or `std::optional` involved, and both protos say so. That also covers the enforcement side — a `WritePB` fence of 0 is not enforced, not just not propagated.

src/yb/common/pgsql_protocol.proto
```diff
@@ -324,6 +324,20 @@ message PgsqlWriteRequestPB {
...
+  // this -- nothing cancels an operation once it is handed to Raft, so a write the caller saw time
+  // out can still commit. 0 means no fence.
+  //
+  // Each leader evaluates the fence independently, so a batch spanning tablets can be partially
+  // applied when it straddles the boundary; non-transactional single-shard writes have no
+  // all-or-nothing to fall back on. Like any hybrid time comparison it is only as tight as
+  // max_clock_skew_usec.
+  //
+  // A tserver too old to know the field ignores it and applies the write unfenced, reporting
+  // ordinary success, so rely on the fence only against a cluster known to understand it.
```
hari90:
```suggestion
```

src/yb/consensus/raft_consensus.cc
```diff
@@ -1376,6 +1376,28 @@ Status RaftConsensus::AppendNewRoundsToQueueUnlocked(
...
+    return Status::OK();
+  }
+  // clock_, not the op's own hybrid time, which AddLeaderPending has not assigned yet. Nor
+  // state_->Clock(), which is the coarse clock and not in the fence's time domain.
+  const auto now = clock_->Now();
+  if (fence > now.ToUint64()) {
+    return Status::OK();
+  }
+  // WRITE_FENCE_EXPIRED rather than a bare Expired: Expired also covers causes that may have
+  // replicated, and a fenced write definitively did not.
```
hari90:
```suggestion
```

src/yb/tablet/write_query.cc
```diff
@@ -122,6 +124,20 @@ void SetupKeyValueBatch(const tserver::WriteRequestMsg& client_request, LWWriteP
     out_request->set_xrepl_origin_id(client_request.xrepl_origin_id());
   }
   out_request->set_batch_idx(client_request.batch_idx());
+  // Lift the ops' write fence into the replicated message, which is all RaftConsensus sees. The
+  // earliest fence wins; 0 means no fence. The ops share a hybrid time, so they cannot be fenced
+  // apart.
+  uint64_t ignore_after_hybrid_time = 0;
+  for (const auto& op : client_request.pgsql_write_batch()) {
```
hari90, line 130: Lets treat 0 as not set, so ignore just that op

src/yb/consensus/raft_consensus.cc
```diff
@@ -1376,6 +1376,28 @@ Status RaftConsensus::AppendNewRoundsToQueueUnlocked(
   return status;
 }
 
+Status RaftConsensus::CheckWriteFenceUnlocked(const ConsensusRoundPtr& round) {
+  const auto& msg = *round->replicate_msg();
+  if (msg.op_type() != OperationType::WRITE_OP) {
+    return Status::OK();
+  }
+  const auto fence = msg.write().ignore_after_hybrid_time();
+  if (!fence) {  // 0 means no fence.
```
hari90:
```suggestion
  if (!fence) { 
```

src/yb/common/pgsql_protocol.proto
hari90: Removed in d89155814f.

src/yb/consensus/raft_consensus.cc
hari90: Removed in d89155814f.

src/yb/consensus/raft_consensus.cc
hari90: Removed in d89155814f.

src/yb/tablet/write_query.cc
hari90: Done in d89155814f — the loop now reads the op's fence once and `continue`s on 0, so "0 is not set, ignore this op" is stated where the op is examined:

```cpp
for (const auto& op : client_request.pgsql_write_batch()) {
  const auto fence = op.ignore_after_hybrid_time();
  if (!fence) {  // Not set for this op.
    continue;
  }
  ignore_after_hybrid_time =
      ignore_after_hybrid_time ? std::min(ignore_after_hybrid_time, fence) : fence;
}
```

Behaviour is unchanged: the value test that was already there (line 132, `if (op.ignore_after_hybrid_time())`) skipped such an op and kept it from collapsing the batch minimum, since proto2 returns 0 for an unset fixed64. The `has_`-based version that did have that bug went away in a7c822dc61. This makes the intent explicit rather than implied by the guard on the accumulator.

Note the pre-existing object-lock paths (`ts_local_lock_manager.cc:333`, `object_lock_info_manager.cc:507`) still use `has_ignore_after_hybrid_time()`, so they would treat an explicit 0 as a fence already passed. Out of scope here — happy to change them separately.

src/yb/thin_client/yb_thin_client.cc
```diff
@@ -198,6 +199,12 @@ ybthin_status_code ClassifyStatus(const Status& status) {
   if (status.IsInvalidArgument() || status.IsNotSupported()) {
     return YBTHIN_INVALID;
   }
+  // The fence's own code, not the Expired category -- which a read deadline and a stale retryable
+  // request id also produce, on writes that may well have replicated.
+  if (tserver::TabletServerError::ValueFromStatus(status) ==
```
spolitov: Does not look like correct indent

src/yb/consensus/raft_consensus.cc
```diff
@@ -1376,6 +1376,26 @@ Status RaftConsensus::AppendNewRoundsToQueueUnlocked(
   return status;
 }
 
+Status RaftConsensus::CheckWriteFenceUnlocked(const ConsensusRoundPtr& round) {
+  const auto& msg = *round->replicate_msg();
+  if (msg.op_type() != OperationType::WRITE_OP) {
+    return Status::OK();
+  }
+  const auto fence = msg.write().ignore_after_hybrid_time();
```
spolitov: The generic comment to the whole diff - operate on HybridTime, use uint64 only for serialization.
There the following methods for it:
```
  static constexpr HybridTime FromPB(uint64_t value) {
    return value ? HybridTime(value) : HybridTime();
  }

  constexpr uint64_t ToPB() const {
    return *this ? v : 0;
  }
```

src/yb/common/pgsql_protocol.proto
```diff
@@ -335,6 +335,17 @@ message PgsqlWriteRequestPB {
...
+  // Fences the write against a caller-held lease: the leader rejects the op if this hybrid time has
+  // already passed when the op is about to be assigned its own. An RPC deadline cannot express
+  // this -- nothing cancels an operation once it is handed to Raft, so a write the caller saw time
+  // out can still commit. 0 means no fence.
+  //
+  // Each leader evaluates the fence independently, so a batch spanning tablets can be partially
+  // applied when it straddles the boundary; non-transactional single-shard writes have no
+  // all-or-nothing to fall back on. Like any hybrid time comparison it is only as tight as
+  // max_clock_skew_usec.
+  optional fixed64 ignore_after_hybrid_time = 32;
```
spolitov, line 348: Add comment that logical part is always 0 for now.

spolitov (changes requested):

spolitov (approved):

Issue #33506, opened 2026-08-25
[DocDB] Snapshot creation flushes RocksDB synchronously on the Raft apply path, blocking the tablet


craigsoules:
`CREATE_ON_TABLET` snapshot operations are applied inline on the Raft apply path while the tablet's ReplicaState mutex is held. The synchronous flush at the start of `TabletSnapshots::Create` is usually the longest part of that window on a write-heavy tablet; while it runs, writes, lease-checked reads, and follower `UpdateConsensus` all block on that tablet. Namespace-wide snapshots (backups, PITR schedules) hit every tablet at once, amplifying this into node-wide latency spikes.

The flush could be started when the operation is prepared, so the apply-time flush only covers the delta written during replication.

PR #33509, opened 2026-08-25
[#33506] DocDB: Preflush snapshots before Raft submission

craigsoules:
## Summary

Snapshot creation flushes while applying a Raft operation, holding the tablet's ReplicaState mutex
and potentially stalling writes, lease-checked reads, and follower replication. Prepare-time
flushing can be overtaken by apply; waiting for it there would block consensus.

Replace prepare-time flushing with an optional leader preflight before submitting `CREATE_ON_TABLET`:

- Pin history, resolve intents, capture the active configuration, and flush every captured replica
  using separate `FlushTablets` RPCs and local flushing. Recheck leadership, term, membership, and
  deadline before submission; failures prevent submission, with no best-effort fallback.
- Use separate bounded orchestration and receiver executors, excluding overlapping work per tablet.
  Caller timeouts do not release admission before physical flush work and error cleanup retire.
- Retain the synchronous all-DB apply-time flush. Preflight does not establish a common flushed OpId
  or exclude later writes, so apply-time work can remain substantial.
- Keep equal-timestamp history registrations independent. Propagate terminal vector/storage failures
  to dependent intents flushes and synchronous/manifest waiters rather than strand shutdown.

Preflight is **default-off**: requiring every captured replica reduces snapshot availability even
when a healthy quorum can still commit writes. Neither the preparer nor `UpdateConsensus` waits for
preflight; ordinary storage contention and write stalls still apply.

| Flag | Default | Scope |
| --- | --- | --- |
| `snapshot_create_flush_before_submit` | `false` | Runtime; enables leader preflight |
| `snapshot_preflush_timeout_ms` | `10000` | Runtime; capped by the incoming RPC deadline |
| `snapshot_preflush_concurrency` | `4` | Startup; outstanding leader preflights per server |
| `tablet_flush_concurrency` | `4` | Startup; admitted admin/local flush batches per server |

## Upgrade/Rollback safety

No wire-format, on-disk-format, or catalog changes. Supported older followers use the existing
`FlushTablets` API; only upgraded, enabled leaders orchestrate preflight. Mixed versions do not
provide a cluster-wide latency guarantee.

Disable `snapshot_create_flush_before_submit` to restore direct submission plus the final synchronous
flush. The branch-only `snapshot_create_flush_on_prepare` flag is removed without a compatibility
shim; remove it from any configuration used to test earlier revisions of this PR.

Shared admin-flush admission and storage error handling apply even with preflight disabled. Storage
filter errors fail closed and make the affected RocksDB read-only, including when paranoid checks
are disabled; disabling preflight does not clear a cached storage failure.

## Test plan

Local validation on the development base, before the publication rebase (release, clang21, arm64;
one exact test filter per invocation):

- [x] 48 targeted tests covering snapshot data, RF=3 write progress, missing/unavailable followers,
  leadership/configuration changes, timeouts, history pins, admission limits, and flush retirement.
- [x] New failure regressions: out-of-order synchronous vector promises, manifest-waiter shutdown,
  memtable-selection rollback, and integrated vector/intents failure with unchanged durable frontiers
  and admission held through cleanup. Vector cases include usearch and hnswlib; the integrated case
  also passed with the YbHnsw usearch backend.
- [x] Three additional repetitions each of `EarlierFailureCompletesSynchronousFlush`,
  `ManifestFailureReleasesCompactionWaiter`, `FailedVectorFlushRetiresIntentsAndAdmission`, and
  `SnapshotTest.PreflushDoesNotBlockReplication`.
- [x] Release daemon build.
- [x] C++ lint and `git diff --check`; branch-wide lint also passed after rebasing onto current master.

The build/test matrix has not been rerun on the rebased PR head. Sanitizers and load benchmarks
were not run.

### DIFF
--- a/src/yb/tablet/tablet_snapshots.cc
+++ b/src/yb/tablet/tablet_snapshots.cc
@@ -229,7 +229,8 @@ TabletSnapshots::~TabletSnapshots() {
   CompleteShutdown();
 }
 
-void TabletSnapshots::SetCleanupPool(ThreadPool* thread_pool, rpc::Scheduler* scheduler) {
+void TabletSnapshots::SetCleanupPool(
+    ThreadPool* thread_pool, rpc::Scheduler* scheduler, ThreadPool* preflush_pool) {
   {
     std::lock_guard lock(cleanup_mutex_);
     if (shutting_down_) {
@@ -237,7 +238,9 @@ void TabletSnapshots::SetCleanupPool(ThreadPool* thread_pool, rpc::Scheduler* sc
     }
     DCHECK(!cleanup_token_);
     cleanup_token_ = thread_pool->NewToken(ThreadPool::ExecutionMode::SERIAL);
-    preflush_token_ = thread_pool->NewToken(ThreadPool::ExecutionMode::CONCURRENT);
+    if (preflush_pool) {
+      preflush_token_ = preflush_pool->NewToken(ThreadPool::ExecutionMode::CONCURRENT);
+    }
     retry_task_tracker_.Bind(scheduler);
   }
 
@@ -397,7 +400,7 @@ void TabletSnapshots::SubmitPreflush() {
   // On the leader, Prepare runs on the preparer thread, but on followers it runs inline in
   // UpdateConsensus while the ReplicaState lock is held (see PreparerImpl::Submit). Even a
   // non-waiting Tablet::Flush call synchronously enters the RocksDB write queue to switch the
-  // memtable, so dispatch it to the cleanup pool instead of running it here. The token is
+  // memtable, so dispatch it to the preflush pool instead of running it here. The token is
   // shut down in CompleteShutdown, so the task cannot outlive this object.
   {
     std::lock_guard lock(cleanup_mutex_);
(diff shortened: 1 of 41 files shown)


src/yb/tablet/tablet_snapshots.cc
```diff
@@ -368,9 +383,52 @@ bool TabletSnapshots::IsLastSnapshotTimeFilePath(const std::string& dir) {
 }
 
 Status TabletSnapshots::Prepare(SnapshotOperation* operation) {
+  if (FLAGS_snapshot_create_flush_on_prepare &&
+      operation->operation() == tserver::TabletSnapshotOpRequestPB::CREATE_ON_TABLET) {
+    SubmitPreflush();
+  }
```
yamen-haddad, line 389: My objection to the async pre-flush is that it isn't deterministic and, at best, we are saving the time to replicate the RAFT op. i.e suppose flush time is `Tf`and the replication round trip is `Tr`. Then, if there is no wait in the flush thread pool at all, we will save `Tf-Tr`. I would love for us to measure how much that is, but my initial estimation is that `Tf>>Tr` and thus the savings are negligible.

We discussed offline and converged on the following:
We will do a pre-flush right before submitting the snapshot operation to RAFT. The pre-flush is initiated at the RPC handler thread of `TabletServiceBackupImpl::TabletSnapshotOp`. So the new workflow will be:

1. Reader registration, ResolveIntents, all as today.
2. Fan out `FlushTabletsto` the followers and start a local `Flush(kSync, kAllDbs)` in parallel. Both take roughly the same time `Tf`.
3. Submit the operation.

yamen-haddad (changes requested):

Issue #33511, opened 2026-08-25
[YSQL] PgRelcachePreloadMem tests fail on many-core hosts


ellabaron-code:
Jira Link: [DB-23182](https://yugabyte.atlassian.net/browse/DB-23182)
## Description
PgRelcachePreloadMemMinimalTest.FreshEmptyDbStaysBounded and
PgRelcachePreloadMemNormalTest.FreshEmptyDbStaysBounded fail on many-core
hosts: backend peak RSS includes tcmalloc per-CPU caches, which scale with
core count. Measured ~88 MB on 8 cores vs ~103 MB on 192 cores for the same
build, so the 100 MB cap fails on large machines.

Fix (PR incoming): raise the cap to 120 MB. The ~560 MB regression the test
guards against still exceeds the new cap by a wide margin.

## Issue Type
kind/failing-test

- [x] I confirm this issue does not contain any sensitive information.

[DB-23182]: https://yugabyte.atlassian.net/browse/DB-23182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33513, opened 2026-08-25
[#33511] YSQL: Raise relcache preload mem test RSS cap to 120 MB

ellabaron-code:
## Summary
The per-backend peak-RSS cap in pg_relcache_preload_mem-test fails on
many-core hosts: backend peak RSS includes tcmalloc per-CPU caches, which
scale with core count. Measured ~88 MB on 8 cores vs ~103 MB on 192 cores
for the same build. Raise the cap from 100 MB to 120 MB; the ~560 MB
regression the test guards against still exceeds it by a wide margin.

## Test plan
- [x] Ran PgRelcachePreloadMem tests on a 192-core Linux host (release):
      failed with the 100 MB cap (peak RSS 103/102 MB), pass with 120 MB.
- [x] Ran on an 8-core Linux host (release): peak RSS 88-89 MB, passes.


kai-franz (approved):
LGTM!

Issue #33516, opened 2026-08-25
[DocDB] Session-level advisory lock txn start time sent as monotonic nanoseconds, hiding locks from pg_locks on long-uptime hosts


ellabaron-code:
Jira Link: [DB-23186](https://yugabyte.atlassian.net/browse/DB-23186)
## Description

`BeginPgSessionLevelTxnIfNecessary` (pg_client_session.cc) sets the session-level (kPgSession) transaction's start time with `MonoTime::Now().ToUint64()` — monotonic **nanoseconds since boot** — while every consumer treats `pg_txn_start_us` as **microseconds since the Unix epoch** (which is what the plain-txn path sends).

On a host that has been up longer than ~20.7 days, nanoseconds-since-boot exceeds current epoch-microseconds, so the session txn looks like it starts in the far future:

- It sorts newest in the coordinator's first_touch index, so `GetOldTransactions`' break-on-young-txn scan can exclude it whenever a young pending txn shares its coordinator — `pg_locks` then shows **none** of the session's advisory locks.
- It is the first transaction evicted by the `pg_locks` `max_num_txns` priority queue (which pops largest start time first).
- It is mis-ordered in wait-queue fairness (waiters resume in `txn_start_us` order).

On short-uptime hosts the same value is smaller than any epoch timestamp, so the txn just looks infinitely old and everything works — which is why CI never catches this.

**Repro:** run `PgObjectLocksTest.ConsecutiveCreateIndexDontDeadlock` on a host with uptime greater than ~21 days; it fails with advisory lock count 0 (expected 3). Verified on hosts with 82 and 210 days of uptime.

Issue Type: kind/bug

- [x] I confirm this issue does not contain any sensitive information.

[DB-23186]: https://yugabyte.atlassian.net/browse/DB-23186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33517, opened 2026-08-25
[#33516] DocDB: Fix session-level advisory lock txn start time units

ellabaron-code:
## Summary

`BeginPgSessionLevelTxnIfNecessary` stamped the session-level (advisory lock) transaction's start time with `MonoTime::Now().ToUint64()` — monotonic nanoseconds since boot — while every consumer treats `pg_txn_start_us` as microseconds since the Unix epoch. On hosts up longer than ~20.7 days that value exceeds current epoch-microseconds, so the txn looks like it starts in the far future: the coordinator's `GetOldTransactions` scan can exclude it (`pg_locks` then shows none of the session's advisory locks), the `pg_locks` `max_num_txns` priority queue evicts it first, and wait-queue fairness mis-orders it. On fresh hosts the same value is smaller than any epoch timestamp, so the txn just looks infinitely old and everything works — which is why CI never catches this.

Fix: send the hybrid clock's physical time (epoch microseconds), matching what the plain-txn path sends.

## Test plan

- [x] `PgObjectLocksTest.ConsecutiveCreateIndexDontDeadlock` on a release build on a host with 82 days of uptime: failed before the fix (advisory lock count 0, expected 3), passes with it.

### DIFF
--- a/src/yb/tserver/pg_client_session.cc
+++ b/src/yb/tserver/pg_client_session.cc
@@ -3598,11 +3598,12 @@ class PgClientSession::Impl {
       txn = transaction_provider_.Take<kSessionKind>(deadline);
       txn->SetLogPrefixTag(kTxnLogPrefixTag, id_);
       txn->InitPgSessionRequestVersion();
-      // Set the start time before initializing the transaction to allow start time to be
-      // propagated to txn coordinator.
+      // Set the start time (epoch microseconds, like pg_txn_start_us) before initializing the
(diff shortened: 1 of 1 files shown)


src/yb/tserver/pg_client_session.cc
basavaraj29: can you fix this as well, thanks

ellabaron-code: Fixed.

basavaraj29 (approved):
thanks!

Issue #33540, opened 2026-08-26
[DocDB] MiniCluster: client connections to broadcast addresses reset on macOS


ellabaron-code:
Jira Link: [DB-23199](https://yugabyte.atlassian.net/browse/DB-23199)
## Description

Issue Type: kind/bug

On macOS, an unbound socket that connects to a loopback alias gets the destination address itself as its source address (src == dst); Linux picks 127.0.0.1.

The MiniCluster connectivity simulation filters incoming connections by source address, so a client dialing a server's broadcast address appears to connect from that blacklisted address and is reset. Since the fix for issue #28539 made yb-admin prefer broadcast addresses over private ones, these tests fail on macOS:

- `XClusterAdminCliTest.TestListCdcStreamsWithBootstrappedStreams`
- `XClusterAdminCliTest.TestDeleteCDCStreamWithBootstrap`
- `XClusterAlterUniverseAdminCliTest.TestAlterUniverseReplicationWithBootstrapId`
- `XClusterAdminCliTest_Large.TestBootstrapProducerPerformance`
- `NotSupportedTabletSplitITest.SplittingWithBootstrappedStream`
- `AreLeadersOnPreferredOnlyTest` (both tests)

with:

```
Network error (yb/util/net/socket.cc:544): recvmsg error: Connection reset by peer (system error 54)
```

Fix: bind outbound sockets to 127.0.0.1 / ::1 when no source address is configured and the remote is loopback (macOS only).

I confirm this issue does not contain any sensitive information.

[DB-23199]: https://yugabyte.atlassian.net/browse/DB-23199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #28539, opened 2025-09-09
[DocDB] Certain yb-admin CDC commands fail when servers have separate broadcast and rpc addresses


iSignal:
Jira Link: [DB-18238](https://yugabyte.atlassian.net/browse/DB-18238)
### Description

In the YB-admin call made to run the cdc rpc, we’re using the private_rpc_address of the first tserver in list of tservers from ListTabletServers response.

Rpc address for the tservers in Istio is not resolvable( its set to POD_IP, 127.0.0.1 ) outside pod.

We switched to run the yb-admin on the first tserver from the list_all_tablet_servers response, so that the private rpc address is resolvable as its run inside the pod. That made it work.
I think we’ll need to fix the yb-admin client code to use broadcast addresses instead of private_rpc_address ? https://github.com/yugabyte/yugabyte-db/blob/6f86a895f9509c44206521464449fc2a627e8499/src/yb/tools/yb-admin_client.cc#L3873
Fetching the rpc address here: https://github.com/yugabyte/yugabyte-db/blob/6f86a895f9509c44206521464449fc2a627e8499/src/yb/tools/yb-admin_client.cc#L4461 

We should use logic similar to

```
const HostPortPB& PublicHostPort(const ServerRegistrationPB& registration) {
  return GetHostPort(registration.broadcast_addresses(),
                     registration.private_rpc_addresses(),
                     PublicAddressAllowed::kTrue);
}
```
in yb client instead.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-18238]: https://yugabyte.atlassian.net/browse/DB-18238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33541, opened 2026-08-26
[#33540] DocDB: Bind outbound loopback connections to 127.0.0.1 on macOS

ellabaron-code:
## Summary

On macOS, an unbound socket that connects to a loopback alias gets the destination address itself as its source address (src == dst); Linux picks 127.0.0.1.

The MiniCluster connectivity simulation filters incoming connections by source address. A client dialing a server's broadcast address therefore appears to connect from that blacklisted address and is reset. Since the fix for issue #28539 made yb-admin prefer broadcast addresses over private ones, several xCluster bootstrap / admin-cli tests and AreLeadersOnPreferredOnlyTest fail on macOS with "Connection reset by peer" (see #33540 for the full list).

Bind the outbound socket to 127.0.0.1 / ::1 when no source address is configured and the remote is loopback, macOS only — Linux behavior is unchanged. Server messengers in MiniCluster set an explicit outbound base address and are unaffected, so the simulated connectivity breaking still works.

## Test plan

On macOS arm64 (release build), all previously failing with "Connection reset by peer", now passing:

- [x] `./yb_build.sh release --cxx-test tools_yb-admin-xcluster-test --gtest_filter '*Bootstrap*'`
- [x] `./yb_build.sh release --cxx-test integration-tests_xcluster-tablet-split-itest --gtest_filter NotSupportedTabletSplitITest.SplittingWithBootstrappedStream`
- [x] `./yb_build.sh release --cxx-test integration-tests_are_leaders_on_preferred_only-itest`

### DIFF
--- a/src/yb/rpc/reactor.cc
+++ b/src/yb/rpc/reactor.cc
@@ -789,6 +789,15 @@ Status Reactor::FindOrStartConnection(const ConnectionId &conn_id,
     auto outbound_address = conn_id.remote().address().is_v6()
         ? messenger_.outbound_address_v6()
         : messenger_.outbound_address_v4();
+#if defined(__APPLE__)
+    // macOS picks the destination as the source for unbound loopback sockets (src == dst);
+    // bind to the loopback address like Linux does.
+    if (outbound_address.is_unspecified() && conn_id.remote().address().is_loopback()) {
+      outbound_address = conn_id.remote().address().is_v6()
+          ? IpAddress(boost::asio::ip::address_v6::loopback())
+          : IpAddress(boost::asio::ip::address_v4::loopback());
+    }
+#endif
     if (!outbound_address.is_unspecified()) {
       auto status = sock.SetReuseAddr(true);
       if (status.ok()) {


iSignal:
> The MiniCluster connectivity simulation filters incoming connections by source address. A client dialing a server's broadcast address therefore appears to connect from that blacklisted address and is reset.

What does connectivity simulation mean? It seems that for this to work reliably, the test needs to set local_ip_for_outbound? Does this also include client -> PG connectivity, for which that param does not work?

ellabaron-code:
Connectivity simulation = `Messenger::BreakConnectivity`: MiniCluster gives each node its own loopback alias, and these tests make a server messenger reject incoming connections by source address (`TEST_ShouldArtificiallyRejectIncomingCallsFrom` in `messenger.cc`). On macOS an unbound client dialing `127.0.0.x` also gets `127.0.0.x` as its source, so it appears to come from a rejected address.

Setting `local_ip_for_outbound_sockets` would work too, but it's process-global: it would need plumbing into every `yb-admin` invocation and every affected test, and in MiniCluster the client shares the process with the servers. The reactor fix only applies when neither that flag nor a messenger outbound address is set, so explicit configuration still wins.

Client→PG connections are out of scope both ways: libpq sockets don't go through the reactor, and nothing filters PG connections by source address — the simulation exists only at the YB RPC layer. The PG→tserver direction does go through the reactor, so it's covered.

spolitov (approved):

PR #33555, opened 2026-08-26
[DEVOPS-3901] build: Add selector keywords to jenkins_jobs.yml

svarnau:
Build list can by differentiated for full official builds and change testing.

Future use may include some build/test combinations that only apply to changes and not full builds, such as smoke/subset testing by combining when keywords with test_opts.

For comment-only change:
Jenkins: skip

### DIFF
--- a/jenkins_jobs.yml
+++ b/jenkins_jobs.yml
@@ -5,11 +5,6 @@
 #   change - run by default for DB changes
 #   full - run for full official build
 #
-# To request optional build/test on PR, include Jenkins: directives
-# in the description.
-# Examples:
-#   Jenkins: build type: tsan
-#   Jenkins: jobs: full
 
 jobs:
   - os: alma9
--- a/jenkins_jobs.yml
+++ b/jenkins_jobs.yml
@@ -50,6 +50,7 @@ jobs:
     build_opts: YB_LINKING_TYPE=full-lto YB_PGO_BOLT=1
     release_artifact: true
     when:
+     - change
      - full
 
   - os: mac14
--- a/jenkins_jobs.yml
+++ b/jenkins_jobs.yml
@@ -3,11 +3,13 @@
 #
 # when selectors:
 #   change - run by default for DB changes
-#   sanitize - optional addition for DB changes
 #   full - run for full official build
 #
-# To request optional build/test on PR, include "Jenkins: jobs: <selector-keyword>"
+# To request optional build/test on PR, include Jenkins: directives
 # in the description.
+# Examples:
+#   Jenkins: build type: tsan
+#   Jenkins: jobs: full
 
 jobs:
   - os: alma9
@@ -15,7 +17,7 @@ jobs:
     build_type: asan
     release_artifact: false
     when:
-     - sanitize
+     - change
      - full
 
   - os: alma8
@@ -23,7 +25,6 @@ jobs:
     build_type: tsan
     release_artifact: false
     when:
-     - sanitize
      - full
 
   - os: alma8
@@ -41,6 +42,7 @@ jobs:
     build_opts: YB_LINKING_TYPE=full-lto YB_PGO_BOLT=1
     release_artifact: true
     when:
+     - change
      - full
 
   - os: alma8
@@ -50,7 +52,6 @@ jobs:
     build_opts: YB_LINKING_TYPE=full-lto YB_PGO_BOLT=1
     release_artifact: true
     when:
-     - change
      - full
 
   - os: mac14
--- a/jenkins_jobs.yml
+++ b/jenkins_jobs.yml
@@ -1,27 +1,47 @@
 # This file determines the jobs that we run in this branch on Jenkins.
 # Default architecture is x86_64
+#
+# when selectors:
+#   change - run by default for DB changes
+#   sanitize - optional addition for DB changes
+#   full - run for full official build
+#
+# To request optional build/test on PR, include "Jenkins: jobs: <selector-keyword>"
+# in the description.
+
 jobs:
   - os: alma9
     compiler: clang21
     build_type: asan
     release_artifact: false
+    when:
+     - sanitize
+     - full
 
   - os: alma8
     compiler: clang21
     build_type: tsan
     release_artifact: false
+    when:
+     - sanitize
+     - full
 
   - os: alma8
     compiler: gcc15
     build_type: fastdebug
     test_opts: YB_TEST_YB_CONTROLLER=0 YB_ENABLE_YSQL_CONN_MGR_IN_JAVA_TESTS=true
     release_artifact: false
+    when:
+     - change
+     - full
 
   - os: alma8
     compiler: clang21
     build_type: release
     build_opts: YB_LINKING_TYPE=full-lto YB_PGO_BOLT=1
     release_artifact: true
+    when:
+     - full
 
   - os: alma8
     compiler: clang21
@@ -29,6 +49,8 @@ jobs:
     architecture: aarch64
     build_opts: YB_LINKING_TYPE=full-lto YB_PGO_BOLT=1
     release_artifact: true
+    when:
+     - full
 
   - os: mac14
     compiler: clang21
@@ -36,6 +58,8 @@ jobs:
     architecture: arm64
     test_opts: YB_TEST_YB_CONTROLLER=0
     release_artifact: true
+    when:
+     - full
 
   - os: mac14
     compiler: clang21
@@ -43,9 +67,13 @@ jobs:
     architecture: x86_64
     test_opts: YB_COMPILE_ONLY=1
     release_artifact: true
+    when:
+     - full
 
   - os: ubuntu22.04
     compiler: clang21
     build_type: debug
     release_artifact: false
     test_opts: YB_COMPILE_ONLY=1
+    when:
+     - full


svarnau:
Given the relative costs, we are considering trimming the pre-commit list to just two flavors.

jenkins_jobs.yml
```diff
@@ -1,51 +1,80 @@
...
 
   - os: alma8
     compiler: clang21
     build_type: release
     architecture: aarch64
     build_opts: YB_LINKING_TYPE=full-lto YB_PGO_BOLT=1
     release_artifact: true
+    when:
+     - change
+     - full
```
jasonyb, line 50: according to internal https://yugabyte.slack.com/archives/C0BSDAULQDP/p1788152221452779, you flipped the two release types
```suggestion
     - change
     - full

- os: alma8
compiler: clang21
build_type: release
architecture: aarch64
build_opts: YB_LINKING_TYPE=full-lto YB_PGO_BOLT=1
release_artifact: true
    when:
     - full
```

jenkins_jobs.yml
```diff
@@ -1,51 +1,80 @@
 # This file determines the jobs that we run in this branch on Jenkins.
 # Default architecture is x86_64
+#
+# when selectors:
+#   change - run by default for DB changes
+#   sanitize - optional addition for DB changes
```
jasonyb: what is the purpose of this "sanitize" category? why not simply have everything non-change be optional additions? Or if "sanitize" is specifically meant to cover sanitizer builds, what is the point?  Seems this is detached from the spec that was agreed upon: asan belongs in "change", not "sanitize"

jenkins_jobs.yml
```diff
@@ -1,51 +1,80 @@
...
+#   full - run for full official build
+#
+# To request optional build/test on PR, include "Jenkins: jobs: <selector-keyword>"
```
jasonyb: this seems not flexible. What if one wants to do a union of two selectors? or simply run the exact jobs they want to? It would be simpler to only have "change" and "full" and users can choose a list of individual build flavors or selectors so that they are not artificially constrained to certain choices.

jasonyb (changes requested):

jasonyb:
Stepping back from the inline comments, I think the keyword set can be collapsed while keeping the capability this PR is after.

- Drop `sanitize`. Per the spec that was agreed upon, asan belongs in `change`. For opting into other flavors, the existing `build type:` and `platform:` flags already let a user name exactly what they want, provided a jenkins-helpers follow-up makes an explicit flag select from the full matrix instead of intersecting with the default change set. As merged in jenkins-helpers#1574, the `when` filter runs before the build type filter, so `Jenkins: build type: asan` alone would schedule nothing under this PR's tags unless the user also knows to write `jobs: sanitize`. That is a trap.
- Drop `full`. Every entry here carries it, so it is pure ceremony, and it introduces a silent failure mode where a job whose `when` list is missing `full` quietly disappears from official release builds. If every entry is implicitly part of full builds, ReleaseAllBuild needs no `when` handling at all, and the only per-job annotation left is change membership, e.g. `when: [change]` or a plain boolean.
- Keep `Jenkins: jobs: full` (or `all`) as the request keyword for running the whole matrix on a change, since that should be the common opt-up case.

Regarding the future smoke/subset idea in the PR description, I don't think `full` tagging is the right vehicle for it anyway. A change-only variant would share os, compiler, and build type with its official twin, and the Phabricator pipeline keys scheduled builds by that tuple, so `jobs: full` on a diff would make the two variants collide with one silently overwriting the other. That feature needs its own design when it becomes real, and dropping `full` now does not foreclose it.

Concretely for this PR, that means tagging only the change-set jobs (membership per the agreed spec and the Slack thread, as flagged inline) and leaving the rest untagged, with the ReleaseAllBuild and filter-ordering adjustments as jenkins-helpers follow-ups.

_automated · Fable 5_

jenkins_jobs.yml
svarnau: It seems to me that hardly anyone uses these directives because they are too confusing. I was thinking simple keyword selectors would make it more obvious even though not fully expressive mix/match.

So, I'll trim it back to simple examples that show selecting an extra build type or selecting all (use the old behavior).

jenkins_jobs.yml
svarnau: removed

jenkins_jobs.yml
svarnau: fixed

jasonyb (approved):
Unless you can otherwise find a case (now or planned near future) where full does not cover every job, I still argue to drop the "when: full" entirely: https://github.com/yugabyte/yugabyte-db/pull/33555#issuecomment-5486929009

svarnau:
> Unless you can otherwise find a case (now or planned near future) where full does not cover every job, I still argue to drop the "when: full" entirely: [#33555 (comment)](https://github.com/yugabyte/yugabyte-db/pull/33555#issuecomment-5486929009)

The use case is running a subset of tests (smoke test) for "change" and all tests for "full". This is a case that has been suggested often. We just don't have any implemented cases of it now.

jasonyb:
> > Unless you can otherwise find a case (now or planned near future) where full does not cover every job, I still argue to drop the "when: full" entirely: [#33555 (comment)](https://github.com/yugabyte/yugabyte-db/pull/33555#issuecomment-5486929009)
> 
> The use case is running a subset of tests (smoke test) for "change" and all tests for "full". This is a case that has been suggested often. We just don't have any implemented cases of it now.

And why does this need to be implemented by a "when: full"?  My earlier comment suggests to have "full" implicitly map to all jobs rather than explicitly tagging all jobs with "full".

svarnau:
> > > Unless you can otherwise find a case (now or planned near future) where full does not cover every job, I still argue to drop the "when: full" entirely: [#33555 (comment)](https://github.com/yugabyte/yugabyte-db/pull/33555#issuecomment-5486929009)
> > 
> > 
> > The use case is running a subset of tests (smoke test) for "change" and all tests for "full". This is a case that has been suggested often. We just don't have any implemented cases of it now.
> 
> And why does this need to be implemented by a "when: full"? My earlier comment suggests to have "full" implicitly map to all jobs rather than explicitly tagging all jobs with "full".

Because the smoke test would be change only, not full.

jasonyb:
> > > > Unless you can otherwise find a case (now or planned near future) where full does not cover every job, I still argue to drop the "when: full" entirely: [#33555 (comment)](https://github.com/yugabyte/yugabyte-db/pull/33555#issuecomment-5486929009)
> > > 
> > > 
> > > The use case is running a subset of tests (smoke test) for "change" and all tests for "full". This is a case that has been suggested often. We just don't have any implemented cases of it now.
> > 
> > 
> > And why does this need to be implemented by a "when: full"? My earlier comment suggests to have "full" implicitly map to all jobs rather than explicitly tagging all jobs with "full".
> 
> Because the smoke test would be change only, not full.

Not sure where we are missing each other.  My point is, for the smoke test, you could say something like "Jenkins: jobs: change", and for the full run, you can say "Jenkins: jobs: full", and that is still compatible with not having to specify full in this yml file because the logic somewhere else can interpret full as all jobs.

The bigger point is that this "jobs" directive introduces a new interaction that ought to be thought through: what if "jobs: full" and "compiler: gcc"?  What if "jobs: change" and "build type: tsan"?  How do the two interact?  What happens if "jobs" is not specified but "build type: tsan" is?

svarnau:
> Not sure where we are missing each other. My point is, for the smoke test, you could say something like "Jenkins: jobs: change", and for the full run, you can say "Jenkins: jobs: full", and that is still compatible with not having to specify full in this yml file because the logic somewhere else can interpret full as all jobs.

Maybe we are mixing two issues. Even if we do not support a "jobs:" keyword, I contend that we should have both full and change keywords, because we need a selection criteria for official release builds. If an official build just includes everything in the file, then there is no way to specify test subsets that should only run for changes and NOT for official builds. In that way I could have a Mac-all-tests with full and a Mac-minimal-tests with change, but not full, since it is a subset of the other.  Likewise, if we decided to run faster c++ tests for changes and run all tests in official build, we need a mechanism to allow it.

svarnau:
> The bigger point is that this "jobs" directive introduces a new interaction that ought to be thought through: what if "jobs: full" and "compiler: gcc"? What if "jobs: change" and "build type: tsan"? How do the two interact? What happens if "jobs" is not specified but "build type: tsan" is?

It is a valid point. The jenkins directive keywords for selecting builds are confusing. I was hoping to simplify it with keywords that show up in this file. But maybe that increases the confusion.  I'm willing to drop support of the jobs keyword, but the intent was to have an easy way to specify the old behavior of running all the builds.

I suspect we do not really need to support all the old complexity of selecting by os/architecture/compiler/etc. People normally just go with the default, and if we gave small set of choices like small/medium/large it would be sufficient. But it is fine with me to stick with old keywords if that is sufficient.

hkimura:
Non blocking comment: csi-diff-summary's "Multi-Build Test Failures" currently uses `BLD_LIMIT = 3`, so with three `change` lanes it now fires only when a test fails in all of them. Previously "3 of 8" was a decent cross-flavor signal, now is it? I actually think that dropping "Multi-Build Test Failures" altogether is one option now, relying more on background checks we've been discussing, so I don't have a strong opinion.

hkimura (approved):

svarnau:
> Non blocking comment: csi-diff-summary's "Multi-Build Test Failures" currently uses `BLD_LIMIT = 3`, so with three `change` lanes it now fires only when a test fails in all of them. Previously "3 of 8" was a decent cross-flavor signal, now is it? I actually think that dropping "Multi-Build Test Failures" altogether is one option now, relying more on background checks we've been discussing, so I don't have a strong opinion.

Good point. I've logged a Jira against that check so that we'll flag the same failure on 2 builds in future. Probably will wait until this change is back ported to make that change.

PR #18567, opened 2023-08-04
Testing Automatic Grammar/Diagram generation

premkumr:
Automatic Generation of BNF diagrams and Grammar
-------------------------------------------------------
- Refer to [Spec doc](https://docs.google.com/document/d/1Qk3wkP0POG4aZao71WT2a6B0Xf7TZ-TqzYnhGHATQcs/edit) for full details
- Adds `ebnf` shortcode with which diagrams, grammar, and tabs can be generated dynamically
- The `ebnf` shortcode takes 2 parameters
    - `type` (default: `tabs`)
         - `tabs`: Generate the grammar + diagram tabs
         - `grammar`: Generate just the grammar
         -  `diagram`: Generate just the diagram
         -  `reference`: Generate the large reference file
    - `localrefs` : comma-separated list of rules that will be linked to the same page. These rules will be anchored as #rulename instead of grammar_diagrams#rulename as they are expected to be defined on the same page in a different diagram/grammar pair. `NOTE:` The grammar_diagrams is the reference page with all the diagrams.
- The rulenames to be processed are specified as the content of the shortcode (The text between the start and end markers) as a comma separated values, eg: 
```{{<ebnf>}}frame_clause,frame_bounds,frame_start{{</ebnf>}}```
- When the `ebnf` file of a specific version (eg `preview`, `stable`, ...) is modified, only the diagrams in that version will be regenerated
- Multiple diagrams/tabs can be added on the same page without any additional parameters. (previously they had to be suffixed with unique numbers) eg:
```
{{<ebnf>}}frame_clause,frame_bounds,frame_start{{</ebnf>}}

{{<ebnf>}}select,select_options{{</ebnf>}}
```

- The `rrdiagrams.jar` has been packaged and published to [npm repository](https://www.npmjs.com/package/yb-rrdiagram)and will be automatically downloaded and installed when Hugo is set up. 
- The diagrams will be cached internally by Hugo & will be regenerated only if the base grammar file is modified.
- Both old and new schemes can coexist. This means that the pair of grammar/diagram files generated for the previous versions (eg:`stable`, `v2.2`, ...)  can continue to be included in the `.md` files till we decide to move to the new scheme permanently.

### Developer Flow
- To add diagrams/tabs/grammar, add the rulenames with the `ebnf` shortcode (as shown below) in the `.md` files.
[...]

PR #33560, opened 2026-08-27
[DOC-1357] Docs: Reference page for YSQL configuration parameters

pgyogesh:
## Summary

YSQL exposes a few hundred YugabyteDB-specific (`yb_`) configuration parameters, but unlike yb-tserver and yb-master flags they have no systematic reference on the docs site. Of the 215 `yb_` parameters in the v2026.1 series, 127 were not mentioned anywhere under `docs/content/stable`. Support, field engineers, and customers tuning these parameters had no documented data type, default, unit, or restart requirement to work from.

This adds a reference page listing the YugabyteDB-specific YSQL configuration parameters a user would tune on a working cluster, with description, type, default, unit, and context, grouped by functional area. Parameter names link to the fuller prose entries under [YSQL configuration parameters](https://docs.yugabyte.com/stable/reference/configuration/yb-tserver/#ysql-configuration-parameters) on the YB-TServer page where one exists, so the two pages complement rather than duplicate each other.

The page is generated, not hand-maintained. `docs/scripts/gen-ysql-configuration-parameters.py` reads `pg_settings` from a running cluster of the release the docs version tracks, so the list is exact for that release. Re-run it per release and commit the result; `--docs-version` generates the page for the other docs versions. This PR covers `stable` (v2026.1.1.1) only; backporting to v2025.2, v2025.1, and v2024.2 is a separate call for the docs team.

Of the 222 parameters the release exposes, the page lists 127. It leaves off developer and test parameters (73), parameters whose own description disclaims user use (9), deprecated parameters (9), and YSQL major version upgrade plumbing (4). Those rules live in `excluded()` and match on the description text rather than on a name list, so a parameter that stops being deprecated or internal reappears on the next regeneration. The generator prints what it dropped and why.

Two source-side defects are worth fixing separately, in `src/postgres`:

- `yb_pg_metrics.log_tcmalloc_stats` has a `short_desc` copy-pasted from `yb_pg_metrics.log_accesses`. It is passed to the webserver as `enable_tcmalloc_logging`.
- Four misspellings in parameter descriptions: `testint`, `statmements`, `transation`/`atleast`, and `herejust`.

The generator carries small correction maps for both so the page does not publish known-wrong text. The corrections are display-only and cannot move a parameter between sections. Fix the strings in the source and drop the entries.

## Test plan

Docs-only change; no runtime behavior is affected.

- [x] Built the site locally with the pinned Hugo 0.154.5 (`npm run fast`) and confirmed the page renders: 10 tables, 148 rows, 127 parameters, no unrendered markdown or shortcodes.
- [x] Verified the page lists exactly the 127 parameters the generator's rules select from `pg_settings` on a v2026.1.1.1 cluster, with none missing, extra, duplicated, or leaked from the excluded groups.
- [x] Verified all 25 cross-page anchors resolve against `yb-tserver.md`, and confirmed in the built HTML that a sample link target (`#yb-bnl-batch-size`) exists on the YB-TServer page.
- [x] Confirmed the new entry appears on the Configuration index page and links correctly.
- [x] Re-ran the generator and confirmed it reproduces the committed page byte for byte.


ddhodge:
@pgyogesh Great thanks. Perhaps the parameters will need a format more similar to TServer All Flags page though, due to the width of some parameter names and the max width of the content many of the tables are too wide for convenient display.

pgyogesh:
@ddhodge Good catch, thanks. Switched the tables to per-parameter entries in the same format the flags pages use: a heading, a tags row with default, type, unit and context, and the description below it. The tags row wraps, so nothing forces a horizontal scroll now.

Two other changes that came out of it:

- Parameters that can only be set in the cluster configuration now carry the existing restart-needed badge, instead of leaving the reader to work it out from the context value.
- Parameter headings are h3 with `rightNav.hideH3`, so the right nav lists the eight areas rather than all 127 parameters.

The only table left is the five-row context legend. Rebuilt locally with Hugo 0.154.5 to confirm the entries and badges render.

jasonyb (changes requested):
Did not take a close look at the code, but the one thing we should really avoid is checking in the generated file into git.  We had much pain dealing with generated rrdiagram files being checked into git, and we later came up with a solution for that in #18567.

Regarding the little I looked at in the code, it seems this will expose several GUCs that we don't necessarily want normal users to know of.  But I'm not sure if there is an automated way to detect which GUCs are such.  Maybe the ones that are not part of the sample pg.conf?

ddhodge:
- Typically we use h5 for flags/parameters/commands headings in these reference style pages
- All TServer flags page dynamically drops the flag content into the page - I don't know how that was implemented but I wonder if we can do the same here
- As an odd sort of (perhaps flawed) compromise RE visibility of All Flags, we removed it from the left nav. Perhaps add a big warning at the top of the page?

pgyogesh:
Thanks both. I've pushed the changes that were clear-cut, and I have findings on the two open design questions that I'd like your call on before rebuilding further.

## Done in this push

@ddhodge:
- Parameter headings are now h5, matching the other reference pages.
- Added an "Advanced parameters" warning at the top.
- The TOC is configured `startLevel = 2 / endLevel = 4`, so h5 keeps the parameters out of the right nav on its own. The `rightNav` override I had added is gone; the right nav now lists just the 10 area headings.

@jasonyb: on parameter exposure, the page is now gated on an explicit allowlist (`docs/scripts/ysql-parameters-allowlist.txt`) rather than on rules alone. Nothing outside that file is published. The generator reports both directions, so a release can't change the published set silently:
- a parameter a release exposes that isn't in the allowlist, and matches no exclusion rule, is withheld and listed for a decision
- an allowlist entry the release no longer exposes is reported as renamed or removed

I verified that by simulating a release that adds one parameter and removes another. Pruning the list is now a matter of deleting lines, so it's a concrete thing to review.

## Findings on the two open questions

**1. `postgresql.conf.sample` won't work as the filter.** It lists 11 `yb_` parameters, 10 of which are already on the page. Using it as the test would drop 117 of the 127.

**2. The All Flags mechanism can't be reused as-is.** All Flags works by fetching a per-release artifact that release engineering publishes, and rendering it with `resources.GetRemote` in `all-flags-list.html`. I checked the published XML for 2026.1.1.1: it has 1183 flags but contains no GUCs. It carries 89 `ysql_yb_*` flag wrappers, which cover only 41 of the 127 parameters. `yb_fetch_row_limit`, `yb_enable_cbo` and `yb_read_from_followers` are all absent. So "do the same here" needs a new artifact published per release, which is a release-engineering change rather than a docs one.

## What I'd propose, and the question

On not checking the generated file into git: I read #18567 as establishing the pattern that the *source* stays in git (the `.bnf` grammar files) while the generated presentation does not. The docs-only equivalent here is to check in a data file (`docs/data/ysql_parameters/<version>.json`, dumped from `pg_settings`) and add a shortcode that renders the entries at build time. That removes the generated markdown from git and gives @ddhodge the dynamic drop-in, without waiting on release engineering.

The alternative is the fuller fix: get a `ysql_parameters` artifact published alongside the flags XML each release, then this page uses `resources.GetRemote` exactly like All Flags. Cleaner long-term, but it blocks this PR on another team.

Which would you prefer? I'm happy to do the data-file + shortcode version now and file a separate issue for the published artifact, but I didn't want to rebuild it twice.

One smaller question for @ddhodge: All Flags stays out of the left nav by omitting `parent:` from its menu entry. You called that a possibly flawed compromise, so I've left this page in the nav for now and added the warning instead. Say the word if you'd rather it match All Flags and come out of the nav too.

jasonyb:
Thanks for running down both options.  Disclaimer up front: I did not get time to fully think this through, so treat the following as a direction to poke holes in rather than a decision.

Stepping back from "how do the docs get a `pg_settings` dump," all three options (hand-run generator, checked-in data file, published release artifact) share the same two weaknesses, and those are what will make this page rot regardless of delivery mechanism:

1. The "is this user-facing" decision lives in `docs/`, as description regexes plus an allowlist, far from the engineer who adds the parameter.  That has to be re-curated every release.
2. The metadata is only observable from a running binary.  `guc.c` has 223 `yb_` GUCs, 128 of them `GUC_NOT_IN_SAMPLE` (applied inconsistently, which is why `postgresql.conf.sample` only has 11), only 4 in `DEVELOPER_OPTIONS`, and one YB-specific flag bit total.  Nobody reviews `short_desc` because nobody sees it, hence the copy-pasted description and the typos.

Compare gflags: `TAG_FLAG(stable|evolving|advanced|hidden|deprecated|pg)` is set at definition, reviewed in code review, dumped by `--dump_flags_xml`, and the docs just render it.  GUCs have none of that.  So the shape I would aim for:

**1. Classification in the source, enforced by a test.**  Add YB GUC flag bits next to `GUC_YB_CUSTOM_STICKY` (something like `GUC_YB_INTERNAL` and `GUC_YB_DEPRECATED`), or a stability tag mirroring the gflag vocabulary.  Your allowlist is the seed for that sweep: the curation is done, it just moves from `docs/scripts/ysql-parameters-allowlist.txt` into `guc.c` and the extension `DefineCustom*Variable` calls.  Precedent for "checked-in list, CI fails if you do not decide" is `src/yb/gflag_allowlist.txt` and `previous_auto_flags.json`.

**2. The per-release artifact is a regress expected file.**  A `yb.orig.guc_inventory.sql` regress test that emits every `yb%` row of `pg_settings` (your `QUERY` columns plus the YB flag bits) as JSON or CSV.  Its `.out` file is then:

- in git at every release tag, fetchable by URL, with no `downloads.yugabyte.com` publishing step.  I checked: GitHub has `2026.1.1.1-b2` (the exact `appVersion` in `currentVersions.json`) and `v2026.1.1.1` as tags on the same SHA, and `raw.githubusercontent.com` serves files at that tag.
- inclusive of extension GUCs, because it runs on a real cluster with the default preload list from `pg_wrapper.cc`.  That is the part `postgres --describe-config` cannot do even with its `GUC_NOT_IN_SAMPLE` filter dropped, so the "new describe-config mode plus release-eng upload" path is more work for a worse result.
- a reviewable diff in whatever source PR adds, renames, removes, or redescribes a GUC.  Typos get caught there, and the generator's "renamed or removed" reporting becomes unnecessary.
- test-only, so it backports cheaply to 2025.2, 2025.1, and 2024.2, after which every patch release on those branches has its own exact file.  That is how the older docs versions get covered without hand-generated data files.

**3. Docs as a pure projection.**  A shortcode modeled on `all-flags-list.html`: map the content folder to `appVersion` via `currentVersions.json`, `resources.GetRemote` the `.out` at the tag, filter on the flag bits, group by `category`.  For `preview` it can read the file from the local checkout since `docs/` is in the same repo.  The cross-links to `yb-tserver.md` can be computed at build time from that page's `.Fragments.Identifiers`, so anchors cannot go stale.  Longer term the shortcode can also join the flags XML (`pg`-tagged `ysql_yb_*` wrappers) to render "also settable as `--ysql_yb_enable_cbo`" on the 41 parameters that have one.

### What this means for this PR

I would not block this PR on any of the above.  The interim data-file plus shortcode version you proposed is the right first step, with one adjustment: have the shortcode take its source as "remote at the tag if available, else `docs/data`".  Then when the source-side work lands and the test is backported, the docs change is deleting data files, and nothing gets built twice.

For the source side I would file two issues: the classification sweep plus inventory test (needs a YSQL owner, not docs), and the two description fixes you already identified, which should just land now.

Caveats I have not worked through: some `boot_val`s may differ by build type and the inventory would need to normalize those or the test flaps between release and debug CI, and anyone adding a GUC now regenerates an expected file, which is the point but is a new step to socialize.

PR #33582, opened 2026-08-27
[DOC-1456] Docs: Document SSH user on Linux versions, not SSH Key Pairs

ddhodge:
## Summary

Cloud provider configuration still told readers they could leave SSH user empty on a Linux version and fall back to SSH Key Pairs. The UI requires SSH user on each Linux version and disables the provider-level SSH User and Port fields.

This updates AWS, GCP, and Azure provider pages (stable, v2025.2, v2025.1, v2024.2) so SSH user is documented on the Linux version: required, passwordless sudo, and not `yugabyte`. On AWS and GCP, stock images use the image's default login user. On Azure, Azure creates the SSH user at VM provision time; for a custom or shared-gallery image, that username must not already exist on the image. SSH Key Pairs now states those fields are disabled. GovCloud no longer recommends CentOS 7; add a supported custom Linux version instead of using YugabyteDB Anywhere-managed images there.

Prepare pages (software requirements and Azure node permissions) match that split.

## Test plan

- [x] No runtime behavior change; visually inspected the provider-page diffs against the current YBA UI (SSH user required on Add Linux Version; SSH User/Port disabled under SSH Key Pairs).

### DIFF
--- a/docs/content/stable/yugabyte-platform/configure-yugabyte-platform/aws.md
+++ b/docs/content/stable/yugabyte-platform/configure-yugabyte-platform/aws.md
@@ -148,7 +148,7 @@ To add your own machine images to the catalog:
 
 1. Enter the Amazon Machine Image (AMI) ID to use for each [provider region](#regions).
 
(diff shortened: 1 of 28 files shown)


nkhogen (approved):
LGTM

PR #33583, opened 2026-08-27
[BACKPORT 2026.1][#31938] YSQL: MAGE extension does work in backup/restore scenarios (#31981)

shubin-yb:
sql_dump in binary-upgrade mode emits

    DROP EXTENSION IF EXISTS mage;

before recreating an empty extension shell, so any backup/restore of a
database that uses the mage (Apache AGE) extension hits this statement
during restore. The mage ProcessUtility hook intercepts every DROP that
mentions "mage" via is_age_drop() and unconditionally routes it through
drop_age_extension(), which enumerates graphs from mag_catalog.ag_graph.
When the extension is not (yet) installed on the restore target, that
lookup errors out with

    ERROR:  table "ag_graph" does not exist

and the whole restore aborts, so a backup of a mage-using database was
effectively unrestorable.

Fix: guard is_age_drop() with get_extension_oid("mage", true). When the
extension is not installed, defer to standard ProcessUtility, which
correctly performs the IF EXISTS no-op (or raises the standard
"extension does not exist" error for a bare DROP EXTENSION mage).
get_extension_oid is preferred over checking the mag_catalog schema: the
schema can linger after a prior DROP EXTENSION (so a schema-based guard
still ran the AGE drop path and still hit the ag_graph lookup).

Tests:

* src/postgres/third-party-extensions/mage/regress/{sql,expected}/
yb.orig.basic.{sql,out}: three direct regress cases appended at the end
of the existing yb.orig.basic test: - DROP EXTENSION IF EXISTS mage;
when not installed -> NOTICE skip
      - DROP EXTENSION mage; when not installed           -> proper
        "extension does not exist" error, not the ag_graph one
      - CREATE EXTENSION mage; then DROP EXTENSION IF EXISTS mage;
        confirms the AGE cleanup path still runs when installed

* java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java:
testMageBackupRestore does an end-to-end YBC backup/restore round trip
on a database with mage installed Adds a class-level
ysql_yb_enable_mage=true tserver flag; harmless to the other tests in
the class, which never CREATE EXTENSION mage.

Verified locally on a 3-node mini-cluster:
  * mage regress (TestPgRegressThirdPartyExtensionsMage): 1/1 pass.
  * TestYbBackup#testMageBackupRestore under YB_TEST_YB_CONTROLLER=1:

---------

Co-authored-by: Karthik Ramanathan <32414766+karthik-ramanathan-3006@users.noreply.github.com>

Original commit: 06d9bf1de623777d6a2c31288ea85a8a8d6ee2ea / #31981

### DIFF
--- a/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
+++ b/java/yb-pgsql/src/test/java/org/yb/pgsql/TestYbBackup.java
@@ -114,6 +114,15 @@ public class TestYbBackup extends BasePgSQLTest {
   }
 
   @Override
+  protected Map<String, String> getTServerFlags() {
+    Map<String, String> flagMap = super.getTServerFlags();
+    // Gates `CREATE EXTENSION mage` -- required by testMageBackupRestore and
+    // harmless to the other tests in this class (they do not touch mage).
+    flagMap.put("ysql_yb_enable_mage", "true");
+    return flagMap;
+  }
+
+  @Override
   protected void customizeMiniClusterBuilder(MiniYBClusterBuilder builder) {
     super.customizeMiniClusterBuilder(builder);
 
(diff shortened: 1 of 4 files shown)


shubin-yb:
Trigger Jenkins

hari90 (approved):

Issue #33614, opened 2026-08-29
Agent dev tooling conflicts with the GitHub PR workflow (force-push and stacked PRs)


kai-franz:
The repo's agent tooling (`.agents/scripts/*`, `.claude/skills/create-pr`, `src/AGENTS.md`) encodes assumptions that the GitHub PR workflow has since changed. Two of them are now actively wrong.

### 1. Unconditional rebase + force-push

`git-push.sh` rebases onto `upstream/<base>` and force-pushes on every run, and `src/AGENTS.md` states "Force-pushing is the expected behavior of every push; don't try to avoid it."

Once a PR is out of draft this is harmful: the rebase renews every SHA, GitHub marks reviewers' existing line comments "outdated", and the diff-since-last-review is lost. The guidance is now to avoid rebase and force-push after a PR is ready for review.

(`.claude/skills/create-pr/SKILL.md` already says "Never force-push without explicit user permission", so the two documents contradict each other today.)

### 2. Stacked PRs are impossible with the current scripts

GitHub can only chain PRs whose head branches live in the main repo, so a stack member cannot sit in a fork. The repo's branch rulesets already accommodate this — `Block Creations`, `Require PR`, `yb_required commit status` and `Allow Stacked PRs` all carry `exclude: ["refs/heads/feature-stack/**/*"]` — and `feature-stack/...` branches already exist upstream.

The tooling blocks it in three places:

- `git-push.sh` hard-refuses any push whose target owner matches the upstream owner.
- `create-pr.sh` always builds a cross-repo head spec (`<user>:<branch>`).
- `create-pr/SKILL.md` instructs the agent never to prompt for a base branch and to always target `master`.

### Scope

Fix both in the scripts and in the agent-facing docs, so that the tooling matches the workflow instead of fighting it.

PR #33615, opened 2026-08-29
[#33614] ClaudeCode: Match dev tooling to the GitHub PR workflow

kai-franz:
## Summary

The GitHub PR workflow has two rules that our tooling did not obey. In both cases the tooling did the opposite of the rule. This change corrects the tooling.

## Rule 1. Do not force-push a pull request that is ready for review

Before this change, `git-push.sh` rebased the branch and force-pushed it every time. A rebase gives a new SHA to each commit. GitHub then marks the line comments of the reviewers as "outdated". The reviewers lose the record of their earlier work.

The script now reads the state of the pull request. Then it selects one of two modes:

- **No pull request, or a draft pull request.** No person reads the branch. The script rebases the branch and force-pushes it. This mode is the same as before.
- **A pull request that is ready for review.** The script does not rebase the branch. First, the script makes sure that the push is a fast-forward. Then the script pushes the branch.

If the push is not a fast-forward, the script stops. The exit code is 4. The message tells you to merge the remote branch. Do not rebase the remote branch. The new `-f` option gives the old behavior. Use `-f` only when the reviewer agrees.

To add newer commits from the base branch to a live pull request, merge them:

```
git merge upstream/master
```

The merge commit is safe. GitHub squash-merges each pull request. Thus the merge commit does not go to the master branch.

## Rule 2. Let a stacked pull request use the main repository

GitHub makes a stack only from branches in the main repository. A branch in your fork cannot be a part of a stack. The branch rulesets of the repository agree with this. They exclude `refs/heads/feature-stack/**/*`.

Three parts of the tooling prevented a stack:

- `git-push.sh` refused each push to the main repository.
- `create-pr.sh` always made a cross-repository head reference (`<user>:<branch>`).
- The skill told the agent to always use the master branch as the base branch.

The tooling now permits a stack:

- `git-push.sh` finds the `feature-stack/` prefix. Then it pushes the branch to the main repository. You do not give an option. All other branches keep the old refusal.
- `create-pr.sh` uses the branch name without a prefix. The `-b` option accepts the parent branch in the stack. The script first makes sure that the parent branch exists.
- The skill has a new section about a stacked pull request.

`git-push.sh` gives a warning if the branch name does not agree with the convention `feature-stack/<feature-name>/<change-name>`. The script does not stop. GitHub controls which names its rulesets permit. A refusal of a permitted name is worse than a warning.

## Changed files

| File | Change |
| --- | --- |
| `.agents/scripts/git-push.sh` | Two push modes. Support for a stack branch. New exit code 4. |
| `.agents/scripts/create-pr.sh` | Head reference and base branch for a stack. Checks of the parent branch. |
| `.agents/scripts/block-git-push.sh` | The message tells you about a stack branch. |
| `.claude/skills/create-pr/SKILL.md` | New sections. Removal of the contradiction about force-push. |
| `src/AGENTS.md` | New rules for force-push, for a merge, and for a stack. |

## Not in this change

Two related items are not in this change. The first item is a skill that shows the status of the checks of a pull request. The second item is a note in the backport command about the Phorge workflow. A comment on this pull request gives more data about both items.

## Test plan

- [x] `bash -n` on all three changed scripts.
- [x] Unit-tested the branch-name classification against 7 names (2 valid stack names, 3 malformed, 2 non-stack), confirming `feature-stackish/a/b` is not treated as a stack.
- [x] Confirmed `set -e` does not abort the script on the `[[ ... ]] && is_stack_branch=true` line when the test fails (the common non-stack path).
- [x] Checked the claims against live repo config rather than assuming: `allow_squash_merge:true` with `allow_merge_commit:false` and `allow_rebase_merge:false` (so a task-branch merge commit never reaches master), and all four branch rulesets carry `exclude: ["refs/heads/feature-stack/**/*"]`.
- [x] Opening this PR exercises the no-PR-yet path: fork detection, rebase onto `upstream/master`, lint, first push.
- [x] Draft-mode follow-up push: pushed a lint fix onto this PR while it was a draft; the script reported `PR #33615 is a draft -- rebase + force-push` and used `--force-with-lease`, confirming the mode is read from `isDraft` and not from a flag.
- [ ] Ready-for-review append-only push, and the exit-4 non-fast-forward message, once this PR leaves draft.
- [ ] End-to-end stacked PR: create `feature-stack/<feature>/<change>`, push to upstream, open a PR against a parent stack branch. **Not yet run** — it writes a real branch to the main repo, so I'd rather do it once a reviewer agrees the approach is right.


kai-franz:
Longer rationale, since most of this change is judgement calls rather than mechanics. This comes out of two review comments on the PR-workflow pilot doc: one on "avoid force-push once the PR is out of draft", one asking for the scripts and the AI skill to allow `feature-stack/` branches.

## Why the push mode is keyed on `isDraft` rather than a flag

The obvious alternative is a `--no-rebase` flag that the author remembers to pass. That fails in exactly the case it matters: the transition from draft to ready-for-review is the moment the cost of a rebase flips from zero to high, and it is also the moment nobody is thinking about push flags. Keying on the PR's own state means the same command — `git-push.sh`, no arguments — is correct at every stage, which is what makes it safe to keep in `src/AGENTS.md` as an unconditional instruction.

The cost is one extra `gh pr list` call per push. That call already existed at the bottom of the script for the summary-sync reminder; it is now hoisted and reused, so there is no new round trip.

## Why append-only mode refuses instead of merging for you

When the push is not a fast-forward, the script could silently `git merge` and continue. It doesn't, for two reasons:

1. A non-fast-forward means either the remote moved or local history was rewritten. Those want different fixes, and the script can distinguish them (`merge-base --is-ancestor` in both directions) but cannot decide between them.
2. Pulling in a moving base mid-review changes what the reviewer sees. That should be a deliberate act by the author, not a side effect of pushing a typo fix.

So exit 4 reports which of the two cases it is and names the exact command. Same reasoning for the "N commits behind master" notice — it reports, it does not act.

## The subtle one: lint revision in append-only mode

`lint.sh --rev upstream/master` was correct only because the script had *just rebased* onto that ref. Without the rebase, the same invocation lints the branch against a base that has moved on independently, dragging in unrelated files that master gained since the branch was cut. Append-only mode therefore lints against `git merge-base upstream/master HEAD`, which is the branch's own delta. This would have been a slow, confusing failure to debug in the field — noisy lint on files you never touched — so it is worth knowing it is deliberate.

## Why the branch-name check warns instead of rejecting

The rulesets exclude `refs/heads/feature-stack/**/*`. I could not determine with confidence whether GitHub's `**` admits `feature-stack/foo` (one segment) or `feature-stack/a/b/c` (three), and guessing wrong in the strict direction means the script refuses a branch name GitHub would have accepted. A false refusal is worse than a false pass here: the false pass gets a clear error from GitHub with our warning already on screen explaining it, while the false refusal blocks a legitimate workflow with no recourse. I first wrote it as a hard reject on exactly-two-segments and backed it out for this reason.

## What I checked against live config rather than assumed

- `allow_squash_merge: true`, `allow_merge_commit: false`, `allow_rebase_merge: false`. This is what makes "merge instead of rebase" safe advice — the merge commit on a task branch never reaches master. If someone later enables merge commits, the guidance in `src/AGENTS.md` needs revisiting.
- All four branch rulesets (`Block Creations`, `Require PR`, `yb_required commit status`, `Allow Stacked PRs`) carry `exclude: ["refs/heads/feature-stack/**/*"]`, and `feature-stack/...` branches already exist upstream. The prefix really is the only sanctioned route to an upstream push, which is why `git-push.sh` keys the exception on it and nothing else.

## A separate finding worth fixing

**`src/AGENTS.md` is not in `bld-db.yml`'s `paths-ignore`.** `.agents/**` and `.claude/**` are, and `bld-none.yml` picks those up, but a markdown file under `src/` falls through to a full DB build. This PR is a live example: it is docs-and-scripts only, and marking it ready for review will trigger `bld-db`. Adding `src/AGENTS.md` (or `src/**/AGENTS.md`) to the `paths-ignore` list in `bld-db.yml` and to `paths` in `bld-none.yml` would fix it. I left it out of this PR to keep the change reviewable — happy to add it here if reviewers prefer.

## Not in this change

- **A `pr-status` skill** — report `yb_required`, which `bld-*` workflows ran, and pull failing-test logs from csiweb for a given PR. This is the "is my red a known flake" loop that Harbormaster covers on the Phorge side and that has no PR-side equivalent. It is the largest remaining ergonomic gap and wants its own issue.
- **A pointer on `/backport-diff`** noting it is the Phorge path and that `/backport-commit` is the PR-pilot equivalent. One line, prevents wrong-tool starts during the transition.

## For reviewers

One rollout question I did not want to decide alone: the append-only behavior is **unconditional**, so it changes push behavior for everyone using these scripts, pilot member or not. I think that is right — it is strictly better for anyone with a live PR regardless of pilot membership, and gating it on team membership would mean the scripts behave differently for different people, which is worse to reason about. But it is a behavior change to a shared script and reviewers may disagree.

Also note this PR resolves a pre-existing contradiction: `src/AGENTS.md` said force-pushing was expected on every push, while `create-pr/SKILL.md` said never force-push without user permission. Both now describe the same rule.

---

Issue #33638, opened 2026-08-31
LogRollingTest.Rolling fails on fast hosts


ellabaron-code:
Jira Link: [DB-23283](https://yugabyte.atlassian.net/browse/DB-23283)
## Description

`LogRollingTest.Rolling` rolls the master's glog file with `--max_log_size=1` and asserts that the fresh log header's `Running duration (h:mm:ss)` line differs from the initial header's (`log_version-test.cc:117`). The duration has one-second granularity, so on a fast host the roll completes within the master's first second of uptime, both headers read `0:00:00`, and the assert fails deterministically.

Fix (PR incoming): sleep just over one second after reading the initial header, so the fresh header's duration is guaranteed to differ.

## Issue Type

kind/failing-test

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23283]: https://yugabyte.atlassian.net/browse/DB-23283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33639, opened 2026-08-31
[#33638] DocDB: Fix LogRollingTest.Rolling on fast hosts

ellabaron-code:
## Summary

`LogRollingTest.Rolling` asserts the fresh log header's `Running duration (h:mm:ss)` line differs from the initial header's after forcing a glog roll with `--max_log_size=1`. The duration has one-second granularity, so on a fast host the roll completes within the master's first second of uptime, both headers read `0:00:00`, and the assert fails deterministically.

Sleep just over one second after reading the initial header, guaranteeing the fresh header's duration differs regardless of machine speed.

## Test plan

- [x] `LogRollingTest.Rolling` passes with the fix on a 192-core host where it previously failed deterministically (in both parallel and serial runs).

### DIFF
--- a/src/yb/integration-tests/log_version-test.cc
+++ b/src/yb/integration-tests/log_version-test.cc
@@ -91,6 +91,8 @@ TEST_F(LogRollingTest, Rolling) {
   ASSERT_NE(header.GetByPrefix(fingerprint), "");
   const auto& initial_duration = header.GetByPrefix(kDurationPrefix);
   ASSERT_NE(initial_duration, "");
+  // Header duration has 1s granularity; ensure a second elapses so the fresh header differs.
+  SleepFor(MonoDelta::FromMilliseconds(1100));
   // In case of log rolling log_path link will be pointed to newly created file
   const auto initial_target = ASSERT_RESULT(env_->ReadLink(log_path));
   auto prev_size = ASSERT_RESULT(env_->GetFileSize(log_path));


src/yb/integration-tests/log_version-test.cc
```diff
@@ -91,6 +91,8 @@ TEST_F(LogRollingTest, Rolling) {
   ASSERT_NE(header.GetByPrefix(fingerprint), "");
   const auto& initial_duration = header.GetByPrefix(kDurationPrefix);
   ASSERT_NE(initial_duration, "");
+  // Header duration has 1s granularity; ensure a second elapses so the fresh header differs.
+  SleepFor(MonoDelta::FromMilliseconds(1100));
```
d-uspenskiy: Unit test doesn't look valid (stable) and it require deeper redesing rather than adding sleep.

For example

```
  while(initial_target == ASSERT_RESULT(env_->ReadLink(log_path))) {
    ...
    const auto current_size = ASSERT_RESULT(env_->GetFileSize(log_path));
    // Make sure log size is changed and it is not much than 2Mb
    // Something goes wrong in other case
    ASSERT_NE(current_size, prev_size);
    ASSERT_LT(current_size, 2_MB);
    prev_size = current_size;
  }
```

`ReadLink` and `GetFileSize` might reference to different files because of log rolling at the middle. As a result there is no reason to assume that file size will be changed (it could be the same for 2 different files). 

Valid (stable) unit test could be something like this:
- generate logs till we get 5 different results from env_->ReadLink(log_path), store each file name into vector
- iterate vector and check that each file size less than 2MB and all the files has same fingerprint
- perform stable sort based on duration value from each file and check that position doesn't changed

d-uspenskiy (changes requested):

src/yb/integration-tests/log_version-test.cc
ellabaron-code: Reworked the test along the lines you suggested — thanks, much cleaner than the sleep. The 192-core host where the original failure reproduced is gone; it passes on a 176-core (but slower) machine and in release and ASAN on Linux. Since the rewrite no longer asserts the durations differ, host speed shouldn't matter anymore anyway.

src/yb/integration-tests/log_version-test.cc
```diff
@@ -85,36 +85,44 @@ TEST_F(LogRollingTest, Rolling) {
...
   const auto fingerprint = "Application fingerprint: " + version;
-  const LogHeader header(log_path);
-  ASSERT_NE(header.GetByPrefix(fingerprint), "");
-  const auto& initial_duration = header.GetByPrefix(kDurationPrefix);
-  ASSERT_NE(initial_duration, "");
-  // In case of log rolling log_path link will be pointed to newly created file
-  const auto initial_target = ASSERT_RESULT(env_->ReadLink(log_path));
-  auto prev_size = ASSERT_RESULT(env_->GetFileSize(log_path));
+  // In case of log rolling log_path link will be pointed to newly created file.
+  // Generate logs until the link has pointed to 5 distinct files.
```
d-uspenskiy, line 92: Too many implementation details in comment. What if 5 in code will be changed to 10 or 6. Comment will be outdated.

In case you think comment is necessary here it should be something like: "Collect several files during log rolling for further checks"

src/yb/integration-tests/log_version-test.cc
```diff
@@ -85,36 +85,44 @@ TEST_F(LogRollingTest, Rolling) {
...
+  std::vector<string> log_files = {ASSERT_RESULT(env_->ReadLink(log_path))};
```
d-uspenskiy, line 93: 1. It is reasonable to use `reserve()` or `boost::container::small_vector` instead of `std::vector`
2. It is better to keep container empty. This will allow to keep the code which fills it in single place (no copy-paste)
```
boost::container::small_vector<std::string, 5> log_files;
...
while(log_file.size() < 5) {
  ...   
  auto target = ASSERT_RESULT(env_->ReadLink(log_path));
  if (log_files.empty() || target != log_files.back()) {
    log_files.push_back(std::move(target));
  }  
}
```

src/yb/integration-tests/log_version-test.cc
```diff
@@ -85,36 +85,44 @@ TEST_F(LogRollingTest, Rolling) {
...
-    prev_size = current_size;
+    auto target = ASSERT_RESULT(env_->ReadLink(log_path));
+    if (target != log_files.back()) {
+      log_files.push_back(std::move(target));
+    }
+  }
+  // Rolled files are immutable, so checking them by name is race-free.
+  size_t prev_duration_sec = 0;
+  for (const auto& file : log_files) {
+    // glog symlink targets are relative to the log directory.
```
d-uspenskiy, line 111: Redundant comment. It is better to use named variable instaed

```
const auto log_file_full_path = JoinPathSegments(logs_dir, file);
```

src/yb/integration-tests/log_version-test.cc
```diff
@@ -85,36 +85,44 @@ TEST_F(LogRollingTest, Rolling) {
...
+    const LogHeader header(JoinPathSegments(logs_dir, file));
+    ASSERT_NE(header.GetByPrefix(fingerprint), "");
+    const auto& duration_line = header.GetByPrefix(kDurationPrefix);
+    ASSERT_NE(duration_line, "");
+    ASSERT_LT(ASSERT_RESULT(env_->GetFileSize(JoinPathSegments(logs_dir, file))), 2_MB);
+    const auto duration_str = duration_line.substr(kDurationPrefix.size());
+    std::smatch match;
```
d-uspenskiy, line 118: The stuff with converting string into seconds is redundant. Duration has the format with leading zeros `0:00:00` so it can be compared lexicographical for almost 10 hours. 

```
const auto duration_str = duration_line.substr(kDurationPrefix.size());
ASSERT_GE(duration_str, prev_duration_str);
prev_duration_str = duration_str;
```

d-uspenskiy (changes requested):

src/yb/integration-tests/log_version-test.cc
ellabaron-code: Fixed.

src/yb/integration-tests/log_version-test.cc
ellabaron-code: Fixed.

src/yb/integration-tests/log_version-test.cc
ellabaron-code: Fixed.

src/yb/integration-tests/log_version-test.cc
ellabaron-code: Fixed.

d-uspenskiy (approved):

PR #33673, opened 2026-09-01
[BACKPORT 2025.2][#33201] DocDB: Drop pending status moves for participants with no completed batches (#33202)

khosrow:
## Summary

`StartPromotionToGlobal` registers every tablet in `tablets_` for a
status-location move.
`DoSendUpdateTransactionPromotingRpcs` then skips any tablet with no
completed batches, but leaves
its entry in `transaction_status_move_handles_`. That entry is only ever
erased by
`UpdateTransactionPromotingDone`, which runs when the rpc completes — so
for a tablet whose rpc is
never sent, it stays forever, and `Commit()` parks on `commit_waiter_`
until the pggate rpc deadline
expires (600 s by default).

A write that fails inside a savepoint reaches commit in exactly this
state: the tablet is registered
in `tablets_` with `num_completed_batches == 0`, and
`CanAbortTransaction` leaves the transaction
running because the failure carries a non-minimum `subtransaction_id`.

`Commit()` now prunes those tablets from
`transaction_status_move_tablets_` and
`transaction_status_move_handles_` before deciding whether to wait.

Pruning rather than sending the rpc anyway:
`ApplyUpdateTransactionPromoting` returns an expired
status for a transaction the participant has not registered, which would
surface as a spurious
abort. The participant is not part of the commit either — `DoCommit()`
drops tablets without
metadata, which is recorded in the same branch that increments
`num_completed_batches`. This is safe
at commit time only, since `CheckCouldCommitUnlocked` rejects a commit
with running requests, so no
further batch can complete.

The new test uses a fixture that lowers
`ysql_client_read_write_timeout_ms`, so that on a regression
the commit fails with an rpc timeout inside the test rather than being
killed by the harness timeout
with no assertion output.

Original commit: 3a325999599f56d78c09117ac08471822aa4d247 / #33202

## Test plan

- [x] New test
`GeoTransactionsPromotionTest.TestCommitWithZeroBatchParticipant` in
`src/yb/yql/pgwrapper/geo_transactions_promotion-test.cc` — verified it
hangs on the unmodified
  tree and passes with the fix:
  ```
  ybd --cxx-test pgwrapper_geo_transactions_promotion-test \
--gtest_filter
GeoTransactionsPromotionTest.TestCommitWithZeroBatchParticipant
  ```
- [ ] Full `pgwrapper_geo_transactions_promotion-test` binary (CI)

Assisted-By: devx/86cd09ab-76cd-4714-97ce-fb4896a579e7

## Merge conflicts

- `src/yb/client/transaction.cc:2298` — the new `DropPendingStatusMovesWithoutMetadata()` method is inserted above a function that master renamed (`DoSendUpdateTransactionStatusLocationRpcs` → `DoSendUpdateTransactionPromotingRpcs`); kept this branch's function name and adapted the new method's doc comment and VLOG message from "UpdateTransaction(PROMOTING) rpc" / `DoSendUpdateTransactionPromotingRpcs()` to this branch's "UpdateTransactionStatusLocation rpc" / `DoSendUpdateTransactionStatusLocationRpcs()` naming. No logic changes vs. the original commit.

### DIFF
--- a/src/yb/client/transaction.cc
+++ b/src/yb/client/transaction.cc
@@ -661,6 +661,7 @@ class YBTransaction::Impl final : public internal::TxnBatcherIf {
         RequestStatusTablet(deadline);
         return;
       }
+      DropPendingStatusMovesWithoutMetadata();
       if (!transaction_status_move_handles_.empty()) {
         DCHECK(!commit_waiter_);
         VLOG_WITH_PREFIX(1) << "Waiting for transaction move RPCs to finish";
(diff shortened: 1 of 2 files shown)


khosrow:
Trigger Jenkins

es1024 (approved):

Issue #33684, opened 2026-09-01
[DocDB] Add a per-tablet metric for overdue WAL fsync


hkimura:
Jira Link: [DB-23321](https://yugabyte.atlassian.net/browse/DB-23321)
### Description

With `durable_wal_write=false` (the default), WAL entries are fsynced in the background, and
`interval_durable_wal_write_ms` (default 1000) is meant to bound how long an acknowledged entry
can stay unsynced. #25239 showed that bound was not enforced on an idle tablet; c25b05ab5ee
fixes it for followers by re-checking the interval on every incoming `UpdateConsensus`.

There is no metric that shows whether a tablet is currently meeting that bound. The existing
log metrics (`log_sync_latency`, `log_wal_size`, `log_bytes_logged`, ...) report work done, not
work that is overdue. This makes the fix hard to verify in a real cluster and impossible to alert
on.

Add a tablet-level gauge, `log_wal_sync_overdue_ms`: the number of milliseconds by which the
oldest unsynced WAL entry has exceeded `interval_durable_wal_write_ms`. It is 0 when there is
nothing unsynced, when the oldest unsynced entry is still within the interval, when the interval
is disabled, or when `durable_wal_write=true`. On a healthy cluster it should stay near 0 on
every tablet; a sustained non-zero value means the tablet's WAL is not being fsynced when it
should be.

The `Log` already tracks the inputs (`periodic_sync_needed_` and
`periodic_sync_earliest_unsync_entry_time_`, `src/yb/consensus/log.h:716-719`); the metric only
needs to expose them. It should be computed at read time, since nothing runs on an idle tablet
to update it, and aggregated with max across tablets so table- and server-level values report
the worst tablet.

The immediate use is a stress test for the #25239 fix before it is backported, so the metric
should be backported to the same release branches as c25b05ab5ee.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23321]: https://yugabyte.atlassian.net/browse/DB-23321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

hkimura:
Backport target 26.1 and 25.2.
24.2 might benefit from it, but I see some potential conflict, so not worth it.

Issue #12570, opened 2022-05-19
[DocDB] Avoid FDataSync calls on the critical write path.


rthallamko3:
Jira Link: [[DB-433]](https://yugabyte.atlassian.net/browse/DB-433)
### Description

Avoid FDataSync calls on the critical write path. There are multiple facets to this.

1. The fdatasync system call is done for log appends at a regular interval (combination of log generation threshold/time threshold). Currently time/size based fdatasync call in [1] is done by the thread that is generating the WAL (in the critical path). Instead, those fdatasync should be deferred to background task/thread.

2. Observe the pre-allocation aspects of the new WAL files (currently async), and ensure that pre-allocation works as expected, for a workload that exhausts 80%+ CPU on 8/16 core VCPUs, there should be 0 allocations of new log files in the critical write path. As an enhanced optimization, consider the following feedback from Sergei - _Also, I think we could do it in some pool basis. I.e. preallocate file in some special dir in the same drive. And share this preallocation for all tablets at this drive. Then just move it to log dir of appropriate tablet._

[1]
Status Log::Sync() {
…
…
       if (durable_wal_write_ || timed_or_data_limit_sync ) {
      periodic_sync_needed_.store(false);
      periodic_sync_unsynced_bytes_ = 0;
      LOG_SLOW_EXECUTION(WARNING, 50, "Fsync log took a long time") {
        RETURN_NOT_OK(active_segment_->Sync());
      }
    }
}

Relevant slack thread: https://yugabyte.slack.com/archives/CPAGT5Z4Y/p1651112089659199

[DB-433]: https://yugabyte.atlassian.net/browse/DB-433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33690, opened 2026-09-01
[#33684] DocDB: Add log_wal_sync_overdue_ms, a per-tablet gauge for overdue WAL fsync

hkimura:
Summary:
Add a tablet-level gauge, log_wal_sync_overdue_ms: the number of milliseconds by which the
oldest unsynced WAL entry has exceeded interval_durable_wal_write_ms. It is 0 when nothing is
unsynced, when the oldest unsynced entry is still within the interval, and when the interval does
not apply (durable_wal_write, or the interval disabled).

The value is computed at read time from the Log's existing periodic-sync state, since nothing
runs on an idle tablet to keep a stored value current. It is aggregated with max across tablets,
like follower_lag_ms, so table- and server-level values report the worst tablet.

Test Plan:
```
LogTest.TestWalSyncOverdueMetric
LogTest.TestWalSyncOverdueMetricUnderDurableWalWrite
LogTest.TestWalSyncOverdueMetricWhenIntervalDisabled
```

### DIFF
--- a/src/yb/consensus/log.cc
+++ b/src/yb/consensus/log.cc
@@ -95,6 +95,8 @@ using namespace yb::size_literals;  // NOLINT.
 using namespace std::literals;  // NOLINT.
 using namespace std::placeholders;
 
+METRIC_DECLARE_gauge_int64(log_wal_sync_overdue_ms);
+
 // Log retention configuration.
 // -----------------------------
 DEFINE_RUNTIME_int32(log_min_segments_to_retain, 2,
@@ -739,6 +741,13 @@ Log::Log(
   set_wal_retention_secs(options_.retention_secs);
   if (table_metric_entity_ && tablet_metric_entity_) {
     metrics_.reset(new LogMetrics(table_metric_entity_, tablet_metric_entity_));
+    // The tablet entity outlives any one Log and is reused when the tablet is opened again on this
+    // server. InstantiateFunctionGauge() would then hand back the previous Log's gauge, already
+    // detached to a constant, so drop it first.
+    tablet_metric_entity_->RemoveFromMetricMap(&METRIC_log_wal_sync_overdue_ms);
+    metrics_->wal_sync_overdue_ms = METRIC_log_wal_sync_overdue_ms.InstantiateFunctionGauge(
+        tablet_metric_entity_, Bind(&Log::WalSyncOverdueMs, Unretained(this)));
+    metrics_->wal_sync_overdue_ms->AutoDetach(&metric_detacher_, 0);
   }
   if (background_synchronizer_wait_state_) {
     background_synchronizer_wait_state_->set_root_request_id(yb::Uuid::Generate());
@@ -749,6 +758,25 @@ Log::Log(
     SET_WAIT_STATUS_TO(background_synchronizer_wait_state_, Idle);
     yb::ash::RaftLogWaitStatesTracker().Track(background_synchronizer_wait_state_);
   }
+}
+
+int64_t Log::WalSyncOverdueMs() const {
+  if (durable_wal_write_ || !interval_durable_wal_write_ || !periodic_sync_needed_) {
+    return 0;
+  }
+  // The appender publishes periodic_sync_needed_, then the timestamp, then the byte count, in that
+  // order, and DoSync() resets needed then bytes. Reading the byte count before the timestamp
+  // means a non-zero count came after the timestamp of the same batch, so a scrape that lands
+  // between the appender's first two stores reads 0 instead of pairing a fresh "needed" with the
+  // previous batch's timestamp. The remaining window is a background DoSync() whose two resets
+  // become visible out of step, which is store-buffer sized; a value read there is at most one
+  // previous cycle old.
+  if (periodic_sync_unsynced_bytes_ == 0) {
+    return 0;
+  }
+  const MonoTime earliest_unsync_entry_time = periodic_sync_earliest_unsync_entry_time_;
+  const auto overdue = MonoTime::Now() - earliest_unsync_entry_time - interval_durable_wal_write_;
+  return std::max<int64_t>(overdue.ToMilliseconds(), 0);
 }
 
 Status Log::Init() {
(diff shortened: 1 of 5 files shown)


hkimura:
trigger jenkins

balajisubramanian-yugabyte (approved):
LGTM overall

src/yb/consensus/log.cc
```diff
@@ -751,6 +760,25 @@ Log::Log(
...
+    return 0;
+  }
+  // The appender publishes periodic_sync_needed_, then the timestamp, then the byte count, in that
+  // order, and DoSync() resets needed then bytes. Reading the byte count before the timestamp
+  // means a non-zero count came after the timestamp of the same batch, so a scrape that lands
+  // between the appender's first two stores reads 0 instead of pairing a fresh "needed" with the
+  // previous batch's timestamp. The remaining window is a background DoSync() whose two resets
+  // become visible out of step, which is store-buffer sized; a value read there is at most one
+  // previous cycle old.
+  if (periodic_sync_unsynced_bytes_ == 0) {
```
ttyusupov, line 774: If `Log::DoSync()` would have failure in underlying fsync (`active_segment_->Sync()`), `periodic_sync_unsynced_bytes_` would be `0` and `WalSyncOverdueMs() ` would return `0` although WAL is not synced. This contradicts metric description "0 when every entry is fsynced".

src/yb/consensus/log_metrics.cc
```diff
@@ -42,6 +42,15 @@ METRIC_DEFINE_gauge_uint64(tablet, log_wal_size, "Size of WAL Files",
                            yb::MetricUnit::kBytes,
                            "Size of wal files");
 
+METRIC_DEFINE_gauge_int64(tablet, log_wal_sync_overdue_ms, "WAL Sync Overdue",
```
ttyusupov, line 45: Would YBA / Prometheus scrape this metric? At which level? Is this intended?

src/yb/consensus/log.cc
```diff
@@ -739,6 +741,13 @@ Log::Log(
   set_wal_retention_secs(options_.retention_secs);
   if (table_metric_entity_ && tablet_metric_entity_) {
     metrics_.reset(new LogMetrics(table_metric_entity_, tablet_metric_entity_));
+    // The tablet entity outlives any one Log and is reused when the tablet is opened again on this
+    // server. InstantiateFunctionGauge() would then hand back the previous Log's gauge, already
+    // detached to a constant, so drop it first.
```
ttyusupov, line 746: Since you've paid attention to potential issue with `InstantiateFunctionGauge()` usage, maybe we already have such issues with other metrics, for example snapshot_oldest_pending_deletion_age_ms? If yes, worth filing them.

src/yb/consensus/log.cc
hkimura: Right, and the cause is one level down. `DoSync()` clears `periodic_sync_needed_` and
`periodic_sync_unsynced_bytes_` before it calls active_segment_->Sync(), and only the background
caller (`DoSyncAndResetTaskInQueue`) re-arms them on failure. On the inline paths (`Sync()` under
kForceFsync, RollOver, Close) a failed fsync returns with the state cleared, so `FindSyncType()`
returns kNoSync and the sync policy itself forgets the pending entries until the next append
re-arms it. This gauge reads that same state, so it inherits the gap. The behaviour predates this
PR (it is from the background-sync change, #12570), and `DoSync()`'s own comment already says the
state should be re-armed on failure.

The window where the process survives a failed fsync is narrow (EIO is fatal via suicide_on_eio,
and a failed fsync on a group with a Raft batch hits the CHECK_OK in `LocalPeerAppendFinished`),
but marker-only groups and the synchronous Append path do survive it, so I filed #33816 for the
sync-policy side: re-arm inside `DoSync()` so every caller retries. Once that is in, the gauge
reports the age of the oldest entry whose fsync failed (the earliest-entry timestamp is not reset
by `DoSync()`).

src/yb/consensus/log_metrics.cc
hkimura: Yes, intended, and it is collected with no YBA change: the tserver has no allowlist of metric
names and YBA's Prometheus job has no drop rules, so it is stored as soon as a tserver runs a
build with the gauge. Under YBA's default NORMAL collection level it lands as one server-level
series per tserver (max across that tserver's tablets), which is the shape to alert on; ALL makes
it table-level, MINIMAL does not collect it.

Exposing it in YBA (panel, alert template) is a separate PLAT change on the YBA release train.
That is how every recent DB metric has gone in: none of the `METRIC_DEFINE` additions I checked
since 2023 touched managed/ in the same commit, and the YBA side followed weeks to months later
when someone wanted a panel or alert.

src/yb/consensus/log.cc
hkimura: Went through all 12 InstantiateFunctionGauge() call sites on master. The server-level and
drive-level gauges (hybrid_clock, logical_clock, untracked_memory, tcmalloc, drive_io_stats,
spinlock_profiling, thread.cc, object_lock_shared_state_manager) are bound to owners that live as
long as the process, several with NeverRetire, so they are fine.
`snapshot_oldest_pending_deletion_age_ms` is the one other tablet-entity function gauge and it has
the same issue: a Tablet re-created with the same id on the same tserver (remote bootstrap
replacing a stale or tombstoned peer) gets the previous Tablet's gauge back from
`FindOrCreateMetric`, already detached to 0, and reports 0 for the life of the new tablet. The
entity outlives the Tablet by at least metrics_retirement_age_ms, so the window is real.

Filed #33817 with the one-line `RemoveFromMetricMap` fix as the interim and the generic option (have
`InstantiateFunctionGauge()` replace the map entry, since a function gauge is tied to one owner)
noted there.

src/yb/consensus/log.cc
ttyusupov: 1. That limitation raised as a result of review should have been reflected in metric description to make it correct.
2. > The window where the process survives a failed fsync is narrow (EIO is fatal via suicide_on_eio,
and a failed fsync on a group with a Raft batch hits the CHECK_OK in LocalPeerAppendFinished),
but marker-only groups and the synchronous Append path do survive it.

What if the fsync that fails with a non-EIO errno (say ENOSPC or EDQUOT) is the one `Sync()`
submitted at 600 ms under `kAsyncFsync`, rather than the inline `kForceFsync` at 1000 ms? Which callback sees that error?

3. Let's keep #33816 short, mentioned only minimal set of properly verified facts. We should discuss and approve the proposal appropriately before mentioning it in the ticket. In its current form #33816 seems to contain not accurate conclusions, please see comments there.

Issue #33267, opened 2026-08-12
[DocDB] Observability for accumulated deleted data and automatic compaction decisions


hkimura:
Jira Link: [DB-22955](https://yugabyte.atlassian.net/browse/DB-22955)
### Description

Tables used as work queues (insert, process, delete, continuously, at high
volume) accumulate deleted rows whose space has not yet been reclaimed. Queries
scanning those key ranges must step over that data until a full compaction
runs. The accumulation stays invisible until queries are already spending most
of their time on it, and because the effect is multiplicative rather than
gradual, it presents as a sudden latency event on a cluster that looked healthy
the day before.

The only signal for this condition today is the ratio of
`docdb_obsolete_keys_found_past_cutoff` to `docdb_keys_found`. It has three
limitations:

1. **It is derived from reads**, so it only exists after queries have already
   paid the cost. A table that is rarely read accumulates with no signal at all
   until the first slow scan discovers it.
2. **It is only produced by the replica that serves the read.** Other replicas
   can be carrying the same accumulation with nothing observing it, so a leader
   change (any restart or maintenance operation) can surface the problem
   immediately on a replica that was never compacted. Related: #17227.
3. **Neither counter is surfaced by default**, and the automatic compaction
   feature's own decisions are not observable either. The per-tablet
   percent-obsolete and window totals it computes on every check interval are
   written only to the INFO log, and only when it decides to fire. Separately,
   full-compaction task metrics place statistics-triggered, scheduled, and
   admin-requested compactions in the same bucket, so metrics cannot tell which
   mechanism ran.

The practical consequence is that an operator cannot answer "is this table
accumulating deleted data", "on which nodes", or "when will automatic
compaction act on it" without collecting raw diagnostics and analysing them by
hand. That analysis is what a recent production investigation required, and the
information needed to answer it existed all along.

### Desired outcome

- A per-table measure of how much deleted-but-unreclaimed data a table is
  carrying, available on **every replica**, that does not depend on a read
  having discovered it.
- Exposed at table and server level, the way tablet metrics already aggregate.
  Per-tablet series are deliberately excluded: the volume is impractical to
  collect at production scale.
- The automatic compaction trigger's decision inputs available as metrics
  rather than log lines: what it measured, what it compared against, and when a
  tablet next becomes eligible.
- Full-compaction activity distinguishable by what triggered it.

Acceptance: from metrics alone, an operator can identify which tables are
accumulating and roughly when automatic compaction will act on them, on any
[...]

PR #33707, opened 2026-09-02
[#33267] DocDB: SST statistics collector -- properties block and registration

hkimura:
Summary:
The RocksDB-facing half of the SST statistics collector, on top of the
chain tracker from the previous change in this stack: property keys (all
under yb.docdb.*), SstStats <-> user_collected_properties serialization
(sparse histograms with a scale tag; 8 comma-separated age-band counts;
capped per-coprefix subtotals), the rocksdb::TablePropertiesCollector
wrapper, and per-tablet registration on the regular RocksDB.

The collector runs while an SST is built (flush or compaction, on every
replica) and stores what the tracker saw in the file's own properties
block. This is the storage-side signal the read-driven full-compaction
trigger lacks: followers and point-get workloads never populate the read
statistics (#33267), so garbage debt goes unmeasured until reads suffer.

NeedCompact() keeps the v1 semantics: tombstone-heavy files get
FileMetaData::marked_for_compaction (MANIFEST breadcrumb only; nothing
consumes it under universal compaction).

Registration is behind docdb_enable_sst_stats_collector, DEFAULT OFF.
Flipping the default is a separate change gated on performance results.

Not in this change: the per-tablet aggregate, the Prometheus gauges and
the full-compaction trigger clause that consume these properties. They
follow separately.

Replaces the Phorge draft D56928.

Test Plan:
./yb_build.sh release --cxx-test sst_stats_collector-test
(properties round trip; the collector end to end through the RocksDB
interface including NeedCompact)

### DIFF
--- a/src/yb/docdb/CMakeLists.txt
+++ b/src/yb/docdb/CMakeLists.txt
@@ -118,6 +118,7 @@ set(DOCDB_DEPS
         PUBLIC vector_index
         PUBLIC yb_ash
         PUBLIC yb_common
+        PUBLIC yb_docdb_properties_collector
         PUBLIC yb_docdb_shared
         PUBLIC yb_dockv
         PUBLIC yb_pggate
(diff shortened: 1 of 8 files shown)


hkimura:
Superseded by #33710: the change was split, with the chain tracker now in #33709 and the collector on branch 04-collector. This PR auto-closed when its head branch was renamed during the restack.

PR #33713, opened 2026-09-02
[BACKPORT 2026.1][PLAT-22125] YBA: Validate YSQL/YCQL auth and password consistency in the operator (#33687)

shubin-yb:
## Summary

The Kubernetes operator ignored the `enableYSQLAuth` and
`enableYCQLAuth` spec fields
entirely and inferred auth solely from whether a password secret was
referenced. Two
misconfigurations were accepted silently and produced a universe that
did not match the CR:

- `enableYSQLAuth: true` with no `ysqlPassword` created a universe with
auth **disabled**.
- A `ysqlPassword` with the flag left `false` created a universe with
auth **enabled**.

This makes the spec flags authoritative and validates the pair when a
universe is created:
auth cannot be enabled without a password, and a password is rejected
without auth. Both
directions are enforced for YSQL and YCQL, and the failure surfaces on
the CR as
`ERROR_CREATING` before any universe is created.

### Why the reconciler and not the CRD

The obvious home for this is `x-kubernetes-validations` on
`universeCrd.yaml`, but a
CRD-level rule would permanently break the operator's universe-import
path. A universe
imported from YBA via `OperatorUtils.createUniverseCr` has auth enabled
with no secret to
reference — the import writes no Secret at all — so the CR it writes
would be rejected by its
own schema and the import would fail with no way to satisfy it.

Enforcement therefore lives in `createUserIntent`, gated on the existing
`isCreate` flag.
Imported CRs resolve to an existing universe and take the edit path, so
they never reach it.
The cost is that the error arrives as CR status rather than a `kubectl
apply` rejection.

### Edit path

No password is read out of the secret on edit. Dropping the password
from the spec, or
pointing it at a secret that is empty or deleted, is now a no-op instead
of aborting the
reconcile of everything else in the spec — previously a deleted secret
threw an NPE while
being tracked as a dependency, and a secret without the expected key
threw outright, so an
unrelated edit (node count, gflags, software version) could not proceed
until the secret was
restored.

The secrets are still registered as dependencies on every edit pass.
That registration is
what `OperatorResourceRestorer` restores from on HA failover and what
marks the Secret
orphaned when the CR is deleted, and for universes whose CRs predate the
`operator_resource`
table it is the only pass that ever writes those rows.

Changing a password through the operator is not supported here and is
tracked as PLAT-22298.

Also hardens `parseSecretForKey`, which dereferenced whichever of `data`
or `stringData` was
absent.

Original commit: c01ae6f32edeab6b1743cbe7902ab114c259b74e / #33687

## Test plan

- [x] `sbt "testOnly
com.yugabyte.yw.common.operator.YBUniverseReconcilerTest"` — 73 passed,
0 failed. 10 new cases, symmetric across YSQL and YCQL: auth+password
from the spec, neither set, all four validation failures, a secret
holding no password, a missing secret, and two edit-path cases (password
removed from the spec; secrets deleted/emptied).
- [x] End-to-end on a Kubernetes cluster with
`kubernetesOperatorEnabled`, using a locally built YBA:
- [x] All five invalid CRs were accepted by the API server and rejected
by the reconciler with the expected per-API message, each CR landing in
`Error Creating`, and the YBA universe list stayed empty — validation
fails before any universe or task is created.
- [x] A single-node RF1 universe with auth enabled on both APIs and both
password secrets present created successfully; connecting with the
secret's password succeeded and a wrong password was rejected,
confirming auth was actually enabled.
- [x] Removing `ysqlPassword`/`ycqlPassword` from that live CR was a
no-op: universe stayed `Ready`, both auth flags stayed enabled in YBA,
both passwords still authenticated, and no password-path errors appeared
in the operator log.
- [x] Deleting both referenced Secret objects outright was also a no-op,
with no NPE — the case that previously aborted the reconcile.

## Merge conflicts

- `managed/src/test/java/com/yugabyte/yw/common/operator/YBUniverseReconcilerTest.java:105` — import block only: 2026.1 has already dropped `java.util.Date`, so kept the branch's list and added the cherry-pick's `java.util.Collections`.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/common/operator/YBUniverseReconciler.java
+++ b/managed/src/main/java/com/yugabyte/yw/common/operator/YBUniverseReconciler.java
@@ -76,7 +76,6 @@ import io.yugabyte.operator.v1alpha1.ybuniversespec.YcqlPassword;
 import io.yugabyte.operator.v1alpha1.ybuniversespec.YsqlPassword;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Base64;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -1280,7 +1279,8 @@ public class YBUniverseReconciler extends AbstractReconciler<YBUniverse> {
     return createTaskParams(ybUniverse, customerUUID, provider);
   }
 
-  private UserIntent createUserIntent(
+  @VisibleForTesting
+  protected UserIntent createUserIntent(
       YBUniverse ybUniverse, UUID customerUUID, boolean isCreate, Provider provider) {
     Optional<Universe> optUniverse = Optional.empty();
     if (!isCreate) {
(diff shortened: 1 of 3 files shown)


shubin-yb:
Trigger Jenkins

posriniv (approved):

amannijhawan (approved):

PR #33741, opened 2026-09-03
[docs] Note that Npgsql pool size applies per node

Sfurti-yb:
Adds one line clarifying that `Maximum Pool Size` applies per node, not cluster-wide, since the smart driver creates one connection pool per node when load balancing is enabled.


aishwarya24 (approved):
LGTM

ashetkar (approved):

Issue #23949, opened 2024-09-16
[YSQL] Surface schema mismatch version errors as 40001 instead of XX000


pao214:
Jira Link: [DB-12848](https://yugabyte.atlassian.net/browse/DB-12848)
### Description

Schema version mismatches are caused due to concurrent DDLs altering the schema version on DocDB. Currently, we return XX000 (internal error) to the user since schema version mismatches do not have a direct counterpart in postgres. Since a schema version mismatch is a retryable (non-internal) error, we should choose a different error code. Thematically, this is a serialization failure and should be surfaced as a 40001 to the user. 

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [X] I confirm this issue does not contain any sensitive information.

[DB-12848]: https://yugabyte.atlassian.net/browse/DB-12848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33748, opened 2026-09-03
[#33837] YSQL: Report index backfill failures with the PostgreSQL SQLSTATE instead of XX000

ellabaron-code:
## Summary

When a DDL from another session bumps a table's DocDB schema version while an index backfill is running, the BACKFILL INDEX statement fails with "schema version mismatch". This is unrelated to the preview concurrent-DDL feature: with `ysql_enable_concurrent_ddl` / object locking enabled the DDL waits on the table lock, so the race cannot happen, and strictly serial DDLs cannot hit it either since CREATE INDEX only returns after the backfill completes. The exposure is the default configuration (feature off), where nothing stops a second session from running a DDL against the table mid-backfill.

The error is flattened to plain text on its way to the master (`IndexInfoPB.backfill_error_message`), and the YBClient inside the tserver hosting the CREATE INDEX backend re-raised it as an `Aborted` status with no error codes attached (`Aborted` is the `Status` code, not `TransactionErrorCode::kAborted`, and plays no part in the resulting SQLSTATE). The session running CREATE INDEX therefore sees XX000 (`internal_error`) instead of the retryable 40001 (`serialization_failure`) that every other schema-mismatch path (read and write DML) already delivers.

Fix: propagate the real failure status end-to-end instead of re-tagging by message text:

- The tserver's `QueryPostgresToDoBackfill` keeps the PostgreSQL error code from the libpq error (e.g. 40001 for schema version mismatch) on the status it returns, instead of dropping it when rebuilding the status.
- The master persists the full status in a new `IndexInfoPB.backfill_status` field, alongside `backfill_error_message`, which stays populated for old readers.
- `YBClient::Data::IsBackfillIndexInProgress` re-raises the persisted status (as `Aborted`, keeping its error codes), so pggate's `FetchErrorCode` maps the schema mismatch to 40001. Against an older master that doesn't populate `backfill_status`, it falls back to the message string as before.
- The backend's unique-violation message rewrite (`YBCMakeStatusErrorData`) now skips statuses without a relation OID. Backfill duplicate-key failures now reach it with error code 23505 but no relation OID attached, and the rewrite crashed the backend dereferencing the missing relation. Such statuses already carry a fully formatted PG error message, which is now raised unchanged — so CREATE UNIQUE INDEX failing on duplicate data surfaces as 23505 (`unique_violation`, matching vanilla PostgreSQL) instead of XX000.

## Upgrade/Rollback safety

`IndexInfoPB.backfill_status` is a new optional field.

- New master, old tserver/client: old readers ignore the field; `backfill_error_message` is still populated, so behavior is unchanged.
- Old master, new client: `backfill_status` is never set; the client uses the `backfill_error_message` fallback (pre-change behavior, XX000).
- Rollback: the field is ignored again and `backfill_error_message` remains authoritative. No migration needed either way.

## Test plan

- [x] `PgSchemaVersionMismatchBackfillTest.BackfillSurfacesAsSerializationFailure` — new regression test; fails with XX000 without the fix. Passed 20 consecutive iterations.
- [x] `pg_index_backfill-test` — full suite. Four duplicate-key tests caught the backend crash and pass with the `YBCMakeStatusErrorData` guard.
- [x] `pg_op_buffering-test` — covers the DML duplicate-key path through the changed unique-violation rewrite.
- [x] `cassandra_cpp_driver-test` — YCQL backfill failure path through the changed master code.

### DIFF
--- a/src/yb/client/client-internal.cc
+++ b/src/yb/client/client-internal.cc
@@ -52,6 +52,7 @@
 #include "yb/client/table_info.h"
 
 #include "yb/common/common_util.h"
+#include "yb/common/pgsql_error.h"
 #include "yb/common/redis_constants_common.h"
 #include "yb/common/schema.h"
 #include "yb/common/schema_pbutil.h"
@@ -1043,7 +1044,13 @@ Status YBClient::Data::IsBackfillIndexInProgress(YBClient* client,
   *backfill_in_progress = true;
   if (!index_info->backfill_error_message().empty()) {
     *backfill_in_progress = false;
-    return STATUS(Aborted, index_info->backfill_error_message());
+    auto status = STATUS(Aborted, index_info->backfill_error_message());
+    // Re-tag so YSQL surfaces a retryable 40001 instead of XX000.
+    if (index_info->backfill_error_message().starts_with("ERROR:  schema version mismatch")) {
+      status = status.CloneAndAddErrorCode(
+          PgsqlRequestStatus(PgsqlResponsePB::PGSQL_STATUS_SCHEMA_VERSION_MISMATCH));
+    }
+    return status;
   } else if (index_info->index_permissions() > IndexPermissions::INDEX_PERM_DO_BACKFILL) {
     *backfill_in_progress = false;
   }
--- a/src/yb/client/client-internal.cc
+++ b/src/yb/client/client-internal.cc
@@ -1041,15 +1041,13 @@ Status YBClient::Data::IsBackfillIndexInProgress(YBClient* client,
   const auto* index_info = VERIFY_RESULT(yb_table_info.index_map.FindIndex(index_id));
 
   *backfill_in_progress = true;
-  if (!index_info->backfill_status().ok()) {
+  if (!index_info->backfill_error_message().empty()) {
     *backfill_in_progress = false;
-    // Preserve the error codes of the status the backfill failed with (such as the PostgreSQL
-    // error code), but keep this function's Aborted contract.
-    return index_info->backfill_status().CloneAndReplaceCode(Status::kAborted);
-  } else if (!index_info->backfill_error_message().empty()) {
-    // The master is older and does not populate backfill_status.
-    *backfill_in_progress = false;
-    return STATUS(Aborted, index_info->backfill_error_message());
+    // backfill_status preserves the error codes (such as the PostgreSQL error code) but may be
+    // stale after a rollback: an old master clears only backfill_error_message on success.
+    return index_info->backfill_status().ok()
+        ? STATUS(Aborted, index_info->backfill_error_message())
+        : index_info->backfill_status().CloneAndReplaceCode(Status::kAborted);
   } else if (index_info->index_permissions() > IndexPermissions::INDEX_PERM_DO_BACKFILL) {
     *backfill_in_progress = false;
   }
--- a/src/yb/client/client-internal.cc
+++ b/src/yb/client/client-internal.cc
@@ -52,7 +52,6 @@
 #include "yb/client/table_info.h"
 
 #include "yb/common/common_util.h"
-#include "yb/common/pgsql_error.h"
 #include "yb/common/redis_constants_common.h"
 #include "yb/common/schema.h"
 #include "yb/common/schema_pbutil.h"
@@ -1042,15 +1041,15 @@ Status YBClient::Data::IsBackfillIndexInProgress(YBClient* client,
   const auto* index_info = VERIFY_RESULT(yb_table_info.index_map.FindIndex(index_id));
 
   *backfill_in_progress = true;
-  if (!index_info->backfill_error_message().empty()) {
+  if (!index_info->backfill_status().ok()) {
     *backfill_in_progress = false;
-    auto status = STATUS(Aborted, index_info->backfill_error_message());
-    // Re-tag so YSQL surfaces a retryable 40001 instead of XX000.
-    if (index_info->backfill_error_message().starts_with("ERROR:  schema version mismatch")) {
-      status = status.CloneAndAddErrorCode(
-          PgsqlRequestStatus(PgsqlResponsePB::PGSQL_STATUS_SCHEMA_VERSION_MISMATCH));
-    }
-    return status;
+    // Preserve the error codes of the status the backfill failed with (such as the PostgreSQL
+    // error code), but keep this function's Aborted contract.
+    return index_info->backfill_status().CloneAndReplaceCode(Status::kAborted);
+  } else if (!index_info->backfill_error_message().empty()) {
+    // The master is older and does not populate backfill_status.
+    *backfill_in_progress = false;
+    return STATUS(Aborted, index_info->backfill_error_message());
   } else if (index_info->index_permissions() > IndexPermissions::INDEX_PERM_DO_BACKFILL) {
     *backfill_in_progress = false;
   }
--- a/src/yb/master/backfill_index.cc
+++ b/src/yb/master/backfill_index.cc
@@ -1080,17 +1080,17 @@ Status BackfillTable::Done(const Status& s, const std::unordered_set<TableId>& f
 }
 
 Status BackfillTable::MarkIndexesAsFailed(
-    const std::unordered_set<TableId>& failed_indexes, const Status& failure_status) {
+    const std::unordered_set<TableId>& failed_indexes, const Status& backfill_status) {
   if (indexes_to_build() == failed_indexes) {
     state_.store(State::kFailed, std::memory_order_release);
     StopLivenessMonitor();
     backfill_job_->SetState(MonitoredTaskState::kFailed);
   }
-  return MarkIndexesAsDesired(failed_indexes, BackfillJobPB::FAILED, failure_status);
+  return MarkIndexesAsDesired(failed_indexes, BackfillJobPB::FAILED, backfill_status);
 }
 
 Status BackfillTable::MarkAllIndexesAsFailed() {
-  return MarkIndexesAsFailed(indexes_to_build(), STATUS(InternalError, "failed"));
+  return MarkIndexesAsFailed(indexes_to_build(), STATUS(Aborted, "failed"));
 }
 
 Status BackfillTable::MarkAllIndexesAsSuccess() {
@@ -1101,10 +1101,10 @@ Status BackfillTable::MarkAllIndexesAsSuccess() {
 
 Status BackfillTable::MarkIndexesAsDesired(
     const std::unordered_set<TableId>& index_ids_set, BackfillJobPB_State state,
-    const Status& failure_status) {
+    const Status& backfill_status) {
   VLOG_WITH_PREFIX(3) << "Marking " << yb::ToString(index_ids_set)
                       << " as " << BackfillJobPB_State_Name(state)
-                      << " due to " << failure_status;
+                      << " due to " << backfill_status;
   if (!index_ids_set.empty()) {
     auto l = indexed_table_->LockForWrite();
     auto& indexed_table_pb = l.mutable_data()->pb;
@@ -1130,9 +1130,9 @@ Status BackfillTable::MarkIndexesAsDesired(
       IndexInfoPB* idx_pb = indexed_table_pb.mutable_indexes(i);
       if (index_ids_set.find(idx_pb->table_id()) != index_ids_set.end()) {
         // Should this also move to the BackfillJob instead?
-        if (!failure_status.ok()) {
-          idx_pb->set_backfill_error_message(failure_status.message().ToBuffer());
-          StatusToPB(failure_status, idx_pb->mutable_backfill_status());
+        if (!backfill_status.ok()) {
+          idx_pb->set_backfill_error_message(backfill_status.message().ToBuffer());
+          StatusToPB(backfill_status, idx_pb->mutable_backfill_status());
         } else {
           idx_pb->clear_backfill_error_message();
           idx_pb->clear_backfill_status();
--- a/src/yb/master/backfill_index.cc
+++ b/src/yb/master/backfill_index.cc
@@ -1057,8 +1057,7 @@ Status BackfillTable::Done(const Status& s, const std::unordered_set<TableId>& f
     LOG_WITH_PREFIX(WARNING) << "failed to backfill the index: " << AsString(failed_indexes)
                             << " due to " << s;
     RETURN_NOT_OK_PREPEND(
-        MarkIndexesAsFailed(failed_indexes, s.message().ToBuffer()),
-        "Couldn't mark indexes as failed");
+        MarkIndexesAsFailed(failed_indexes, s), "Couldn't mark indexes as failed");
     return CheckIfDone();
   }
 
@@ -1081,31 +1080,31 @@ Status BackfillTable::Done(const Status& s, const std::unordered_set<TableId>& f
 }
 
 Status BackfillTable::MarkIndexesAsFailed(
-    const std::unordered_set<TableId>& failed_indexes, const string& message) {
+    const std::unordered_set<TableId>& failed_indexes, const Status& failure_status) {
   if (indexes_to_build() == failed_indexes) {
     state_.store(State::kFailed, std::memory_order_release);
     StopLivenessMonitor();
     backfill_job_->SetState(MonitoredTaskState::kFailed);
   }
-  return MarkIndexesAsDesired(failed_indexes, BackfillJobPB::FAILED, message);
+  return MarkIndexesAsDesired(failed_indexes, BackfillJobPB::FAILED, failure_status);
 }
 
 Status BackfillTable::MarkAllIndexesAsFailed() {
-  return MarkIndexesAsFailed(indexes_to_build(), "failed");
+  return MarkIndexesAsFailed(indexes_to_build(), STATUS(InternalError, "failed"));
 }
 
 Status BackfillTable::MarkAllIndexesAsSuccess() {
   const auto index_ids = indexes_to_build();
   RETURN_NOT_OK(master_->xcluster_manager()->MarkIndexBackfillCompleted(index_ids, epoch_));
-  return MarkIndexesAsDesired(index_ids, BackfillJobPB::SUCCESS, "");
+  return MarkIndexesAsDesired(index_ids, BackfillJobPB::SUCCESS, Status::OK());
 }
 
 Status BackfillTable::MarkIndexesAsDesired(
     const std::unordered_set<TableId>& index_ids_set, BackfillJobPB_State state,
-    const string message) {
+    const Status& failure_status) {
   VLOG_WITH_PREFIX(3) << "Marking " << yb::ToString(index_ids_set)
                       << " as " << BackfillJobPB_State_Name(state)
-                      << " due to " << message;
+                      << " due to " << failure_status;
   if (!index_ids_set.empty()) {
     auto l = indexed_table_->LockForWrite();
     auto& indexed_table_pb = l.mutable_data()->pb;
@@ -1131,10 +1130,12 @@ Status BackfillTable::MarkIndexesAsDesired(
       IndexInfoPB* idx_pb = indexed_table_pb.mutable_indexes(i);
       if (index_ids_set.find(idx_pb->table_id()) != index_ids_set.end()) {
         // Should this also move to the BackfillJob instead?
-        if (!message.empty()) {
-          idx_pb->set_backfill_error_message(message);
+        if (!failure_status.ok()) {
+          idx_pb->set_backfill_error_message(failure_status.message().ToBuffer());
+          StatusToPB(failure_status, idx_pb->mutable_backfill_status());
         } else {
           idx_pb->clear_backfill_error_message();
+          idx_pb->clear_backfill_status();
         }
         idx_pb->clear_is_backfill_deferred();
 
--- a/src/yb/qlexpr/index.cc
+++ b/src/yb/qlexpr/index.cc
@@ -19,6 +19,7 @@
 #include "yb/common/common.messages.h"
 #include "yb/common/common.pb.h"
 #include "yb/common/schema.h"
+#include "yb/common/wire_protocol.h"
 
 #include "yb/gutil/casts.h"
 
@@ -94,6 +95,8 @@ IndexInfo::IndexInfo(const IndexInfoPB& pb)
       indexed_range_column_ids_(ColumnIdsFromPB(pb.indexed_range_column_ids())),
       index_permissions_(pb.index_permissions()),
       backfill_error_message_(pb.backfill_error_message()),
+      backfill_status_(pb.has_backfill_status() ? StatusFromPB(pb.backfill_status())
+                                                : Status::OK()),
       num_rows_read_from_table_for_backfill_(pb.num_rows_read_from_table_for_backfill()),
       num_rows_backfilled_in_index_(pb.num_rows_backfilled_in_index()),
       birth_time_(pb.birth_time()),
@@ -145,6 +148,9 @@ void IndexInfo::ToPB(IndexInfoPB* pb) const {
   }
   pb->set_index_permissions(index_permissions_);
   pb->set_backfill_error_message(backfill_error_message_);
+  if (!backfill_status_.ok()) {
+    StatusToPB(backfill_status_, pb->mutable_backfill_status());
+  }
   pb->set_num_rows_read_from_table_for_backfill(num_rows_read_from_table_for_backfill_);
   pb->set_num_rows_backfilled_in_index(num_rows_backfilled_in_index_);
   if (birth_time_ != 0) {
--- a/src/yb/qlexpr/index.cc
+++ b/src/yb/qlexpr/index.cc
@@ -150,6 +150,8 @@ void IndexInfo::ToPB(IndexInfoPB* pb) const {
   pb->set_backfill_error_message(backfill_error_message_);
   if (!backfill_status_.ok()) {
     StatusToPB(backfill_status_, pb->mutable_backfill_status());
+  } else {
+    pb->clear_backfill_status();
   }
   pb->set_num_rows_read_from_table_for_backfill(num_rows_read_from_table_for_backfill_);
   pb->set_num_rows_backfilled_in_index(num_rows_backfilled_in_index_);
--- a/src/yb/tablet/tablet.cc
+++ b/src/yb/tablet/tablet.cc
@@ -3395,21 +3395,26 @@ Result<std::tuple<std::string, uint64_t, double>> QueryPostgresToDoBackfill(
     const auto libpq_error_msg = AuxilaryMessage(result.status()).value();
     LOG(WARNING) << "libpq query \"" << query << "\" returned " << result.status() << ": "
                  << libpq_error_msg;
+    const auto pg_error_code = PgsqlError::ValueFromStatus(result.status());
+    // Keep the SQLSTATE so that an error which reaches the CREATE INDEX backend is raised with
+    // the PostgreSQL error code it failed with, rather than XX000.
+    const auto keep_pg_error_code = [&pg_error_code](Status status) {
+      return pg_error_code ? status.CloneAndAddErrorCode(PgsqlError(*pg_error_code)) : status;
+    };
     // The 2 spaces after ERROR: is necessary to match the error message.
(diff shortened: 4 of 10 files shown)


src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+  auto conn = ASSERT_RESULT(ConnectToDB("test_colo6"));
+
+  ASSERT_OK(conn.Execute(
+      "CREATE TABLE t (k INT PRIMARY KEY, v1 INT) WITH (COLOCATION = true)"));
+  ASSERT_OK(conn.Execute("INSERT INTO t SELECT i, i FROM generate_series(1, 100) i"));
+
+  Status create_index_status;
+  TestThreadHolder thread_holder;
+  thread_holder.AddThreadFunctor([&conn, &create_index_status] {
+    create_index_status = conn.Execute("CREATE INDEX idx ON t (v1)");
```
pao214: Running concurrent DDL is not supported without setting ysql_enable_concurrent_ddl=true. With the feature disabled, concurrent DDL can also run into data corruption errors, etc. So, this is not a supported mode and the test should NOT run DDLs concurrently.

Is there a repro where both DDLs run serially?

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
 INSTANTIATE_TEST_CASE_P(, PgAlterTableTest, testing::ValuesIn(kStorageFormatArray), ParamsToString);
 
+class PgSchemaVersionMismatchBackfillTest : public LibPqTestBase {
+ public:
+  void UpdateMiniClusterOptions(ExternalMiniClusterOptions* opts) override {
+    LibPqTestBase::UpdateMiniClusterOptions(opts);
+    // Disable table locks so the ALTER can land mid-backfill instead of queueing behind it.
+    opts->extra_tserver_flags.emplace_back("--enable_object_locking_for_table_locks=false");
+    opts->extra_tserver_flags.emplace_back("--ysql_enable_concurrent_ddl=false");
+    AppendFlagToAllowedPreviewFlagsCsv(opts->extra_tserver_flags, "ysql_enable_concurrent_ddl");
```
pao214: If disabling the concurrent DDL feature and running DDL concurrently, please enable the following flag

yb_fail_catalog_write_on_catalog_version_mismatch

Moreover, disable auto analyze so that auto analyze does not trigger this GUC.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+    opts->extra_tserver_flags.emplace_back("--ysql_yb_ddl_transaction_block_enabled=false");
+    opts->extra_tserver_flags.emplace_back("--ysql_yb_enable_ddl_savepoint_support=false");
```
pao214: we can combine these options into one function. Something like DisableConcurrentDDL or something of the sort.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+    // No retries: a retried chunk's fresh backfill connection would succeed and hide the error.
+    opts->extra_master_flags.emplace_back("--index_backfill_rpc_max_retries=0");
+    // Throttle to one-row BACKFILL statements so the ALTER lands with statements still to run.
+    opts->extra_tserver_flags.emplace_back("--backfill_index_write_batch_size=1");
+    opts->extra_tserver_flags.emplace_back("--backfill_index_rate_rows_per_sec=10");
```
pao214: ThrottleIndexBackfill?

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+    opts->extra_tserver_flags.emplace_back("--ysql_prefetch_limit=1");
+  }
+};
+
+// A mismatch hit by BACKFILL INDEX must reach the CREATE INDEX client as 40001, not XX000.
+TEST_F(PgSchemaVersionMismatchBackfillTest, BackfillSurfacesAsSerializationFailure) {
+  auto setup_conn = ASSERT_RESULT(Connect());
+  ASSERT_OK(setup_conn.Execute("CREATE DATABASE test_colo6 WITH COLOCATION = true"));
```
pao214: why is the database colocated? Please remove colocation if not necessary. Please add a comment if necessary.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+    opts->extra_tserver_flags.emplace_back("--ysql_enable_concurrent_ddl=false");
+    AppendFlagToAllowedPreviewFlagsCsv(opts->extra_tserver_flags, "ysql_enable_concurrent_ddl");
+    opts->extra_tserver_flags.emplace_back("--ysql_yb_ddl_transaction_block_enabled=false");
+    opts->extra_tserver_flags.emplace_back("--ysql_yb_enable_ddl_savepoint_support=false");
+    // No retries: a retried chunk's fresh backfill connection would succeed and hide the error.
+    opts->extra_master_flags.emplace_back("--index_backfill_rpc_max_retries=0");
+    // Throttle to one-row BACKFILL statements so the ALTER lands with statements still to run.
+    opts->extra_tserver_flags.emplace_back("--backfill_index_write_batch_size=1");
+    opts->extra_tserver_flags.emplace_back("--backfill_index_rate_rows_per_sec=10");
+    opts->extra_tserver_flags.emplace_back("--ysql_prefetch_limit=1");
```
pao214: Please do NOT use deprecated flags. Does ysql_yb_fetch_row_limit do the work? Whats the purpose of this flag? Is it also to throttle backfill? Then, it should go isnide the ThrottleIndexBackfill helper function

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+        return true;
+      }
+    }
+    return false;
+  }, MonoDelta::FromSeconds(60), "backfill is executing"));
+  // Let a few one-row chunks complete so the backfill connection has cached the table.
+  SleepFor(MonoDelta::FromSeconds(2));
+
+  auto ddl_conn = ASSERT_RESULT(ConnectToTsForDB(*cluster_->tablet_server(1), "test_colo6"));
+  ASSERT_OK(ddl_conn.Execute("SET yb_test_fail_next_ddl = 1"));
```
pao214, line 252: why is this necessary? Please remove the line if not necessary. Please add a comment if necessary.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
 
 INSTANTIATE_TEST_CASE_P(, PgAlterTableTest, testing::ValuesIn(kStorageFormatArray), ParamsToString);
 
+class PgSchemaVersionMismatchBackfillTest : public LibPqTestBase {
```
pao214, line 201: Can we run test with -n 20 to ensure that there is no flakiness in the test?

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+      if (VERIFY_RESULT(ts_conn.FetchRow<PGUint64>(
+              "SELECT count(*) FROM pg_stat_activity"
+              " WHERE query LIKE 'BACKFILL INDEX%'")) > 0) {
+        return true;
+      }
+    }
+    return false;
+  }, MonoDelta::FromSeconds(60), "backfill is executing"));
+  // Let a few one-row chunks complete so the backfill connection has cached the table.
+  SleepFor(MonoDelta::FromSeconds(2));
```
pao214: Please check SlowdownBackfillForTests() and see if that helps.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+
+  auto ddl_conn = ASSERT_RESULT(ConnectToTsForDB(*cluster_->tablet_server(1), "test_colo6"));
+  ASSERT_OK(ddl_conn.Execute("SET yb_test_fail_next_ddl = 1"));
+  ASSERT_NOK(ddl_conn.Execute("ALTER TABLE t ADD COLUMN v2 INT"));
+
+  thread_holder.JoinAll();
+  ASSERT_NOK(create_index_status);
+  LOG(INFO) << "Client-visible error: " << create_index_status;
+  ASSERT_STR_CONTAINS(create_index_status.ToString(), "schema version mismatch");
```
pao214: What if the error is optimized away in the future? Can we instead check for the absence of an internal error? We can also assert that the test result is either a pass or a serialization error? Please allow the index backfill pass case in the test.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+    create_index_status = conn.Execute("CREATE INDEX idx ON t (v1)");
+  });
+
+  // The backfill connection lands on the colocation tablet's leader, so poll every tserver.
+  std::vector<PGConn> ts_conns;
+  for (size_t i = 0; i < cluster_->num_tablet_servers(); ++i) {
+    ts_conns.push_back(ASSERT_RESULT(
+        ConnectToTsForDB(*cluster_->tablet_server(i), "test_colo6")));
+  }
+  ASSERT_OK(WaitFor([&ts_conns]() -> Result<bool> {
```
pao214: Can we replace this pg_stat_activity poll using yb_test_block_index_phase?

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,73 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+  void UpdateMiniClusterOptions(ExternalMiniClusterOptions* opts) override {
+    LibPqTestBase::UpdateMiniClusterOptions(opts);
+    // Disable table locks so the ALTER can land mid-backfill instead of queueing behind it.
+    opts->extra_tserver_flags.emplace_back("--enable_object_locking_for_table_locks=false");
+    opts->extra_tserver_flags.emplace_back("--ysql_enable_concurrent_ddl=false");
+    AppendFlagToAllowedPreviewFlagsCsv(opts->extra_tserver_flags, "ysql_enable_concurrent_ddl");
+    opts->extra_tserver_flags.emplace_back("--ysql_yb_ddl_transaction_block_enabled=false");
+    opts->extra_tserver_flags.emplace_back("--ysql_yb_enable_ddl_savepoint_support=false");
+    // No retries: a retried chunk's fresh backfill connection would succeed and hide the error.
+    opts->extra_master_flags.emplace_back("--index_backfill_rpc_max_retries=0");
```
pao214, line 208: Do retries resolve the problem then? e01d18cce67 also seems to suggest so.

pao214:
> When a concurrent DDL bumps a table's DocDB schema version while an index backfill is running

Does this happen when the concurrent DDL feature is on? If not, this MUST be clarified -- the test sets `--ysql_enable_concurrent_ddl=false` and `--enable_object_locking_for_table_locks=false`, so the wording and the repro seem to disagree. Can we also clarify that concurrent DDL is a preview feature, off by default, and not an expected operation? Does this issue happen when (a) the concurrent DDL feature is disabled and (b) DDLs are executed serially? That is a significant question about the impact of this issue.

Was this seen in prod? Note that concurrent DDL is not supported in production.

> That error is flattened to plain text on its way to the master

Why? Can we retain all the elements of the error? The error has other sub-fields as well which might prove important -- `PgsqlError`, `RelationOid`, `AuxilaryMessage`. Can we not flatten the error, and thread the `Status` all the way through?

Possible context: 8bc12044c4e bubbles up the error as a message. Back then we did not use `pgwrapper::PGConn` (that hunk calls `PQexec` / `PQresultErrorMessage` directly), so there was no `Status` to propagate -- which is why the error became a string. That is no longer true.
cc @jasonyb

> and the client re-raised it as an untagged Aborted status

The client terminology is overloaded in this sentence. Is the first client the tserver proxy? And the second client the pg session that runs CREATE INDEX? Let's also clarify that the `Aborted` here is `Status::Code` and not `TransactionErrorCode::kAborted` -- and that it plays no part in the resulting SQLSTATE.

Can we use a different issue number for this? #23949 is meant for DMLs and not INDEX BACKFILL. In fact, the issue is that the error is being flattened and not propagated. Schema version mismatch is just one victim -- `CREATE UNIQUE INDEX` on duplicate data reports XX000 instead of 23505 through the same path, with no race involved.

pao214 (changes requested):
Reviewed commit summary and design in the previous comment.

ellabaron-code:
It does not repro with `ysql_enable_concurrent_ddl` / object locking enabled: the `ALTER` then waits on the table lock until `CREATE INDEX` finishes, which is why the test disables those flags. It also does not repro with serial DDLs, since `CREATE INDEX` only returns after backfill completes. The exposure is the default config (concurrent DDL off) with a second session issuing a DDL against the table while the backfill runs. Reworded the description to state this and to not read as the preview feature.

No info on prod occurrences — I was guided by the issue description only.

Agreed, it's doable: the tserver→master RPC already carries the full `AppStatusPB` — the flattening happens on the master, which persists only `status.message()` into `IndexInfoPB.backfill_error_message`, and the original status does carry `PgsqlError` and `AuxilaryMessage` from libpq. I'll add an optional `AppStatusPB` next to `backfill_error_message` in `IndexInfoPB`, populate it on the master, and have the client rebuild the `Status` from it, keeping the string for mixed-version clusters.

Reworded the description to disambiguate the two "client"s and to state that `Aborted` is the `Status` code, not `TransactionErrorCode::kAborted`, and plays no part in the resulting SQLSTATE.

I disagree — #23949 doesn't mention DML; it asks generically for schema version mismatches to be surfaced as 40001 instead of XX000, which is exactly what this PR does. That said, happy to open a new issue if you prefer it tracked separately.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Done — added both flags.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Done — extracted into a ThrottleIndexBackfill helper.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Done — combined into a DisableConcurrentDDL helper.Done — combined into a `DisableConcurrentDDL` helper.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Not necessary — removed. The table is now in the default database with `SPLIT INTO 1 TABLETS` (one tablet keeps the whole backfill on a single connection).

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Yes, `ysql_yb_fetch_row_limit` does the work — switched to it, and moved it into `ThrottleIndexBackfill`. It is part of the throttling: without it the first BACKFILL statement's read pages in 1024 rows, so the whole table gets backfilled in one statement regardless of `backfill_index_write_batch_size`.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: The test needs the backfill connection to send a stale schema version, and this `SET` is what creates that state. The `ALTER` bumps the DocDB schema version first and bumps the catalog version at commit; the catalog version bump is what tells other backends (including the backfill connection) to refresh their caches. `yb_test_fail_next_ddl` fails the `ALTER` between the two: the DocDB schema version stays bumped, the catalog version never bumps, so the backfill connection keeps its cached schema and every subsequent `BACKFILL` statement hits the mismatch.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: I checked — it doesn't help here. `TEST_slowdown_backfill_by_ms` sleeps once at the top of `BackfillIndexesForYsql`, before the backfill PG connection is created. The test needs the failed `ALTER` to land after that connection has run at least one `BACKFILL` statement (so its cache holds the old schema) but before the remaining ones — a connection created after the `ALTER` loads fresh metadata and succeeds. So the backfill has to be stretched across many statements, which is what the throttle flags do; delaying the start doesn't open that window.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Done — the test now allows the pass case, and a failure must carry `SQLSTATE 40001` (checked via the PG error code on the status, not the message text).

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: I don't think it can — `yb_test_block_index_phase` pauses the CREATE INDEX backend at phase boundaries (before/after backfill), never in the middle of the backfill itself. The test needs a point where the backfill connection has already run at least one `BACKFILL` statement (its cache holds the old schema) and still has more to run; blocking before the backfill starts would make the failed `ALTER` land before the connection exists, and a fresh connection loads current metadata and succeeds. The poll now waits for the `BACKFILL` query text in `pg_stat_activity` to change, which proves a statement boundary passed on that connection.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Mostly, yes — that's by design from the commit you cite (the 2022 master-failover fix that made backfill retry on schema version mismatch): the mismatch from a `BACKFILL` statement becomes `TryAgain`, and the retried chunk's fresh backfill connection reloads metadata and succeeds. But the retry budget is bounded (`index_backfill_rpc_max_retries`, default 150, plus the backfill deadline), and when it runs out the error still escapes to the CREATE INDEX client — today as `XX000`. This fix is about the SQLSTATE it carries when that happens, not about preventing it. The test sets retries to 0 only to reach that path deterministically instead of looping through 150 retries.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Ran it with -n 20 after the redesign — all 20 iterations passed.

pao214:
> I disagree — https://github.com/yugabyte/yugabyte-db/issues/23949 doesn't mention DML; it asks generically for schema version mismatches to be surfaced as 40001 instead of XX000, which is exactly what this PR does. That said, happy to open a new issue if you prefer it tracked separately.

Please open a new issue. The intention of the old issue is not to solve this case. Otherwise, we will have to reopen this issue each time schema version mismatch is not surfaced at 40001.

ellabaron-code:
Opened #33837 and repointed the PR at it.

src/yb/client/client-internal.cc
```diff
@@ -1041,7 +1041,13 @@ Status YBClient::Data::IsBackfillIndexInProgress(YBClient* client,
   const auto* index_info = VERIFY_RESULT(yb_table_info.index_map.FindIndex(index_id));
 
   *backfill_in_progress = true;
-  if (!index_info->backfill_error_message().empty()) {
+  if (!index_info->backfill_status().ok()) {
```
pao214: Can we instead do this?

```cpp
if (!index_info->backfill_error_message().empty()) {
  *backfill_in_progress = false;
  return index_info->backfill_status().ok()
      ? STATUS(Aborted, index_info->backfill_error_message())
      : index_info->backfill_status().CloneAndReplaceCode(Status::kAborted);
}
```

The exact code is upto you but lets use !index_info->backfill_error_message().empty() as the condition for determining whether error exists. Because on a rollback to old master, old master does not clear backfill_status().

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -14,6 +14,7 @@
```
pao214, line 14: Does CREATE UNIQUE INDEX on duplicate data now surface 23505 instead of XX000? We need a test for that.

src/yb/master/backfill_index.cc
```diff
@@ -1081,31 +1080,31 @@ Status BackfillTable::Done(const Status& s, const std::unordered_set<TableId>& f
 }
 
 Status BackfillTable::MarkIndexesAsFailed(
-    const std::unordered_set<TableId>& failed_indexes, const string& message) {
+    const std::unordered_set<TableId>& failed_indexes, const Status& failure_status) {
```
pao214: Can we rename it to backfill_status? if status = OK(), pass. if status != OK() it is a fail. Failure status is a little confusing.

src/yb/qlexpr/index.cc
```diff
@@ -145,6 +148,9 @@ void IndexInfo::ToPB(IndexInfoPB* pb) const {
   }
   pb->set_index_permissions(index_permissions_);
   pb->set_backfill_error_message(backfill_error_message_);
+  if (!backfill_status_.ok()) {
```
pao214, line 151: This is a conditional set while backfill_error_message does not set the value conditionally.

We should either set the status uncodnitionally or clear it in else branch.

src/yb/master/backfill_index.cc
```diff
@@ -1081,31 +1080,31 @@ Status BackfillTable::Done(const Status& s, const std::unordered_set<TableId>& f
...
     StopLivenessMonitor();
     backfill_job_->SetState(MonitoredTaskState::kFailed);
   }
-  return MarkIndexesAsDesired(failed_indexes, BackfillJobPB::FAILED, message);
+  return MarkIndexesAsDesired(failed_indexes, BackfillJobPB::FAILED, failure_status);
 }
 
 Status BackfillTable::MarkAllIndexesAsFailed() {
-  return MarkIndexesAsFailed(indexes_to_build(), "failed");
+  return MarkIndexesAsFailed(indexes_to_build(), STATUS(InternalError, "failed"));
```
pao214: Is InternalError the correct code? Did we want to return Aborted?

pao214 (dismissed):
When reviewing, I realized that the DST team is a better reviewer for this. Adding basava and Amit as reviewers instead. I will let them decide the direction of this diff.

src/yb/client/client-internal.cc
ellabaron-code: Done.

src/yb/master/backfill_index.cc
ellabaron-code: Renamed

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Yes, with this change the duplicate-key backfill failure comes back as `23505`. Extended the existing `PgIndexBackfillTest.Unique` to assert on the error code.

src/yb/qlexpr/index.cc
ellabaron-code: Good catch, clearing it in the else branch now.

src/yb/master/backfill_index.cc
ellabaron-code: Changed to `Aborted`. Note it makes no client-visible difference: the client replaces the code with `Aborted` anyway, and the SQLSTATE is derived from the `PgsqlError` tag rather than the status code, so this path stays `XX000` either way.

amitanandaiyer (approved):
will defer to patnaik on the pg side/error codes.

dst/backfill related changes looks good to me.

src/yb/tablet/tablet.cc
```diff
@@ -3395,21 +3395,26 @@ Result<std::tuple<std::string, uint64_t, double>> QueryPostgresToDoBackfill(
...
+    // Keep the SQLSTATE so that an error which reaches the CREATE INDEX backend is raised with
+    // the PostgreSQL error code it failed with, rather than XX000.
+    const auto keep_pg_error_code = [&pg_error_code](Status status) {
+      return pg_error_code ? status.CloneAndAddErrorCode(PgsqlError(*pg_error_code)) : status;
+    };
     // The 2 spaces after ERROR: is necessary to match the error message.
     constexpr auto kSchemaMismatchSubstring = "ERROR:  schema version mismatch";
     if (libpq_error_msg.starts_with(kSchemaMismatchSubstring)) {
-      return STATUS(TryAgain, libpq_error_msg);
+      return keep_pg_error_code(STATUS(TryAgain, libpq_error_msg));
```
pao214: optional: Can we instead call

```cpp
return STATUS(TryAgain, libpq_error_msg, libpq_error_codes, size_t(0));
```

where

```cpp
libpq_error_codes = libpq_status.ErrorCodesSlice()
```

Rationale: if there are other tags like PgsqlError that we want to add in the future, this takes care of it.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
```diff
@@ -170,6 +171,98 @@ std::string ParamsToString(const testing::TestParamInfo<StorageFormat>& param_in
...
+    LibPqTestBase::UpdateMiniClusterOptions(opts);
+    DisableConcurrentDDL(opts);
+    ThrottleIndexBackfill(opts);
+    // No retries: a retried chunk's fresh backfill connection would succeed and hide the error.
+    opts->extra_master_flags.emplace_back("--index_backfill_rpc_max_retries=0");
+  }
+};
+
+// A mismatch hit by BACKFILL INDEX must reach the CREATE INDEX client as 40001, not XX000.
+TEST_F(PgSchemaVersionMismatchBackfillTest, BackfillSurfacesAsSerializationFailure) {
```
pao214: nit: Isn't this a test primarily about index backfill? pg_index_backfill-test.cc sounds like a more natural place to add this test.

pao214:
Looks mostly good.

title suggestion: [#33837] YSQL: Report index backfill failures with the PostgreSQL SQLSTATE instead of XX000

Rationale: The main impact of this change is PostgreSQL compatibility of CREATE INDEX error. This is important for postgres wire compatibility since upstream postgres does not return XX000 for backfill errors.

ellabaron-code:
Done — took the suggested title.

src/yb/tablet/tablet.cc
ellabaron-code: Done. Nice — this also carries any other tags on the `libpq` status, not just the `SQLSTATE`.

src/yb/yql/pgwrapper/pg_alter_table-test.cc
ellabaron-code: Moved to `pg_index_backfill-test.cc` as-is. Kept the fixture standalone rather than deriving from `PgIndexBackfillTest`, since that suite is parameterized on table locks and this test needs them off.

PR #33749, opened 2026-09-03
[PLAT-22333] YBA: Fix as_root upgrade failure after an older yba-ctl rewrote the installer state

shubin-yb:
## Summary

Upgrading YBA Installer from 2025.2.x to 2026.1.x could fail with

```
FATAL invalid reconfigure during upgrade: cannot change as_root from false
```

even though `yba-ctl.yml` says `as_root: true`. The state file in these cases has `run_schemas` containing 13 but no `config.as_root` key:

1. A 2026.1 `yba-ctl preflight --upgrade` / `upgrade` runs migration 13, stores `config.as_root` and marks it run, then fails on some other precheck (no rollback on preflight failure).
2. The operator runs the **installed 2025.2** binary (`yba-ctl reconfigure` always ends in `StoreState`). Its `Config` struct has no `AsRoot`, so `json.Unmarshal` drops the key and the file is written back without it; `run_schemas` is a plain `[]int` and survives.
3. The next 2026.1 run skips migration 13, `Config.AsRoot` loads as the zero value `false`, and `ValidateReconfig` fatals. Every retry fails until the state file is hand-edited.

Changes in `pkg/ybactlstate`:

* The `migrations` table holds `migration{run, stateField}`. Migrations that write into the state file (9 `initialized`, 11 `config.self_signed_cert`, 12 `services`, 13 `config.as_root`) declare the json path they populate; `handleMigration` re-runs one that is marked run but whose field is missing from the loaded document, logging `state file is missing config.as_root, re-running migration 13`. This is the repair for state files already damaged by shipped binaries.
* `State` keeps the document it was loaded from; `MarshalJSON` merges the struct's output over it (struct wins, unknown keys carried through, numbers kept verbatim). Once the writer is a build with this change, an older/newer yba-ctl mix can no longer drop fields.

`cmd/init.go`: `handleRootCheck` reloads viper after writing `as_root` into `yba-ctl.yml`, so migration 13 in the same process copies the value just written rather than the unset one.

## Upgrade/Rollback safety

* On load, a state file whose `run_schemas` includes a state-writing migration but lacks that migration's field triggers a re-run of that migration (and a `StoreState`). Only migrations 9, 11, 12 and 13 declare a field; all four are idempotent.
* The state file is now written with sorted keys and carries through keys this binary does not know. Older yba-ctl versions read it unchanged (they ignore unknown keys); an older binary that rewrites the file still drops the keys it does not know, exactly as before — a build with this change repairs that on its next run.
* Workaround for installs already stuck on a build without this change: add `"as_root": true` (root install) or `false` to the `config` object in `/opt/yba-ctl/.yba_installer.state`, then retry the upgrade.

## Test plan

- [x] `go test ./...` in `managed/yba-installer` (new: `TestHandleMigrationRerunsLostStateField`, `TestLoadStateRerunsLostAsRootMigration`, `TestLoadStateKeepsStoredAsRoot`, `TestStateJsonCarriesUnknownFields`, `TestStateJsonWithoutLoadedDocument`).
- [x] Reproduced on a test VM: install 2025.2.5.2-b5 → stock 2026.1.1.1-b2 `preflight --upgrade` (fails on disk) → 2025.2 `yba-ctl reconfigure` → stock 2026.1.1.1-b2 `upgrade` fails with `cannot change as_root from false`; state shows `run_schemas [..14]` and no `config.as_root`.
- [x] Same flow with this change built on top of the 2026.1.1.1-b2 tag, after also injecting unknown keys (`config.future_flag`, `future_section.id: 12345678901234567890`) into the damaged state: `upgrade` logs `re-running migration 13`, completes, and the final state has `as_root: true` plus both injected keys intact (id not rounded). `yba-ctl status` afterwards leaves the file unchanged.

### DIFF
--- a/managed/yba-installer/cmd/init.go
+++ b/managed/yba-installer/cmd/init.go
@@ -135,6 +135,8 @@ func handleRootCheck(cmdName string) {
 			log.Warn("Failed to set as_root in config file, please set it manually")
 			log.Fatal("Failed to set as_root in config file: " + err.Error())
 		}
+		// Reload so the as_root state migration copies the value just written, not the unset one.
+		common.InitViper()
 		return
 	} else if user.Uid == "0" && !viper.GetBool("as_root") {
 		log.Fatal("running as root user with 'as_root' set to false is not supported")
--- a/managed/yba-installer/pkg/ybactlstate/json.go
+++ b/managed/yba-installer/pkg/ybactlstate/json.go
@@ -1,6 +1,7 @@
 package ybactlstate
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 )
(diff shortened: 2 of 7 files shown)


chidmuthu (approved):

shubin-yb:
trigger jenkins

PR #33751, opened 2026-09-03
[BACKPORT 2026.1.1][PLAT-22333] YBA: Fix as_root upgrade failure after an older yba-ctl rewrote the installer state

shubin-yb:
Upgrading 2025.2.x to 2026.1.x could fail with "invalid reconfigure during
upgrade: cannot change as_root from false". A 2026.1 yba-ctl run (e.g. a
preflight that then failed on another check) stores config.as_root in the
state file and marks migration 13 as run. If the still-installed 2025.2 binary
then rewrites the state file (reconfigure always does), json.Unmarshal drops
the key it does not know while run_schemas survives, so the next 2026.1 run
skips the migration and loads as_root as the zero value.

Migrations that write into the state file now declare the json path they
populate and are re-run when that field is missing from the loaded document.
State also keeps the document it was loaded from and MarshalJSON carries
unknown keys through, so a mix of yba-ctl versions can no longer drop fields.
handleRootCheck reloads viper after backfilling as_root into yba-ctl.yml so a
migration in the same process does not copy the stale unset value.

Claude-Session: https://claude.ai/code/session_01CJShf6RcFVvALX6pSTpbek

Original commit: 329bc68234486c81e564863becbdb9a4b30d7fbd / #33749

## Merge conflicts

- `managed/yba-installer/pkg/ybactlstate/migration.go:401` — kept the cherry-picked `migrations` table but dropped its `perfAdvisorConfig` entry; migration 15 (`migratePerfAdvisorConfig`) does not exist on 2026.1.1. The resulting table is the one verified end-to-end on a 2026.1.1.1-b2 build.

### DIFF
--- a/managed/yba-installer/cmd/init.go
+++ b/managed/yba-installer/cmd/init.go
@@ -134,6 +134,8 @@ func handleRootCheck(cmdName string) {
 			log.Warn("Failed to set as_root in config file, please set it manually")
 			log.Fatal("Failed to set as_root in config file: " + err.Error())
 		}
+		// Reload so the as_root state migration copies the value just written, not the unset one.
+		common.InitViper()
 		return
 	} else if user.Uid == "0" && !viper.GetBool("as_root") {
 		log.Fatal("running as root user with 'as_root' set to false is not supported")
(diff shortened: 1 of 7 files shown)


shubin-yb:
Trigger Jenkins

chidmuthu (approved):

Issue #33753, opened 2026-09-03
[ClaudeCode] Add review-pg-import skill for reviewing upstream PostgreSQL point-imports


kai-franz:
## Description

A point-import from upstream PostgreSQL conflicts twice, differently: once against prior
point-imports on the `yugabyte/postgres` side, and again against YB modifications when the
subtree merges into `yugabyte-db`. Reviewing the first half is mechanical but easy to do
badly — the common failure is reading the cherry-pick diffs, finding them plausible, and
taking the author's conflict-resolution notes at face value rather than checking the claims
they make about both trees.

Add a `review-pg-import` skill covering the `yugabyte/postgres` side of such a review:

- Establishing the target `yb-pg<version>` branch, the corresponding upstream
  `REL_<N>_STABLE` back-patch branch, and the true review scope.
- A mechanical patch-fidelity pass classifying every commit as identical / line-offsets-only
  / genuinely different, so only the last group needs reading.
- Verifying each factual claim in the resolution notes against the tree rather than trusting
  it.
- Change-set completeness, for both CVE batches (enumerable via `Security:` trailers) and
  ordinary imports (no trailer, so a structural search is needed).
- Import-scope policy, cherry-pick footers, and authorship metadata.

Out of scope: the `yugabyte-db` side of the import, and performing an import rather than
reviewing one.

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

kai-franz:
Added to internal repo instead

PR #33754, opened 2026-09-03
[#33753] agents: Add review-pg-import skill

kai-franz:
## Summary

A point-import from upstream PostgreSQL conflicts twice, differently: once against prior
point-imports on the `yugabyte/postgres` side, and again against YB modifications when the
subtree merges into `yugabyte-db`. Reviewing the first half is mechanical but easy to do
badly — the common failure is reading the cherry-pick diffs, finding them plausible, and
taking the author's conflict-resolution notes at face value rather than checking the claims
they make about both trees.

Adds `.claude/skills/review-pg-import/`, covering the `yugabyte/postgres` side of such a
review:

- Establishing the target `yb-pg<version>` branch, the corresponding upstream
  `REL_<N>_STABLE` back-patch branch, and the true review scope.
- A mechanical patch-fidelity pass classifying every commit as identical / line-offsets-only
  / genuinely different, so only the last group needs reading.
- Verifying each factual claim in the resolution notes against the tree rather than trusting
  it.
- Change-set completeness, for both CVE batches (enumerable via `Security:` trailers) and
  ordinary imports (no trailer, so a structural search is needed).
- Import-scope policy, cherry-pick footers, and authorship metadata.

It wraps, and points at,
[merge-with-upstream-repositories](https://docs.yugabyte.com/stable/contribute/core-database/merge-with-upstream-repositories)
§ "Squash point-imports" / "Review cherry-picks", which stays the source of truth.

Out of scope: the `yugabyte-db` side of the import, and performing an import rather than
reviewing one.

The skill resolves both git remote names at runtime (the `yugabyte/postgres` fork and true
upstream PostgreSQL) instead of hardcoding them, and adds whichever the checkout is missing,
so it works regardless of local remote naming.

## Test plan

Documentation-only; no runtime behavior change.

- [x] Skill loads and is discoverable in a Claude Code session from the committed path.
- [x] Remote-resolution snippet exercised against a real `yugabyte/postgres` checkout: it
      resolves the fork remote by URL and correctly detects the absence of an upstream
      PostgreSQL remote (falling through to the `git remote add`).
- [ ] Not re-run: the review procedure itself, which was previously exercised on two
      upstream point-import reviews.

### DIFF
--- a/.claude/skills/review-pg-import/SKILL.md
+++ b/.claude/skills/review-pg-import/SKILL.md
@@ -1,0 +1,334 @@
+---
+name: review-pg-import
+description: >-
+  Review a point-import into an upstream-fork repo such as yugabyte/postgres:
+  commits cherry-picked from upstream PostgreSQL onto a yb-pg<version> branch,
+  as a GitHub PR, a remote branch, or a local commit range.  Covers cherry-pick
+  footer and back-patch-branch checks, mechanical patch-fidelity comparison
+  against each upstream commit, verifying conflict-resolution notes instead of
+  trusting them, change-set completeness for both CVE batches and ordinary
+  imports, and the import-scope policy.  Use when asked to review, verify, or
+  check such a PR, branch, or set of cherry-picks.
+---
+
+# Reviewing PostgreSQL point-imports
+
+Authoritative process doc:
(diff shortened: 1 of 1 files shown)


kai-franz:
Making this an internal plugin instead

PR #33762, opened 2026-09-04
[BACKPORT 2026.1][PLAT-22035] YBA: Resolve dependent CR universe references by yba-resource-id (#33745)

shubin-yb:
## Summary

A YBUniverse custom resource is named in a dependent resource's spec
(`Backup`,
`BackupSchedule`, `PitrConfig`, `RestoreJob`, `DrConfig`,
`SupportBundle`) by its
`metadata.name`, and `OperatorUtils.getUniverseFromNameAndNamespace`
turns that into a YBA
universe. It did so purely by name, via
`YBUniverseReconciler.getUniverseName`, which since
PLAT-21329 always derives `<resourceName>-<hash(name+namespace+uid)>`
from the resource
metadata. An imported universe keeps the name it already had in YBA, so
the lookup never
matches it and every dependent reconcile fails with `No universe found
with name ...`.

The resolver now goes through the `io.yugabyte.operator/yba-resource-id`
annotation first
and falls back to both historical naming schemes - the resolution order
`YBUniverseReconciler.resolveExistingUniverse` already applies to the
YBUniverse CR itself.
That reconcile writes the annotation back onto the resource, so it is
set by the time a
dependent CR reconciles. The two resolvers stay separate: the
reconciler's owns
self-healing and ambiguity reporting, this one is read-only, and it
scopes the UUID lookup
to the requesting customer (the name lookup enforces that;
`Universe.maybeGet` does not).

`createBackupScheduleCr` had the mirror-image bug on the write side: it
put the YBA universe
name into `BackupSchedule` `spec.universe`, which consumers read back as
a YBUniverse
resource name. Those coincide only for an imported universe; for an
operator-created one the
YBA name carries the hash suffix. It now writes
`kubernetesResourceDetails.name`, matching
`createBackupCr`, and fails loudly if the universe has no resource
details rather than
emitting a CR that points at nothing.

Original commit: a42f253942ec1798a4073526c723c28be81d00c7 / #33745

## Test plan

- [x] `build-support/lint.sh --rev upstream/master` - clean;
google-java-format reports no diff on
      either file.
- [ ] `sbt "testOnly
com.yugabyte.yw.common.operator.utils.OperatorUtilsTest"` - runs in CI;
a local
run was not possible (`com.yugabyte#yba-client-v2;1.8.3` does not
resolve in my environment).
New cases cover annotation resolution, both name fallbacks, a stale
annotation, and the
customer boundary; the two existing `createBackupScheduleCr` cases now
assert the YBUniverse
      resource name in `spec.universe`.

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>

## Merge conflicts

Resolved by hand on 2026.1; the cherry-pick did not apply cleanly.

- `managed/src/main/java/com/yugabyte/yw/common/operator/utils/OperatorUtils.java:303` - the
  `if (ybUniverse == null)` guard in `getUniverseFromNameAndNamespace` is unchanged context in the
  original commit but does not exist on 2026.1, so it conflicted. Took the original's post-state:
  the backport adds those four lines, and the method now matches master. Behavior is the same
  either way - `getUniverseFromCr` null-checks its argument - but without the guard 2026.1 would
  keep resolving a missing custom resource one frame later and lose the debug log.
- `managed/src/test/java/com/yugabyte/yw/common/operator/utils/OperatorUtilsTest.java:37` and
  `:88` - the new `KubernetesResourceDetails`, `YBUniverse` and `YBUniverseSpec` imports sort next
  to KMS-config imports that 2026.1 does not have. Kept only the three new imports, in the same
  sorted positions.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/common/operator/utils/OperatorUtils.java
+++ b/managed/src/main/java/com/yugabyte/yw/common/operator/utils/OperatorUtils.java
@@ -318,19 +318,79 @@ public class OperatorUtils {
         .map(PlatformInstance::getUuid);
   }
 
-  public Universe getUniverseFromNameAndNamespace(
-      Long customerId, String universeName, String namespace) throws Exception {
+  /**
+   * Resolves the YBA universe backing a YBUniverse custom resource, given that resource's name and
+   * namespace.
+   *
+   * @param customerId the customer the universe must belong to
+   * @param crName the {@code metadata.name} of the YBUniverse custom resource, as named by the
+   *     {@code universe} field of a dependent resource's spec
+   * @param namespace the namespace holding the custom resource
+   * @return the universe, or null if either the custom resource or its universe does not exist
+   */
+  public Universe getUniverseFromNameAndNamespace(Long customerId, String crName, String namespace)
+      throws Exception {
     KubernetesResourceDetails ybUniverseResourceDetails = new KubernetesResourceDetails();
-    ybUniverseResourceDetails.name = universeName;
+    ybUniverseResourceDetails.name = crName;
     ybUniverseResourceDetails.namespace = namespace;
     YBUniverse ybUniverse = getYBUniverse(ybUniverseResourceDetails);
-    String name = YBUniverseReconciler.getUniverseName(ybUniverse);
-    log.debug("Getting universe from name: {}", name);
-    Optional<Universe> universe = Universe.maybeGetUniverseByName(customerId, name);
-    if (universe.isPresent()) {
-      return universe.get();
+    if (ybUniverse == null) {
+      log.debug("YBUniverse '{}' not found in namespace '{}'", crName, namespace);
+      return null;
     }
-    return null;
+    return getUniverseFromCr(customerId, ybUniverse).orElse(null);
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


shubin-yb:
Trigger Jenkins

posriniv (approved):

shubin-yb:
The `yba-unit-tests` failure came from the 2026.1 merge, not the backported change: both sides appended to the end of `OperatorUtilsTest` (this branch's `getUniverseFromCr` cases, 2026.1's KMS config cases from 7623dea6b34), and the merge dropped the closing brace of `testCreateBackupScheduleCrWithoutResourceDetailsThrows` while leaving the three new imports out of sort order. Fixed in c33b86e9415; the file now parses and google-java-format reports no diff.

Trigger Jenkins

shubin-yb:
Rebased onto the current 2026.1 tip (9d64e17edf9) and force-pushed, replacing the merge commit and its fixup with a single backport commit. The end-of-class conflict against 2026.1's KMS config tests is resolved in the commit itself, so the tree is unchanged from what was verified on the previous head: both files parse, google-java-format reports no diff, and `lint.sh --rev origin/2026.1` is clean.

Trigger Jenkins

Issue #33557, opened 2026-08-26
[DocDB] Add yb-admin commands to verify an xCluster pair holds the same data


balajisubramanian-yugabyte:
Jira Link: [DB-23217](https://yugabyte.atlassian.net/browse/DB-23217)
## Description

There is no way to check that an xCluster target actually holds the same rows as its source.

Comparing table hashes by hand becomes possible once the hashes are comparable, but doing it correctly means pinning one read time that both universes can read, confirming the two catalogs agree before and after hashing, and telling replication lag apart from real data loss. Done by hand or in a script, the failure mode is a false report of data loss, which is the most expensive wrong answer available.

### Proposed

Two `yb-admin` commands:

- `verify_xcluster_slice` compares one table pair over one key range, at a single hybrid time resolved from the target's xCluster safe time, and reports `MATCH`, `TRANSIENT`, `DIVERGED`, `STARVED`, `SCHEMA_MISMATCH` or `INFRA` as JSON.
- `verify_xcluster_group` sweeps a whole replication group by calling the above repeatedly. It discovers the table pairs from the group itself, bounds the work per invocation, and prints a resume position so a scheduled run continues where the previous one stopped.

Only `DIVERGED` claims data was lost. A side that could not hash yet, whether from replication lag, history retention, or a timeout, reports `STARVED` and is never compared. Catalogs that disagree, or that change while hashing, report `SCHEMA_MISMATCH`. Sides that hashed under different schemes report `INFRA` rather than a difference.

Also adds `get_table_schema`, which prints one table's schema as JSON so an operator can see what a `SCHEMA_MISMATCH` refers to.

## Issue Type

kind/new-feature

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23217]: https://yugabyte.atlassian.net/browse/DB-23217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

hsdfat:
Operator input from a production deployment, in case it helps shape the verdicts.

[...]

PR #33764, opened 2026-09-04
[#33557] Docs: Add yb-admin verify_xcluster_slice and verify_xcluster_group reference

balajisubramanian-yugabyte:
## Summary

Reference documentation for the two cross-cluster verification commands,
`verify_xcluster_slice` and `verify_xcluster_group`, plus a cross-reference from
`get_table_hash`'s `--max_rows_per_scan` note to the command whose YB-TServer support it
arrived with.

The text covers what an operator has to decide or react to:

- Which tables a sweep verifies, including the ones a replication group only names
  indirectly, and how a table found on one universe but not the other is reported.
- The flags that shape a sweep, and the fact that none of them reduce what is verified — a
  sweep always covers the whole group. `--max_rows_per_scan` bounds a single scan;
  `--max_concurrent_ranges` only decides how much of the sweep is in flight, and so how much
  load lands on both universes.
- That ranges of a table are verified concurrently but slices within a range are sequential,
  each starting where the last stopped — so concurrency buys nothing on a single-tablet
  table, and slice lines are not in key order once more than one range is in flight.
- The JSON each command emits, the verdicts, and how they rank.
- The exit-status difference between the two: `verify_xcluster_group` exits non-zero for any
  summary but `kMatch` so cron need only check `$?`, while `verify_xcluster_slice`
  deliberately exits 0 even on `kDiverged`, because it is the primitive a driver walks a
  table with and the driver reads the verdict to decide where to go next.

**Sequencing.** These commands are not in master yet, so this stays a draft until they land.
It is also stacked on #33763 — this branch edits a note that PR adds, so both commits appear
here until #33763 merges, after which this diff reduces to the verification reference alone.

## Test plan

- [x] Documentation only; no runtime behavior change.
- [x] Verified every flag named in the new text is defined in `src/yb/tools/yb-admin_cli.cc`:
      `--source_master_addresses`, `--max_rows_per_scan`, `--max_concurrent_ranges`,
      `--verify_skip_tables`, `--verify_recheck_attempts`, `--verify_recheck_backoff_ms`.
      Confirmed no flag is named that the code does not define.
- [x] Verified the documented flag semantics and defaults match the flag help text, and that
      the verdict names and their documented ranking match `VerdictSeverity` in
      `src/yb/tools/xcluster_verify.cc`.
- [x] Verified the documented JSON field names match the keys emitted by
      `src/yb/tools/xcluster_verify.cc`.
- [x] `lint.sh --rev origin/master` clean.


balajisubramanian-yugabyte:
Consolidated into #33763, which now carries both commits. Closing this one to keep the docs change under a single review.

PR #33780, opened 2026-09-05
[docs] Clarify multi-cluster support is v2.25.2.0 or later

Sfurti-yb:
Clarifies that multi-cluster support applies to v2.25.2.0 and later releases, rather than reading as that specific patch only. The v2024.x backport versions are unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Sfurti-yb:
Closing — folding this into #33741 instead.

PR #33792, opened 2026-09-07
[DocDB][Vector Index] quantization support.

shaharuk-yb:


shaharuk-yb:
Superseded by #33806: identical change as a single commit off current master, with the unrelated `git-push.sh` tooling commit dropped. No reviews were pending here.

Issue #32967, opened 2026-07-29
[DocDB] Observability: per-drive write throughput and fsync latency, so slow storage is diagnosable


hkimura:
Jira Link: [DB-22687](https://yugabyte.atlassian.net/browse/DB-22687)
## Motivation

Slow or degraded storage under a tserver is one of the harder things to diagnose
with the metrics we ship today, and it shows up regularly: writes take seconds
instead of milliseconds, transaction status checks and other deadline-bound RPCs
time out, and the immediate question is whether YugabyteDB is overloading the
device or the device has gone bad underneath us. Answering that currently takes
days of correlating support bundles, OS-level metrics and log grepping, and it
often ends without a defensible number either way.

The reason it is hard is that we report how *long* our fsyncs take but never how
much data we handed to the device over the same window. A latency number with no
volume next to it is not interpretable: 500 ms of fsync is a broken device if we
pushed 64 KB through it and unremarkable if we pushed 64 MB. From a support
bundle we currently cannot tell those apart.

## What we can see today

- `log_sync_latency` (event stats, microseconds): time inside the WAL fsync.
  Note this is a **table**-level metric, so it aggregates every tablet of that
  table on the node, across whatever drives those tablets live on. It cannot be
  attributed to a device even in principle.
- `Fsync log took a long time: real 10.813s` log lines (WAL, threshold 50 ms,
  rate-limited to one per second) and `sync call for <file> took ...`
  (`LOG_SLOW_EXECUTION`, threshold 1000 ms) for other files.
- `log_bytes_logged` (tablet-level counter): bytes appended to the WAL, with no
  relationship to any fsync and not aggregated per drive.
- A thread-local `IOStatsContext` (`src/yb/util/stats/iostats_context.h`) that
  already counts `bytes_written`, `write_nanos`, `fsync_nanos` and
  `range_sync_nanos`. It is RocksDB-only, gated on perf level, and never
  exported as a server metric; today it only shows up in compaction/flush job
  log lines.
- A `drive` metric entity already exists (`src/yb/fs/fs_manager.cc`, keyed by a
  `drive_path` attribute) but carries only `drive_fault`.

## Why per-tablet or per-fsync accounting is the wrong shape

It is tempting to pair each fsync with the bytes that fsync flushed. That does
not measure what we need:

- `fsync(fd)` is per file descriptor, and the kernel has no notion of tablets;
  the tablet mapping is ours, derived from the path. So a per-tablet number is
[...]

Issue #31746, opened 2026-05-19
[DocDB] Unify filesystem abstractions between yb_util and rocksdb libraries


spolitov:
Jira Link: [DB-21472](https://yugabyte.atlassian.net/browse/DB-21472)
### Description

The yb_util and rocksdb libraries currently contain duplicate filesystem-related classes — parallel definitions of the same abstractions in `yb::` and `rocksdb::` namespaces. We should provide a unified set of filesystem-related classes in env, eliminating the duplication.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-21472]: https://yugabyte.atlassian.net/browse/DB-21472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33811, opened 2026-09-08
[BACKPORT 2026.1][#32967] DocDB: per-drive write throughput and fsync latency metrics

hkimura:
Summary:
We report how long our fsyncs take but never how much data went to the device over the same
window, and an fsync duration with no volume beside it is not interpretable: 500 ms is a broken
device for 64 KB and unremarkable for 64 MB. log_sync_latency does not answer it either, being
table-level and so aggregated across whatever drives a table's tablets happen to sit on.

Add per-drive write-IO accounting on the existing 'drive' metric entity. New
util/drive_io_stats.{h,cc} holds the counters, a process-global registry keyed by drive root, and
the metric definitions. The design is in drive_io_stats.h's file comment: the single choke point
(yb::PosixWritableFile, behind both the Raft WAL and everything RocksDB writes through the
YugabyteDB Env), why the unit is the drive rather than the tablet or file, the separate
buffered/direct entry points and the unsynced-bytes contract, and the util/fs layering. What each
counter means is in its METRIC_DEFINE description.

Sync durations and byte settlements are recorded only on success: a failed fsync restores the
unsynced debt and counts nothing, matching the failed-write path.

Slow-sync log lines now carry the pending byte count as well, so a support bundle with no metrics
in it is still actionable.

Fixed in passing, because this change makes it reachable: FsManagerOpts::metric_registry was a
raw MetricRegistry* with no initializer, so tools and unit tests left it holding garbage. Now
defaults to nullptr, with both call sites null-checked.

Out of scope: PosixRWFile, which is on neither the WAL nor the SST path, and a WAL-versus-SST
breakdown of the byte counter.

Test Plan:
util/drive_io_stats-test, 10 tests, new: path attribution, the unsynced-bytes contract across the
buffered, direct, and closed-without-syncing paths, and the export onto the 'drive' metric entity.
Its file comment has the layout and the two traps worth knowing before adding a case.

fs/fs_manager-test, 29 tests (3 new), covering which roots SetUpDriveIoMetrics registers: both WAL
and data roots as distinct drives, nothing at all with --export_drive_io_metrics off, and never a
root that was dropped for a fault.

All passing on fastdebug/clang21. No default behavior change on the write path beyond the
accounting itself: with --export_drive_io_metrics off nothing is registered, and every instrumented
site reduces to one clock read and a null check.

Not yet exercised: a real multi-drive cluster with the metrics scraped, and a dashboard panel or
alert rule expressed in throughput terms.

Original commit: f7ece6587aff4134987e28df2f8813b103e8a2b8 / D56396

Reviewers: huapeng.yuan, esheng, #db-approvers

Reviewed By: huapeng.yuan, esheng, #db-approvers

Subscribers: svc_phabricator, ybase

Differential Revision: https://phorge.dev.yugabyte.com/D56396

## Backport notes

Clean cherry-pick, no conflicts. 1082 insertions, identical to the original commit.

Rebased onto the now-landed dependency and squashed to a single commit, replacing the earlier
hand-ported version of this PR. That version predated the backport of #31746 (unify filesystem
abstractions) and had to duplicate the per-drive accounting into both writable-file classes, since
this branch still had two. With #31746 landed there is one merged `yb::PosixWritableFile` again, so
the original commit applies as-is and none of that adaptation is needed. Reviewer comments on the
previous revision are resolved by its removal rather than by individual fixes.

### Verification

fastdebug/clang19, this branch's pinned toolchain: `util/drive_io_stats-test` 11/11 and
`fs/fs_manager-test` 13/13 pass. `build-support/lint.sh` reports 0 errors, 0 warnings.

### DIFF
--- a/src/yb/util/file_system_posix.cc
+++ b/src/yb/util/file_system_posix.cc
@@ -32,6 +32,7 @@
 #include "yb/util/coding-inl.h"
 #include "yb/util/coding.h"
 #include "yb/util/debug/trace_event.h"
+#include "yb/util/drive_io_stats.h"
 #include "yb/util/errno.h"
 #include "yb/util/logging.h"
 #include "yb/util/malloc.h"
@@ -420,7 +421,10 @@ PosixWritableFile::PosixWritableFile(const std::string& fname, int fd,
       sync_on_close_(sync_on_close),
       filesize_(initial_size),
       pre_allocated_size_(0),
-      pending_sync_(false) {
+      pending_sync_(false),
+      // Files opened before their drive is registered (FsManager's own startup writes) go
+      // uncounted.
+      drive_stats_(DriveIoStatsRegistry::Instance().Find(fname)) {
 #ifdef ROCKSDB_FALLOCATE_PRESENT
   allow_fallocate_ = options.allow_fallocate;
   fallocate_with_keep_size_ = options.fallocate_with_keep_size;
@@ -430,6 +434,20 @@ PosixWritableFile::PosixWritableFile(const std::string& fname, int fd,
 PosixWritableFile::~PosixWritableFile() {
   if (fd_ >= 0) {
     WARN_NOT_OK(PosixWritableFile::Close(), "Failed to close " + filename_);
+  }
+  // Covers paths that abandon the file without a successful Close(); the exchange in
+  // ReleaseUnsyncedBytes makes the double call harmless.
+  ReleaseUnsyncedBytes();
+}
+
+void PosixWritableFile::ReleaseUnsyncedBytes() {
+  // No need to zero the per-file counter when the file is going away.
+  if (drive_stats_ == nullptr) {
+    return;
+  }
+  const auto residual = unsynced_bytes_.exchange(0, std::memory_order_relaxed);
+  if (residual != 0) {
+    drive_stats_->ReleaseUnsyncedBytes(residual);
   }
 }
 
@@ -540,6 +558,10 @@ Status PosixWritableFile::Close() {
     }
   }
   fd_ = -1;
+  // After the optional sync above, so a synced close has nothing left to release. Anything still
+  // pending here was never fsynced by us (sync_on_close_ defaults to false, and RocksDB's writer
+  // does not sync on close either) and must come off the drive gauge, or it stays there forever.
+  ReleaseUnsyncedBytes();
   return s;
 }
 
@@ -554,8 +576,14 @@ Status PosixWritableFile::Flush(FlushMode mode) {
   if (mode == FLUSH_SYNC) {
     flags |= SYNC_FILE_RANGE_WAIT_AFTER;
   }
-  if (sync_file_range(fd_, 0, 0, flags) < 0) {
+  const auto start = MonoTime::Now();
+  const auto rc = sync_file_range(fd_, 0, 0, flags);
+  if (rc < 0) {
+    // Failed operations go uncounted.
     return STATUS_IO_ERROR(filename_, errno);
+  }
+  if (drive_stats_) {
+    drive_stats_->RecordRangeSync(MonoTime::Now() - start);
   }
 #else
   if (fsync(fd_) < 0) {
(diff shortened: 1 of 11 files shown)


src/yb/util/file_system_posix.cc
```diff
@@ -361,24 +365,46 @@ PosixWritableFile::~PosixWritableFile() {
...
+  const auto residual = unsynced_bytes_.exchange(0, std::memory_order_relaxed);
+  if (residual != 0) {
+    drive_stats_->ReleaseUnsyncedBytes(residual);
+  }
 }
 
 Status PosixWritableFile::Append(const Slice& data) {
   const char* src = data.cdata();
   size_t left = data.size();
+  const auto start = yb::MonoTime::Now();
```
es1024: yb:: unnecessary

src/yb/util/file_system_posix.cc
```diff
@@ -436,26 +462,43 @@ Status PosixWritableFile::Close() {
...
+  ReleaseUnsyncedBytes();
   return s;
 }
 
 // write out the cached data to the OS cache
 Status PosixWritableFile::Flush() { return Status::OK(); }
 
 Status PosixWritableFile::Sync() {
+  const auto synced_bytes = unsynced_bytes_.exchange(0, std::memory_order_relaxed);
+  const auto start = yb::MonoTime::Now();
```
es1024: yb:: unnecessary

src/yb/util/file_system_posix.cc
```diff
@@ -436,26 +462,43 @@ Status PosixWritableFile::Close() {
...
   if (fsync(fd_) < 0) {
+    // A failed sync settles nothing: restore the unsynced debt and count nothing.
+    unsynced_bytes_.fetch_add(synced_bytes, std::memory_order_relaxed);
     return STATUS_IO_ERROR(filename_, errno);
   }
+  if (drive_stats_) {
+    drive_stats_->RecordSync(synced_bytes, yb::MonoTime::Now() - start);
+  }
```
es1024, line 636: Avoid unnecessary changes over original commit in backport.

src/yb/util/file_system_posix.cc
```diff
@@ -498,12 +541,16 @@ Status PosixWritableFile::Allocate(uint64_t offset, uint64_t len) {
...
+  const auto start = yb::MonoTime::Now();
+  const auto rc = sync_file_range(
+      fd_, static_cast<off_t>(offset), static_cast<off_t>(nbytes), SYNC_FILE_RANGE_WRITE);
+  if (rc != 0) {
     return STATUS_IO_ERROR(filename_, errno);
   }
+  if (drive_stats_) {
+    drive_stats_->RecordRangeSync(yb::MonoTime::Now() - start);
+  }
+  return Status::OK();
```
es1024: Avoid unnecessary changes over original commit in backport.

src/yb/util/file_system_posix.h
```diff
@@ -91,6 +94,37 @@ class PosixWritableFile : public WritableFile {
...
+  // claims to be (see the drive_bytes_unsynced description).
+  //
+  // Accessed with memory_order_relaxed, like the drive counters it feeds. It publishes no other
+  // memory, and every update is a read-modify-write on this one variable, so concurrent updates
+  // still compose correctly without any barrier.
+  std::atomic<uint64_t> unsynced_bytes_{0};
+
+  // Hands whatever this file still holds unsynced back to the drive gauge without counting a
+  // sync. Closing without syncing is the normal case, so without this the gauge only climbs.
+  void ReleaseUnsyncedBytes();
```
es1024: Avoid unnecessary changes over original commit in backport.

es1024 (changes requested):

src/yb/util/file_system_posix.cc
hkimura: Outdated as of squash

src/yb/util/file_system_posix.cc
hkimura: Outdated as of squash

src/yb/util/file_system_posix.cc
hkimura: Outdated as of squash

src/yb/util/file_system_posix.cc
hkimura: Outdated as of squash

es1024 (approved):

PR #33840, opened 2026-09-09
[PLAT-22308] YBA: Fail fast with a clear error when the master leader is missing

shubin-yb:
## Summary

`Universe.getMasterLeaderNode()` is documented as nullable — it returns `null` whenever `getMasterLeaderHostText()` yields `""`, either because no master leader is currently elected or because the `YBClient` lookup threw and was swallowed. Five call sites dereferenced the result with no null check, so a missing leader surfaced as an opaque `NullPointerException` instead of an actionable error.

The reported path is a Kubernetes universe edit:

`EditKubernetesUniverse` → `createWaitForDataMoveTask()` → `WaitForDataMove` → `verifyNoTabletsOnBlacklistedTservers()` → `UniverseTaskBase.dumpDbEntities()`

which builds a `HostAndPort` directly off the nullable node.

This adds `Universe.getMasterLeaderNodeOrThrow()`, which raises a `RuntimeException` naming the universe, and routes the five escaping sites through it:

| Call site | Previously dereferenced |
|---|---|
| `UniverseTaskBase.dumpDbEntities()` | `.cloudInfo.private_ip` |
| `YbcManager.getPreferenceBasedYBCNodeIPsList()` | `.cloudInfo.private_ip` |
| `ManageCatalogUpgradeSuperUser.run()` | passed to `getDataDirectoryPath` / `getNodeHomeDir` |
| `RunYsqlUpgrade.findAliveTServer()` | `.getRegion()` |
| `KubernetesCommandExecutor` (ysql major upgrade `PGPASSFILE` path) | passed to `getDataDirectoryPath` |

`ManageCatalogUpgradeSuperUser` also resolved the master leader a second time for the non-Kubernetes home directory — a redundant RPC that could return `null` independently of the first call. It now reuses the local.

Call sites that already handle `null` keep the nullable accessor: `HealthChecker`, `PGUpgradeTServerCheck` and `CommonUtils.getServerToRunYsqlQuery` all null-check and fall back. `TabletReportComponent` keeps it too — its dereference sits inside a `try`/`catch` that logs and continues, so the NPE never escapes there.

### Behavior change

Every path touched here already failed when the master leader was missing; they now fail with a message naming the universe instead of an NPE. No path that previously succeeded now fails.

Two of them use the leader only as a *preference* — `RunYsqlUpgrade.findAliveTServer()` uses it as a region hint, and `YbcManager.getPreferenceBasedYBCNodeIPsList()` as a first-choice backup node — so they could in principle degrade gracefully instead of throwing. Both previously raised an NPE, so failing fast is not a regression, but they are reasonable candidates for a follow-up that falls back to another node.

## Test plan

Two cases added to `UniverseTest`:

- `testGetMasterLeaderNodeOrThrowReturnsLeader` — resolves the correct node when `getLeaderMasterHostAndPort()` returns a known node's private IP.
- `testGetMasterLeaderNodeOrThrowWithoutLeader` — asserts `getMasterLeaderNode()` still returns `null` (nullable contract preserved) and that `getMasterLeaderNodeOrThrow()` throws a `RuntimeException` whose message names the universe.

- [x] `sbt "testOnly com.yugabyte.yw.models.UniverseTest"`, run on this branch rebased onto current master — `0 failed, 0 ignored, 46 total`, both new cases included.
- [x] `google-java-format` 1.17.0 — all 7 changed files clean.
- [x] `build-support/lint.sh --rev upstream/master` — 0 errors.

The rebase onto master resolved three conflicts, all upstream movement rather than semantic overlap:

- `ManageCatalogUpgradeSuperUser` — master added an early return when YSQL is disabled. Kept it, so the new throw is reached only when YSQL is actually enabled.
- `RunYsqlUpgrade` — master renamed the parameter type `YBClient` → `YBClientApi`. Kept master's signature.
- `UniverseTest` — master appended its own `getNodePrefixesForCustomer` / `findUniverseUuidsByNodePrefix` tests at the same point in the file. Kept both sets; the suite going from 38 to 46 tests reflects that.

### DIFF
--- a/managed/src/main/java/com/yugabyte/yw/models/Universe.java
+++ b/managed/src/main/java/com/yugabyte/yw/models/Universe.java
@@ -1256,7 +1256,7 @@ public class Universe extends Model {
   public NodeDetails getMasterLeaderNodeOrThrow() {
     NodeDetails masterLeaderNode = getMasterLeaderNode();
     if (masterLeaderNode == null) {
-      throw new RuntimeException(
+      throw new IllegalStateException(
           "Could not find the master leader node in universe " + getUniverseUUID());
     }
     return masterLeaderNode;
--- a/managed/src/main/java/com/yugabyte/yw/models/Universe.java
+++ b/managed/src/main/java/com/yugabyte/yw/models/Universe.java
@@ -1245,6 +1245,24 @@ public class Universe extends Model {
   }
 
   /**
+   * Find the current master leader node, failing loudly when it is missing. Prefer this over {@link
+   * #getMasterLeaderNode()} on paths that cannot make progress without a master leader, so that a
+   * missing leader surfaces as an actionable error instead of a NullPointerException.
+   *
+   * @return NodeDetails of the master leader, never null
+   * @throws RuntimeException if the universe has no reachable master leader
+   */
+  @JsonIgnore
+  public NodeDetails getMasterLeaderNodeOrThrow() {
+    NodeDetails masterLeaderNode = getMasterLeaderNode();
+    if (masterLeaderNode == null) {
+      throw new RuntimeException(
+          "Could not find the master leader node in universe " + getUniverseUUID());
+    }
+    return masterLeaderNode;
+  }
+
+  /**
    * Find the current master leader in the universe
    *
    * @return a String of the private_ip of the current master leader in the universe or an empty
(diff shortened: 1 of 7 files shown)


managed/src/main/java/com/yugabyte/yw/models/Universe.java
```diff
@@ -1244,6 +1244,24 @@ public NodeDetails getMasterLeaderNode() {
...
+   * missing leader surfaces as an actionable error instead of a NullPointerException.
+   *
+   * @return NodeDetails of the master leader, never null
+   * @throws RuntimeException if the universe has no reachable master leader
+   */
+  @JsonIgnore
+  public NodeDetails getMasterLeaderNodeOrThrow() {
+    NodeDetails masterLeaderNode = getMasterLeaderNode();
+    if (masterLeaderNode == null) {
+      throw new RuntimeException(
```
nkhogen, line 1259: nit: IllegalStateException

nkhogen (approved):
LGTM. Left a minor comment.

shahrooz1997 (approved):
LGTM.

Issue #33874, opened 2026-09-10
[DocDB] Automatic tablet split can be admitted while a replica add (remote bootstrap) of the same tablet is in flight; split children inherit the PRE_VOTER


hkimura:
Jira Link: [DB-23487](https://yugabyte.atlassian.net/browse/DB-23487)
### Description

The master is supposed to refuse an automatic split for a tablet that has a `PRE_VOTER` replica ([`CheckLiveReplicasForSplit`](https://github.com/yugabyte/yugabyte-db/blob/1d650612a168505d14c06ec09730dfa3a961d04e/src/yb/master/tablet_split_manager.cc#L563-L590), added for #8526). In practice the check is blind for almost the whole duration of a load-balancer-initiated remote bootstrap, because the master's replica map loses the `PRE_VOTER` member type while the destination is copying:

1. When the leader reports the committed config with the new peer, the master rebuilds the replica map from that config, and the destination's entry carries `member_type: PRE_VOTER` ([`UpdateTabletReplicasAfterConfigChange`](https://github.com/yugabyte/yugabyte-db/blob/1d650612a168505d14c06ec09730dfa3a961d04e/src/yb/master/master_heartbeat_service.cc#L1366-L1428)). At this point the check works.
2. The destination registers the copying peer and includes it in every incremental tablet report for the rest of the bootstrap ([`GenerateTabletReport`](https://github.com/yugabyte/yugabyte-db/blob/1d650612a168505d14c06ec09730dfa3a961d04e/src/yb/tserver/ts_tablet_manager.cc#L3225-L3230) re-adds all tablets being remote bootstrapped to the dirty set). The peer has no consensus yet, so the report has state `NOT_STARTED` and no `committed_consensus_state`.
[...]

Issue #33295, opened 2026-08-14
[DocDB] Log at INFO when an over-threshold tablet is rejected as an automatic split candidate


ttyusupov:
Jira Link: [DB-22973](https://yugabyte.atlassian.net/browse/DB-22973)
### Problem

Every rejection path in the automatic tablet split manager logs at `VLOG(3)` or `VLOG(4)`, so on a production master running the default verbosity there is no way to learn *why* a tablet that is far above its split threshold is not being split. The only INFO-level signal the subsystem emits is the positive one (`Scheduled split for tablet_id: ...`). Absence of that line is indistinguishable between "nothing is over threshold", "the candidate was rejected", and "the split manager is not running".

Concretely, in `TabletSplitManager::DoSplitting`:

- table-level rejections (`ValidateSplitCandidateTable`, `XReplValidateSplitCandidateTable`) → `VLOG(3) "Skipping table for splitting"`
- per-tablet rejections (`ValidateAutomaticSplitCandidateTablet`, i.e. `GetLeaderReplicaDriveInfo` / `ValidateSplitCandidateTablet` / `ShouldSplitValidCandidate` / `CheckLiveReplicasForSplit` / post-split-data check / `CanSplitMoreOnReplicas`) → `VLOG(4) "Should not split tablet $0. $1"`
- `table->GetTablets()` failure → a **bare `continue` with no log at any level**

### Proposal

1. **Log rejected over-threshold candidates at INFO, throttled.** In `DoSplitting`, collect rejected tablets whose leader SST size exceeds the applicable threshold (or simply `tablet_force_split_threshold_bytes`, which is unambiguous), and emit the top N by size with their rejection `Status` via `YB_LOG_EVERY_N_SECS(INFO, 900)`. This is the pattern already accepted for this code path in #31038 (throttled INFO for outstanding post-split compactions), so volume concerns are bounded and precedented.

2. **Cover the branches #31038 does not.** #31038 logs only the `AddCompactingSplit` branch, which requires the tablet to have a split parent. A **depth-0** tablet that fails the same post-split-data condition (or `CheckLiveReplicasForSplit`, or any other check) inside `ValidateAutomaticSplitCandidateTablet` is still dropped at `VLOG(4)` — the message text was improved there, but its visibility was not. The table-level `VLOG(3)` skip and the unlogged `GetTablets()` `continue` are likewise untouched.

### Issue Type

kind/enhancement

[...]

PR #33876, opened 2026-09-10
[#33874] DocDB: Refuse split while a replica is being remote bootstrapped

hkimura:
## Summary

`CheckLiveReplicasForSplit` is meant to keep the master from splitting a tablet while one of its replicas is being remote bootstrapped, but it only rejected replicas whose member type is `PRE_VOTER`. While a peer is being remote bootstrapped, the destination reports it without a consensus state, and the master records it with member type `UNKNOWN_MEMBER_TYPE` and state `NOT_STARTED` or `BOOTSTRAPPING`, so a tablet with RF running voters plus one bootstrapping replica passed the check. The children of such a split are created with the parent's committed Raft config, bootstrapping peer included, so each child then needs its own remote bootstrap, and the parent cannot be deleted until that bootstrap finishes.

The check now rejects any replica in `NOT_STARTED` or `BOOTSTRAPPING` state regardless of member type and requires every remaining non-read-replica to be a `RUNNING` voter. It runs from `ShouldSplitValidCandidate`, which is evaluated both when a candidate is picked and again in `DoSplitTablet` right before the children are registered, so a replica add that starts in between is caught too. This narrows the exposure from the whole bootstrap to the heartbeat lag between a config change committing on the leader and the master hearing about it; the complete fix is a Raft-side guard that refuses a `SPLIT_OP` while the committed config contains a non-voter, tracked in #33874 as a follow-up.

Until now the default `may_have_orphaned_post_split_data` of a never-reported replica happened to block these splits on the candidate path; D57694 removes that for non-voters, so this change should land before or together with it. A tablet whose non-voter never finishes bootstrapping now stays unsplittable, visible only at VLOG(4); #33295 covers logging such rejections at INFO.

`DoNotSplitTabletDoingRBS` is rewritten rather than joined by a new test: its previous body removed a follower before starting the bootstrap, so its refusal came from the under-replication branch and the bootstrapping replica was never what blocked the split. The new body keeps RF running voters, so on the previous check it fails.

## Test Plan

```
./yb_build.sh --cxx-test tablet-split-itest --gtest_filter AutomaticTabletSplitAddServerITest.DoNotSplitTabletDoingRBS
./yb_build.sh --cxx-test tablet-split-itest --gtest_filter 'AutomaticTabletSplitAddServerITest.*'
```

The rewritten test adds a fourth replica as `PRE_VOTER` with the bootstrap paused, waits for the destination's own heartbeat and for the original replicas to report `RUNNING` again, and checks that the split is refused because of the bootstrapping peer, then refused as over-replicated after promotion, then completes once a follower is removed. Verified that it fails against master without the fix (the check returns OK with the master's entry logged as `UNKNOWN_MEMBER_TYPE`, `NOT_STARTED`).

### DIFF
--- a/src/yb/integration-tests/tablet-split-itest.cc
+++ b/src/yb/integration-tests/tablet-split-itest.cc
@@ -2360,6 +2360,100 @@ TEST_F(AutomaticTabletSplitAddServerITest, DoNotSplitTabletDoingRBS) {
   ASSERT_OK(WaitForTabletSplitCompletion(2));
 }
 
+TEST_F(AutomaticTabletSplitAddServerITest, DoNotSplitFullyReplicatedTabletDoingRBS) {
+  // A tablet that already has RF running voters must not be split while a fourth replica is being
+  // remote bootstrapped as PRE_VOTER. While the copy is in progress the destination reports the
+  // peer without a consensus state, so the master records it with member type UNKNOWN_MEMBER_TYPE
+  // rather than PRE_VOTER, and the check must not depend on seeing PRE_VOTER. Unlike
+  // DoNotSplitTabletDoingRBS, no voter is removed first, so the under-replication path does not
+  // apply here.
+
+  CreateSingleTablet();
+  ASSERT_OK(WriteRows());
+
+  BuildTServerMap();
+
+  const auto catalog_mgr = ASSERT_RESULT(catalog_manager());
+  const auto tablet = ASSERT_RESULT(GetSingleTestTabletInfo(catalog_mgr));
+  const auto tablet_id = tablet->id();
+  const auto leader_idx = ASSERT_RESULT(GetLeaderIdx(tablet));
+  const auto leader_ts = cluster_->mini_tablet_server(leader_idx)->server();
+  const auto leader_id = leader_ts->permanent_uuid();
+  const auto follower_idx = (leader_idx + 1) % 3;
+  const auto follower_id = cluster_->mini_tablet_server(follower_idx)->server()->permanent_uuid();
+  LOG(INFO) << "Source tablet id " << tablet_id;
+
+  // Start RBS of a fourth replica but pause it before downloading the WAL.
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_pause_rbs_before_download_wal) = true;
+  AddTabletToNewTServer(tablet_id, leader_id, consensus::PeerMemberType::PRE_VOTER);
+
+  const auto new_ts_idx = cluster_->num_tablet_servers() - 1;
+  const auto new_ts_id = cluster_->mini_tablet_server(new_ts_idx)->server()->permanent_uuid();
+
+  // Wait for the destination's own heartbeat. From here on the replica is NOT_STARTED in the
+  // master's map and its member type is whatever the consensus-less report left there.
+  ASSERT_OK(itest::WaitUntilTabletInState(tablet, new_ts_id, tablet::NOT_STARTED));
+  {
+    const auto replicas = tablet->GetReplicaLocations();
+    const auto it = replicas->find(new_ts_id);
+    ASSERT_NE(it, replicas->end());
+    LOG(INFO) << "Bootstrapping replica as seen by the master: " << it->second.ToString();
+    ASSERT_NE(it->second.member_type, consensus::PeerMemberType::VOTER);
+  }
+
+  // The config change rebuilt the replica map, which leaves the original replicas in state UNKNOWN
+  // until each reports again. Wait for them, so that the refusal below can only be due to the
+  // bootstrapping replica and not to a voter that is momentarily "not running".
+  ASSERT_OK(WaitFor([&]() -> Result<bool> {
+    for (const auto& [ts_uuid, replica] : *tablet->GetReplicaLocations()) {
+      if (ts_uuid != new_ts_id &&
+          (replica.member_type != consensus::PeerMemberType::VOTER ||
+           replica.state != tablet::RaftGroupStatePB::RUNNING)) {
+        return false;
+      }
+    }
+    return true;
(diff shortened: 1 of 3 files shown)


src/yb/integration-tests/tablet-split-itest.cc
```diff
@@ -2360,6 +2360,100 @@ TEST_F(AutomaticTabletSplitAddServerITest, DoNotSplitTabletDoingRBS) {
...
+  ASSERT_EQ(itest::GetNumTabletsOfTableOnTS(leader_ts, table_->id()), 1);
+
+  // Let the RBS finish. The new peer is promoted to VOTER, which makes the tablet over-replicated,
+  // so still no split.
+  ANNOTATE_UNPROTECTED_WRITE(FLAGS_TEST_pause_rbs_before_download_wal) = false;
+  ASSERT_OK(WaitFor([&]() -> Result<bool> {
+    const auto replicas = tablet->GetReplicaLocations();
+    const auto it = replicas->find(new_ts_id);
+    return it != replicas->end() && it->second.member_type == consensus::PeerMemberType::VOTER &&
+           it->second.state == tablet::RaftGroupStatePB::RUNNING;
```
cuivienor: Could this wait require all four replicas to be `VOTER`/`RUNNING` after promotion?

`UpdateTabletReplicasAfterConfigChange` rebuilds the map for the promotion and initializes non-reporting peers to `UNKNOWN`. The new replica can therefore satisfy this wait while an original replica is still `UNKNOWN`; `CheckLiveReplicasForSplit` then returns "At least one tablet peer not running" instead of "over replicated", making the following assertion timing-dependent. The earlier wait for the original replicas already handles the same state reset.

cuivienor (commented):
One possible synchronization issue in the new regression test.

src/yb/master/catalog_manager.cc
basavaraj29, line 3111:
```
  auto replicas = tablet_info.GetReplicaLocations();
  RETURN_NOT_OK(CheckLiveReplicasForSplit(
      tablet_info.id(), *replicas,
      CatalogManagerUtil::GetReplicationFactor(table_replication_info)));
```

basavaraj29: this will avoid the duplication.

src/yb/master/catalog_manager.cc
```diff
@@ -3452,6 +3452,23 @@ Status CatalogManager::DoSplitTablet(
...
+      status = CheckLiveReplicasForSplit(
+          source_tablet_info->tablet_id(), *source_tablet_info->GetReplicaLocations(),
+          CatalogManagerUtil::GetReplicationFactor(replication_info));
+      if (!status.ok()) {
+        return STATUS_FORMAT(
+            InvalidArgument,
+            "Tablet split candidate $0 is no longer a valid split candidate: $1",
+            source_tablet_info->tablet_id(),
+            status);
+      }
```
basavaraj29: can remove this

src/yb/master/tablet_split_manager.cc
basavaraj29: can remove

src/yb/master/tablet_split_manager.cc
```diff
@@ -554,33 +554,55 @@ std::unordered_set<TabletServerId> GetReplicasWithOutstandingCompaction(
...
+      continue;
+    }
+    if (replica.state != tablet::RaftGroupStatePB::RUNNING) {
+      return STATUS_FORMAT(NotSupported,
+                           "At least one tablet peer not running, "
+                           "tablet_id: $0, peer_uuid: $1, member type: $2, current RAFT state: $3",
+                           tablet_id, ts_uuid,
+                           consensus::PeerMemberType_Name(replica.member_type),
+                           RaftGroupStatePB_Name(replica.state));
+    }
```
basavaraj29: this covers the `IsStarting()` case too. may be get rid of the explicit `IsStarting` case since the error here is not much different from the error there?

src/yb/integration-tests/tablet-split-itest.cc
hkimura: Right, fixed: the wait now requires all four replicas to be running voters before the over-replication check.

src/yb/master/catalog_manager.cc
hkimura: Done. The check now lives in ShouldSplitValidCandidate right after the replication info is computed, so both the candidate path and the re-validation in DoSplitTablet get it.

src/yb/master/catalog_manager.cc
hkimura: Removed.

src/yb/master/tablet_split_manager.cc
hkimura: Removed, along with the now-unused replication_factor.

src/yb/master/tablet_split_manager.cc
hkimura: Merged the two into one condition with one message. I kept the IsStarting part in front of the read-replica skip on purpose: a bootstrap in progress should block whatever the member type is, while a read replica in any other non-running state (its tserver down, or UNKNOWN right after a config change) should stay ignored as before, otherwise a dead read-replica node would block splits indefinitely.

src/yb/master/tablet_split_manager.cc
```diff
@@ -910,7 +928,6 @@ void TabletSplitManager::DoSplitting(
                                        << StatusToString(replication_info.status());
       continue;
     }
```
basavaraj29: we don't use this here anymore. can we move it to `ShouldSplitValidCandidate`?

basavaraj29 (approved):
one minor comment

src/yb/master/tablet_split_manager.cc
hkimura: Removed. ShouldSplitValidCandidate already fetches the replication info per tablet, so the per-table fetch and its skip were dead.

Issue #33767, opened 2026-09-04
[DocDB] SST statistics: per-tablet in-memory aggregate with coverage counter


hkimura:
Jira Link: [DB-23376](https://yugabyte.atlassian.net/browse/DB-23376)

Follow-through on the SST statistics collector (#33267, PRs #33705/#33706 merged, #33709/#33710 in review): the collector stores per-file garbage statistics in each SST's properties block, but nothing aggregates them per tablet yet.

Add a per-tablet in-memory aggregate that sums the additive scalars (entries, tombstones, chain/subdoc/row counts, dead rows, reclaimable entries and bytes, droppable-age bands) across the tablet's live files:

- Maintained from `RegularRocksDbListener`: add the new file's stats on flush and compaction outputs, subtract compaction inputs when present in the event.
- Periodic resync from `DB::GetPropertiesOfAllTables`, merging by hand via `SstStatsFromProperties` (`TableProperties::Add` drops `user_collected_properties`, so the built-in aggregation cannot be used).
- A coverage counter: live files that predate the collector (no `yb.docdb.collector_version` property). A ratio computed over uncovered files silently reads near zero, so every consumer needs this alongside the sums.
- Expose the aggregate on the tablet status page (raw numbers; histogram rendering is tracked separately).

This is item A1 of the collector follow-through plan; the Prometheus gauges and the full-compaction trigger clause read this aggregate.

[DB-23376]: https://yugabyte.atlassian.net/browse/DB-23376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33887, opened 2026-09-11
[#33767] DocDB: per-tablet in-memory aggregate for SST statistics

balajisubramanian-yugabyte:
## Summary

The SST statistics collector (#33267) stores per-file garbage statistics in each SST's properties block, but nothing aggregates them. A per-file record only answers questions about one SST; the consumers that follow — the `docdb_sst_*` Prometheus gauges and the full-compaction trigger clause — need the tablet-wide number.

`SstStatsAggregator` sums the additive scalars over one tablet's live files: entries, tombstones, chain/subdoc/row counts, dead rows, reclaimable entries and bytes, and the droppable-age bands.

It is maintained two ways, and both are needed. The tablet's `RegularRocksDbListener` adds flush and compaction outputs and subtracts compaction inputs, so a full compaction that reclaims the garbage is visible to the trigger at once rather than a resync interval later. A periodic resync from `DB::GetPropertiesOfAllTables` catches the file-set changes that arrive with no event at all: DB open, remote bootstrap, snapshot restore, split inheritance. The resync merges by hand because `TableProperties::Add` drops `user_collected_properties`, and it discards its own result if a listener event landed while it was running, so a slow resync cannot clobber newer incremental state.

The distributions are deliberately left out. Bucket-wise merging is exact, but a shrinking file set needs subtraction, and five resident 145-bucket vectors cost ~5.8 KB per tablet against ~250 bytes for the scalars. Tablet-level distributions are built on demand instead.

The aggregate also has to be honest about what it does not know, or a consumer cannot tell "no garbage" from "no data" — a ratio computed over files the collector never saw silently reads near zero. `uncovered_files` counts live files written before the collector existed (no `yb.docdb.collector_version` property), `partial_files` those with an incomplete record, and `unsubtracted_files` signals that a file left the set but its stats could not be subtracted, leaving the aggregate over-counted until the next resync.

The raw numbers are exposed on the per-tablet status page. Nothing consumes the aggregate programmatically yet; the gauges and the trigger clause follow separately.

### What to act on

- Gated on the existing `docdb_enable_sst_stats_collector`, still **default off**. With the flag off the aggregate is never constructed and the poller sits idle, so this is inert on an upgraded cluster until the flag is flipped.
- New `docdb_sst_stats_resync_interval_sec` (`NON_RUNTIME`, default 300) sets the resync cadence.

## Test plan

- [x] `sst_stats_aggregator-test` (8 cases) — aggregate arithmetic including saturating subtraction; per-file parsing of covered, uncovered and partial files; flush/compaction bookkeeping including a failed compaction and a compaction of an uncounted file; the resync-overtaken-by-flush race.
- [x] `compaction-test --gtest_filter='SstStatsAggregateTest.*:SstStatsCoverageTest.*'` (2 cases) — the incrementally maintained aggregate checked against a recount of the live files after flushes and a manual compaction; `uncovered_files` for files written before the collector was enabled, and their transition to covered after a later compaction.

Both suites were green on the collector branch this change was originally stacked on. Now that #33709/#33710 have merged, the branch was rebased onto `master` (conflict-free) and both suites are being re-run against it; the rebase also required a thirdparty bump, so treat the two runs above as pending re-confirmation on master.


src/yb/tablet/tablet.cc
```diff
@@ -1257,6 +1265,8 @@ Status Tablet::OpenRegularDB(const rocksdb::Options& common_options) {
   if (FLAGS_docdb_enable_sst_stats_collector) {
     regular_rocksdb_options.table_properties_collector_factories.push_back(
         docdb::MakeSstStatsCollectorFactory());
+    // Before DB::Open, so that the listener installed below always sees it.
+    sst_stats_ = std::make_unique<docdb::SstStatsAggregator>();
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Good catch. Truncate and snapshot restore can replace the aggregator while the status page still holds its raw pointer.

I changed sst_stats_ to a mutex-protected shared_ptr, and sst_stats() now returns a shared copy. This keeps the old aggregator alive until an existing reader finishes. A concurrent reopen may produce one stale status-page reading, but it can no longer cause a use-after-free.

src/yb/tablet/tablet.cc
```diff
@@ -4837,6 +4847,23 @@ uint64_t Tablet::GetCurrentVersionNumSSTFiles() const {
   }, 0);
 }
 
+Status Tablet::ResyncSstStats() {
+  if (!sst_stats_) {
+    return Status::OK();
+  }
+  return sst_stats_->Resync(
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. The scoped RocksDB operation inside the snapshot callback protects access to regular_db_, but it does not protect the aggregator before or after that callback.

ResyncSstStats now holds a shared_ptr to the aggregator for the entire resync. If a truncate or snapshot restore replaces it concurrently, the old resync completes safely into the old instance, which is then discarded. I kept the RocksDB operation scoped to the callback because it only needs to protect the potentially blocking database reads.

src/yb/docdb/properties_collector/sst_stats_aggregator-test.cc
```diff
@@ -0,0 +1,286 @@
...
+  EXPECT_EQ(sum.total_entries, 160);
+  EXPECT_EQ(sum.reclaimable_entries, 50);
+  EXPECT_EQ(sum.droppable_age_entries[3], 50);
+  EXPECT_EQ(sum.covered_files, 2);
+
+  sum -= two;
+  EXPECT_EQ(sum, one);
+
+  // Saturating: a subtraction that was never matched by an addition must not wrap.
+  sum -= two;
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. The test deliberately exercises underflow, so DCHECK_GE makes it abort in debug builds.

I removed the assertion and retained saturating subtraction. Underflow can happen even when file tracking is correc. for example, if properties parse when a file is added but cannot be read when it is removed, its two calculated contributions differ. Saturating prevents that from becoming a value near 2^64. The header comment now explains this case.

src/yb/tserver/ts_tablet_manager.cc
```diff
@@ -499,6 +509,21 @@ void TSTabletManager::VerifyTabletData() {
...
 }
 
+void TSTabletManager::ResyncSstStats() {
+  for (const TabletPeerPtr& peer : GetTabletPeers()) {
+    auto tablet = peer->shared_tablet_maybe_null();
+    if (!tablet) {
+      continue;
+    }
+    // Expected to fail on a tablet that starts shutting down mid-pass; the next pass covers it.
+    const auto status = tablet->ResyncSstStats();
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. A full sweep can perform disk reads across every tablet and should not occupy a messenger IO thread.

The poller now only submits work to a dedicated single-thread sst-stats-resync pool. The pool exists only when the collector is enabled. An atomic guard skips a scheduled pass if the previous pass is still running, preventing a backlog of sweeps. Shutdown stops the poller before waiting for the pool.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
```diff
@@ -0,0 +1,215 @@
...
+  result.droppable_age_bytes = stats->droppable_age_bytes;
+  result.covered_files = 1;
+  result.covered_raw_bytes = raw_bytes;
+  result.partial_files = stats->chain_valid ? 0 : 1;
+  return result;
+}
+
+void SstStatsAggregator::AddFile(
+    uint64_t file_number, const rocksdb::TableProperties& properties) {
+  ++event_seqno_;
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. A replayed flush for an already-counted file does not change the aggregate, so it should not advance event_seqno_.

I moved the sequence-number increment after the successful insert. applied the same correction to removal of an already-absent file. Also added ResyncSurvivesReplayedFlush, which verifies that a duplicate event no longer causes a valid resync result to be discarded.

src/yb/docdb/properties_collector/sst_stats_aggregator.h
```diff
@@ -0,0 +1,163 @@
...
+  // not hold over the sum.
+  uint64_t partial_files = 0;
+  // Files dropped from the aggregate without their statistics, because the compaction event that
+  // consumed them did not carry their properties. The sums stay high by those files' contribution
+  // until the next resync clears the count.
+  uint64_t unsubtracted_files = 0;
+
+  // Identities over the chain-tracked population, valid only while partial_files is zero. See
+  // SstStats for what each measures.
+  uint64_t shadowed_entries() const { return chain_entries - num_subdoc_keys; }
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. Saturating aggregate fields independently can temporarily break the usual Ec >= K >= R ordering. I changed all three derived accessors to saturating subtraction and added a test covering the broken-order case.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
```diff
@@ -0,0 +1,214 @@
...
+  SstStatsAggregate aggregate;
+  std::unordered_set<uint64_t> counted_files;
+  counted_files.reserve(live_files.size());
+  for (const auto& file : live_files) {
+    counted_files.insert(file.name_id);
+    const auto* file_properties = Lookup(properties, file.BaseFilePath());
+    if (file_properties == nullptr) {
+      // Properties that could not be read at all: still count the file, so that a consumer sees
+      // the tablet is not fully measured. Its entries and bytes are simply unknown.
+      ++aggregate.uncovered_files;
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. I am changing the production snapshot path to tolerate a per-file properties-read failure: unreadable files remain in the live-file list but are omitted from the properties map, so this branch counts them as uncovered and the first resync can still complete. Existing RocksDB callers will retain fail-fast behavior.

src/yb/tablet/tablet.cc
```diff
@@ -4837,6 +4851,31 @@ uint64_t Tablet::GetCurrentVersionNumSSTFiles() const {
...
+  return sst_stats->Resync(
+      [this](std::vector<rocksdb::LiveFileMetaData>* live_files,
+             rocksdb::TablePropertiesCollection* properties) -> Status {
+        // Reading a properties block can hit disk, so this must not hold component_lock_ or block
+        // the start of a RocksDB shutdown.
+        auto scoped_operation = CreateScopedRWOperationNotBlockingRocksDbShutdownStart();
+        RETURN_NOT_OK(scoped_operation);
+        SCHECK(regular_db_, IllegalState, "No regular DB to read SST statistics from");
+        regular_db_->GetLiveFilesMetaData(live_files);
+        return regular_db_->GetPropertiesOfAllTables(properties);
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. I am adding an explicit per-file error-handling mode to `GetPropertiesOfAllTables` and using its skip mode only from `Tablet::ResyncSstStats`. The aggregator can then install the readable files, count each omitted live file in `uncovered_files`, and clear stale incremental state instead of remaining unusable after one properties-read failure.

src/yb/tablet/tablet.cc
```diff
@@ -1257,6 +1265,12 @@ Status Tablet::OpenRegularDB(const rocksdb::Options& common_options) {
   if (FLAGS_docdb_enable_sst_stats_collector) {
     regular_rocksdb_options.table_properties_collector_factories.push_back(
         docdb::MakeSstStatsCollectorFactory());
+    // Before DB::Open, so that the listener installed below always sees it. A reopen of a live
+    // tablet (truncate, snapshot restore) replaces the previous aggregator, which stays alive for
+    // as long as any reader still holds it.
+    auto sst_stats = std::make_shared<docdb::SstStatsAggregator>();
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed. I am clearing the tablet's current `sst_stats_` in `CompleteShutdownStorages` alongside the regular DB. Existing readers remain safe because they hold shared ownership of the old instance; new readers see no aggregate until `OpenRegularDB` installs one for the replacement DB.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
```diff
@@ -0,0 +1,214 @@
...
+  RETURN_NOT_OK(snapshot(&live_files, &properties));
+
+  SstStatsAggregate aggregate;
+  std::unordered_set<uint64_t> counted_files;
+  counted_files.reserve(live_files.size());
+  for (const auto& file : live_files) {
+    counted_files.insert(file.name_id);
+    const auto* file_properties = Lookup(properties, file.BaseFilePath());
+    if (file_properties == nullptr) {
+      // Properties that could not be read at all: still count the file, so that a consumer sees
```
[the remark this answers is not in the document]
balajisubramanian-yugabyte: Agreed with the current-code analysis. This is being addressed together with the all-or-nothing findings: `Tablet::ResyncSstStats` will request per-file error tolerance, making an unreadable properties block a real reason for a live file to be absent from the map. I am also clarifying that such a file contributes to `uncovered_files`, while its entry and raw-byte counts are unknown.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
```diff
@@ -0,0 +1,214 @@
...
+Status SstStatsAggregator::Resync(const SnapshotFn& snapshot) {
+  uint64_t seqno_before;
+  {
+    std::lock_guard lock(mutex_);
+    seqno_before = event_seqno_;
+  }
+
+  std::vector<rocksdb::LiveFileMetaData> live_files;
+  rocksdb::TablePropertiesCollection properties;
+  RETURN_NOT_OK(snapshot(&live_files, &properties));
```
hkimura, line 181: One thought on the steady-state cost of the sweep. `Version::GetTableProperties` serves a file from the table cache only if it is already resident; otherwise it opens the file and reads the properties block directly, and does not insert it into the cache afterwards (version_set.cc:608-650). So on a tserver whose table cache does not hold every SST, each pass re-reads the footer and properties block of every non-resident file of every tablet, every interval, even when nothing changed.

Since the resync exists to catch file-set changes the listener never saw, it could probably skip the properties read when there is nothing to catch: compare the live file numbers from `GetLiveFilesMetaData` (in-memory) against `counted_files_`, and if they match and `unsubtracted_files == 0`, just refresh `last_resync_micros` and return. A counted file's contribution never changes after it is written, so an equal set means the aggregate is already exact. The first pass after open still does the full read because `counted_files_` starts empty.

It would mean splitting the snapshot into a live-files step and a properties step, so this is fine as a follow-up if you would rather keep this PR as is.

src/yb/docdb/properties_collector/sst_stats_aggregator.h
```diff
@@ -0,0 +1,163 @@
...
+  void OnCompactionCompleted(const rocksdb::CompactionJobInfo& info);
+
+  // Replaces the aggregate with the sum over the whole live file set. `snapshot` fills the live
+  // file list and the properties collection keyed by base file path, as
+  // DB::GetLiveFilesMetaData / DB::GetPropertiesOfAllTables return them; it runs without the lock
+  // held because reading a properties block can hit disk.
+  //
+  // The snapshot is dropped rather than installed if a flush or compaction landed while it ran:
+  // it predates that event, so installing it would undo an update the incremental path already
+  // applied exactly. A tablet busy enough to skip every resync is one whose file set only ever
```
hkimura, line 140: The justification here holds once the listener has seen every live file, but right after open it is the other way around: a tablet that came up with many inherited files and is also flushing or compacting steadily is exactly the one whose resync gets overtaken, and each drop leaves `last_resync_micros` at zero for another whole interval.

Probably rare in practice since the snapshot is short, but a bounded immediate retry on an overtaken snapshot (a couple of attempts inside `Resync`, or a distinguishable status so the sweep can revisit the tablet at the end of the pass) would remove the dependence on timing, and this comment could then be simplified.

src/yb/tablet/tablet.cc
```diff
@@ -1257,6 +1265,12 @@ Status Tablet::OpenRegularDB(const rocksdb::Options& common_options) {
...
+    std::lock_guard lock(sst_stats_mutex_);
+    sst_stats_ = std::move(sst_stats);
```
hkimura, line 1273: Since this handoff is what closes the use-after-free that came up earlier in the review, it might be worth a test that exercises the reopen path: for example, truncate a tablet in the compaction test and check that a resync afterwards installs an aggregate over the new, empty file set, and that `sst_stats()` no longer returns the pre-truncate instance. Not blocking, just so a later refactor of `OpenRegularDB` cannot quietly drop it.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
```diff
@@ -0,0 +1,214 @@
...
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied.  See the License for the specific language governing permissions and limitations
+// under the License.
+//
+
+#include "yb/docdb/properties_collector/sst_stats_aggregator.h"
+
+#include "yb/docdb/properties_collector/sst_stats_collector.h"
+
+#include <algorithm>
```
hkimura, line 18: Nit: Should this go to Line 16? I think the convention (assuming we follow the Google C++ style) is
1. ".h of the component itself", in this case `"yb/docdb/properties_collector/sst_stats_aggregator.h"`.
2. algorithm etc without the "" alphabetically ordered.
3. other ".h" alphabetically sorted.

hkimura (approved):
Only nit comments, so LGTM.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
balajisubramanian-yugabyte: Good point.  Resync now reads the live-file list first and skips GetPropertiesOfAllTables when the counted file set matches and there are no incomplete reads. The first resync still reads all properties. I added tests for both the fast path and recovery after an unreadable properties block.

src/yb/docdb/properties_collector/sst_stats_aggregator.h
balajisubramanian-yugabyte: Added one immediate retry when a flush or compaction overtakes the snapshot. If the retry is also overtaken, the snapshot is dropped until the next scheduled pass. The unit test now verifies the retry installs a complete snapshot.

src/yb/tablet/tablet.cc
balajisubramanian-yugabyte: Sure. Added a compaction integration test that truncates a tablet and verifies:  sst_stats()  that returns a new aggregator instance. Resync installs an empty aggregate for the reopened DB. New flushes update only the replacement aggregator. A retained shared_ptr to the old aggregator remains safe.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
balajisubramanian-yugabyte: sure. Fixed. The component header is first, followed by <algorithm>, then project headers.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
```diff
@@ -0,0 +1,241 @@
...
+      continue;
+    }
+    aggregate_ = aggregate;
+    counted_files_ = std::move(counted_files);
+    properties_incomplete_ = properties_incomplete;
+    last_resync_micros_ = GetCurrentTimeMicros();
+    return Status::OK();
+  }
+
+  VLOG(1) << "Dropping SST statistics resync overtaken twice by a flush or compaction";
```
hkimura: On a hot tablet with a few thousand files, the window from `seqno_before` to the final check spans
a properties read per file. Any flush or compaction in that window discards the snapshot. With two
attempts per interval, such a tablet may never get `last_resync_micros != 0`, and by the header's
contract nothing reads its aggregate. VLOG(1) keeps that invisible.

`AddFile` and `RemoveFile` are already idempotent by file number, so the discard may not be needed.
Journal the events that land during a resync (file number, contribution, add or remove), install
the snapshot unconditionally, and replay the journal on top:

- an install that predates the live-file read is in the snapshot, and the replay no-ops;
- an install that postdates it is missing from the snapshot, and the replay applies it.

Both come out exact. The retry loop and `kMaxAttempts` go away, and the journal lives only for one
resync. The mixed case (properties newer than the live list) still lands on
`properties_incomplete_` as today.

If you prefer the current shape, a throttled WARNING while a tablet has never completed a resync
would at least make the starvation visible.

src/yb/tserver/ts_tablet_manager.cc
```diff
@@ -1014,6 +1067,11 @@ Status TSTabletManager::Start() {
   StartScheduledTask(
       data_size_metric_updater_.get(), "Data size metric updater",
       FLAGS_data_size_metric_updater_interval_sec * 1s);
+  if (sst_stats_resync_poller_) {
+    StartScheduledTask(
```
hkimura, line 1072: `Poller::Schedule` uses the interval for the first tick too. After every tserver restart, every
aggregate is unreadable for five minutes, and a tablet opened later (remote bootstrap, split child)
waits up to that long. Likely fine for the trigger given its cooldown, so this is a question: is a
resync from `OpenRegularDB`, or a short first delay, worth it? The table readers are already open
right after DB open, so that pass is the cheap one. If not, one sentence in the flag description
saying the aggregate is blind for one interval after open would save the next reader the question.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
```diff
@@ -0,0 +1,241 @@
...
+  aggregate_ -= SstFileContribution(*properties);
+}
+
+void SstStatsAggregator::OnFlushCompleted(const rocksdb::FlushJobInfo& info) {
+  const auto file_number = FileNumber(info.file_path);
+  if (file_number == 0) {
+    return;
+  }
+  std::lock_guard lock(mutex_);
+  AddFile(file_number, info.table_properties);
```
hkimura: `FlushJob::WriteLevel0Table` drops a zero-size output from the manifest but still returns OK, and
`NotifyOnFlushCompleted` fires with that file number. The aggregator then counts a file that is not
live: `covered_files` off by one, a phantom in `counted_files_`, and the next resync forced down the
full path. It self-heals at that resync. Skipping the event when `num_entries == 0` looks safe,
since a live file cannot have zero entries. Empty flushes are rare on the regular DB, so leaving it
to the resync is also reasonable.

src/yb/integration-tests/compaction-test.cc
```diff
@@ -541,6 +544,140 @@ TEST_F(CompactionTest, ManualCompactionProducesOneFilePerDb) {
...
+  }
+
+  // Asserts that what the listener accumulated matches what a read of the whole live file set
+  // computes, and returns the aggregate.
+  docdb::SstStatsAggregate CheckAgainstLiveFiles(const tablet::TabletPtr& tablet) {
+    const auto stats = tablet->sst_stats();
+    EXPECT_NE(stats, nullptr);
+    const auto from_events = stats->Get().aggregate;
+    EXPECT_OK(tablet->ResyncSstStats());
+    EXPECT_EQ(stats->Get().aggregate, from_events);
```
hkimura: Nit: `ResyncSstStats` returns OK when it discarded an overtaken snapshot, and then the aggregate is
untouched, so this equality holds without a resync having happened. The workload is stopped at
these call sites, so it will not misfire today. Capturing `last_resync_micros` before the call and
expecting it to advance would make the "listener == whole file set" claim self-evident.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
balajisubramanian-yugabyte: Went with the journal. Events that land during a resync are saved and replayed on the new snapshot, matched by file number, so a duplicate is a no-op and the retry loop is gone.

src/yb/tserver/ts_tablet_manager.cc
balajisubramanian-yugabyte: Ok, thats a good point. Documented it rather than resyncing at open since dont want disk reads on open path.

src/yb/docdb/properties_collector/sst_stats_aggregator.cc
balajisubramanian-yugabyte: Skipped now. num_entries == 0 is RocksDB's own condition: BuildTable only fills properties for a non-empty output. Compaction outputs come from the manifest edit, so they need no guard. Test: IgnoresEmptyFlushOutput.

src/yb/integration-tests/compaction-test.cc
balajisubramanian-yugabyte: Fixed and good catch. the helper now requires last_resync_micros to advance, so the comparison only runs after a resync completed.

Issue #33932, opened 2026-09-14
[DocDB] XClusterDDLReplicationTest.CheckpointMultipleDatabases fails on many-core hosts


ellabaron-code:
Jira Link: [DB-23531](https://yugabyte.atlassian.net/browse/DB-23531)
## Description

`XClusterDDLReplicationTest.CheckpointMultipleDatabases` creates `NumCPUs() * 2` databases and checkpoints them all in one replication group; checkpointing sets up the xCluster DDL replication extension in every database, so the database count is also the number of concurrent DDLs. On many-core hosts (176 cores → 352 databases) the test fails.

With `ysql_enable_concurrent_ddl` on by default (#29490), hundreds of concurrent extension-setup DDLs saturate the tserver service queue. Transaction status heartbeats are among the dropped RPCs, so the coordinator expires the DDL transactions and extension setup fails. The lock-waiting DDLs also fire a storm of deadlock-detection probes that adds to the RPC volume, but disabling deadlock detection does not fix the test — the queue overflow alone still expires the transactions. Even surviving runs exceed the 600s test framework budget on slower cores.

Fix (PR incoming): cap the database count at 64, keeping the test's multi-database purpose while bounding DDL concurrency.

## Issue Type

kind/failing-test

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23531]: https://yugabyte.atlassian.net/browse/DB-23531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #29490, opened 2025-11-26
[YSQL] Switch on yb_enable_concurrent_ddl by default


pkj415:
Jira Link: [DB-19293](https://yugabyte.atlassian.net/browse/DB-19293)
### Description

Set yb_fallback_to_legacy_catalog_read_time off by default

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-19293]: https://yugabyte.atlassian.net/browse/DB-19293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #26617, opened 2025-04-01
[DocDB][DDL Replication] create_xcluster_checkpoint cmd fails with Timed out waiting for Table Creation if there are multiple DBs


Arjun-yb:
Jira Link: [DB-15994](https://yugabyte.atlassian.net/browse/DB-15994)
### Description

Version: `2.25.2.0-b248`

**Steps**:
Create 3 databases at both sides
Creates 2 tables per DB(3*2 = 6 tables)
Then execute `create_xcluster_checkpoint` for those 3 DBs at source

cmd failed with
`AssertionError: Failed to setup replication: ["Waiting for checkpointing of database(s) to complete\n", "\n", "Error running create_xcluster_checkpoint: Internal error (yb/master/xcluster/xcluster_outbound_replication_group.cc:338): Failed to checkpoint namespace 00004002000030008000000000000000: Timed out (yb/util/backoff_waiter.cc:53): Timed out waiting for Table Creation\n"]`

More details are in the respective JIRA ticket

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-15994]: https://yugabyte.atlassian.net/browse/DB-15994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33933, opened 2026-09-14
[#33932] xCluster: Cap CheckpointMultipleDatabases at 64 databases

ellabaron-code:
## Summary

The test creates `NumCPUs() * 2` databases and checkpoints them all in one replication
group. Checkpointing sets up the xCluster DDL replication extension in every database, so
the database count is also the number of concurrent DDLs — on a 176-core host that is 352.

Since `ysql_enable_concurrent_ddl` became the default (#29490), each extension setup runs as
a concurrent DDL, and hundreds of them at once saturate the tserver service queue.
Transaction status heartbeats are among the dropped RPCs, so the coordinator expires the
DDL transactions and extension setup fails, asserting the test. The lock-waiting DDLs also
fire a storm of deadlock-detection probes that adds to the RPC volume, but disabling
deadlock detection is not a fix: the queue overflow alone still expires the transactions
and produces the same failure. And even when a run survives, this many databases pushes
the runtime past the 600s test framework budget on slower cores.

Capping the count at 64 keeps the test's purpose — checkpointing many databases in one
group, which is what the original regression (#26617) required — while bounding the DDL
concurrency on large machines. Hosts with fewer than 32 cores are unaffected.

## Test plan

- [x] Ran `XClusterDDLReplicationTest.CheckpointMultipleDatabases` (release) on a 176-core host: timed out at `NumCPUs() * 2` databases, passes with the cap.


src/yb/integration-tests/xcluster/xcluster_ddl_replication-test.cc
```diff
@@ -309,7 +309,9 @@ TEST_F(XClusterDDLReplicationTest, CheckpointMultipleDatabases) {
   ASSERT_OK(SetUpClusters());
 
   std::vector<NamespaceName> namespaces{namespace_name};
-  for (int i = 0; i < base::NumCPUs() * 2; i++) {
+  // More DBs than cores, but capped: too many concurrent DDLs overflow the TS service queue.
+  const int num_dbs = std::min(base::NumCPUs() * 2, 64);
```
neera-mital, line 313: nit: The core count dependency is now vestigial. The original test used two times the core count deliberately, because the async task pool was capped at core count and the test needed to overflow it. That cap was removed in all build types, and the pool is now unlimited with a warning comment pointing at issue 26617. So there is no longer a reason for the database count to track cores at all.  This is a suggestion, not a blocker.

src/yb/integration-tests/xcluster/xcluster_ddl_replication-test.cc
```diff
@@ -309,7 +309,9 @@ TEST_F(XClusterDDLReplicationTest, CheckpointMultipleDatabases) {
   ASSERT_OK(SetUpClusters());
 
   std::vector<NamespaceName> namespaces{namespace_name};
-  for (int i = 0; i < base::NumCPUs() * 2; i++) {
+  // More DBs than cores, but capped: too many concurrent DDLs overflow the TS service queue.
```
neera-mital, line 312: nit:Enough DBs to checkpoint many at once, capped so concurrent extension-setup DDLs do not overflow the tserver service queue (#33932)

neera-mital (approved):

PR #33973, opened 2026-09-15
Pg19 master catchup c25b05a

fizaaluthra:


src/postgres/src/backend/utils/activity/pgstat_io.c
```diff
@@ -388,6 +388,8 @@ pgstat_tracks_io_bktype(BackendType bktype)
 		case YB_MATVIEW_REFRESH_DDL:
 		case YB_RELCACHE_INIT_BACKEND:
 		case YB_GLOBAL_VIEW_BACKEND:
+		case YB_XCLUSTER_DDL_QUEUE_BACKEND:
+		case YB_XCLUSTER_SETUP_BACKEND:
```
jasonyb: These two cases match what is already in this list, so no objection to them.  The list itself is missing its disclaimer though.  Could you add, above `case YB_YSQL_CONN_MGR:` (line 383):

```c
		/* YB_TODO_PG19MERGE: opt YB-only BackendTypes out of IO tracking for now. */
```

The merge that created both lists (4c01ae04c6d) records the intent in its message:

> - src/postgres/src/backend/utils/activity/pgstat_backend.c:
>     - Added YB_TODO_PG19MERGE: opt YB-only BackendTypes out of backend-stats tracking for now.
>
> - src/backend/utils/activity/pgstat_io.c:
>     - Same as above.

The comment only landed in `pgstat_backend.c`, so as it stands this switch reads as a settled per-type classification rather than a provisional opt-out.

Worth recording in that TODO that the two functions are not symmetric.  `pgstat_count_backend_io_op` returns early when the type is untracked (pgstat_backend.c:77), but `pgstat_count_io_op` instead asserts `pgstat_tracks_io_op` (pgstat_io.c:74), and `pgstat_io_flush_cb` asserts `pgstat_bktype_io_stats_valid` (pgstat_io.c:228).  So false here is a claim that the process performs no tracked IO at all, not a request to hide the row.  `YB_YSQL_CONN_MGR` is set on every backend serving user sessions through Connection Manager (yb_ysql_conn_mgr_helper.c:966), and temp relations are not YB relations, so they run the normal local-buffer stack and reach `pgstat_count_io_op*` in localbuf.c:215/300/464.  A temp table workload on a conn-mgr debug cluster looks like it would trip the assert.  Code path only, I have not run it.

src/postgres/src/backend/utils/adt/pgstatfuncs.c
```diff
@@ -673,69 +673,93 @@ pg_stat_get_activity(PG_FUNCTION_ARGS)
...
+					nulls[13] = true;
+				values[14] = Int32GetDatum(beentry->yb_st_cm_client_port >= 0
+										   ? beentry->yb_st_cm_client_port
+										   : -1);
 			}
 			else
 			{
-				if (beentry->st_clientaddr.addr.ss_family == AF_INET ||
-					beentry->st_clientaddr.addr.ss_family == AF_INET6)
+				/* YB: non-CM connection: standard upstream client-address reporting */
```
jasonyb: why was this comment changed from master?  and why were the other comments from master that follow deleted?

src/postgres/src/backend/utils/activity/pgstat_io.c
jasonyb, line 406: These two cases match what is already in this list, so no objection to them.  The list itself is missing its disclaimer though.  Could you add, above `case YB_YSQL_CONN_MGR:` (line 383):

```c
		/* YB_TODO_PG19MERGE: opt YB-only BackendTypes out of IO tracking for now. */
```

The merge that created both lists (4c01ae04c6d) records the intent in its message:

> - src/postgres/src/backend/utils/activity/pgstat_backend.c:
>     - Added YB_TODO_PG19MERGE: opt YB-only BackendTypes out of backend-stats tracking for now.
>
> - src/backend/utils/activity/pgstat_io.c:
>     - Same as above.

The comment only landed in `pgstat_backend.c`, so as it stands this switch reads as a settled per-type classification rather than a provisional opt-out.

Worth recording in that TODO that the two functions are not symmetric.  `pgstat_count_backend_io_op` returns early when the type is untracked (pgstat_backend.c:77), but `pgstat_count_io_op` instead asserts `pgstat_tracks_io_op` (pgstat_io.c:74), and `pgstat_io_flush_cb` asserts `pgstat_bktype_io_stats_valid` (pgstat_io.c:228).  So false here is a claim that the process performs no tracked IO at all, not a request to hide the row.  `YB_YSQL_CONN_MGR` is set on every backend serving user sessions through Connection Manager (yb_ysql_conn_mgr_helper.c:966), and temp relations are not YB relations, so they run the normal local-buffer stack and reach `pgstat_count_io_op*` in localbuf.c:215/300/464.  A temp table workload on a conn-mgr debug cluster looks like it would trip the assert.  Code path only, I have not run it.

src/postgres/src/backend/utils/misc/guc.c
```diff
@@ -8764,6 +8806,183 @@ assign_yb_dist_tracecontext(const char *newval, void *extra)
...
+	 */
+	if (!YbIsClientYsqlConnMgr() && !yb_is_parallel_worker)
+	{
+		GUC_check_errmsg("yb_conn_mgr_client_addr can only be set by "
+						 "YSQL Connection Manager");
+		return false;
+	}
+
+	/* Basic sanity: clean the string of non-ASCII bytes */
+	clean = pg_clean_ascii(*newval, MCXT_ALLOC_NO_OOM);
```
jasonyb, line 8833: Not a port error, the rewrite matches upstream's `check_application_name` (variable.c:1272) verbatim, and I traced all three `call_string_check_hook` sites (guc.c:1793, 3732, 4760) to confirm the pointer handed to `guc_free` is always a `guc_strdup`'d buffer nobody else owns.  Flagging it only because this is the one place in the guc group where behavior changes and not just location.

On master, PG15's `pg_clean_ascii` overwrote each non-printable byte with `?` in place, so the value length never changed.  PG19's version (upstream 45b1a67a0fcb) escapes each one as `\xNN` into a fresh buffer, up to four times longer.  A malformed `yb_conn_mgr_client_addr` therefore now surfaces escaped rather than as `?`, and the expansion can overrun the `strlcpy` targets in backend_status.h:238-240 (`NI_MAXHOST` for addr, `NAMEDATALEN` for hostname), which truncates rather than overflows.

Harmless for well-formed values from Odyssey.  Worth one look if the `pg_stat_activity` output for a hostile or broken value matters.  Same change applies to `check_yb_conn_mgr_client_hostname` at line 8925.

src/postgres/src/backend/utils/misc/guc.c
```diff
@@ -8764,6 +8806,183 @@ assign_yb_dist_tracecontext(const char *newval, void *extra)
...
+										 NULL, 0,
+										 NI_NAMEREQD);
+				freeaddrinfo(gai_result);
+
+				if (ret != 0)
+					ereport(WARNING,
+							(errmsg_internal("pg_getnameinfo_all() failed: %s",
+											 gai_strerror(ret))));
+				else
+					SetConfigOption("yb_conn_mgr_client_hostname",
```
jasonyb, line 8889: Worth a second look, though I believe it is fine.

This nested `SetConfigOption` existed on master too, but there the nested `check_yb_conn_mgr_client_hostname` only sanitized in place and never touched GUC memory.  After the PG19 rewrite it runs `pg_clean_ascii` + `guc_strdup` + `guc_free` while an outer `set_config_option` is still in flight, on an authentication-time path.

Reading it, the nested call is a complete `set_config_option` and the allocation is ordinary GUC memory, so it looks safe.  Any assert-enabled build is enough to exercise it: `guc_free` asserts `GetMemoryChunkContext(ptr) == GUCMemoryContext`, and `--enable-cassert` brings `CLOBBER_FREED_MEMORY` plus `MEMORY_CONTEXT_CHECKING` (pg_config_manual.h:268-279), which is what would catch a bad free or an overrun here.  ASAN would not: this is all palloc/pfree inside `GUCMemoryContext`, and PG's allocators carry Valgrind annotations only, with no ASAN poisoning.

jasonyb (approved):
> src/postgres/src/bin/pg_dump/pg_dump.c:
> master's side is the merge-base declaration, unchanged.

That is false.  There will always be an offending commit on both sides if a merge conflict shows up.  `git show --diff-merges=remerge` shows a merge conflict in that area.  If you dig into it, the reason for the conflict is the adjacent line change made by 62a60a60f20ef1f18125532858b76ad315006731 to add the `char       *sanitized;` line.

> src/postgres/src/include/utils/backend_status.h:
> No YB pg19 change at this point; the conflict is positional (upstream reshaped the surrounding struct).

Again, there will always be an offending commit on both sides if a merge conflict shows up.  I found the initial pg19 merge deleted the empty line right above `} PgBackendStatus;` which is the same area that master commit f576915 adds its lines.  If there is a merge conflict, then please find the offending commits on both sides every time.

src/postgres/src/backend/utils/misc/guc.c
fizaaluthra: going to call this out in the summary

src/postgres/src/backend/utils/misc/guc.c
fizaaluthra: going to land the merge and check this asynchronously

PR #33993, opened 2026-09-16
[PLAT-22545] YBA: Add Python 3.14 to supported versions

shubin-yb:
## Summary

Adds Python 3.14 to the supported interpreter set for YBA, and bumps the one dependency pin that has no wheel for it.

### Supported version lists

**`managed/devops/bin/common.sh`** — adds `python3.14` to `PYTHON3_EXECUTABLES` and `PYTHON3_VERSIONS`, plus the matching `cp-314` tags in `LINUX_PLATFORMS` and `MACOS_PLATFORMS`.

The platform arrays have to move together with the version list: `pex/build_pex.sh` expands them positionally (`${PYTHON3_VERSIONS[@]/#/--python=}` alongside `${LINUX_PLATFORMS[@]/#/--platform=}`, one `--platform` per `--python`), so adding a version without its platform tag leaves the pex build unable to resolve wheels for it.

**`common.ValidatePython`** — probes `python3.14` and raises the accepted range to 3.14, keeping the probe order aligned with `common.sh`.

### MarkupSafe 2.0.1 → 3.0.3

2.0.1's last wheel is `cp310`, so 3.11 through 3.14 have no prebuilt wheel at all. It still installs, by building from sdist and degrading to the pure-Python implementation, but `build_pex.sh` passes `--platform=manylinux_2_28_x86_64-cp-3XX-cp3XX` and pex cannot build from an sdist for a foreign platform — it only survives because `--resolve-local-platforms` lets pex substitute a matching local interpreter. 3.0.3 is the first release shipping `cp314` wheels.

Jinja2 3.1.6 only requires `MarkupSafe>=2.0`, so the bump is unconstrained.

With this in, MarkupSafe is no longer a gap on any supported version. Probing every pin against PyPI for cp310–cp314 on `manylinux_2_28_x86_64` leaves `psycopg2` as the only pin without a prebuilt wheel, and it is sdist-only on *every* version including 3.10 — unchanged by this PR.

### Known remaining gap, not addressed here

`boto` 2.49.0 fails to import on Python 3.12 and later, so it is already broken on two versions listed as supported today. Two independent causes: it vendors six 1.9.0, whose `_SixMetaPathImporter` implements only the meta-path `find_module`/`load_module` protocol that 3.12 removed, and `boto/plugin.py` (imported from `boto/__init__.py`) imports the `imp` module also removed in 3.12. Verified directly — imports cleanly on 3.11, fails on 3.12 with identical package sets.

It is reached only from `yb_backup.py`'s `get_instance_profile_credentials()` (the S3 IAM instance-profile path), which already defers the import into the function body with a comment noting the conflict, so it fails at call time rather than import time. boto 2.49.0 is the final release of a retired project, so there is no version to bump to — that call needs to move to botocore, which the same function already uses for its IMDSv2 branch. Left for a follow-up.

## Test plan

Verified locally:

- [x] `bash -n managed/devops/bin/common.sh` passes; each of 3.10–3.14 appears exactly twice across the four arrays (executables/versions, and the linux/macos platform tags), so the positional pairing `build_pex.sh` relies on stays intact. Edited lines stay under 100 chars.
- [x] `gofmt` clean on `pkg/common/python.go`.
- [x] Wheel availability probed directly against PyPI for cp310/311/312/313/314 on `manylinux_2_28_x86_64` (plus the older manylinux tags valid on a 2_28 target). Before the bump, MarkupSafe had no wheel for 3.11–3.14; after it, `psycopg2` is the only pin without a prebuilt wheel, and it is sdist-only on every version including 3.10 — so no version regresses relative to 3.10.
- [x] Installed the edited frozen set into a clean CPython 3.12 venv: `pip check` reports no broken requirements, and `pip freeze` produces nothing outside the frozen file, so the pin set is still a complete closure.
- [x] `markupsafe` 3.0.3 escapes correctly and `jinja2.Template('{{ x }}').render(...)` still auto-escapes against it.

Still needs a CI/build-machine run — cannot be done on the dev box:

- [ ] Full `pex/build_pex.sh` run producing a pexEnv.pex covering cp310–cp314, on a machine with all five interpreters installed.
- [ ] Exercise the resulting pex against a real Python 3.14 interpreter. 3.13 and 3.14 are not installed on the dev box, so the runtime evidence above comes from 3.12; wheel availability for 3.13/3.14 was checked against PyPI directly, but a real 3.14 run is the actual confirmation.
- [ ] `yba-installer` preflight against a host with Python 3.14 to confirm `ValidatePython` accepts it, and that the error string reads correctly when no supported interpreter is present.


amannijhawan (approved):

nkhogen (commented):
LGTM. 

nit FYI only: Found another reference 
https://github.com/yugabyte/yugabyte-db/blob/master/managed/node-agent/resources/templates/server/clock-sync.sh.j2#L25

I do not think we need to keep all the versions in the j2 file just python or python3 should be enough. Also, it is exporting this version only for a few simple usages.

Issue #32524, opened 2026-07-07
[YSQL] yb_index_check should have guardrails for tables with RLS


angeladev333:
Jira Link: [DB-22229](https://yugabyte.atlassian.net/browse/DB-22229)
### Description

**Problem**
Currently, `yb_index_check` can fail or return incorrect, inconsistent results when run against a table that has Row-Level Security (RLS) enabled, if the executing user lacks RLS bypass permissions.

In PG, the standard guardrails for similar functions only assert that the user is either a superuser or possesses execution privileges. However, this model is flawed for index validation: users without RLS bypass capabilities shouldn't be allowed to implicitly validate index consistency on data rows they do not have structural permissions to see.

To prevent inaccurate results or silent gaps in index checking, we need to fail fast and surface a clear RLS-related error early in the execution block. 

**Proposal**
We need to introduce a new guardrail check within yb_index_check. The logic should adhere to the following:

- Superusers: Granted full execution capabilities as always (no change).
- Execution Permissions: Granted to everyone as always (no change).
- RLS Guardrail (New Diff Change): * If the target table has RLS enabled, the engine must inspect whether the executing user possesses RLS bypass permissions (e.g., BYPASSRLS or superuser status). If the user does not have RLS bypass permissions, the execution must abort immediately and throw a descriptive error.

### Issue Type

kind/bug

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-22229]: https://yugabyte.atlassian.net/browse/DB-22229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

angeladev333:
related ticket: https://github.com/yugabyte/yugabyte-db/issues/30287 :

Create the following schema as a superuser (yugabyte):
```
CREATE TABLE secure_notes (id SERIAL PRIMARY KEY,  tenant_name TEXT, note_content TEXT);
CREATE INDEX idx_note_content ON secure_notes(note_content);

-- Enable and force row-level security
ALTER TABLE secure_notes ENABLE ROW LEVEL SECURITY;
ALTER TABLE secure_notes FORCE ROW LEVEL SECURITY;

CREATE POLICY tenant_isolation ON secure_notes USING (tenant_name = current_user);

yugabyte=> \d secure_notes
[...]

PR #34005, opened 2026-09-17
[BACKPORT 2024.2][#32524] YSQL: Guardrail yb_index_check() callers subject to row-level security

karthik-ramanathan-3006:
Summary:
`yb_index_check()` validates an index against every physical row of its base relation. Its scan paths are raw executor plans that bypass the rewriter (and therefore row-level security), so a caller who is subject to RLS could use the function to learn the existence and count of rows they are not permitted to see and may otherwise get incorrect or inconsistent results.

This adds a guardrail that runs before any scan work. `lock_index_check_relations()` walks the index tree, takes `AccessShareLock` on every index and base table in it, and returns the leaf indexes to scan. `validate_index_check_rls()` then runs over that list and aborts immediately with an insufficient-privilege error if RLS applies to the caller on any base table that would be read. The decision reuses `check_enable_rls()` (with noError=true), which returns `RLS_ENABLED` precisely when the caller is neither a superuser, nor holds `BYPASSRLS`, nor is the exempt owner (owners are exempt unless FORCE ROW LEVEL SECURITY is set). Primary indexes are skipped, since YB has no separate PK index and those leaves are never scanned.

Tables without RLS are unaffected, and superuser / BYPASSRLS / exempt-owner callers continue to run the check unchanged. The guardrail applies to both single- and multi-snapshot modes, and to partitioned indexes: `find_all_inheritors()` recurses through the partitioned index tree and each leaf is checked against its own base table, since only leaves hold data and RLS is not inherited. This replaces `partitioned_index_check()`, which enumerated a single level of children at scan time; the tree is now enumerated once by the locking walk, `do_index_check()` is responsible for one leaf, and `yb_index_check()` contains the loop and the error budget.

The verdict has to still hold when the scan runs, so the AccessShareLock taken on a leaf's base table before the check is held continuously until that leaf's scan finishes. Every DDL that can change the verdict (`ALTER TABLE ... ENABLE/FORCE ROW LEVEL SECURITY` and `ALTER TABLE ... OWNER TO`) needs AccessExclusiveLock, which those locks conflict with. Locks are released as the walk proceeds: non-leaf indexes and their base tables are unlocked as soon as the tree is enumerated, each leaf is unlocked by `do_index_check()` when its check ends, and any leaf left unscanned once the error budget is spent is unlocked in the same loop. Role-side changes (`ALTER ROLE ... BYPASSRLS`, role membership) take no relation lock and are outside what relation locking can order against; PostgreSQL evaluates authorization at statement start elsewhere and accepts the same race.

A weak table lock is sufficient because the set of leaves is fixed by `lock_index_check_relations()` and the scan loop only visits that list. A partition attached after the walk is absent from the list instead of being scanned unchecked, so ATTACH PARTITION (which takes only ShareUpdateExclusiveLock on the parent) does not have to be blocked for the duration of what may be a full scan of every partition. Ordinary DML is unaffected throughout, so the check is not a table-wide blocker.

The error follows the standard PostgreSQL style: a primary "permission denied for function yb_index_check" message with a DETAIL naming the table and a HINT pointing at the BYPASSRLS privilege.

Original commit: 865ad07cc7b8ed87c66b187ddd4a5d052b210b7f / D55361

Test Plan:
Added regression coverage in yb.orig.yb_index_check.sql (expected output in
yb.orig.yb_index_check.out and yb.orig.yb_index_check_1.out) exercising, on an
RLS + FORCE table:
  - superuser: check passes (both modes)
  - a role with BYPASSRLS and execute privilege: check passes (both modes)
  - a role with only EXECUTE (via PUBLIC) and no RLS bypass: rejected (both
    modes); a role with no privileges hits the same guardrail

Run:
  ./yb_build.sh debug --java-test 'org.yb.pgsql.TestPgRegressYbIndexCheck'
  ./yb_build.sh debug --java-test 'org.yb.pgsql.TestPgRegressYbIndexCheckSingleSnapshot'

Reviewers: kramanathan, aagrawal

Reviewed By: kramanathan

Subscribers: yql, ybase

Tags: #jenkins-ready

Differential Revision: https://phorge.dev.yugabyte.com/D55361

## Merge conflicts

Non-trivial port onto 2024.2's older `yb_index_check` (void `yb_index_check(oid)`, no multi-snapshot / log_num_errors / `do_index_check`).

- `src/postgres/src/backend/utils/misc/yb_index_check.c` — re-implemented `lock_index_check_relations()` / `validate_index_check_rls()` against `yb_index_check_internal`; replaced `partitioned_index_check()` with a `find_all_inheritors()` walk; leaves keep `AccessShareLock` until `yb_index_check_internal()` finishes. Did not pull master's logging / set-returning / multi-snapshot machinery.
- `src/postgres/src/test/regress/sql/yb_index_check.sql` and `expected/yb_index_check.out` — ported the RLS tests onto the existing non-`yb.orig` files and the `yb_index_check(oid)` signature (dropped dual-mode `true`/`false` pairs). Skipped nested DEFAULT-partition tests that use `pg_partition_tree` (PG12+; 2024.2 is PG11).
- `src/postgres/src/test/regress/expected/yb.orig.yb_index_check.out` — kept deleted; `yb.orig` tests do not exist on 2024.2
- `src/postgres/src/test/regress/expected/yb.orig.yb_index_check_log_num_errors.out` — kept deleted
- `src/postgres/src/test/regress/yb_index_check_multi_snapshot_schedule` — kept deleted


karthik-ramanathan-3006:
Trigger Jenkins

PR #34010, opened 2026-09-17
[#34011] YSQL: Use the child partition's tablespace for secondary indexes that YSQL automatically creates

austenLacy:
## Summary

re: https://github.com/yugabyte/yugabyte-db/issues/34011

YSQL currently creates child partition indexes in the parent index tablespace or the session default tablespace even if the child partition defines its own tablespace. This can put an index in a different placement from its table partition which can lead to varying performance and resiliency characteristics. 

Instead we should use the child partition's tablespace for secondary indexes when YSQL creates those indexes automatically during these operations:

 - `CREATE INDEX` on a partitioned table
 - `CREATE TABLE ... PARTITION OF`
 - `ALTER TABLE ... ATTACH PARTITION` when no matching index exists

This rule also applies to nested partitions and partitions that use the database default tablespace. Primary keys and indexes that users create directly on partitions keep their current tablespace rules.

## SQL reproduction

```sql
\set ON_ERROR_STOP on

SELECT json_build_object(
    'num_replicas', 1,
    'placement_blocks', json_build_array(json_build_object(
        'cloud', cloud,
        'region', region,
        'zone', zone,
        'min_num_replicas', 1
    ))
) AS repro_placement
FROM yb_servers()
WHERE node_type = 'primary'
ORDER BY host
LIMIT 1
\gset

CREATE TABLESPACE repro_partition_ts
    WITH (replica_placement = :'repro_placement');
CREATE TABLESPACE repro_parent_ts
    WITH (replica_placement = :'repro_placement');

CREATE TABLE repro_t (a int, b int) PARTITION BY LIST (a);

-- Create an index for an existing partition.
CREATE TABLE repro_p1 PARTITION OF repro_t
    FOR VALUES IN (1) TABLESPACE repro_partition_ts;
CREATE INDEX repro_idx ON repro_t (b) TABLESPACE repro_parent_ts;

-- Create a partition after the parent index exists.
CREATE TABLE repro_p2 PARTITION OF repro_t
    FOR VALUES IN (2) TABLESPACE repro_partition_ts;

-- Attach a table that has no matching index.
CREATE TABLE repro_p3 (a int, b int) TABLESPACE repro_partition_ts;
ALTER TABLE repro_t ATTACH PARTITION repro_p3 FOR VALUES IN (3);

SELECT t.relname AS partition_table,
        s.spcname AS index_tablespace,
        i.reltablespace = t.reltablespace AS same_tablespace
FROM pg_inherits AS h
JOIN pg_index AS p ON p.indexrelid = h.inhrelid
JOIN pg_class AS i ON i.oid = p.indexrelid
JOIN pg_class AS t ON t.oid = p.indrelid
JOIN pg_tablespace AS s ON s.oid = i.reltablespace
WHERE h.inhparent = 'repro_idx'::regclass
ORDER BY t.relname;
```

Results should look something like

```
┌────────────────┬────────────────────┬─────────────────┐
│ Version        │ index_tablespace   │ same_tablespace │
├────────────────┼────────────────────┼─────────────────┤
│ Before the change │ repro_parent_ts    │ f               │
├────────────────┼────────────────────┼─────────────────┤
│ With the change   │ repro_partition_ts │ t               │
└────────────────┴────────────────────┴─────────────────┘
```


austenLacy:
trigger jenkins

Issue #34012, opened 2026-09-17
[DocDB] Unnecessary UUID mutex blocks master catalog RPCs


jordanwheeler:
Jira Link: [DB-23624](https://yugabyte.atlassian.net/browse/DB-23624)
### Description

 A `threadz` dump from a production master leader captured Read, GetTableSchema, and GetTableLocations RPC workers blocked in `FsManager::uuid()`. These requests were waiting for the filesystem metadata mutex before their catalog operations could proceed.

`ScopedLeaderSharedLock::Lock()` [fetches the UUID unconditionally](https://github.com/yugabyte/yugabyte-db/blob/026fae880da97ef00f3c4bc96d0438c18b946eb0/src/yb/master/scoped_leader_shared_lock.cc#L104-L112). `FsManager::uuid()` [acquires `metadata_mutex_`](https://github.com/yugabyte/yugabyte-db/blob/026fae880da97ef00f3c4bc96d0438c18b946eb0/src/yb/fs/fs_manager.cc#L852-L855), even though the UUID is only used to format an error when the master is in shell mode. Successful catalog RPCs take this mutex unnecessarily.

Instead, consider using the already-initialized server instance UUID (`master_->instance_pb().permanent_uuid()`) inside the shell mode branch. This preserves the error response and removes the filesystem mutex dependency from these leadership checks.

### Issue Type

kind/enhancement

### Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23624]: https://yugabyte.atlassian.net/browse/DB-23624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #34013, opened 2026-09-17
[#34012] DocDB: Avoid FsManager uuid mutex in ScopedLeaderSharedLock::Lock

jordanwheeler:
Master RPC leadership checks [called `FsManager::uuid()` unconditionally](https://github.com/yugabyte/yugabyte-db/blob/026fae880da97ef00f3c4bc96d0438c18b946eb0/src/yb/master/scoped_leader_shared_lock.cc#L104-L112), which [acquires `metadata_mutex_`](https://github.com/yugabyte/yugabyte-db/blob/026fae880da97ef00f3c4bc96d0438c18b946eb0/src/yb/fs/fs_manager.cc#L852-L855), even though only a shell-mode error needs the UUID. 

A production `threadz` capture showed `Read`, `GetTableSchema`, and `GetTableLocations` RPC workers blocked on this mutex before their catalog operations could proceed. Every catalog `Read` passes through this check, so contention can delay queries on both new and established connections whenever they need catalog data from the master.

This change uses the immutable server instance UUID inside that error branch, removing the filesystem mutex dependency from both paths.

Validated with a release build and 14 focused tests across `master-test`, `master_cluster_test`, and `master_config-itest`, covering catalog RPCs, concurrent DDL/schema reads, heartbeats, startup/shell mode, and master membership/failover.


Hasan-Al-Khazraji (approved):
LGTM 👌 (still my 5th day... lol)

spolitov (approved):

Issue #1602, opened 2019-06-21
Clusters started with initDB snapshot will have the same cluster_uuid


WesleyW:
Using the `--use_initial_sys_catalog_snapshot` gflag will override both the built-in UUID generator and the pass in `--cluster_uuid` field with the saved UUID in the snapshot. Either the snapshot should not be saving this UUID or there should be a way to update `cluster_uuid`.

Issue #1595, opened 2019-06-21
Proposal for labels in this repo


rkarthik007:
**Please DO NOT CREATE NEW LABELS**. You should open a new issue with the details of the label that is needed and label that issue `community/guideline`.

Below is a proposal for labelling the various issues on the site. This has been discussed with a few folks. All labels not in the list below will be removed.

| Generic | Description |
| --- | --- |
| good first issue | Good issue for first timers
| help wanted | Issues where help is much appreciated
| duplicate | Issue is a duplicate. Add info in a comment.

| Functional Areas | Description |
| --- | --- |
| area/kubernetes (was k8s) | Kunernetes
| area/sql (was ysql) | YSQL
| area/docdb (was ybase) | DocDB
| area/ui (was yugaware) | UI and UX
| area/documentation | Related to documentation

| Community Related | Description |
| --- | --- |
| community/guideline | This issue is related to community guidelines
| community/tutorial | This is about a community contributed tutorial 
| community/needs-cla | The contributor of this issue/PR needs to sign a CLA

| Kinds of Bugs | Description |
| --- | --- |
| kind/bug (was bug) | 
| kind/enhancement (was enhancement) | 
| kind/failing-test (was testing) | 
| kind/improve-ux (was improve-ux) | 
| kind/new-feature (was new-feature) | 
| kind/question (was question) | 

| Priority | Description |
| --- | --- |
| priority/critical | 
| priority/high (was high-pri) | 
| priority/medium | 
| priority/low | 

| Size/complexity of a feature | Description |
| --- | --- |
| size/XS | Denotes a PR that changes 0-9 lines, ignoring generated files.
| size/S | Denotes a PR that changes 10-29 lines, ignoring generated files.
| size/M | Denotes a PR that changes 30-99 lines, ignoring generated files.
| size/L | Denotes a PR that changes 100-499 lines, ignoring generated files.
| size/XL | Denotes a PR that changes 500-999 lines, ignoring generated files.
| size/XXL | Denotes a PR that changes 1000+  lines, ignoring generated files.

rkarthik007:
Folks, 

Just an FYI that this is starting to get implemented.

Issue #1591, opened 2019-06-20
[YCQL] Unable to do explicit ORDER BY sort on clustered columns of index


dwheeler:
When creating a keyspace with the following schema
```cql
CREATE TABLE test_table (
	a text,
	b text,
	c int,
	PRIMARY KEY (a, b))
WITH CLUSTERING ORDER BY (b ASC) AND default_time_to_live = 0;

CREATE INDEX test_table_index ON test_table (b, c) INCLUDE (a) 
WITH CLUSTERING ORDER BY (c DESC) 
  AND transactions = {'enabled' : FALSE, 'consistency_level' : 'user_enforced'};
```
Queries that use the secondary index and provide an order by will fail with the error:
```cql
SELECT * FROM test_table WHERE b = 'b2'  ORDER BY c DESC;
ERROR: Invalid Arguments. All hash columns must be set if order by clause is present.
```
Leaving the order clause out succeeds 
```cql
SELECT * FROM test_table WHERE b = 'b2';
```

kmuthukk:
Thx @dwheeler for reporting.
@nocaway - could you take an initial look of what's involved to address this?

nocaway:
This is a bug in YugaByte semantic analysis.
- The error check was processing SELECT statement against the table description without considering the secondary indexes of the table.
- The decision on which index to use for a query is determined at a later time.

Implementation.
- Currently, when WHERE clause is processed, values of columns are divided to list<hash_column>, list<range_column>, and list<regular_column> according to the table description.  These list are then used to check against ONLY the PRIMARY index. Afterward, YugaByte walks the lists to decide which INDEX to use.

- The fix would be changing the list<hash_column_values> to array< list<hash_column_values> >.  Each entry of the array would be associated with an index.  PRIMARY index is associated with array[0].  Index id 1 would be associated with array[1], and so on. The array can then be used to analyze the query against each INDEX.

nocaway:
This has been fixed by commit https://github.com/YugaByte/yugabyte-db/commit/592fcfab8a8bd6a78fbf6438f307ed9af3b5b97f

Instead of making copies of column_list (schema) for the indexes as suggested in my previous comment, the existing code made one new copy for the entire SELECT statement, replace the target-table with its chosen table_index, and rerun analysis on the entire SELECT against the chosen table_index.  I decided to keep this design of copying the entire SELECT statement in this bug fix as changing the current design would be beyond the scope of this bug fix.

[...]

Issue #118, opened 2018-03-22
Limit the total amount of memory consumed by network read buffers


rkarthik007:
When there are a very large number of outstanding operations (very large IO spikes), YugaByte DB could run out of memory a few times due to allocating too many read buffers. Limit the total amount of memory allocated by read buffers and slow down request processing in that case.

rkarthik007:
Is this done @spolitov? If so could we please add the commit id and close... thanks!

spolitov:
There is related diff, that allows to specify such limit based on available RAM.
So I think it would be better to close this issue when that diff is also landed.

spolitov:
Done in 0fc413a1a905000649c3a1208af06f2362d63994

PR #34029, opened 2026-09-18
[DEVOPS-3691] build: csi_report: run a lane's new tests extra times, tagged as such

hkimura:
## Summary

A test that is flaky from the day it lands gets exactly one sample on the build that introduces it. When it fails weeks later there is no healthy history behind it on that lane, so "this commit broke it" and "it was always a coin flip" are indistinguishable, and someone spends an afternoon finding that out. The flavors where flaky-at-birth is most likely (tsan, aarch64, mac) never run a change before it lands, so the birth build on the scheduled lane is the one cheap chance to catch it: the build is already sitting on the workers.

This PR gives the test harness a `--new_test_repetitions N` option (default 0, off). After a lane's main pass, each test that the previous launch of the same lane did not report, and that passed here, is run N-1 more times as one more Spark job after the main pass, in the shape the fail repetitions already use (its tasks run in parallel, as theirs do). The extra executions are reported to CSI (ReportPortal) as retries of the same item tagged `retry_kind=new_test_repetition`. Three or more failures in 20 is then a flaky-at-birth call with a defensible false-positive rate (~1e-3 against a 1% flake baseline), made while the author still has the context.

Nothing changes on any lane until `YB_NEW_TEST_REPETITIONS` is set (jenkins-helpers, separate PR): `yb-jenkins-test.sh` passes `--new_test_repetitions 0` and the pass returns immediately. Consumers: the CSI summary (`csi/lib.groovy`, `csi/sum.groovy`), csiweb (detective-webapp), and Grissom (agent-k), each of which gets a one-rule reader change in its own PR before the variable is set anywhere.

## Why this shape and not X

- **Why the harness asks CSI, and nothing generates or passes a test list around.** The previous launch of the lane IS the "before" list, already recorded. `csi_report.py` already holds the CSI credentials and already queries ReportPortal about its own launch (`launch_qid`, `query_test`), so it reads the previous launch's item list itself after its own pass. That removes a Groovy producer, a shell bridge and a `--known_list` argument, and leaves `ReleaseBuild.groovy` untouched, which its compiled method size makes a hard requirement (jenkins-helpers #1596, #1602).
- **Why in-launch rather than a separate probe launch per lane per build.** A separate launch keeps lane launches byte-identical but costs ~18 extra pipeline runs a day whose per-run overhead (workspace, build download) dominates the test time it buys. In-launch costs nothing but the test executions, and tagging the repetitions makes them separable for every reader.
- **Why not `--num_repetitions` on a subset.** `YB_CSI_REPS` is a single global per launch; raising it for a subset would report every one of the ~12k tests in the launch as `retry_kind=repetition`, which readers treat as "first attempt unknown".
- **Why not the per-change pipeline.** tsan, aarch64 and mac never run per-change, and per-change launches are excluded from every baseline computation downstream. Worth doing for developer feedback, but it does not answer this question.

## What is in the change

**`python/yugabyte/run_tests_on_spark.py`**

- `--new_test_repetitions N` (int, default 0). 0 or 1 disables it. Wired from `YB_NEW_TEST_REPETITIONS` by `build-support/jenkins/yb-jenkins-test.sh`, in the `else` branch where `--fail_repetitions` is passed, so runs with `--num_repetitions > 1` are untouched.
- `run_new_test_repetitions(args, conf, results, env_vars, main_pass_cancelled)`, called after the fail-repetition block and before the CSI suites are closed, inside `try/except` that logs and continues. It:
  1. returns unless `reps > 1`, `num_repetitions == 1`, and none of `--test_list`, `--ignore_list`, `--max_tests`, `--test_filter_re`, `--test_conf` is set (every one of those means this run is not the lane's own full pass over its own test list);
  2. returns if the main pass was cancelled at the failure threshold. `main_pass_cancelled` is whether the main pass's job group (`all-tests-on-diff`) is in `g_cancelled_job_groups`, read right after that pass; the re-run jobs that follow keep their own groups in the same set;
  3. asks `csi_report.previous_launch_unique_ids()` for the baseline; `None` means do nothing;
  4. returns if the baseline holds fewer than `NEW_TEST_MIN_KNOWN_RATIO = 0.9` times this run's test count;
  5. takes the tests not in the baseline **that passed here**, one entry per test (a Spark task that reported and was then resubmitted leaves two results), sorted; past `NEW_TEST_LIMIT = 50` a seeded (`random.Random(0)`) sample of 50, reproducible from the build log. A test that exited 0 but skipped itself (`GTEST_SKIP`, a platform gate; CSI has it as `skipped`) is not a pass and is not repeated: `TestResult` gains a `skipped` flag, set where the worker decides the CSI status;
  6. submits attempts 2..N for each through `run_tests_job_with_resubmits(rerun=False, env_vars=env + {YB_CSI_NEW_TEST: '1'}, job_group=NEW_TESTS_ON_DIFF_JOB_GROUP)`, the same environment as the main pass so the workers report into the same CSI suites, and its own Spark job group (`new-tests-on-diff`) so the failure monitor cancels only this job and its cancellation is not the main pass's;
  7. logs `New tests on this lane: P passed, F failed (...), S skipped` and, per test with a failure, `New test failed K of its N runs in this build: <descriptor>`.
  A failure in the extra runs never changes the build's exit code. This job must stay the last of the run: executor processes are reused and `set_env_on_spark_worker` never unsets, so a job dispatched after it would inherit the marker (a comment says so).

**`python/yugabyte/csi_report.py`**

- `previous_launch_unique_ids() -> Optional[Set[str]]`. Three GETs plus one per 5000 items, so six for a ~14k-item lane, flat in practice:
  1. `GET /launch/<uuid>` for this launch's `name`, `version` and `bld` attributes; without `version` and `bld` this is a per-change launch (its predecessor is another revision of the same change, not the lane's previous state) and the answer is `None`;
  2. `GET /launch?filter.eq.name=<name>&filter.has.compositeAttribute=version:<version>&page.sort=id,desc&page.size=10`; the first entry with a lower id whose status is `PASSED` or `FAILED` is the baseline (`IN_PROGRESS` and `INTERRUPTED` are rejected; none means `None`);
  3. `GET /item?filter.eq.launchId=<prev>&filter.eq.type=SUITE` via `previous_launch_reached_every_test()`: every suite the harness planned tests into (one carrying an `All`, `RegEx` or `Requested` count attribute; a parent suite a job may nest them under is ignored) must carry `All` and report at least that many in `statistics.executions.total`, else `None`; none planned is `None` too;
  4. `GET /item?filter.eq.launchId=<prev>&filter.eq.type=step&page.size=5000`, three pages for a ~14k-item lane, collecting `uniqueId`; more than `ITEM_QUERY_LIMIT = 30000` items, or any failed request, is `None`.
  It returns `None`, never a partial set, on any doubt.
- `previous_launch_reached_every_test(csi, launch_id) -> bool`: the check above, the comparison `csi/lib.groovy`'s `rerun_list` (jenkins-helpers) already makes before trusting a launch's test list.
- `previous_launch_unique_ids` is a sibling of `test_ids_in_launches` (5a30035a329): both page a CSI item list into a set of uniqueIds, for different launches. They could share a pager later.
- `run_tests_job_with_resubmits` gains an optional `job_group`, for a caller with a Spark job of its own; the default derivation from `rerun` is unchanged.
- `classify_execution(rerun, attempt, reps, attempt_index, new_test=False)`: a fourth branch returning `(True, 'new_test_repetition', False)`, ordered after `fail_repetition` and before `task_resubmit`, so a Spark resubmit inside the new-test job keeps the tag and the "first attempt passed" implication stays exact for every execution of that job.
- `create_test`: a new-test execution does not call `query_test` before reporting, the same skip a fail repetition gets. Both run serially after the main pass, in which the first attempt completed and reported; the query exists for repetitions (which run alongside their first attempt) and resubmits (which follow a worker death).
- `ITEM_PAGE_SIZE = 5000`: the CSI server honors it and answers faster per item than at smaller sizes (measured 2026-09-17: 13,920 items in 3 pages at ~3 s each, against 14 pages at ~7.5 s for 1000).
- The flag reaches the worker as `YB_CSI_NEW_TEST`, set on that job's environment only, exactly how `YB_CSI_REPS` already works. No signature change on the worker path.

**`build-support/jenkins/yb-jenkins-test.sh`**: one line passing `--new_test_repetitions "${YB_NEW_TEST_REPETITIONS:-0}"` next to `--fail_repetitions`.

## Design decisions

- **The unit of "already exists" is (launch name, exact `version` attribute)**, i.e. per lane, per minor release and patch component (`2026.1-alma8-clang21-tsan` + `2026.1.3.0`). Not per family: `master-<lane>` is shared by master (`2.31.0.0`) and the pg19 branch (`2.31.0.1900`), so a family-level baseline would compare a master build against a pg19 launch whenever pg19 built last, and branch-specific tests would look new on every build.
- **No comparable predecessor means skip, never "everything is new".** The first build after a version bump, the first build of a new lane or a new branch, an interrupted or truncated predecessor: all skip with one log line. The cost is one build per lane per version bump with the status quo n=1 for tests born in it. Deliberately no fallback to an older version: it would reintroduce the pg19 folding above.
- **Only tests that PASSED the main pass are repeated.** A new test that failed already gets its fail repetitions, so nothing is lost, and the restriction is what makes the tag exact, the dual of `fail_repetition`:

  | tag on the top-level item | what the first attempt did |
  |---|---|
  | `fail_repetition` | failed (fail reps only run after a failure) |
  | `new_test_repetition` | passed (new-test reps only run after a pass) |

  That is a one-line rule for every reader, and it is what keeps a new test that flakes in run 7 of 20 out of the first-attempt baseline instead of quietly poisoning it.
- **A false "new" can only waste executions, never corrupt data.** Because only passing tests are repeated, the tag means "first attempt passed" whether or not the test was really new. Every detection error degrades cost, bounded by the 50-test cap (950 executions, ~7% of one launch, once), and self-heals on the next build against a correct predecessor.
- **The completeness check is per suite, not a ratio.** A main pass that stops at `YB_MAX_NUM_TEST_FAILURES` finishes its launch `FAILED` with the unreached tests unreported; under a 10% shortfall a count ratio would still accept it and the cap would sample 50 of ~1,300 old tests. Planned-vs-reported per suite catches any shortfall. The 90% ratio stays as a second line.
- **Bounded and non-gating by construction.** At most 50 tests per build; the extra runs start after the pass is over; their failures and any exception in the whole pass leave the exit code alone. A lane that would have passed still passes.

## What it deliberately does not do

- It does not run on per-change (Phabricator / PR) launches, on `ReTestMultiple` / `RELEASE_JOB` re-runs, on a rerun of named tests, on an affected-tests run, or on a cancelled main pass.
- It does not gate anything. A flaky new test shows up in the build's CSI summary as a flaky item (a retried item with a failure); it does not trip "New Consistent Failures" or "Worse than Baseline", both of which need every try to fail.
- It does not try to pair renames (one uniqueId dies, a similar one is born in the same build): those get repeated once, at the cost above.
- It does not enable anything. That is the jenkins-helpers change, after the readers are deployed.

## Cost

About 5 new tests on a typical build, so ~95 extra executions on a ~12-14k-test lane: under 1%. Worst case at the cap, 950, about 7% of that one build. Plus six CSI GETs on the driver after the tests have run (a few seconds each for the item pages), and no per-execution CSI query on the workers.

Known and shared with the fail repetitions: the attempts of one test share the `yb-test-logs/<test>__error.log` path (it is derived from the descriptor without the attempt index), so two attempts of the same test on one worker can overwrite each other's harness error log. The test's own log carries the `__attempt_N` suffix and does not collide.

## Testing

- `pytest python/yugabyte/test_csi_report.py python/yugabyte/test_run_tests_on_spark.py`: 118 passed (the 51 cases 5a30035a329 brought included). New cases cover: the `retry_kind` decision table row and its precedence over `task_resubmit`; which launch is chosen as the baseline (newer, in-progress, interrupted ones rejected; the version filter asserted); a truncated (reported < planned) and a partial (`RegEx`-planned) previous launch yielding no baseline and reading no items; a parent suite ignored and a null count attribute not trusted; the read being a handful of requests; a new-test repetition reporting without a previous-item query, as a fail repetition does; the oversized-item-query bail-out; and every guard in `run_new_test_repetitions` (off, `num_repetitions > 1`, each partial-run option, cancelled main pass, no baseline, short baseline, the 50-test sample, a test that skipped itself not repeated, two results for one test repeated once, only-passing-tests-repeated with `rerun=False` and `YB_CSI_NEW_TEST=1` in the job's environment).
- `pycodestyle --max-line-length=100` and `mypy` clean on the changed Python files; `bash -n` on the shell file.
- Not exercised here: a real lane run with the variable set. The first enabled lane's build log carries the whole signal (`New tests on this lane: ...`, `New test failed K of its N runs ...`), and the CSI item's retries carry the tag.

## Rollout order

The same wave `fail_repetition` went through (yugabyte-db #33682 -> jenkins-helpers #1591 -> detective-webapp #118 -> jenkins-helpers #1595):

1. this PR (inert);
2. jenkins-helpers: `csi/lib.groovy first_attempt_status` reads the tag as a passing first attempt, `csi/sum.groovy` keeps such items out of `reran_passed` (yugabyte/jenkins-helpers#1622);
3. detective-webapp: the same rule in `lib/detective/csi_attempts.rb` (yugabyte/detective-webapp#120);
4. agent-k: Grissom's retry-kind inference and `attempt_source` (yugabyte/agent-k#162);
5. jenkins-helpers: `YB_NEW_TEST_REPETITIONS=20` on the integration lanes (not mac, not `RELEASE_JOB` re-runs, not per-change builds) (yugabyte/jenkins-helpers#1623).

Between 1 and 5 nothing changes anywhere.

CI note: the change is inert in a per-change build (the flag is 0 and a per-change launch has no lane baseline), so the run for the required status is restricted to one lane and one small test rather than the full matrix; what it proves is that the harness still starts, parses the new option and finishes.

Jenkins: cpp only
Jenkins: build type: fastdebug
Jenkins: test regex: tests-util/status-test


svarnau (changes requested):
Looks okay. One question about the limiting of new tests.

> @svarnau reviewed 6 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/yugabyte/yugabyte-db/34029#-:-P2FG2WZ1-lIztwZvI52:bx5rxuw)* status: all files reviewed, 1 unresolved discussion (waiting on hkimura).

___
*[`python/yugabyte/run_tests_on_spark.py` line 145 at r1](https://reviewable.io/reviews/yugabyte/yugabyte-db/34029#-P2FCg-A9Gy53gYMwWSb:-P2FCg-A9Gy53gYMwWSc:bwt5o0e) ([raw file](https://github.com/yugabyte/yugabyte-db/blob/1cabb5ba072eeef611113c31ba848af4d247c75a/python/yugabyte/run_tests_on_spark.py#L145)):*
> ```Python
> # lane; one that ended early reported a fraction of its tests, and every test it never reached
> # would otherwise read as new.
> NEW_TEST_MIN_KNOWN_RATIO = 0.9
> ```

Why are we picking a ratio here instead of a hard count like the other limits?  With 13,000 tests, we can add 1,300 new ones in a single PR and we'll run each 20 times? Yikes!  Maybe limit to 100 new tests?

hkimura (commented):
Replied there.

> @hkimura made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/yugabyte/yugabyte-db/34029#-:-P2FfKj853FxOgii_c6a:b4cdl12)* status: all files reviewed, 1 unresolved discussion (waiting on svarnau).

___
*[`python/yugabyte/run_tests_on_spark.py` line 145 at r1](https://reviewable.io/reviews/yugabyte/yugabyte-db/34029#-P2FCg-A9Gy53gYMwWSb:-P2Ff8RH42o3rejrMgXY:boz8guy) ([raw file](https://github.com/yugabyte/yugabyte-db/blob/1cabb5ba072eeef611113c31ba848af4d247c75a/python/yugabyte/run_tests_on_spark.py#L145)):*
<details><summary><i>Previously, svarnau (Steve Varnau) wrote…</i></summary><blockquote>

Why are we picking a ratio here instead of a hard count like the other limits?  With 13,000 tests, we can add 1,300 new ones in a single PR and we'll run each 20 times? Yikes!  Maybe limit to 100 new tests?
</blockquote></details>

There is a hard count too, and it is lower than 100: NEW_TEST_LIMIT = 50, just above this one. Past 50 new tests a seeded sample of 50 is repeated, so a PR adding 1,300 tests costs 50 x 19 = 950 extra executions on each lane, not 1,300 x 19. The ratio answers a different question: whether the previous launch is complete enough to be the baseline at all. A list covering less than 90% of this run's tests is not used, and nothing is repeated. Reworded the comment so it doesn't read as the cap.

svarnau (approved):

hkimura (commented):
> @hkimura made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/yugabyte/yugabyte-db/34029)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on svarnau).

___
*[`python/yugabyte/run_tests_on_spark.py` line 145 at r1](https://reviewable.io/reviews/yugabyte/yugabyte-db/34029#-P2FCg-A9Gy53gYMwWSb:-P2G0Jk10d2iIzSprKeP:buulnas) ([raw file](https://github.com/yugabyte/yugabyte-db/blob/1cabb5ba072eeef611113c31ba848af4d247c75a/python/yugabyte/run_tests_on_spark.py#L145)):*
<details><summary><i>Previously, hkimura (Hideaki Kimura) wrote…</i></summary><blockquote>

There is a hard count too, and it is lower than 100: NEW_TEST_LIMIT = 50, just above this one. Past 50 new tests a seeded sample of 50 is repeated, so a PR adding 1,300 tests costs 50 x 19 = 950 extra executions on each lane, not 1,300 x 19. The ratio answers a different question: whether the previous launch is complete enough to be the baseline at all. A list covering less than 90% of this run's tests is not used, and nothing is repeated. Reworded the comment so it doesn't read as the cap.
</blockquote></details>

I assume your Approval implies this discsussion is resolved, but please explicitly Resolve on Reviewable next time. Otherwise Reviewable still says "unstable". It doesn't auto-resolve everything up on the reviewer's Approve.

PR #34037, opened 2026-09-18
[#34052] xCluster: Add --source_certs_dir_name for cross-universe verification

balajisubramanian-yugabyte:
## Summary

`verify_xcluster_slice` and `verify_xcluster_group` are the only `yb-admin` commands that connect to a second universe directly from the client. Every other cross-universe command (`setup_universe_replication`, `create_xcluster_checkpoint`, and friends) sends an RPC to the local master, which reaches the peer using the certificates staged under `certs_for_cdc_dir`.

That direct client connection built its secure context from `--certs_dir_name`, which names the universe the tool was pointed at. When the two universes were provisioned with separate certificate authorities, the source handshake was rejected and the command failed after the client's RPC timeout with `Unable to connect to source masters`. TLS-enabled stress runs hit this, and there was no flag that could fix it: naming the source's certificates in `--certs_dir_name` only moved the failure to the target connection.

`--source_certs_dir_name` applies to the source connection and only it. Leaving it empty preserves the existing behavior of falling back to `--certs_dir_name`, which is correct when both universes share a certificate authority. Operators whose universes have separate authorities point it at the source's certificates, which xCluster already stages on the target's nodes at `<certs_for_cdc_dir>/<replication_group_id>`.

## Upgrade/Rollback safety

No wire-format, catalog, or on-disk change. The new `yb-admin` flag is client-side only and defaults to empty, which is the pre-change behavior, so an upgraded `yb-admin` against an older cluster and an older `yb-admin` against an upgraded cluster both behave exactly as before. Rollback is dropping the binary back; the flag is only read by the two verify commands, and any script passing it would need it removed.

## Test plan

- [x] `yb_build.sh release --cxx-test yb-admin-test --gtest_filter AdminCliTestWithYSQL.TestVerifyXClusterRejectsInvalidArguments` — passes (50s).

The added case names certificates for the source against a plaintext cluster. The source connection turns TLS and the master rejects it (`Invalid connection header: 1603...`, a TLS ClientHello arriving at a plaintext RPC port), while the target connection, made without certificates, still succeeds. The test therefore fails if the flag is ignored or if it leaks into the target client.

Also reproduced the original bug by hand before the fix: a TLS master under one CA, `yb-admin` pointed at a second CA, reproducing the same handshake rejection and 60-second timeout seen in the stress runs.


src/yb/tools/yb-admin-test.cc
```diff
@@ -2078,6 +2079,17 @@ TEST_F(AdminCliTestWithYSQL, TestVerifyXClusterRejectsInvalidArguments) {
   ASSERT_STR_CONTAINS(
       ASSERT_RESULT(ok_root["detail"].GetString()),
       "unable to resolve the target's xCluster safe time");
+
+  // --source_certs_dir_name has to reach the source connection and only it. This cluster serves
+  // plaintext, so naming certificates for the source makes that one connection attempt TLS and it
+  // fails, while the target connection, made without certificates, still succeeds. Were the flag
+  // ignored, or applied to both, the failure would not be this one.
+  auto source_certs = CallAdmin(
+      "--source_certs_dir_name", GetCertsDir(),
```
hkimura, line 2088: This pins the routing (source, not target) but not that a source with the right CA accepts the connection. Could a positive case go in the xcluster fixtures? `xcluster_db_scoped-test.cc` already drives both commands through `CallAdmin`, `xcluster-test.cc` has the TLS setup, and `SetupUniverseReplication` stages certs under `<certs_for_cdc_dir>/<group>`. Pointing `--source_certs_dir_name` at `test_certs/CA2` there would fail the source side alone. That would also cover `verify_xcluster_group`, which takes the new flag but isn't exercised anywhere yet. Fine as a follow-up.

src/yb/tools/tools_utils.cc
```diff
@@ -28,15 +28,23 @@ namespace yb::tools {
...
 Result<std::unique_ptr<rpc::SecureContext>> CreateSecureContextIfNeeded(
-    rpc::MessengerBuilder& messenger_builder) {
-  auto certs_dir_name = FLAGS_certs_dir_name;
+    rpc::MessengerBuilder& messenger_builder, const std::string& certs_dir) {
+  // The certificate flags describe the universe this tool was pointed at, so when certs_dir names
+  // a second universe neither the directory nor the node name beside it applies. Leaving the node
+  // name empty asks only for ca.crt, matching the cross-universe clients in client/xcluster_client
+  // and master/xcluster_rpc_tasks, which is why <certs_for_cdc_dir>/<group> holds nothing else.
+  auto certs_dir_name = certs_dir;
+  std::string cert_name;
```
hkimura, line 37: When the source universe has `node_to_node_encryption_use_client_certificates=true`, wouldn't it refuse both clients, not just this one (master's cross-universe client seems to pass an empty name then)? Let's check on this approach with @hari90  ,who reviewed https://github.com/yugabyte/yugabyte-db/commit/c0b84c779aad8aa8caad835da79a338a2c3cc872 .

hkimura (approved):
Two minor issues to follow up. LGTM.

src/yb/tools/yb-admin-test.cc
balajisubramanian-yugabyte: Yes, good point. added a test case rather than followup.

src/yb/tools/tools_utils.cc
balajisubramanian-yugabyte: Good catch . it seems to affects both. But the master's client doesn't refuse, it  seems to crash . It asks for client certificates while passing an empty node name, and an empty name only loads ca.crt. So it thinks it has a certificate to hand over, and reaching for it reads off the end of an empty list.

Instead of crashing it, fixed with clear error. The secure tests all still pass, including YbTools, which runs with the flag on.

@hari90 can you  please check this approach

src/yb/tools/yb-admin_cli.cc
```diff
@@ -82,6 +82,13 @@ DEFINE_NON_RUNTIME_int64(timeout_ms, 1000 * 60, "RPC timeout in milliseconds");
 
 // Command-specific flags
 DEFINE_NON_RUNTIME_bool(exclude_dead, false, "Exclude dead tservers from output");
+DEFINE_NON_RUNTIME_string(source_certs_dir_name, "",
+    "Directory with certificates for connecting to the source universe, used by the "
```
hari90, line 86: whats a source universe?
Lets rename this to xcluster_source_certs_dir_name

balajisubramanian-yugabyte: Ok. sounds better. will rename it.

Issue #34073, opened 2026-09-21
[DocDB] Write fence is checked before the op's hybrid time is assigned, so a write can commit past its fence


mdbridge:
Jira Link: [DB-23681](https://yugabyte.atlassian.net/browse/DB-23681)
## Description

The write fence added in #33175 (`PgsqlWriteRequestPB.ignore_after_hybrid_time`, commit 26d436e257a) is checked in `RaftConsensus::CheckWriteFenceUnlocked` before the op is assigned its own hybrid time.  The check reads `clock_->Now()` and rejects if the fence has already passed; the op's commit hybrid time is then read from the clock again later, in `MvccManager::AddLeaderPending`.  Nothing bounds the gap between the two reads, so a write can pass the check and still commit at a hybrid time later than its fence.

Both reads happen on the same thread under the ReplicaState lock with no I/O between them, so the gap is normally microseconds.  But the fence's purpose is lease handoff: the outgoing holder's writes must not take effect after its window ends, and "take effect" is decided by the commit hybrid time, not by when the check ran.  A new holder that reads at time R after the fence and then acts on what it read can have a stalled write from the old holder land at a hybrid time between R and the new holder's own write.  Within one tablet the ReplicaState lock keeps admission ordered, so the exposure is to reads and to writes on other tablets.  A stall long enough to matter (past the skew allowance a lease protocol waits out) would have to happen inside a lock-held critical section, so this is unlikely in practice, but it is not excluded.

**Fix:** compare the fence against the hybrid time actually assigned to the op, i.e., choose `ht` in `AddLeaderPending` first and then reject if `ht > fence`.  Rejecting at that point needs new plumbing: a failure from `Operation::AddedToLeader` currently returns out of `DoAppendNewRoundsToQueueUnlocked` without rolling back the op id allocated by `NewIdUnlocked`, unlike the `AddPendingOperation` failure path just below it.  The same post-assignment gate is needed by the planned xCluster StartPersisting/StopPersisting mechanism, so the two should share it.

## Issue Type

kind/bug

## Warning: Please confirm that this issue does not contain any sensitive information

- [x] I confirm this issue does not contain any sensitive information.

[DB-23681]: https://yugabyte.atlassian.net/browse/DB-23681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #34076, opened 2026-09-21
[#34073] DocDB: Judge the write fence against the op's assigned hybrid time

mdbridge:
## Summary

The `ignore_after_hybrid_time` write fence (#33175) exists so a lease holder's writes cannot take effect after its window ends.  Taking effect is decided by the op's commit hybrid time, which `MvccManager::AddLeaderPending` assigns.  The fence was compared against a separate, earlier clock reading in `RaftConsensus::CheckWriteFenceUnlocked`.  Both reads happen on one thread under the ReplicaState lock, but nothing bounds the stall between them, so a write could pass the check and still commit at a hybrid time past its fence.  Within one tablet admission stays ordered; the exposure is to reads and to writes on other tablets after a takeover.

**What changed**

- `MvccManager::AddLeaderPending` takes the fence and returns `Expired` / `WRITE_FENCE_EXPIRED` when the hybrid time it just chose is not before the fence, without adding the op.  `WriteOperation` supplies the fence through a new `Operation::WriteFence` hook; the pre-assignment check in `RaftConsensus` is removed.
- A failure from `AddedToLeader` now fails only its own round: the op id is rolled back and the round is notified through `ReplicaState`, which also undoes the retryable-request registration.  Previously such a failure (reachable only when the tablet was shutting down) failed the whole batch and left the allocated op id behind.  `OperationDriver` clears its op id copy on that path so the abort is not reported as having been pending.
- The batch lease check now looks at the last message actually appended rather than the last round.  A round rejected earlier in the loop carries no hybrid time, so checking it tested the lease at time zero.
- New test flag `TEST_inject_mvcc_delay_before_add_leader_pending_ms` stalls before the hybrid time is chosen, to reproduce the race.

## Test plan

- [x] `PgThinClientTest.WriteFencedByIgnoreAfterHybridTime`, extended with a write whose fence passes during an injected stall before its hybrid time is chosen; it must be rejected as fenced and leave no row and no retryable-request registration.
- [x] `PgThinClientTest.ReplayOfReplicatedWriteWinsOverExpiredFence` and `PgThinClientTest.AlreadyReplicatedWriteReportsSuccess`, which exercise the fence's interaction with retryable requests.
- [x] `mvcc-test` (all 6) and `raft_consensus-test` (all 9), release build.


src/yb/consensus/raft_consensus.cc
```diff
@@ -1310,8 +1310,8 @@ Status RaftConsensus::AppendNewRoundToQueueUnlocked(const scoped_refptr<Consensu
   return AppendNewRoundsToQueueUnlocked({ round }, &processed_rounds);
 }
 
-Status RaftConsensus::CheckLeasesUnlocked(const ConsensusRoundPtr& round) {
```
hari90, line 1313: Let's undo changes to this function. Is not related to write fence

src/yb/tablet/operations/operation_driver.cc
```diff
@@ -210,7 +210,12 @@ Status OperationDriver::AddedToLeader(const OpId& op_id, const OpId& committed_o
   CHECK(!GetOpId().valid());
   op_id_copy_.store(op_id, boost::memory_order_release);
 
-  RETURN_NOT_OK(operation_->AddedToLeader(op_id, committed_op_id));
+  if (auto s = operation_->AddedToLeader(op_id, committed_op_id); !s.ok()) {
+    // Consensus rolls the op id back and the operation was never added as pending; a valid op id
+    // here would make HandleFailure report it as having been.
+    op_id_copy_.store(OpId(), boost::memory_order_release);
```
hari90, line 216: @spolitov  can you please double check this

hari90 (approved):

PR #34081, opened 2026-09-21
[PLAT-18087] YBA: Allow skipping storage config validation via runtime config

shubin-yb:
## Summary

Creating a storage configuration validates it against the backup storage itself -
credentials, bucket existence, and read/write/list/delete permissions. There was no way to
bypass that check, so storage YBA cannot reach directly (but the DB nodes can) could not be
configured at all.

Adds `yb.storage_config.skip_validation`, a global runtime config defaulting to `false` and
tagged `INTERNAL` so it stays out of the UI and out of `RUNTIME-FLAGS.md`. When set,
`CustomerConfigValidator.validateConfig` skips the registered `ConfigDataValidator` for
storage configs - the validators that reach out to the backup storage.

The skip is deliberately narrow: it sits *after* field-level bean validation and is gated on
`ConfigType.STORAGE`, so nothing that is local to YBA gets disarmed along with it.

- **Field-level validation of the payload** (`@NotNull` backup location, key and secret
  lengths, location length constraint) still runs. These are local checks, and an
  unreachable-storage escape hatch has no business skipping them.
- **Non-storage config types** (alerts preferences, password policy) are unaffected - their
  validators never touch the network, so there is nothing to escape.
- **Config name conflict** still runs. It is backed by the DB constraint
  `config_name_unique (customer_uuid, config_name)`, so skipping it would only turn a clean
  409 into a constraint-violation 500.
- **Read-only backup location** still runs. Skipping it would let an edit repoint a config
  that existing backups resolve through, orphaning them.

`CustomerConfigService.create()/edit()` is the only write path - the v2 handler is read-only
- so this single gate covers the whole API surface.

Operator note: this complements the universe-scoped
`yb.backup.skip_config_based_preflight_validation`, which unblocks *using* a storage config.
This one unblocks *creating* it; a deployment hitting this problem likely needs both.

## Upgrade/Rollback safety

No upgrade or rollback impact by default. The new key defaults to `false`, so behavior is
unchanged unless an operator explicitly sets it.

On rollback to a YBA version without this key, a `yb.storage_config.skip_validation` row
left in `runtime_config_entry` is simply never read - unrecognized runtime config paths are
inert.

Storage configs created while the flag was set persist normally across upgrade and rollback.
Their payloads were still field-validated; what was skipped is the check that the credentials
and bucket actually work, so a bad one surfaces at backup time rather than at config creation
time. That is inherent to the escape hatch, not to the version change.

## Test plan

- [x] `sbt "testOnly com.yugabyte.yw.models.helpers.CustomerConfigValidatorTest com.yugabyte.yw.common.config.ConfKeysTest"` - 92 passed, 0 failed
- [x] `testValidateConfig_SkipStorageValidationRuntimeConfig` - a storage config whose credentials are rejected still fails with the flag off, and is accepted once it is set
- [x] `testValidateConfig_SkipStorageValidationKeepsFieldValidation` - with the flag set, an S3 config missing `BACKUP_LOCATION` is still rejected by bean validation
- [x] `testValidateConfig_SkipStorageValidationLeavesNonStorageConfigsAlone` - with the flag set, an alerts config carrying an invalid email is still rejected by its own validator
- [x] `testValidateConfig_SkipStorageValidationKeepsNameConflictCheck` - with the flag set, a duplicate config name is still rejected
- [x] `ConfKeysTest` exercises the new key automatically: registered as mutable, default resolves from `reference.conf`, accepts a valid value, rejects an invalid one
- [x] `sbt javafmt Test/javafmt` and `build-support/lint.sh --rev upstream/master` both clean


nkhogen (approved):
LGTM

PR #34087, opened 2026-09-22
[BACKPORT 2026.1][#33916] DocDB: Collect thin client stack traces through libunwind (#33917)

hari90:
## Summary

Fixes #33916. `ybthin_client_create` switches stack trace collection to
libunwind before anything in
the client creates a thread. Sanitizer builds keep their default.

With the default glibc `backtrace()`, `ybthin_client_create` can
deadlock and never return on a host
whose allocator unwinds from inside `malloc` — jemalloc with heap
profiling, a leak checker —
reported at roughly 1 in 8 process starts. `Thread::Create` warms up the
stack trace library on
first use; that warm-up unwinds through the host's libgcc, which sorts
this `.so`'s registered
`.eh_frame` lazily inside a `malloc` held under `object_mutex`. A host
allocator that unwinds from
inside that `malloc` re-enters `_Unwind_Find_FDE` and blocks on the
mutex its own thread holds.

libunwind keeps its own FDE cache and is already statically linked into
the `.so` — verified against
the shipped 2026.1.2.0 tarball: no `unw_*` undefined symbols, no
`libunwind.so` dependency. No new
dependency, no packaging change.

Skipping the warm-up instead would not fix it; that defers the same FDE
sort to the first real
`StackTrace::Collect`.

## Trade-off

libunwind can SIGSEGV collecting a trace in a BOLT-ed binary, which is
why the non-sanitizer
default is `false`. This gives up BOLT for this `.so`.
`src/yb/thin_client/yb_release` does not pass `--bolt`, so nothing
regresses today.

## Test Plan

`yb_thin_client-itest`.

`ClientCreateSelectsLibunwindForStackTraces` creates a client twice in
one process and asserts the
flag after each, so a once-per-process write cannot pass. Reproducing
the deadlock needs a
jemalloc-profiling host, so the test asserts the condition that prevents
it rather than the absence
of the hang.

## Follow-ups

Backport to 2026.1.2 is on `yb_thin_client_2026.1.2`; the thin client
tarballs already on
`s3://downloads.yugabyte.com/yb_thin_client/` predate this fix.

## Upgrade/Rollback safety

No wire, storage, or catalog change.

---------

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>

Original commit: 1dadd9e2813b46598ed6a35b364425cfc084da2b / #33917


hari90:
Trigger Jenkins

es1024 (approved):

PR #34103, opened 2026-09-22
[PLAT-22303] YBA: Capture max and p95 Postgres backend PSS

vipul-yb:
## Summary

Existing PSS Memory for `yb-postmaster` is a node-wide sum of proportional set size, so a single backend that spikes memory is hard to see. Collect max and p95 of per-backend PSS in `node_health.py` from the PSS already gathered in `check_process_stats`, and add two Per-process Metrics panels (MB) so operators can spot fat backends without another `/proc` walk.

## Test plan

- [x] `sbt "test:testOnly com.yugabyte.yw.models.MetricConfigTest"` (includes `testMetricsYaml`)
- [x] `sbt "test:testOnly com.yugabyte.yw.controllers.GrafanaGenTest"`
- [x] `./build-support/lint.sh --rev origin/master` (warnings only, pre-existing)
- [ ] After YBA deploys the new `node_health.py`: confirm `yb_postgres_backend_pss_max_kb` and `yb_postgres_backend_pss_p95_kb` are present and max ≥ p95


anmalysh-yb (approved):
LGTM

